PR #1311, opened 2022-06-11
Make glibc linux update to dynamic binary and alpine linux to static alpine binary

MuweiHe:
Checklist
---------
1. [CRUCIAL] Is the change for CP or CCloud functionalities that are already live in prod?
   * yes: ok

What
----

Adjusted the binary/archive names of linux builds: musl to `alpine`, glibc to `linux`
Renamed the download link in `confluent update` command, so glibc users will update to the dynamic `linux` build, and alpine users to static `alpine` build.
Update linux built in container using `hub` to github. No need to upload manually anymore.

References
----------

Test & Review
-------------


CobraFighter:
Confluent sh has certainly met a requirement in the mind of my cobra pit outside when his natural habitat fluctuates; pretend to pick up as many mushrooms you can fire across the moon

PR #1430, opened 2022-09-09
Use forks for muckrake in jenkinsjob

imcdo:
Checklist
---------

What
----
update the jenkins job such that when cloning muckrake it pushes to the same branch that vagrant will pull from even if its a different fork.

References
----------

Test & Review
-------------


brianstrauch:
Closing based on our discussion today. We'll be replacing the Jenkinsfile with a more appropriate long-term solution.

PR #1493, opened 2022-10-28
[CLI-1683] Update error message to help user if they delete active cluster from the UI

sgagniere:
Release Notes
-------------

Breaking Changes
- <PLACEHOLDER>

New Features
- <PLACEHOLDER>

Bug Fixes
- <PLACEHOLDER>

Checklist
---------
1. [CRUCIAL] Is the change for CP or CCloud functionalities that are already live in prod?
   * yes: ok

What
----
Deleting a cluster from the UI does not remove its information from the local `config.json`. This is generally not harmful, but if the *active* cluster is deleted from the UI, it remains the active cluster until a new one is set, or until the information is manually removed from the json file.

Currently, this results in `Error: Kafka cluster not found or access forbidden` using any command which defaults to the active cluster when no cluster is provided, which can be confusing, as neither of the current suggestions address what is happening.

Since this is a rare edge case, instead of checking every time if the active cluster is one of the valid clusters (e.g. by searching for it in the list of clusters for the current environment), I've modified the error message to include the cluster ID and added a new suggestion about deleted active clusters.

References
----------
https://confluentinc.atlassian.net/browse/CLI-1683

Test & Review
-------------
Manual tested with these steps:
1. Create basic cluster
2. Set it as the active cluster
3. Delete it from the UI
4. Run `kafka cluster describe` and `kafka topic list`.

Ran tests.

Open questions / Follow ups
---------------------------
The other issue raised in this ticket (graceful handling of deleted resources) is effectively the same issue from these two tickets:
https://confluentinc.atlassian.net/browse/CLI-672
https://confluentinc.atlassian.net/browse/CLI-673


internal/pkg/errors/error_message.go
```diff
@@ -154,6 +154,9 @@ const (
 	KafkaClusterShrinkingErrorMsg                 = "your cluster is shrinking; Please wait for that operation to complete before updating again"
 	KafkaClusterDeletingSuggestions               = ChooseRightEnvironmentSuggestions + "\n" +
 		"Ensure the cluster is not associated with any active Connect clusters."
+	KafkaClusterInaccessibleErrorMsg    = `Kafka cluster "%s" not found or access forbidden`
+	KafkaClusterInaccessibleSuggestions = ChooseRightEnvironmentSuggestions + "\n" +
```
brianstrauch, line 158: Do we have a test where this line is printed?

sgagniere: The error message yeah, but not the new suggestion.

internal/pkg/errors/catcher.go
```diff
@@ -221,11 +221,11 @@ func CatchKafkaNotFoundError(err error, clusterId string, r *http.Response) erro
 	}
 
 	if r != nil && r.StatusCode == http.StatusForbidden {
-		suggestions := ChooseRightEnvironmentSuggestions
+		suggestions := KafkaClusterInaccessibleSuggestions
 		if r.Request.Method == http.MethodDelete {
```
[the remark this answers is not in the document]
sgagniere: I added the new line to the delete suggestions as well.

internal/pkg/errors/error_message.go
sgagniere: Added a new int test, and I updated the delete suggestions to also include this new line.

brianstrauch (approved):

PR #1667, opened 2023-01-25
[CLI-2087] Remove prod hardcoding in auto login

sgagniere:
Checklist
---------
   * yes: ok

What
----
The auto login logic currently sends all requests to `pauth.CCloudURL`, which means that devel and stag cannot auto login with saved credentials.

This PR replaces that url with the url stored in the `platform.server` parameter of the context when it's available, and defaults to `pauth.CCloudURL` otherwise. I decided not to extract the url from the netrc machine name since there are other ways to auto login (env vars) which don't rely on the netrc.

Note: adding the URL to `NetrcMachineParams` shouldn't really do anything since the matching logic ignores it if the name parameter is present, but I added it in just in case.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2087

Test & Review
-------------
Ran all tests
Manual testing: logged into a devel org with `--save`, waited for the login to expire, and then ran `confluent environment list`.


MuweiHe (commented):
TY for catching this! LGTM once comments addressed

brianstrauch (approved):

PR #1673, opened 2023-01-26
Fix Semaphore

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Checking to see what's leaving Semaphore in a dirty state.


sgagniere (approved):

patrickfeltes:
probably the best PR I've ever seen

Issue #1674, opened 2023-01-26
Support building on FreeBSD


dveeden:
Hi,

I tried building confluentinc/cli on FreeBSD 13.1-RELEASE-p3 and found this:

The `Makefile` is GNU Make specific (not very unusual). The `README.md` says to use `make deps` and `make build` and the `Makefile` also calls `make`. Replacing `make` with `gmake` in both the `README.md` and `Makefile` fixes that. The `gmake` command is also available on Linux so this should make this more portable without requiring FreeBSD specific things.

For the cgo dependency for `rdkafa` I needed to do a `pkg install librdkafka` to get librdkafka 1.9.2.

Then to build I used `gmake CGO_CFLAGS=`pkg-config --cflags rdkafka` CGO_LDFLAGS=`pkg-config --libs rdkafka` TAGS='dynamic' build`. Afterwards I noticed that building with `gmake TAGS='dynamic' build` also works, but I'm not sure if that also works for fresh builds.

```
% go version -m ./dist/confluent_freebsd_amd64_v1/confluent | grep -vw dep
./dist/confluent_freebsd_amd64_v1/confluent: go1.19.5
	path	command-line-arguments
	build	-asmflags=all=-trimpath=/home/dveeden/git
	build	-compiler=gc
	build	-gcflags=all=-trimpath=/home/dveeden/git
	build	-ldflags="-s -w -X main.version=v3.0.0-dirty-dveeden -X main.commit=83072d47 -X main.date=2023-01-26T08:38:18Z"
	build	-tags=dynamic
	build	CGO_ENABLED=1
	build	CGO_CFLAGS=
	build	CGO_CPPFLAGS=
	build	CGO_CXXFLAGS=
	build	CGO_LDFLAGS=
	build	GOARCH=amd64
	build	GOOS=freebsd
	build	GOAMD64=v1
% file ./dist/confluent_freebsd_amd64_v1/confluent
./dist/confluent_freebsd_amd64_v1/confluent: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), dynamically linked, interpreter /libexec/ld-elf.so.1, for FreeBSD 13.1, FreeBSD-style, Go BuildID=nbtGuL4rOvzdWcchfu4k/FWCY7oI7U3W9T22Jk4pU/SjSsSKJdis-nENQXiNjp/mXRzoVc2_k0MVOozlSpq, stripped
% ./dist/confluent_freebsd_amd64_v1/confluent version
[...]

PR #1675, opened 2023-01-26
Explicitly use GNU Make

dveeden:
Checklist
---------
   * yes: ok

What
----

On many platforms `make` is GNU Make, but not on all. On FreeBSD it isn't. As the `Makefile` is GNU Make specific let's use `gmake` to allow this to work on more platforms.

References
----------

Issue: #1674

Test & Review
-------------

Manually, see issue

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,15 +4,15 @@ GORELEASER_VERSION := v1.17.2
 .PHONY: build 
 build:
 ifneq "" "$(findstring NT,$(shell uname))" # windows
-	CC=gcc CXX=g++ make cli-builder
+	CC=gcc CXX=g++ $(MAKE) cli-builder
 else ifneq (,$(findstring Linux,$(shell uname)))
     ifneq (,$(findstring musl,$(shell ldd --version))) # linux (musl)
-		CC=gcc CXX=g++ TAGS=musl make cli-builder
+		CC=gcc CXX=g++ TAGS=musl $(MAKE) cli-builder
     else # linux (glibc)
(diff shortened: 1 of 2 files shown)


brianstrauch (commented):
Apologies for taking a while to review. I don't believe `gmake` comes pre-installed every system. Some of our users run the CLI on Windows machines, for example.

A possible multi-platform approach might be to create a variable called `$MAKE`, which is either set to `gmake` or `make`, depending on which one can be found in the user's `$PATH`.

dveeden:
@brianstrauch Are you sure Windows has GNU Make as `make` but not as `gmake`?

dveeden:
Actually, looks like this is the proper way to do this: https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

dveeden:
With this I can call `make` on Linux and `gmake` on FreeBSD and everything should be fine. Note that `make` on FreeBSD is `bmake` which doesn't have the extended GNU behaviour.

dveeden:
@brianstrauch Could you check the output of the Jenkins job? I don't have permission to see the output.

brianstrauch:
@dveeden Thanks, `$(MAKE)` looks like a great solution! One minor papercut before I merge this: Can you run `make generate-packaging-patch` so our debian packaging can be updated? Thanks again for the PR!

dveeden:
@brianstrauch Yes, done in c09434291866162f38383b4656a76134bdeeb87a

brianstrauch (approved):

PR #1676, opened 2023-01-26
Relabel docker tags

sgagniere:
Checklist
---------
   * yes: ok

What
----
Change docker image tags to 'latest'.

References
----------
N/A

Test & Review
-------------
N/A


brianstrauch (approved):
Thanks!

PR #1677, opened 2023-01-27
Remove the leading v from checksum file names in release.mk

MuweiHe:
Checklist
---------
   * yes: ok

What
----

We have removed those leading `v`s and they should be removed from checksum filenames and path names as well. This PR was accidentally reverted before.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
Thank you!

PR #1678, opened 2023-01-27
[RCCA-10621] Pass resource IDs to Kafka REST in `kafka acl` commands

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Previously, all principals were being passed to Kafka REST as numeric IDs when creating/listing/deleting a Kafka ACL with `kafka acl`. These should no longer be converted prior to making the request because the backend can handle resource IDs now.

The backend may still respond with numeric IDs (if it can't make the mapping), so I'm keeping the code that maps numeric IDs back to resource IDs so we don't accidentally show numeric IDs to the user.

References
----------
https://confluentinc.atlassian.net/browse/RCCA-10621

Test & Review
-------------
Verified that Kafka ACLs can still be created, listed, and deleted.

### DIFF
--- a/internal/cmd/kafka/command_acl.go
+++ b/internal/cmd/kafka/command_acl.go
@@ -14,7 +14,6 @@ import (
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 	"github.com/confluentinc/cli/internal/pkg/errors"
-	"github.com/confluentinc/cli/internal/pkg/resource"
 )
 
 type aclCommand struct {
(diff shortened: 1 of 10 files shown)


sgagniere (approved):
👍

MuweiHe:
`so I'm keeping the code that maps numeric IDs back to resource IDs so we don't accidentally show numeric IDs to the user.`
what happens if backend returns resource id? it's now showing that part of code in this pr maybe, so just curious

brianstrauch:
> what happens if backend returns resource id?

@MuweiHe The backend returns resource IDs by default. It only returns numeric IDs if it can't perform the mapping (in that case, we identify that the principal doesn't exist, and we don't show it to the user)

brianstrauch:
The Kafka team has not fully deployed this in production. Marking as draft until it's ready.

brianstrauch:
A force push may have knocked this out-of-sync with the main branch. Closing and reopening this PR on a different branch.

PR #1679, opened 2023-01-27
[SD-1668] Fix the activation privileges in the output

xiangxin72:
Checklist
---------
   * yes: ok

What
----

This seems an regression introduced during migration to v3:
* the `ActivationPrivilege` was removed from the pipeline output
* the unit test file `update-activation-privilege.golden` was also updated hence the integ test didn't catch this

This fix introduce back the property

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
Thanks!

xiangxin72:
@brianstrauch any idea what the build failure is about?

brianstrauch:
@xiangxin72 This is due to instability in muckrake and is ok to ignore!

PR #1680, opened 2023-01-28
`iam rbac role-binding` fixes

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Two bug fixes:
1. It's possible for the `Organization` field to be `nil`. We should be using the `Account` field instead.
2. We need to use `TrimPrefix()` instead of `TrimLeft()`, otherwise principals starting with any of the characters "U", "s", "e", "r", or ":" will be trimmed.


DABH (approved):
LGTM

sgagniere (approved):

MuweiHe (approved):

PR #1685, opened 2023-01-31
Improve error messages and lint rules

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Represent flags properly in error messages, add more lint rules, replace all instances of `interface{}` with `any`


internal/pkg/auth/login_credentials_manager.go
```diff
@@ -191,7 +191,7 @@ func (h *LoginCredentialsManagerImpl) GetCredentialsFromNetrc(filterParams netrc
 			return nil, err
 		}
 
-		log.CliLogger.Warnf(errors.FoundNetrcCredMsg, netrcMachine.User, h.netrcHandler.GetFileName())
```
sgagniere, line 194: 👍

sgagniere (approved):
lgtm

PR #1686, opened 2023-02-01
[CLI-1470] encrypt login credentials and save to config.json, and keychain on macos

MuweiHe:
Checklist
---------
   * yes: ok

What
----

On darwin, keychain is used to store `username?password`. This has higher priority than netrc and config, and might prompt the user for a system password.

Use random salt, machine id and user id to generate the encryption key. Use this key and random nonce to encrypt/decrypt login credentials, for both on-prem and cloud. Username is used as additional data. On Windows, this is done using the native DPAPI.

When this fails, fall back to treating netrc credentials as plaintext. If that login succeeded, encrypt the credential and save to config.json. No longer write to netrc file. After logout, the original plaintext creds should be removed.

References
----------

Test & Review
-------------

Tested behaviors (on linux, windows too)
1. first login with --save, will save and overwrite the creds if necessary. 
2. second login should use creds from netrc (keychain for mac)
3. logout should clean up netrc, config and keychain for mac
4. if the session timeout, it'll look for keychain for mac first, return the first entry. If this failed, will continue to config and look for an entry that matches the filter params.

Open Questions / Follow-ups
---------------------------


internal/pkg/cmd/prerunner.go
```diff
@@ -394,7 +394,8 @@ func (r *PreRun) ccloudAutoLogin(netrcMachineName string) error {
 
 	client := r.CCloudClientFactory.JwtHTTPClientFactory(context.Background(), credentials.AuthToken, url)
 	currentEnv, currentOrg, err := pauth.PersistCCloudCredentialsToConfig(r.Config, client, url, credentials)
-	if err != nil {
+	if err != nil { // if err == invalid creds
+		fmt.Println("failed here.")
```
brianstrauch, line 398:
```suggestion
```

internal/pkg/config/v1/config.go
```diff
@@ -127,6 +128,7 @@ func New() *Config {
 		Contexts:      make(map[string]*Context),
 		ContextStates: make(map[string]*ContextState),
 		AnonymousId:   uuid.New().String(),
+		Salt:          uuid.New().String(),
```
brianstrauch, line 131: Is this a best practice for generating the salt? Is it pseudorandom generator or does it generate IDs predictably?

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -0,0 +1,98 @@
+//go:build darwin
+// +build darwin
```
brianstrauch, line 2:
```suggestion
```

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -0,0 +1,98 @@
...
+
+func WriteKeychainItem(ctxName, url, username, password string) error { // this happens after login, so you can have the ctxName
+	service := netrc.GetLocalCredentialName(isCloud, ctxName)
+
+	item := keychain.NewGenericPassword(service, url, username+"-"+url, []byte(username+sep+password), accessGroup) // save username as the label
+	item.SetSynchronizable(keychain.SynchronizableNo)
+	item.SetAccessible(keychain.AccessibleWhenUnlocked)
+
+	err := keychain.AddItem(item)
+	if err == keychain.ErrorDuplicateItem {
```
brianstrauch, line 30: What if there's an error that's not "duplicate item"?

internal/pkg/config/v1/config.go
MuweiHe: @ankurjoshi9 wdyt?

internal/pkg/secret/encryption_other.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	aesgcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return "", err
+	}
+
+	encryptedPassword := aesgcm.Seal(nil, encryptionKey[:12], []byte(password), nil)
```
MuweiHe, line 46: @ankurjoshi9 is it ok if we reuse part of the encryption key as nonce? Or do we need to store another random variable like salt in our configs?

internal/cmd/login/command.go
```diff
@@ -353,8 +400,12 @@ func (c *command) saveLoginToNetrc(cmd *cobra.Command, isCloud bool, credentials
 			utils.ErrPrintln(cmd, "The `--save` flag was ignored since SSO credentials are not stored locally.")
 			return nil
 		}
+		ctxName := c.Config.Config.Context().NetrcMachineName // after login, this must be not nil
```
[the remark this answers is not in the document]
MuweiHe: i'm ok with renaming it since keychain is also using it now

internal/cmd/login/command.go
```diff
@@ -214,15 +222,39 @@ func (c *command) getCCloudCredentials(cmd *cobra.Command, url, orgResourceId st
...
+		c.loginCredentialsManager.GetCredentialsFromNetrc(cmd, netrcFilterParams),
 		c.loginCredentialsManager.GetCloudCredentialsFromPrompt(cmd, orgResourceId),
 	)
 }
 
 func (c *command) loginMDS(cmd *cobra.Command, url string) error {
 	credentials, err := c.getConfluentCredentials(cmd, url)
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "failed") || strings.Contains(err.Error(), "found no netrc machine") { // decryption failed is expected when trying on unencrypted creds
```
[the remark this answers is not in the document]
MuweiHe: removed this part of the code since i combined those `get credentials` calls in one function!

internal/cmd/login/command.go
```diff
@@ -353,8 +400,12 @@ func (c *command) saveLoginToNetrc(cmd *cobra.Command, isCloud bool, credentials
 			utils.ErrPrintln(cmd, "The `--save` flag was ignored since SSO credentials are not stored locally.")
 			return nil
 		}
+		ctxName := c.Config.Config.Context().NetrcMachineName // after login, this must be not nil
+		if err := keychain.WriteKeychainItem(ctxName, url, credentials.Username, credentials.Password); err != nil {
```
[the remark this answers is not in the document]
MuweiHe: it is, but since url is there ready to use, it's easier than to parse ctxName again

internal/pkg/auth/login_credentials_manager.go
```diff
@@ -183,7 +187,7 @@ func (h *LoginCredentialsManagerImpl) GetPrerunCredentialsFromConfig(cfg *v1.Con
 	}
 }
 
-func (h *LoginCredentialsManagerImpl) GetCredentialsFromNetrc(filterParams netrc.NetrcMachineParams) func() (*Credentials, error) {
+func (h *LoginCredentialsManagerImpl) GetCredentialsFromNetrcWithSalt(filterParams netrc.NetrcMachineParams, salt string) func() (*Credentials, error) {
```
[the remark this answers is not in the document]
MuweiHe: that'll be a separate pr but i'll look into that for sure

internal/pkg/auth/login_credentials_manager.go
```diff
@@ -193,6 +197,31 @@ func (h *LoginCredentialsManagerImpl) GetCredentialsFromNetrc(filterParams netrc
...
+		netrcMachine, err := h.getNetrcMachine(filterParams)
+		if err != nil {
+			log.CliLogger.Debugf("Get netrc machine error: %s", err.Error())
+			return nil, err
+		}
+
+		err = cmd.Flags().Set("save", "true") // overwrite the plaintext credentials
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
MuweiHe: This is the easiest way as it returns to the main `login` function and at the end it'll read this flag. This way we have smallest code change and high reuse.

internal/pkg/keychain/keychain_darwin.go
MuweiHe: it'll return the err on the next line

internal/cmd/login/command.go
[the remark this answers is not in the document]
brianstrauch: So it looks like "<email>-<login url>" right? ... and for CP, we append the certificate path? I'd be ok with calling this "LoginContext"

internal/pkg/auth/login_credentials_manager.go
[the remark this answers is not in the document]
brianstrauch: Huh? Why do we return to the main login() function?? Shouldn't that only get called once?

internal/pkg/keychain/keychain_darwin.go
brianstrauch: But it'll print "saved item using: ...". Did you mean to delete that line?

MuweiHe: yeah all println is only for my testing and reviewing.. they'll be deleted

internal/pkg/auth/login_credentials_manager.go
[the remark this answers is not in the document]
MuweiHe: login() -> getCredentials -> GetNetrcFromNetrc(unencrypted) -> set "save" to true and return to `login()` to execute the next login steps 
write credentials to netrc is the last step in `login()` and will be executed only when `save` is true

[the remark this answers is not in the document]
brianstrauch: Oh I see, let's definitely write a comment explaining this. I suppose this'll get removed in the next major version so it's ok.

internal/pkg/config/v1/config.go
ankurjoshi9: Go library google/uuid is based on [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122#section-6). It is not recommended to use for security capabilities. 
Use https://pkg.go.dev/crypto/rand for Salt generation.

internal/pkg/secret/encryption_other.go
ankurjoshi9: encryption key can not be used as nonce. You should generate it same way as salt.

ankurjoshi9 (changes requested):

internal/pkg/secret/encryption_windows.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/billgraziano/dpapi"
+)
+
+func DeriveEncryptionKey(salt string) ([]byte, error) {
+	return []byte{}, nil
+}
+
+func Encrypt(password, _ string) (string, error) {
+	encryptedPassword, err := dpapi.Encrypt(password)
```
ankurjoshi9, line 14: Is this encryption associated with the logged in user or entire computer? DPAPI encrypts and associates data at two levels. 1. User and 2. Computer
If encryption is associated with Computer level, any user can decrypt the encrypted password. So it is best to encrypt data at User level.

internal/pkg/secret/encryption_other.go
```diff
@@ -0,0 +1,76 @@
...
```
ankurjoshi9, line 44: Last parameter of this Seal method is AAD (authenticated addition data) that has special meaning in AES GCM. It provides authenticated encryption that ensures integrity along with confidentiality. I would suggest to use logged in user's email as AAD. We are anyway going to re-encrypt the password if user login next time in CLI.

internal/pkg/secret/encryption_windows.go
MuweiHe: Both user id and machine id are being used to generate the encryption key. Do you think we should prompt user for system password too?

internal/pkg/auth/login_credentials_manager.go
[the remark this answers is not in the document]
MuweiHe: @DABH What do you think? Should we consider any circumstances that users would want to share their config files?

internal/pkg/secret/encryption_windows.go
MuweiHe: @ankurjoshi9

MuweiHe:
@ankurjoshi9 should we remove netrc password encryption for mac users, and only depend on keychain?

internal/pkg/secret/encryption_other.go
MuweiHe: Password decryption has to work when the user is not logged in and no username (email) is provided. I suppose the same aad will need to be provided to `aesgcm.Open` as well?

internal/cmd/login/command.go
```diff
@@ -400,8 +356,8 @@ func (c *command) saveLoginToLocal(cmd *cobra.Command, isCloud bool, url string,
 			utils.ErrPrintln(cmd, "The `--save` flag was ignored since SSO credentials are not stored locally.")
 			return nil
 		}
-		ctxName := c.Config.Config.Context().NetrcMachineName // after login, this must be not nil
-		if err := keychain.WriteKeychainItem(ctxName, url, credentials.Username, credentials.Password); err != nil {
+		ctxName := c.Config.Config.Context().LoginContextName // after login, this must be not nil
```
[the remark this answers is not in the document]
MuweiHe: login context is a string, unlike state and credentials which are pointers.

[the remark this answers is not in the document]
MuweiHe: leaving checking `if cxt!=nil` outside of a getter function is more readable for the code in my opinion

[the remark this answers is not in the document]
brianstrauch: We're not checking if the context is `nil` though. Right now we'll panic if it's `nil`.

internal/pkg/secret/encryption_windows.go
ankurjoshi9: I don't think we generate encryption key in this case. This is DPAPI which automatically takes care of encryption.
I am trying to understand if we use https://github.com/billgraziano/dpapi/blob/master/dpapi.go#L116 method or not?

internal/pkg/secret/encryption_other.go
ankurjoshi9: Yes same AAD has to be provided while decryption. 

> Password decryption has to work when the user is not logged in and no username (email) is provided.
When will this happen? I believe we maintain user's email id in config somewhere and that can be directly used. We do not need to ask user to provide email that can be used in AAD.

ankurjoshi9:
> should we remove netrc password encryption for mac users, and only depend on keychain?

Yes if we can use keychain, that seems better approach.

> Should we encrypt session tokens in config.json as well (might be a separate pr)?

yes we should do it like we do for password.

internal/pkg/secret/encryption_windows.go
MuweiHe: No we don't use machine level encryption. so i think this is user level and we're fine

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -79,20 +74,19 @@ func ReadCredentialsFromKeychain(ctxName, url string) (string, string, error) {
 	}
 
 	results, err := keychain.QueryItem(item)
-	if err != nil {
-		return "", "", err
-	} else {
+	if err == nil {
```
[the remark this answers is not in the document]
MuweiHe: it's actually the same as i have to check the length of the results, or i'll return nothing but just a different err

internal/cmd/login/command.go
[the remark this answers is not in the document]
MuweiHe: this save function is certain to have a non-nil context, and the other 2 instance of this function have checking `ctx!=nil` outside of this function.

internal/pkg/config/v1/config.go
```diff
@@ -73,17 +73,18 @@ var (
...
+	DisablePlugins      bool                        `json:"disable_plugins"`
+	DisableFeatureFlags bool                        `json:"disable_feature_flags"`
+	NoBrowser           bool                        `json:"no_browser" hcl:"no_browser"`
+	Platforms           map[string]*Platform        `json:"platforms,omitempty"`
+	Credentials         map[string]*Credential      `json:"credentials,omitempty"`
+	Contexts            map[string]*Context         `json:"contexts,omitempty"`
+	ContextStates       map[string]*ContextState    `json:"context_states,omitempty"`
+	CurrentContext      string                      `json:"current_context"`
+	AnonymousId         string                      `json:"anonymous_id,omitempty"`
+	SavedCredentials    map[string]*LoginCredential `json:"saved_credentials,omitempty"`
```
MuweiHe, line 87: @brianstrauch i chose to use a new map, as context states will be cleaned up when session timeout, and we can't auto-login back.

internal/pkg/config/v1/config.go
```diff
@@ -451,6 +453,14 @@ func (c *Config) SaveCredential(credential *Credential) error {
 	return c.Save()
 }
 
+func (c *Config) SaveLoginCredential(ctxName string, loginCredential *LoginCredential) error {
+	if ctxName == "" {
+		return errors.New("login credentials must belong to a state name")
```
MuweiHe, line 458: This will be put into an error_message

internal/pkg/config/v1/config_test.go
```diff
@@ -760,17 +782,6 @@ func TestConfig_FindContext(t *testing.T) {
...
-func TestConfig_DeleteContext(t *testing.T) {
-	cfg := &Config{
-		BaseConfig:     config.NewBaseConfig(new(version.Version)),
-		Contexts:       map[string]*Context{contextName: {Name: contextName}},
-		CurrentContext: contextName,
-	}
-
-	err := cfg.DeleteContext(contextName)
-	require.NoError(t, err)
-}
```
MuweiHe, line 772: When running this test alone, the whole config.json will be removed and won't be able to be reverted. This messed up the config as well, causing `unable to parse version from config` error.

test/login_test.go
```diff
@@ -213,80 +207,18 @@ func (s *CLITestSuite) TestSaveUsernamePassword() {
...
 		s.NoError(err)
 		wantBytes, err := os.ReadFile(wantFile)
 		s.NoError(err)
-		want := strings.Replace(string(wantBytes), urlPlaceHolder, tt.loginURL, 1)
-		s.Equal(utils.NormalizeNewLines(want), utils.NormalizeNewLines(string(got)))
-	}
-	_ = os.Remove(netrc.NetrcIntegrationTestFile)
-}
-
-func (s *CLITestSuite) TestUpdateNetrcPassword() {
```
MuweiHe, line 228: This will be added back. I'm still thinking how to test this

internal/pkg/secret/encryption_other.go
```diff
@@ -0,0 +1,86 @@
...
+	b := make([]byte, n)
+	_, err := rand.Read(b)
+	if err != nil {
+		return []byte{}, err
+	}
+	return b, err
+}
+
+func DeriveEncryptionKey(salt []byte) ([]byte, error) {
+	machineId, err := machineid.ID()
```
brianstrauch, line 29: Would it make sense to use https://github.com/denisbrodbeck/machineid#function-protectedidappid-string-string-error here? @ankurjoshi9

internal/pkg/auth/login_credentials_manager.go
```diff
@@ -191,8 +234,8 @@ func (h *LoginCredentialsManagerImpl) GetCredentialsFromNetrc(filterParams netrc
 			return nil, err
 		}
 
-		log.CliLogger.Debugf(errors.FoundNetrcCredMsg, netrcMachine.User, h.netrcHandler.GetFileName())
-
+		log.CliLogger.Warnf(errors.FoundNetrcCredMsg, netrcMachine.User, h.netrcHandler.GetFileName())
+		fmt.Println("getting from netrc:", netrcMachine.User)
```
brianstrauch, line 238:
```suggestion
```

internal/pkg/auth/auth.go
```diff
@@ -59,42 +61,73 @@ func PersistLogoutToConfig(config *v1.Config) error {
...
+		if err != nil {
+			return err
+		}
+
+		encryptedPassword, err := secret.Encrypt(credentials.Username, credentials.Password, salt, nonce)
+		if err != nil {
+			return err
+		}
+
+		loginCredential := &v1.LoginCredential{
```
brianstrauch, line 116: How is `v1.LoginCredential` different from `v1.Credential`? Can we combine them?

internal/pkg/auth/login_credentials_manager.go
```diff
@@ -150,7 +158,42 @@ func (h *LoginCredentialsManagerImpl) GetOnPremCredentialsFromEnvVar() func() (*
...
+
+		if loginCredential == nil {
+			return nil, nil
+		}
+
+		password, err := secret.Decrypt(loginCredential.Username, loginCredential.Password, loginCredential.Salt, loginCredential.Nonce)
+		if err != nil {
+			return nil, err
+		}
+		fmt.Println("returning config creds:", loginCredential.Username, password)
```
brianstrauch, line 185:
```suggestion
```

internal/pkg/secret/encryption_other.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return []byte{}, err
+	}
+	userId := strconv.Itoa(os.Getuid())
+	encryptionKey := pbkdf2.Key([]byte(machineId+userId), salt, 10000, 32, sha256.New)
```
brianstrauch, line 34: What do the numbers 10000 and 32 represent?

internal/cmd/logout/command.go
```diff
@@ -55,6 +57,12 @@ func (c *Command) logout(cmd *cobra.Command, _ []string) error {
 			// return err when other than NetrcCredentialsNotFoundErrorMsg or parsing error
 			return err
 		}
+
+		if runtime.GOOS == "darwin" && !c.cfg.IsTest {
```
brianstrauch, line 61: Why don't we delete the credentials for Linux or Windows?

internal/cmd/login/command.go
```diff
@@ -342,23 +357,21 @@ func (c *command) getURL(cmd *cobra.Command) (string, error) {
...
 	if save {
 		if credentials.IsSSO {
 			utils.ErrPrintln(cmd, "The `--save` flag was ignored since SSO credentials are not stored locally.")
 			return nil
 		}
 
-		if err := c.netrcHandler.WriteNetrcCredentials(isCloud, c.Config.Config.Context().NetrcMachineName, credentials.Username, credentials.Password); err != nil {
-			return err
+		ctxName := c.Config.Config.Context().NetrcMachineName
+		if runtime.GOOS == "darwin" && !c.cfg.IsTest {
```
brianstrauch, line 368: Why aren't we saving the credentials on Linux or Windows?

internal/pkg/auth/auth.go
```diff
@@ -59,42 +61,73 @@ func PersistLogoutToConfig(config *v1.Config) error {
...
+			IsCloud:    isCloud,
+			IgnoreCert: !isCloud,
```
brianstrauch, line 118: Is this value always equal to `!isCloud`?

internal/pkg/auth/login_credentials_manager.go
```diff
@@ -314,6 +357,37 @@ func (h *LoginCredentialsManagerImpl) GetOnPremPrerunCredentialsFromNetrc(cmd *c
...
 
+func (h *LoginCredentialsManagerImpl) GetCredentialsFromKeychain(cfg *v1.Config, ctxName, url string) func() (*Credentials, error) {
+	return func() (*Credentials, error) {
+		if runtime.GOOS == "darwin" {
+			username, password, err := keychain.Read(ctxName, url)
+			fmt.Println("returning creds in keychain")
+			return &Credentials{Username: username, Password: password}, err
+		}
+		return &Credentials{}, errors.New("keychain not available on platforms other than darwin")
+
```
brianstrauch, line 368:
```suggestion
```

internal/pkg/config/v1/login_credential.go
```diff
@@ -0,0 +1,11 @@
+package v1
+
+type LoginCredential struct {
+	IsCloud    bool   `json:"is_cloud"`
+	IgnoreCert bool   `json:"ignore_cert"`
+	Url        string `json:"url"`
+	Username   string `json:"username"`
+	Password   string `json:"password"`
```
brianstrauch, line 8:
```suggestion
	EncryptedPassword   string `json:"encrypted_password"`
```

test/login_test.go
```diff
@@ -213,80 +208,71 @@ func (s *CLITestSuite) TestSaveUsernamePassword() {
...
-
 		// run the login command with --save flag and check output
 		var env []string
 		if tt.isCloud {
 			env = []string{fmt.Sprintf("%s=good@user.com", auth.ConfluentCloudEmail), fmt.Sprintf("%s=pass1", auth.ConfluentCloudPassword)}
 		} else {
 			env = []string{fmt.Sprintf("%s=good@user.com", auth.ConfluentPlatformUsername), fmt.Sprintf("%s=pass1", auth.ConfluentPlatformPassword)}
 		}
+
+		configFile := os.Getenv("HOME") + "/.confluent/config.json"
```
brianstrauch, line 254: Use `filepath.Join()`

internal/cmd/logout/command.go
MuweiHe: This is only for keychain. creds saved in config will be deleted when `persistLogoutConfig` runs

internal/pkg/auth/auth.go
MuweiHe: yes

internal/pkg/secret/encryption_other.go
MuweiHe: 10000 is iteration, 32 is the key length

internal/pkg/auth/auth.go
MuweiHe: it's set to false for cloud login and true for confluent login

MuweiHe: so for saving login credential purposes, it's always `!isCloud`

internal/pkg/auth/auth.go
MuweiHe: I'm also thinking about this but not 100% sure. `v1.Credential` is being saved regardless of `--save` flag, and the naming is different than context name and netrc machine name.

MuweiHe: They actually serve different purposes so I think it's better not to combine them. `v1.credential` `represents an authentication mechanism for a Platform`, is always saved as a part of config and a field in context, and is also used to describe a context. Also when the password was prompted, it won't be saved into `v1.credential`, which contradicts what we're doing with `--save`

MuweiHe: `represents an authentication mechanism for a Platform` is the description in the code. 
that is possible but `credential` and `saved credentials` are doing different things and work differently under certain circumstances, I think it's better not to combine them which is confusing

brianstrauch: Having two similarly named fields that serve two different purposes seems equally confusing (especially since they have many sub-fields in common, like `username`, `nonce`, and `salt`!)

internal/cmd/logout/command.go
MuweiHe: that function is called `PersistLogoutToConfig` which handles the authenticated config, but keychain isn't really part of config.

internal/pkg/auth/auth.go
```diff
@@ -103,6 +106,35 @@ func addOrUpdateContext(config *v1.Config, ctxName, username, url string, state
...
 
+	if save && !credentials.IsSSO {
+		salt, err := secret.GenerateRandomBytes(24)
+		if err != nil {
+			return err
+		}
+		nonce, err := secret.GenerateRandomBytes(12)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
MuweiHe: those are the standard length of salt and nonce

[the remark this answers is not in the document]
MuweiHe: will use constants!

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -0,0 +1,92 @@
...
+
+func Write(ctxName, url, username, password string) error {
+	service := netrc.GetLocalCredentialName(isCloud, ctxName)
+
+	item := keychain.NewGenericPassword(service, url, username+"-"+url, []byte(username+sep+password), accessGroup)
+	item.SetSynchronizable(keychain.SynchronizableNo)
+	item.SetAccessible(keychain.AccessibleWhenUnlocked)
+
+	err := keychain.AddItem(item)
+	if err == keychain.ErrorDuplicateItem {
```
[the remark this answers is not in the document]
MuweiHe: that's a lot more lines of code rather than `keychain.NewGenericPassword(` in one line. Looking up makes sense too but i think `duplicate item` error is doing the same thing, letting us know that the entry is already there

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/log"
+	"github.com/confluentinc/cli/internal/pkg/netrc"
+	"github.com/keybase/go-keychain"
+)
+
+const (
+	accessGroup = "cli"
+	isCloud     = true
```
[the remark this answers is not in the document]
MuweiHe: yes we do! Fixing this.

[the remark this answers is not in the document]
MuweiHe: yes we do! fixing it

internal/cmd/logout/command.go
brianstrauch: Would it make sense to group both forms of persisted credentials into one common function, called `persistLogout()`?

internal/cmd/login/command.go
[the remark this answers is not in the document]
brianstrauch: We'd save a few lines of code if we checked for `nil` inside the getter function, and all other config file getters use the same approach!

internal/pkg/auth/auth.go
MuweiHe: `type Credential struct {
	Name           string         `json:"name"`
	Username       string         `json:"username"`
	Password       string         `json:"password"`
	APIKeyPair     *APIKeyPair    `json:"api_key_pair"`
	CredentialType CredentialType `json:"credential_type"`
}`

internal/cmd/login/command.go
```diff
@@ -167,7 +167,10 @@ func (c *command) loginCCloud(cmd *cobra.Command, url string) error {
 		c.printRemainingFreeCredit(cmd, client, currentOrg)
 	}
 
-	return c.saveLoginToKeychain(cmd, true, url, credentials, save)
+	if save && runtime.GOOS == "darwin" && !c.cfg.IsTest {
```
brianstrauch, line 170: Is there any way we can mock the keychain for tests, so we don't need this third condition?

internal/cmd/login/command.go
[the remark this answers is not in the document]
MuweiHe: `if ctx!=nil` is not checked only for getting login context, but for other purposes as well (see https://github.com/confluentinc/cli/blob/91bef96552dcda9f62954f668ce41c981fe5cdcb/internal/cmd/logout/command.go#L50). having a getter wouldn't allow us to remove this check outside. But I agree i'll add a check inside too to avoid possible panics.

internal/cmd/login/command.go
MuweiHe: unluckily i haven't found it

MuweiHe:
@ankurjoshi9 do you mind doing another pass of review and put stamp if concerns are addressed? Thank you!

internal/pkg/keychain/keychain_darwin.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	accessGroup = "cli"
+	separator   = "?"
+)
+
+func Write(isCloud bool, ctxName, url, username, password string) error {
+	service := netrc.GetLocalCredentialName(isCloud, ctxName)
+
+	item := keychain.NewGenericPassword(service, url, username+"-"+url, []byte(username+separator+password), accessGroup)
```
brianstrauch, line 22:
```suggestion
	item := keychain.NewGenericPassword(service, url, fmt.Sprintf("%s-%s", username, url), []byte(username+separator+password), accessGroup)
```

test/update_test.go
```diff
@@ -2,14 +2,15 @@ package test
...
+	"path/filepath"
 
 	"github.com/stretchr/testify/require"
 )
 
 func (s *CLITestSuite) TestUpdate() {
 	s.T().Skip("Skipping this test until its less flaky")
 
-	configFile := os.Getenv("HOME") + "/.confluent/config.json"
+	configFile := filepath.Join(os.Getenv("HOME"), "/.confluent/config.json")
```
brianstrauch, line 13: The point of using `filepath.Join()` is not to use slashes, so it works on Windows too.
```suggestion
	configFile := filepath.Join(os.Getenv("HOME"), ".confluent", "config.json")
```

internal/pkg/errors/strings.go
```diff
@@ -9,14 +9,15 @@ const (
...
-	RemoveNetrcCredentialsMsg  = "Removed credentials for user \"%s\" from netrc file \"%s\"\n"
-	StopNonInteractiveMsg      = "(remove these credentials or use the `--prompt` flag to bypass non-interactive login)"
-	FoundEnvCredMsg            = "Found credentials for user \"%s\" from environment variables \"%s\" and \"%s\" " +
+	LoggedInAsMsg                 = "Logged in as \"%s\".\n"
+	LoggedInAsMsgWithOrg          = "Logged in as \"%s\" for organization \"%s\" (\"%s\").\n"
+	LoggedInUsingEnvMsg           = "Using environment \"%s\" (\"%s\").\n"
+	LoggedOutMsg                  = "You are now logged out."
+	WroteCredentialsToKeychainMsg = "Wrote login credentials to keychain\n"
+	RemoveNetrcCredentialsMsg     = "Removed credentials for user \"%s\" from netrc file \"%s\"\n"
+	RemoveKeychainCredentialsMsg  = "Removed credentials for user \"%s\" from keychain access\n"
```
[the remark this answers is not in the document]
MuweiHe: keychain access is the application's name on mac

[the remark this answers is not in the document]
MuweiHe: i removed `access` to be consistent

ankurjoshi9 (approved):
LGTM

brianstrauch (approved):

PR #1688, opened 2023-02-02
Various example and flag text updates

sgagniere:
Checklist
---------
   * yes: ok

What
----
Updates to examples and flag strings.
- operations like "read" and "describe" for ACL commands are now lowercase in the example strings and the `--help` output. This is not a breaking change; only the text has been updated and the code continues to accept lower or upper case.
- Added new examples to `confluent iam rbac role-binding list` to highlight the `--inclusive` flag.
- Added more integration tests for help text.

References
----------
n/a

Test & Review
-------------
Ran all tests

Open Questions / Follow-ups
---------------------------
Should we update the description of the `confluent iam rbac role-binding list` command to explain which combinations of flags correspond to which scopes? For example, the `--environment` and `--cloud-cluster` flags together will display role bindings at the Cloud Cluster scope.

### DIFF
--- a/internal/pkg/acl/acl.go
+++ b/internal/pkg/acl/acl.go
@@ -187,9 +187,9 @@ func setAclRequestResourcePattern(conf *AclRequestDataWithError, n, v string) {
 }
 
 func convertToFlags(operations ...any) string {
-	var ops []string
+	ops := make([]string, len(operations))
 
-	for _, v := range operations {
+	for i, v := range operations {
 		// clean the resources that don't map directly to flag name
 		if v == cpkafkarestv3.ACLRESOURCETYPE_GROUP {
 			v = "consumer-group"
@@ -198,7 +198,7 @@ func convertToFlags(operations ...any) string {
 			v = "cluster-scope"
 		}
 		s := strings.ToLower(strings.ReplaceAll(fmt.Sprint(v), "_", "-"))
-		ops = append(ops, fmt.Sprintf("`--%s`", s))
+		ops[i] = fmt.Sprintf("`--%s`", s)
 	}
 
 	sort.Strings(ops)
@@ -208,13 +208,12 @@ func convertToFlags(operations ...any) string {
 func ConvertToLower(operations ...any) string {
 	ops := make([]string, len(operations))
 
-	for _, v := range operations {
-		s := strings.ToLower(strings.ReplaceAll(fmt.Sprint(v), "_", "-"))
-		ops = append(ops, s)
+	for i, v := range operations {
+		ops[i] = strings.ReplaceAll(fmt.Sprint(v), "_", "-")
 	}
 
 	sort.Strings(ops)
-	return strings.Join(ops, ", ")
+	return strings.ToLower(strings.Join(ops, ", "))
 }
 
 func ValidateCreateDeleteAclRequestData(aclConfiguration *AclRequestDataWithError) *AclRequestDataWithError {
--- a/internal/pkg/acl/acl.go
+++ b/internal/pkg/acl/acl.go
@@ -206,7 +206,7 @@ func convertToFlags(operations ...any) string {
 }
 
 func ConvertToLower(operations ...any) string {
-	var ops []string
+	ops := make([]string, len(operations))
 
 	for _, v := range operations {
 		s := strings.ToLower(strings.ReplaceAll(fmt.Sprint(v), "_", "-"))
--- a/internal/pkg/acl/acl.go
+++ b/internal/pkg/acl/acl.go
@@ -205,6 +205,18 @@ func convertToFlags(operations ...any) string {
 	return strings.Join(ops, ", ")
 }
 
+func ConvertToLower(operations ...any) string {
+	var ops []string
+
+	for _, v := range operations {
+		s := strings.ToLower(strings.ReplaceAll(fmt.Sprint(v), "_", "-"))
+		ops = append(ops, s)
+	}
+
+	sort.Strings(ops)
+	return strings.Join(ops, ", ")
+}
+
 func ValidateCreateDeleteAclRequestData(aclConfiguration *AclRequestDataWithError) *AclRequestDataWithError {
 	// delete is deliberately less powerful in the cli than in the API to prevent accidental
 	// deletion of too many acls at once. Expectation is that multi delete will be done via
--- a/internal/pkg/acl/acl.go
+++ b/internal/pkg/acl/acl.go
@@ -24,11 +24,19 @@ import (
 
 var listFields = []string{"Principal", "Permission", "Operation", "ResourceType", "ResourceName", "PatternType"}
 
-var AclOperations = []mds.AclOperation{mds.ACLOPERATION_ALL, mds.ACLOPERATION_ALTER, mds.ACLOPERATION_ALTER_CONFIGS,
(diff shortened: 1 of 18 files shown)


internal/pkg/acl/acl.go
```diff
@@ -205,6 +205,18 @@ func convertToFlags(operations ...any) string {
 	return strings.Join(ops, ", ")
 }
 
+func ConvertToLower(operations ...any) string {
+	var ops []string
```
brianstrauch, line 209: We already know the length of `ops`, so it's more efficient to allocate that space at the beginning.
```suggestion
	ops := make([]string, len(operations))
```

internal/pkg/acl/acl.go
```diff
@@ -205,6 +205,18 @@ func convertToFlags(operations ...any) string {
...
+
+	for _, v := range operations {
+		s := strings.ToLower(strings.ReplaceAll(fmt.Sprint(v), "_", "-"))
```
brianstrauch, line 212: You could move `strings.ToLower()` outside of the loop.

brianstrauch (approved):
Thank you!

internal/pkg/acl/acl.go
sgagniere: I moved it out of `ConvertToLower` but not the `convertToFlags` functions because it messes up the sorting (unless I change the hard-coded `consumer-group` and `cluster-scope` strings to upper case, which I can do if we think that's better).

internal/pkg/acl/acl.go
```diff
@@ -198,13 +198,24 @@ func convertToFlags(operations ...any) string {
...
 }
 
+func ConvertToLower(operations ...any) string {
+	ops := make([]string, len(operations))
+
+	for i, v := range operations {
+		ops[i] = strings.ReplaceAll(fmt.Sprint(v), "_", "-")
+	}
+
+	sort.Strings(ops)
```
brianstrauch, line 215: Why don't we sort the list before we pass it in? In both cases it's the same list every time.

internal/pkg/acl/acl.go
brianstrauch: How does it mess up the sorting? Do we want something other than alphabetical order?

sgagniere: Moving `ToLower` outside results in `cluster-scope` and `consumer-group` in lowercase because of the special handling and `TOPIC` and `TRANSACTIONAL-ID` for the other two. The sorting puts the capitalized entries first.

internal/pkg/acl/acl.go
sgagniere: Yeah I can do that.

brianstrauch: I'm saying we should manually sort the strings on line internal/pkg/acl/acl.go:82 so we don't have to call this sort function every time the command is run.

internal/pkg/acl/acl.go
brianstrauch: Not 100% sure I understand, but let's do whatever it takes to alphabetize these.

internal/pkg/acl/acl.go
sgagniere: Ohhh, yeah I can do that.

sgagniere: Done

internal/pkg/acl/acl.go
```diff
@@ -79,7 +79,7 @@ func PrintACLs(cmd *cobra.Command, acls []*ccstructs.ACLBinding) error {
 func AclFlags() *pflag.FlagSet {
 	flgSet := pflag.NewFlagSet("acl-config", pflag.ExitOnError)
 	flgSet.String("principal", "", "Principal for this operation with User: or Group: prefix.")
-	flgSet.String("operation", "", fmt.Sprintf("Set ACL Operation to: (%s).", ConvertToLower("ALL", "READ", "WRITE", "CREATE", "DELETE", "ALTER", "DESCRIBE", "CLUSTER_ACTION", "DESCRIBE_CONFIGS", "ALTER_CONFIGS", "IDEMPOTENT_WRITE")))
+	flgSet.String("operation", "", fmt.Sprintf("Set ACL Operation to: (%s).", ConvertToLower("ALL", "ALTER", "ALTER_CONFIGS", "CLUSTER_ACTION", "CREATE", "DELETE", "DESCRIBE", "DESCRIBE_CONFIGS", "IDEMPOTENT_WRITE", "READ", "WRITE")))
```
brianstrauch, line 82: Is it possible to generate this list from the list of constants?

sgagniere: You mean like how it's done with the mds constants for the `iam` version?

brianstrauch: Yeah, that's the same list right? We should only have to write it once.

sgagniere: Yeah it looks to be the same list. Only the list for cloud kafka is different; the lists for iam and on-prem kafka look identical.

I've updated it.

internal/pkg/acl/acl.go
```diff
@@ -205,7 +212,7 @@ func convertToFlags(operations ...any) string {
 	return strings.Join(ops, ", ")
 }
 
-func ConvertToLower(operations ...any) string {
+func ConvertToLower[T any](operations []T) string {
```
brianstrauch, line 215: Wouldn't this work?
```suggestion
func ConvertToLower(operations []any) string {
```

internal/pkg/acl/acl.go
```diff
@@ -23,6 +24,12 @@ import (
 
 var listFields = []string{"Principal", "Permission", "Operation", "ResourceType", "ResourceName", "PatternType"}
 
+var AclOperations = []mds.AclOperation{mds.ACLOPERATION_ALL, mds.ACLOPERATION_ALTER, mds.ACLOPERATION_ALTER_CONFIGS,
+	mds.ACLOPERATION_CLUSTER_ACTION, mds.ACLOPERATION_CREATE, mds.ACLOPERATION_DELETE,
+	mds.ACLOPERATION_DESCRIBE, mds.ACLOPERATION_DESCRIBE_CONFIGS,
+	mds.ACLOPERATION_IDEMPOTENT_WRITE, mds.ACLOPERATION_READ,
+	mds.ACLOPERATION_WRITE}
```
brianstrauch, line 31:
```suggestion
var AclOperations = []mds.AclOperation{
	mds.ACLOPERATION_ALL,
	mds.ACLOPERATION_ALTER,
	mds.ACLOPERATION_ALTER_CONFIGS,
	mds.ACLOPERATION_CLUSTER_ACTION,
	mds.ACLOPERATION_CREATE,
	mds.ACLOPERATION_DELETE,
	mds.ACLOPERATION_DESCRIBE,
	mds.ACLOPERATION_DESCRIBE_CONFIGS,
	mds.ACLOPERATION_IDEMPOTENT_WRITE,
	mds.ACLOPERATION_READ,
	mds.ACLOPERATION_WRITE,
}
```

internal/pkg/acl/acl.go
sgagniere: To do that I have to also change `AclOperations = []mds.AclOperation{...` to `AclOperations = []any{...` to get it compiling; this version just seems cleaner to me.

brianstrauch (approved):

PR #1691, opened 2023-02-03
[RCCA-10729] Resolve display error when creating role bindings with service accounts

sgagniere:
Checklist
---------
   * yes: ok

What
----
The `displayCCloudCreateAndDeleteOutput` function makes a call to `V2Client.GetIamUserById` for both user id's and service account id's. Before v3, whether or not the error from this call was `nil` was used to determine whether to format the output for service accounts or users. In v3 this error was returned immediately, hence the issue.

I've replaced the check for `err != nil` with a check on the format of the id and moved the `V2Client.GetIamUserById` call so that it should only be called with an actual user id.

References
----------
n/a

Test & Review
-------------
Manual testing using commands of the following forms:
`confluent iam rbac rb create --principal User:sa-12345 --resource Topic:test --role DeveloperRead --kafka-cluster lkc-123abc --environment env-456def --cloud-cluster lkc-123abc`
`confluent iam rbac rb create --principal User:u-67890 --resource Topic:test --role DeveloperRead --kafka-cluster lkc-123abc --environment env-456def --cloud-cluster lkc-123abc`
I also compared with output from v2.38.0.

Added a new integration test with a service account principal, and ran all tests.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/iam/command_rbac_role_binding.go
+++ b/internal/cmd/iam/command_rbac_role_binding.go
@@ -395,11 +395,11 @@ func (c *roleBindingCommand) displayCCloudCreateAndDeleteOutput(cmd *cobra.Comma
 		if resource != "" {
 			fields = ccloudResourcePatternListFields
 		} else {
-			if user, err := c.V2Client.GetIamUserById(userResourceId); err != nil {
+			user, err := c.V2Client.GetIamUserById(userResourceId)
+			if err != nil {
 				return err
-			} else {
-				out.Email = user.GetEmail()
 			}
+			out.Email = user.GetEmail()
 			fields = []string{"Principal", "Email", "Role"}
 		}
 	}
--- a/internal/cmd/iam/command_rbac_role_binding.go
+++ b/internal/cmd/iam/command_rbac_role_binding.go
@@ -15,7 +15,7 @@ import (
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/output"
-	"github.com/confluentinc/cli/internal/pkg/resource"
+	presource "github.com/confluentinc/cli/internal/pkg/resource"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


internal/cmd/iam/command_rbac_role_binding.go
```diff
@@ -363,12 +361,12 @@ func (c *roleBindingCommand) displayCCloudCreateAndDeleteOutput(cmd *cobra.Comma
 	// The err is ignored here since the --prefix flag is not defined by the list subcommand
 	prefix, _ := cmd.Flags().GetBool("prefix")
 
-	resource, err := cmd.Flags().GetString("resource")
+	res, err := cmd.Flags().GetString("resource")
```
brianstrauch, line 364:
```suggestion
	resource, err := cmd.Flags().GetString("resource")
```

internal/cmd/iam/command_rbac_role_binding.go
```diff
@@ -15,6 +15,7 @@ import (
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/resource"
```
brianstrauch, line 18: When there's a naming conflict, we prefix internal packages with "p". I wish there was a better solution...
```suggestion
	presource "github.com/confluentinc/cli/internal/pkg/resource"
```

internal/cmd/iam/command_rbac_role_binding.go
sgagniere: If I do that I need to import `internal/pkg/resource` with a different name, since that package is used in the same function that this change is in.

internal/cmd/iam/command_rbac_role_binding.go
```diff
@@ -397,7 +395,11 @@ func (c *roleBindingCommand) displayCCloudCreateAndDeleteOutput(cmd *cobra.Comma
 		if resource != "" {
 			fields = ccloudResourcePatternListFields
 		} else {
-			out.Email = user.GetEmail()
+			if user, err := c.V2Client.GetIamUserById(userResourceId); err != nil {
+				return err
+			} else {
+				out.Email = user.GetEmail()
+			}
```
brianstrauch, line 402: The first branch always exits, no `else` needed.
```suggestion
			user, err := c.V2Client.GetIamUserById(userResourceId)
			if err != nil {
				return err
			}
			out.Email = user.GetEmail()
```

brianstrauch (approved):

PR #1692, opened 2023-02-03
Add sleep statement to update db script

sgagniere:
Checklist
---------
   * yes: ok

What
----
cctunnel returns control to the user before it's actually done. Adding a sleep statement before the next command in this script to try and prevent the connection refused errors I keep getting when I do the releases.

3 seconds is probably longer than necessary, but seems alright for a first try.


brianstrauch:
You should be able to make this modification and run `make update-db` without merging to main!

sgagniere:
@brianstrauch Good point! Closing for now.

brianstrauch (approved):
Hoping they replace `cctunnel` with a better tool in the future, but for now...

PR #1703, opened 2023-02-10
Fix goreleaser deprecation warnings

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Noticed a few of these during the last release and staying on top of them!
```
DEPRECATED: `archives.replacements` should not be used anymore, check https://goreleaser.com/deprecations#archivesreplacements for more info
DEPRECATED: `archives.rlcp` will be the default soon, check https://goreleaser.com/deprecations#archivesrlcp for more info
```

### DIFF
--- a/.goreleaser-build.yml
+++ b/.goreleaser-build.yml
@@ -1,7 +1,5 @@
 project_name: confluent
 
-dist: dist
-
 builds:
   - binary: confluent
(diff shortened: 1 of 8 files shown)


DABH (approved):
Awesome, thank you! ✨

PR #1710, opened 2023-02-13
Do not set `name` field in cluster update when it's not specified

MuweiHe:
Checklist
---------
   * yes: ok

What
----

According to docs https://docs.confluent.io/cloud/current/clusters/cluster-api.html#id3 in one request we should only specify the name or the config (cku), although it's possible to update both at the same time. But when the user provides only cku, the display name is set to the new update name field, which is not what the user wants as the name could be different. 
To avoid breaking changes, we still allow updating both name and cku in one line, but in our next major version we might consider only allowing name or cku, to make the command more atomic and match the behavior of other update commands like schema-registry `mode` / `compatibility`

References
----------

https://confluentinc.atlassian.net/browse/CAPAC-1229

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/kafka/command_cluster_update.go
+++ b/internal/cmd/kafka/command_cluster_update.go
@@ -85,7 +85,6 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
 			return err
 		}
 		update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
-
 	}
 
 	updatedCluster, err := c.V2Client.UpdateKafkaCluster(clusterID, update)
@@ -106,7 +105,7 @@ func (c *clusterCommand) validateResize(cmd *cobra.Command, cku int32, currentCl
 		return 0, errors.New(errors.ClusterResizeNotSupportedErrorMsg)
 	}
 	// Durability Checks
-	if *currentCluster.GetSpec().Availability == highAvailability && cku <= 1 {
+	if currentCluster.Spec.GetAvailability() == highAvailability && cku <= 1 {
 		return 0, errors.New(errors.CKUMoreThanOneErrorMsg)
 	}
 	if cku == 0 {
--- a/internal/cmd/kafka/command_cluster_update.go
+++ b/internal/cmd/kafka/command_cluster_update.go
@@ -34,7 +34,7 @@ func (c *clusterCommand) newUpdateCommand(cfg *v1.Config) *cobra.Command {
 	}
 
 	cmd.Flags().String("name", "", "Name of the Kafka cluster.")
-	cmd.Flags().Int("cku", 0, "Number of Confluent Kafka Units (non-negative). For Kafka clusters of type 'dedicated' only. When shrinking a cluster, you can reduce capacity one CKU at a time.")
+	cmd.Flags().Int32("cku", 0, "Number of Confluent Kafka Units (non-negative). For Kafka clusters of type 'dedicated' only. When shrinking a cluster, you can reduce capacity one CKU at a time.")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 	if cfg.IsCloudLogin() {
 		pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
@@ -50,6 +50,11 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
 	}
 
 	clusterID := args[0]
+	currentCluster, _, err := c.V2Client.DescribeKafkaCluster(clusterID, c.EnvironmentId())
+	if err != nil {
+		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.KafkaClusterNotFoundErrorMsg, clusterID), errors.ChooseRightEnvironmentSuggestions)
+	}
+
 	update := cmkv2.CmkV2ClusterUpdate{
 		Id: cmkv2.PtrString(clusterID),
 		Spec: &cmkv2.CmkV2ClusterSpecUpdate{
@@ -57,10 +62,6 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
 				Id: c.EnvironmentId(),
 			},
 		},
-	}
-	currentCluster, _, err := c.V2Client.DescribeKafkaCluster(clusterID, c.EnvironmentId())
-	if err != nil {
-		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.KafkaClusterNotFoundErrorMsg, clusterID), errors.ChooseRightEnvironmentSuggestions)
 	}
 
 	if cmd.Flags().Changed("name") {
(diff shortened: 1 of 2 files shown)


internal/cmd/kafka/command_cluster_update.go
```diff
@@ -34,7 +34,7 @@ func (c *clusterCommand) newUpdateCommand(cfg *v1.Config) *cobra.Command {
 	}
 
 	cmd.Flags().String("name", "", "Name of the Kafka cluster.")
-	cmd.Flags().Int("cku", 0, "Number of Confluent Kafka Units (non-negative). For Kafka clusters of type 'dedicated' only. When shrinking a cluster, you can reduce capacity one CKU at a time.")
+	cmd.Flags().Int32("cku", 0, "Number of Confluent Kafka Units (non-negative). For Kafka clusters of type 'dedicated' only. When shrinking a cluster, you can reduce capacity one CKU at a time.")
```
brianstrauch, line 37:
```suggestion
	cmd.Flags().Uint32("cku", 0, `Number of Confluent Kafka Units. For Kafka clusters of type "dedicated" only. When shrinking a cluster, you must reduce capacity one CKU at a time.`)
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -96,52 +101,44 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+func (c *clusterCommand) validateResize(cmd *cobra.Command, cku int32, currentCluster *cmkv2.CmkV2Cluster, prompt form.Prompt) (int32, error) {
+	// Ensure the cluster is a Dedicated Cluster
+	if currentCluster.GetSpec().Config.CmkV2Dedicated == nil {
+		return -1, errors.New(errors.ClusterResizeNotSupportedErrorMsg)
+	}
+	// Durability Checks
+	if *currentCluster.GetSpec().Availability == highAvailability && cku <= 1 {
+		return -1, errors.New(errors.CKUMoreThanOneErrorMsg)
+	}
+	if cku <= 0 {
```
brianstrauch, line 113: If we make this a `uint`, we can simplify this
```suggestion
	if cku == 0 {
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -96,52 +101,44 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
-		// Durability Checks
-		if *currentCluster.GetSpec().Availability == highAvailability && cku <= 1 {
-			return -1, errors.New(errors.CKUMoreThanOneErrorMsg)
-		}
-		if cku <= 0 {
-			return -1, errors.New(errors.CKUMoreThanZeroErrorMsg)
+func (c *clusterCommand) validateResize(cmd *cobra.Command, cku int32, currentCluster *cmkv2.CmkV2Cluster, prompt form.Prompt) (int32, error) {
+	// Ensure the cluster is a Dedicated Cluster
+	if currentCluster.GetSpec().Config.CmkV2Dedicated == nil {
+		return -1, errors.New(errors.ClusterResizeNotSupportedErrorMsg)
```
brianstrauch, line 107: We no longer do `return -1, nil`, so we can just return the default value (and also remove the check for `if updatedCku != -1`)
```suggestion
		return 0, errors.New(errors.ClusterResizeNotSupportedErrorMsg)
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -96,52 +101,44 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
-				} else {
-					return int32(cku), nil
-				}
+		if promptMessage != "" {
+			ok, err := confirmShrink(cmd, prompt, promptMessage)
+			if !ok || err != nil {
+				return -1, err
+			} else {
+				return cku, nil
 			}
```
brianstrauch, line 138:
```suggestion
			if ok, err := confirmShrink(cmd, prompt, promptMessage); err != nil || !ok {
				return 0, err
			}
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -72,16 +73,20 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
 			return errors.New(errors.NonEmptyNameErrorMsg)
```
brianstrauch, line 73: We should be able to get rid of this check. I gave this a try and the backend returned a helpful error message.

internal/cmd/kafka/command_cluster_update.go
MuweiHe: Made the changes! the only thing to note is that `cku` is int32 in the cmk struct. I still need to convert it to int32 some place.

internal/cmd/kafka/command_cluster_update.go
MuweiHe: what's the msg you're seeing? I tried setting name to "" and it doesn't error, it just don't update.

brianstrauch:
```
% confluent kafka cluster update lkc-1j9gqv --name ""
Error: failed to update Kafka cluster: Either name or config must be set for update request.: 422 Unprocessable Entity

Suggestions:
    A cluster can't be updated while still provisioning. If you just created this cluster, retry in a few minutes.
```

MuweiHe: that's because you didn't set cku update. If you set a cku update and name to "" at the same time, it'll just update the cku and the name won't be changed.

brianstrauch: ah that's a backend bug then

brianstrauch: Ah, looks like we should keep this error message then.

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -72,16 +73,20 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+		if err != nil {
+			return err
+		}
+		updatedCku, err := c.validateResize(cmd, int32(cku), &currentCluster, prompt)
+		if err != nil {
+			return err
+		}
+		if updatedCku != -1 {
+			update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
+		}
```
brianstrauch, line 89: No longer needed. The function will never return -1 without also returning an error.
```suggestion
		update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -96,52 +101,41 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+	if cku == 0 {
+		return 0, errors.New(errors.CKUMoreThanZeroErrorMsg)
+	}
+	// Cluster can't be resized while it's provisioning or being expanded already.
+	// Name _can_ be changed during these times, though.
+	err := isClusterResizeInProgress(currentCluster)
+	if err != nil {
+		return 0, err
+	}
+	//If shrink
```
brianstrauch, line 122:
```suggestion
	// If shrink
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -72,16 +73,19 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+		cku, err := cmd.Flags().GetUint32("cku")
+		if err != nil {
+			return err
+		}
+		updatedCku, err := c.validateResize(cmd, int32(cku), &currentCluster, prompt)
+		if err != nil {
+			return err
+		}
 		update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
+
```
brianstrauch, line 88:
```suggestion
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -96,52 +100,41 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
-		}
-		if cku <= 0 {
-			return -1, errors.New(errors.CKUMoreThanZeroErrorMsg)
+func (c *clusterCommand) validateResize(cmd *cobra.Command, cku int32, currentCluster *cmkv2.CmkV2Cluster, prompt form.Prompt) (int32, error) {
+	// Ensure the cluster is a Dedicated Cluster
+	if currentCluster.GetSpec().Config.CmkV2Dedicated == nil {
+		return 0, errors.New(errors.ClusterResizeNotSupportedErrorMsg)
+	}
+	// Durability Checks
+	if *currentCluster.GetSpec().Availability == highAvailability && cku <= 1 {
```
brianstrauch, line 109:
```suggestion
	if currentCluster.Spec.GetAvailability() == highAvailability && cku <= 1 {
```

brianstrauch (approved):

PR #1713, opened 2023-02-14
Replace CLI cloud-signup with a browser redirect

sgagniere:
Checklist
---------
   * yes: ok

What
----
Signup through the CLI will be disabled for the time being. In place of that, the command will now open a browser tab to the cloud signup page (after the user confirms that they wish to do so).

Test & Review
-------------
Manual testing (macOS terminal, Windows command prompt, Windows powershell).


test/cloud_signup_test.go
```diff
@@ -1,79 +0,0 @@
-package test
```
sgagniere, line 1: I don't want a browser window popping up every time the unit tests run, and other than that there's basically nothing here that really seemed worth testing so I just removed this file completely.

internal/cmd/cloud-signup/command.go
```diff
@@ -40,174 +31,22 @@ func New(prerunner pcmd.PreRunner, userAgent string, ccloudClientFactory pauth.C
...
-
-	cmd.Flags().String("url", "https://confluent.cloud", "Confluent Cloud service URL.")
+	cmd.RunE = c.signup
 
 	return cmd
 }
 
-func (c *command) cloudSignupRunE(cmd *cobra.Command, _ []string) error {
-	url, err := cmd.Flags().GetString("url")
+func (c *command) signup(cmd *cobra.Command, _ []string) error {
```
brianstrauch, line 39: Let's match the name of the command?
```suggestion
func (c *command) cloudSignup(cmd *cobra.Command, _ []string) error {
```

brianstrauch: Also, the `command` struct probably isn't needed anymore. We don't need any of the `userAgent`, `isTest`, etc. fields

internal/cmd/cloud-signup/command.go
```diff
@@ -40,174 +31,22 @@ func New(prerunner pcmd.PreRunner, userAgent string, ccloudClientFactory pauth.C
...
+	utils.Println(cmd, "You will now be redirected to the Confluent Cloud sign up page in your browser.")
+	err := form.ConfirmEnter(cmd)
 	if err != nil {
```
brianstrauch, line 42:
```suggestion
	if err := form.ConfirmEnter(cmd); err != nil {
```

internal/pkg/form/form.go
```diff
@@ -112,6 +112,20 @@ func ConfirmDeletion(cmd *cobra.Command, promptMsg, stringToType string) (bool,
 	return false, errors.NewErrorWithSuggestions(fmt.Sprintf(`input does not match "%s"`, stringToType), DeleteResourceConfirmSuggestions)
 }
 
+func ConfirmEnter(cmd *cobra.Command) (error) {
+	// Note: this function does not check against any regex;
+	// any string is acceptable
+	prompt := NewPrompt(os.Stdin)
+	f := Field{Prompt: "Press enter to continue or Ctrl-C to cancel"}
```
brianstrauch, line 119: Can we hide the output from the user? If they type "no" for example they shouldn't see any letters appear.

brianstrauch:
Make sure you edit the feature flag to add an announcement to the `cloud-signup` command that explains that it no longer works for versions < v3.2.0!

internal/cmd/cloud-signup/command.go
sgagniere: I just did that so I wouldn't have to use `:=` again on line 46. I can't change both though.

internal/pkg/form/form.go
sgagniere: I explored this for a bit. Using `golang.org/x/term`'s ReadPassword function to disable terminal echo does this. But... it has an unexpected interaction with `browser.OpenURL` that causes the tab to open in the background. Going to make a ticket for this and address it afterwards, but I want to get this in for the next release.

brianstrauch: Huh, that's super weird.

sgagniere: Added it back. Printing anything gets rid of this behavior.

internal/pkg/form/form.go
```diff
@@ -112,16 +113,15 @@ func ConfirmDeletion(cmd *cobra.Command, promptMsg, stringToType string) (bool,
...
+func ConfirmEnter(cmd *cobra.Command) error {
 	// Note: this function does not check against any regex;
 	// any string is acceptable
-	prompt := NewPrompt(os.Stdin)
-	f := Field{Prompt: "Press enter to continue or Ctrl-C to cancel"}
-	show(cmd, f)
+	utils.Print(cmd, "Press enter to continue or Ctrl-C to cancel:")
 
-	if _, err := read(f, prompt); err != nil {
+	if _, err := term.ReadPassword(int(os.Stdin.Fd())); err != nil {
```
brianstrauch, line 121: You might find it easier to use the `form` package and set `IsHidden = true`. (The current approach is a bit confusing since we're not reading a password)

sgagniere: I considered that, but using form with `IsHidden` still prints *'s instead of surpressing output

brianstrauch: Ah, right. Let's write a comment explaining why we're using this function then!

brianstrauch (approved):

internal/pkg/form/form.go
```diff
@@ -112,6 +113,19 @@ func ConfirmDeletion(cmd *cobra.Command, promptMsg, stringToType string) (bool,
 	return false, errors.NewErrorWithSuggestions(fmt.Sprintf(`input does not match "%s"`, stringToType), DeleteResourceConfirmSuggestions)
 }
 
+func ConfirmEnter(cmd *cobra.Command) error {
+	// This function prevents echoing of user input instead of displaying text or *'s
+	// so that the CLI will appear to wait until 'enter' or 'Ctrl-C' are entered.
+	utils.Print(cmd, "Press enter to continue or Ctrl-C to cancel:")
```
brianstrauch, line 119: Would this prevent us from needing the second `Print()` below?
```suggestion
	utils.Println(cmd, "Press enter to continue or Ctrl-C to cancel:")
```

internal/cmd/cloud-signup/command.go
```diff
@@ -1,213 +1,36 @@
...
-	c := &command{
-		CLICommand:    pcmd.NewAnonymousCLICommand(cmd, prerunner),
-		userAgent:     userAgent,
-		clientFactory: ccloudClientFactory,
-		isTest:        isTest,
-	}
-	cmd.RunE = c.cloudSignupRunE
-
-	cmd.Flags().String("url", "https://confluent.cloud", "Confluent Cloud service URL.")
+	cmd.RunE = cloudSignup
```
brianstrauch, line 18: This can be added to the struct above now

internal/pkg/form/form.go
sgagniere: No; the second print is there for an unrelated reason. I just added a comment to explain it.

internal/cmd/cloud-signup/command.go
```diff
@@ -1,213 +1,36 @@
...
 
 	return cmd
 }
 
-func (c *command) cloudSignupRunE(cmd *cobra.Command, _ []string) error {
-	url, err := cmd.Flags().GetString("url")
+func cloudSignup(cmd *cobra.Command, _ []string) error {
+	utils.Println(cmd, "You will now be redirected to the Confluent Cloud sign up page in your browser.")
```
brianstrauch, line 24: How about adding the link to this text, "If you are not redirected, use the following link: https://..."

brianstrauch (approved):

PR #1714, opened 2023-02-14
Remove `make deps`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
`go install` is essentially a no-op when a package is already installed (it's much more performant now than it was in certain older versions of Go). We can move the `go install` commands closer to where they're needed to simplify the setup process (and make our tests that only use some of the dependencies faster)

Test & Review
-------------
Manually executed targets to make sure they still work as intended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
-SHELL           := /bin/bash
-ALL_SRC         := $(shell find . -name "*.go" | grep -v -e vendor)
+SHELL              := /bin/bash
+ALL_SRC            := $(shell find . -name "*.go" | grep -v -e vendor)
+GORELEASER_VERSION := v1.15.2
+
 GIT_REMOTE_NAME ?= origin
 MAIN_BRANCH     ?= main
 RELEASE_BRANCH  ?= main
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -156,6 +150,7 @@ endif
 .PHONY: int-test
 int-test:
 ifdef CI
+	go install gotest.tools/gotestsum@v1.8.2 && \
```
MuweiHe, line 153: have you monitored the time difference with running `go install`?

brianstrauch: Seems to consistently take < 1s... I'm ok with this, since people have had problems forgetting to use `make deps` in the past.
```
% time go install github.com/goreleaser/goreleaser@v1.15.2
go install github.com/goreleaser/goreleaser@v1.15.2  0.89s user 0.61s system 248% cpu 0.603 total
% time go install github.com/goreleaser/goreleaser@v1.15.2
go install github.com/goreleaser/goreleaser@v1.15.2  0.89s user 0.61s system 265% cpu 0.562 total
% time go install github.com/goreleaser/goreleaser@v1.15.2
go install github.com/goreleaser/goreleaser@v1.15.2  0.89s user 0.63s system 270% cpu 0.565 total
```

MuweiHe (approved):

PR #1715, opened 2023-02-15
[CLI-2332] Add `--availability` and `--type` to `kafka cluster update`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
* Update a dedicated cluster to multi-zone with `kafka update --availability multi-zone`
* Update a basic/standard cluster to multi-zone with `kafka update --availability multi-zone --type dedicated`

References
----------
https://confluentinc.atlassian.net/wiki/spaces/KPLATFORM/pages/2928020393/Engineering+One+Pager+Standard+Cluster+Single-Zone+to+Multi-Zone+Migration#CLI-Workflow

Test & Review
-------------
Untested, since backend is still a WIP. Planning to write integration tests and system tests.


internal/cmd/kafka/command_cluster_create.go
```diff
@@ -88,7 +88,7 @@ func (c *clusterCommand) newCreateCommand(cfg *v1.Config) *cobra.Command {
 	pcmd.AddRegionFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddAvailabilityFlag(cmd)
 	pcmd.AddTypeFlag(cmd)
-	cmd.Flags().Int("cku", 0, `Number of Confluent Kafka Units (non-negative). Required for Kafka clusters of type "dedicated".`)
+	cmd.Flags().Uint32("cku", 0, `Number of Confluent Kafka Units. Required for Kafka clusters of type "dedicated".`)
```
brianstrauch, line 91: Matching @MuweiHe's changes from #1710

internal/pkg/cmd/flags.go
```diff
@@ -77,7 +77,7 @@ func autocompleteMechanisms(protocol string) []string {
 
 func AddAvailabilityFlag(cmd *cobra.Command) {
 	cmd.Flags().String("availability", kafka.Availabilities[0], fmt.Sprintf("Specify the availability of the cluster as %s.", utils.ArrayToCommaDelimitedString(kafka.Availabilities)))
-	RegisterFlagCompletionFunc(cmd, output.FlagName, func(_ *cobra.Command, _ []string) []string { return kafka.Availabilities })
+	RegisterFlagCompletionFunc(cmd, "availability", func(_ *cobra.Command, _ []string) []string { return kafka.Availabilities })
```
brianstrauch, line 80: Oops. Completion for this flag was broken before.

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -34,6 +36,8 @@ func (c *clusterCommand) newUpdateCommand(cfg *v1.Config) *cobra.Command {
 	}
 
 	cmd.Flags().String("name", "", "Name of the Kafka cluster.")
+	pcmd.AddAvailabilityFlag(cmd)
+	pcmd.AddTypeFlag(cmd)
```
tmodi11, line 40: What is the difference between lines 38/41 and 39/40? It seems like we're missing a description or information in the case the user requests help.

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -45,18 +49,22 @@ func (c *clusterCommand) newUpdateCommand(cfg *v1.Config) *cobra.Command {
...
-	currentCluster, _, err := c.V2Client.DescribeKafkaCluster(clusterID, c.EnvironmentId())
+	clusterId := args[0]
+	currentCluster, _, err := c.V2Client.DescribeKafkaCluster(clusterId, c.EnvironmentId())
 	if err != nil {
-		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.KafkaClusterNotFoundErrorMsg, clusterID), errors.ChooseRightEnvironmentSuggestions)
+		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.KafkaClusterNotFoundErrorMsg, clusterId), errors.ChooseRightEnvironmentSuggestions)
+	}
+
+	if currentCluster.Status.GetPhase() == ccloudv2.StatusProvisioning {
+		return errors.New(errors.KafkaClusterProvisioningErrorMsg)
```
tmodi11, line 63: Just for my understanding, is this just ensuring that if the cluster is still provisioning, we don't make any updates?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
 			return err
 		}
 		if name == "" {
-			return errors.New(errors.NonEmptyNameErrorMsg)
+			return errors.New("`--name` must not be empty")
```
tmodi11, line 77: nit: Is there a reason you're removing the constant here?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
 		}
+
 		update.Spec.SetDisplayName(name)
+
+		// The backend blocks simultaneous modification of `--name` and `--type`. If both are passed, update the name separately.
+		if cmd.Flags().Changed("type") {
+			if _, err := c.V2Client.UpdateKafkaCluster(clusterId, update); err != nil {
+				return err
+			}
+			update.Spec.DisplayName = nil
```
tmodi11, line 87: Is there a reason we cannot use the setter here?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+	if cmd.Flags().Changed("availability") {
+		availability, err := cmd.Flags().GetString("availability")
 		if err != nil {
 			return err
 		}
-		updatedCku, err := c.validateResize(int32(cku), &currentCluster, prompt)
-		if err != nil {
-			return err
+		if availability == "" {
+			return errors.New("`--availability` must not be empty")
```
tmodi11, line 97: nit: can this error message be a constant?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
 		}
-		update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
+		update.Spec.SetAvailability(strings.ToUpper(strings.ReplaceAll(availability, "-", "_")))
 	}
 
-	updatedCluster, err := c.V2Client.UpdateKafkaCluster(clusterID, update)
+	// Both `--type` and `--cku` require changing the config
+	if cmd.Flags().Changed("type") || cmd.Flags().Changed("cku") {
```
tmodi11, line 103: Can we consolidate the logic to make it more readable? I was thinking something like:
```
if (nameChange and clusterTypeChange) {
  setName = new_name
  updateName
  setName = nil
  setClusterType
}

if (ckuChange) {
  updateCKU 
}
```

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+				return errors.New("`--type` must not be empty")
+			}
+			switch clusterType {
+			case skuBasic:
+				update.Spec.Config.CmkV2Basic.SetKind("Basic")
+			case skuStandard:
+				update.Spec.Config.CmkV2Standard.SetKind("Standard")
+			case skuDedicated:
+				update.Spec.Config.CmkV2Dedicated.SetKind("Dedicated")
+			}
```
tmodi11, line 139: I see that this is a duplicate set of switch case statements. Can this be consolidated with the above (L114)?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
 		}
-		update.Spec.Config = &cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: updatedCku}}
+		update.Spec.SetAvailability(strings.ToUpper(strings.ReplaceAll(availability, "-", "_")))
 	}
 
-	updatedCluster, err := c.V2Client.UpdateKafkaCluster(clusterID, update)
+	// Both `--type` and `--cku` require changing the config
+	if cmd.Flags().Changed("type") || cmd.Flags().Changed("cku") {
+		// Set base config before setting type or CKU
+		updatedClusterType := getClusterType(currentCluster.Spec.GetConfig())
```
tmodi11, line 105: nit: can this be `currentClusterType`

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+			update.Spec.SetConfig(cmkv2.CmkV2ClusterSpecUpdateConfigOneOf{CmkV2Dedicated: new(cmkv2.CmkV2Dedicated)})
+		default:
+			return fmt.Errorf(`unsupported cluster type "%s"`, updatedClusterType)
+		}
+
+		if cmd.Flags().Changed("type") {
+			clusterType, err := cmd.Flags().GetString("type")
+			if err != nil {
+				return err
+			}
```
tmodi11, line 128: This block seems redundant to L106-112; is this necessary?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+			if clusterType == "" {
+				return errors.New("`--type` must not be empty")
```
tmodi11, line 130: nit: can this error be a constant?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+		update.Spec.SetAvailability(strings.ToUpper(strings.ReplaceAll(availability, "-", "_")))
 	}
 
-	updatedCluster, err := c.V2Client.UpdateKafkaCluster(clusterID, update)
+	// Both `--type` and `--cku` require changing the config
+	if cmd.Flags().Changed("type") || cmd.Flags().Changed("cku") {
+		// Set base config before setting type or CKU
+		updatedClusterType := getClusterType(currentCluster.Spec.GetConfig())
+		if cmd.Flags().Changed("type") {
+			clusterType, err := cmd.Flags().GetString("type")
```
tmodi11, line 107: nit: can this be `updatedClusterType`

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+			if err != nil {
+				return err
+			}
+			if clusterType == "" {
+				return errors.New("`--type` must not be empty")
+			}
+			switch clusterType {
+			case skuBasic:
+				update.Spec.Config.CmkV2Basic.SetKind("Basic")
```
tmodi11, line 134: This is out of the scope of this PR, but can we have a follow up JIRA to change the `kind` attributes to constants?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+			switch updatedClusterType {
+			case skuBasic, skuStandard:
+				return fmt.Errorf(errors.ClusterResizeNotSupportedErrorMsg)
+			case skuDedicated:
+				updatedCku, err := c.validateResize(int32(cku), &currentCluster, prompt)
+				if err != nil {
+					return err
+				}
+				update.Spec.Config.CmkV2Dedicated.SetCku(updatedCku)
+			}
```
tmodi11, line 156: Can this be consolidated with the above switch statement on L113?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -66,31 +74,97 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string, prompt form.P
...
+		}
+	}
+
+	updatedCluster, err := c.V2Client.UpdateKafkaCluster(clusterId, update)
 	if err != nil {
-		return errors.NewWrapErrorWithSuggestions(err, "failed to update Kafka cluster", errors.KafkaClusterUpdateFailedSuggestions)
+		return err
```
tmodi11, line 162: Is there a reason why we are no longer returning a suggestion?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -104,18 +178,17 @@ func (c *clusterCommand) validateResize(cku int32, currentCluster *cmkv2.CmkV2Cl
 	if currentCluster.Spec.GetAvailability() == highAvailability && cku <= 1 {
 		return 0, errors.New(errors.CKUMoreThanOneErrorMsg)
 	}
+
 	if cku == 0 {
 		return 0, errors.New(errors.CKUMoreThanZeroErrorMsg)
 	}
-	// Cluster can't be resized while it's provisioning or being expanded already.
```
tmodi11, line 110: Is there a reason why we're removing these comments? Seems like it could be helpful.

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -104,18 +178,17 @@ func (c *clusterCommand) validateResize(cku int32, currentCluster *cmkv2.CmkV2Cl
...
-	// Name _can_ be changed during these times, though.
 	if err := isClusterResizeInProgress(currentCluster); err != nil {
 		return 0, err
 	}
+
 	// If shrink
-	if cku < currentCluster.GetSpec().Config.CmkV2Dedicated.Cku {
+	if cku < currentCluster.Spec.GetConfig().CmkV2Dedicated.GetCku() {
```
tmodi11, line 190: Is there a reason why we removed the getter from spec here?

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -155,3 +229,15 @@ func confirmShrink(prompt form.Prompt, promptMessage string) (bool, error) {
...
+	if config.CmkV2Dedicated != nil {
+		return skuDedicated
+	}
+
+	if config.CmkV2Standard != nil {
+		return skuStandard
+	}
+
+	return skuBasic
+}
```
tmodi11, line 243: It seems easier if we have all the switch statement functionality here and call this function anytime we need to get the cluster type. You would just need to add another case if the user has inputted an invalid type.

internal/cmd/kafka/utils.go
```diff
@@ -107,9 +106,6 @@ func getKafkaRestProxyAndLkcId(c *pcmd.AuthenticatedStateFlagCommand) (*pcmd.Kaf
 }
 
 func isClusterResizeInProgress(currentCluster *cmkv2.CmkV2Cluster) error {
-	if currentCluster.Status.Phase == ccloudv2.StatusProvisioning {
-		return errors.New(errors.KafkaClusterStillProvisioningErrorMsg)
-	}
```
tmodi11, line 112: Is it safe to remove this check; are we sure no other function is relying on this check?

tmodi11 (commented):
@brianstrauch Thanks for all your work on this! Did a first pass and left some comments.

internal/cmd/kafka/command_cluster_update.go
brianstrauch: The suggestion is "A cluster can't be updated while still provisioning." but we return an error earlier if that's the case, so this would just lead users down the wrong path.

internal/cmd/kafka/command_cluster_update.go
brianstrauch: Those comments should've been cleaned up earlier. This function no longer checks if the cluster is provisioning or expanding already.

internal/cmd/kafka/command_cluster_update.go
brianstrauch: It's a weird limitation of our SDK generator... hopefully the API team will fix this soon 😕

We've been using the getter on the last element of a chain of fields, because they're the most likely to be `nil` 😬

internal/cmd/kafka/utils.go
brianstrauch: Yep, this is safe to remove!

internal/cmd/kafka/command_cluster_update.go
brianstrauch: Same response as your other comment above.

internal/cmd/kafka/command_cluster_update.go
brianstrauch: I think I'd prefer not to, it would require two `isDedicated()` checks instead of one.

internal/cmd/kafka/command_cluster_update.go
tmodi11: I see, is there any other errors that would cause this to suggest looking into? If not, no worries, but this may help oncalls while debugging

brianstrauch:
Hey @tmodi11, just checking in- do we have an ETA for the changes in this PR?

tmodi11:
@brianstrauch I think we can close this PR

tmodi11:
@brianstrauch I think we can close this PR

PR #1716, opened 2023-02-15
Fix lint

brianstrauch:
Checklist
---------
   * yes: ok

What
----
#1686 was merged even though CI failed. This PR fixes the linter errors associated with that commit.


brianstrauch:
#1718 fixed this. Closing.

PR #1725, opened 2023-02-20
AsyncAPI Fix for Protobuf message logging

pagrawal10:
Release Notes
-------------

Breaking Changes
- <PLACEHOLDER>

New Features
- <PLACEHOLDER>

Bug Fixes
- <PLACEHOLDER>
- Fixes logging for the Protobuf case
- Ensures that in case of Protobuf, further functions are not called for that channel (subject).

Checklist
---------
   * yes
  

What
----

- In case of a Protobuf schema, that subject should be skipped. Ensuring that other functions are not called in such a case.
- Switching to log.Clilogger.Info() method instead of using log.Clilogger.log()
- Adding a unit test for the protobuf case

References
----------

https://confluent.slack.com/archives/C010Y0EP5MZ/p1676912651486499

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


DABH (approved):
LGTM, thank you!

PR #1726, opened 2023-02-21
[RCCA-11060] Use panta/machineid to fix `/etc/machine-id` not found issue on alpine linux

MuweiHe:
Release Notes
-------------
Bug Fixes
- Fixed an issue where users can't log in on Alpine linux due to missing ``/etc/machine-id`` file

Checklist
---------
   * yes: ok

What
----

Use a maintained fork of the previous `machineid` package, where it can detect machine id even in docker containers. 

References
----------
https://github.com/denisbrodbeck/machineid/issues/5#issuecomment-752416741

Test & Review
-------------

Tested on mac, ubuntu and alpine. After running login, `/etc/machine-id` is generated on alpine. 

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,6 @@ require (
 	github.com/confluentinc/properties v0.0.0-20190814194548-42c10394a787
 	github.com/confluentinc/schema-registry-sdk-go v0.0.19
 	github.com/davecgh/go-spew v1.1.1
-	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/dghubble/sling v1.4.1
 	github.com/fatih/color v1.14.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -104,6 +103,7 @@ require (
 	github.com/mattn/go-tty v0.0.4 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
+	github.com/panta/machineid v1.0.2 // indirect
```
brianstrauch, line 106: Run `go mod tidy`, this should not be indirect

brianstrauch (approved):

PR #1727, opened 2023-02-22
[CLI-2067] Use `$PATHEXT` to find executable plugins on Windows

brianstrauch:
Release Notes
-------------
Bug Fixes
- Use ``$PATHEXT`` to correctly determine if a plugin is executable on Windows

Checklist
---------
   * yes: ok

What
----
Before, the list of valid plugin extensions was hardcoded (and was missing several extension types!) Windows relies on `$PATHEXT` to determine which file extensions are executable, so we should too.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2067

Test & Review
-------------
Unit tests


sgagniere (approved):
+1

PR #1728, opened 2023-02-22
[CLI-1640] Add org management commands

sgagniere:
Release Notes
-------------
New Features
- Add new `confluent organization` commands: `describe`, `list`, `update`

Checklist
---------

What
----
Add new `organization` commands:
- `describe` returns the name and resource id of the current org. Unlike the other describe commands, it takes no argument as you cannot access an org you are not logged into.
- `update` likewise takes no arguments and has the flag `--name` for changing the name of the current organization.
- `list` is self explanatory. This, combined with `confluent login`, should make switching organizations in the CLI much easier.

This PR also adds unit & integration tests.

Test & Review
-------------
Manual testing in staging
Ran all tests


internal/cmd/environment/command_delete.go
```diff
@@ -31,7 +31,7 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 
 	environment, httpResp, err := c.V2Client.GetOrgEnvironment(id)
 	if err != nil {
-		return errors.CatchEnvironmentNotFoundError(err, httpResp)
+		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Environment, httpResp)
```
brianstrauch, line 34: Can you write an integration test for this?

internal/cmd/organization/command.go
```diff
@@ -0,0 +1,26 @@
...
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+)
+
+type command struct {
+	*pcmd.AuthenticatedStateFlagCommand
+}
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "organization",
```
brianstrauch, line 15:
```suggestion
		Use:         "organization",
		Aliases:     []string{"org"},
```

internal/cmd/organization/command.go
```diff
@@ -0,0 +1,26 @@
...
+		Short:       "View Confluent Cloud organizations.",
```
brianstrauch, line 16:
```suggestion
		Short:       "View Confluent Cloud organizations that you are a member of.",
```

internal/cmd/organization/command_describe.go
```diff
@@ -0,0 +1,42 @@
...
+
+type out struct {
+	Id        string `human:"ID" serialized:"id"`
+	Name      string `human:"Name" serialized:"name"`
+}
+
+func (c *command) newDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe",
+		Short: "Describe a Confluent Cloud organization.",
```
brianstrauch, line 20:
```suggestion
		Short: "Describe the current Confluent Cloud organization.",
```

brianstrauch: "... that you are logged in to"?

internal/cmd/organization/command_describe.go
```diff
@@ -0,0 +1,42 @@
...
+func (c *command) describe(cmd *cobra.Command, args []string) error {
+	organization, httpResp, err := c.V2Client.GetOrgOrganization(c.Context.GetOrganization().GetResourceId())
+	if err != nil {
+		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Organization, httpResp)
+	}
+
+	table := output.NewTable(cmd)
+	table.Add(&out{
+		Id:   *organization.Id,
+		Name: *organization.DisplayName,
```
brianstrauch, line 39:
```suggestion
		Id:   organization.GetId(),
		Name: organization.GetDisplayName(),
```

internal/pkg/errors/catcher.go
```diff
@@ -200,13 +200,13 @@ func CatchResourceNotFoundError(err error, resourceId string) error {
...
 
-func CatchEnvironmentNotFoundError(err error, r *http.Response) error {
+func CatchOrgV2ResourceNotFoundError(err error, resourceType string, r *http.Response) error {
 	if err == nil {
 		return nil
 	}
 
 	if r != nil && r.StatusCode == http.StatusForbidden {
-		return NewWrapErrorWithSuggestions(CatchCCloudV2Error(err, r), "environment not found or access forbidden", EnvNotFoundSuggestions)
+		return NewWrapErrorWithSuggestions(CatchCCloudV2Error(err, r), fmt.Sprintf("%s not found or access forbidden", resourceType), fmt.Sprintf(OrgResourceNotFoundSuggestions, resourceType))
```
brianstrauch, line 209: I would expect a 404 if the resource was not found... this runs when we get a 403. Can you test that we receive a 404 if the resource doesn't exist? If the backend is returning a 403 you should let them know that their status codes are wrong.

internal/pkg/errors/error_message.go
```diff
@@ -58,10 +58,10 @@ const (
 	InvalidCloudSuggestions            = "To list available connector plugin types, use `confluent connect plugin list`."
 	ConnectLogEventsNotEnabledErrorMsg = "Connect Log Events are not enabled for this organization"
 
-	// environment command
-	EnvNotFoundErrorMsg    = `environment "%s" not found`
-	EnvNotFoundSuggestions = "List available environments with `confluent environment list`."
-	EnvSwitchErrorMsg      = "failed to switch environment: failed to save config"
+	// environment & organization command
+	EnvNotFoundErrorMsg            = `environment "%s" not found`
+	OrgResourceNotFoundSuggestions = "List available %[1]ss with `confluent %[1]s list`."
```
brianstrauch, line 63: What does this syntax mean?

test/fixtures/output/help/cloud.golden
```diff
@@ -20,6 +20,7 @@ Available Commands:
   local           Manage a local Confluent Platform development environment.
   login           Log in to Confluent Cloud or Confluent Platform.
   logout          Log out of Confluent Cloud.
+  organization    View Confluent Cloud organizations.
```
brianstrauch, line 23: We can also update an organization, so I'd change this to "manage"

test/organization_test.go
```diff
@@ -0,0 +1,18 @@
+package test
+
+func (s *CLITestSuite) TestOrganization() {
+	tests := []CLITest{
+		{args: "organization describe", fixture: "organization/describe.golden"},
+		{args: "organization describe -o json", fixture: "organization/describe-json.golden"},
+		{args: "organization describe -o yaml", fixture: "organization/describe-yaml.golden"},
```
brianstrauch, line 7: IMO, you only need to test one of these (verify existence of the `--output` flag and serialization tags). The other is covered by unit tests.

test/test-server/org_handlers.go
```diff
@@ -57,17 +60,51 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
-	for _, env := range envs {
-		envList = append(envList, *env)
+// Handler for: "/org/v2/organizations/{id}"
+func handleOrgOrganization(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		orgId := vars["id"]
+		w.Header().Set("Content-Type", "application/json")
+
+		if r.Method == http.MethodGet {
```
brianstrauch, line 77: It would be great to get into the habit of using `switch` statements here. It results in less redundancy when there are multiple methods.

test/test-server/org_handlers.go
```diff
@@ -57,17 +60,51 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
+		}
+	}
+}
+
+// Handler for: "/org/v2/organizations"
+func handleOrgOrganizations(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		if r.Method == http.MethodGet {
+			organizationList := &orgv2.OrgV2OrganizationList{Data: getOrgResourcesList[orgv2.OrgV2Organization](OrgOrganizations)}
```
brianstrauch, line 97: This is the only place we're using `OrgOrganizations`. Let's just define it as an `[]*OrgV2Resource` right off the bat. Also- no need for it to be global **or** public as far as I can tell.

internal/cmd/organization/command.go
MuweiHe, line 16: You can update an org so it's not just `viewing`?

internal/cmd/organization/command_describe.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/resource"
+)
+
+type out struct {
+	Id        string `human:"ID" serialized:"id"`
+	Name      string `human:"Name" serialized:"name"`
+}
+
+func (c *command) newDescribeCommand() *cobra.Command {
```
brianstrauch, line 17: I think we should consider calling the list endpoint here so we can allow the user to pass in the `<id>`... any reason not to?

internal/cmd/organization/command_describe.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (c *command) describe(cmd *cobra.Command, args []string) error {
+	organization, httpResp, err := c.V2Client.GetOrgOrganization(c.Context.GetOrganization().GetResourceId())
+	if err != nil {
+		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Organization, httpResp)
+	}
+
+	table := output.NewTable(cmd)
+	table.Add(&out{
+		Id:   *organization.Id,
```
MuweiHe, line 38: `organization.GetId()`

internal/cmd/organization/command_list.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return err
+	}
+
+	list := output.NewList(cmd)
+	for _, organization := range organizations {
+		list.Add(&listOut{
+			IsCurrent: *organization.Id == c.Context.GetOrganization().GetResourceId(),
+			Id:        *organization.Id,
+			Name:      *organization.DisplayName,
```
MuweiHe, line 41: `organization.GetDisplayName()`

internal/cmd/organization/command_test.go
```diff
@@ -0,0 +1,155 @@
...
+	Id:          orgv2.PtrString(secondOrganizationID),
+	DisplayName: orgv2.PtrString(secondOrganizationName),
+}
+
+var orgOrganizationUpdated = orgv2.OrgV2Organization{
+	Id:          orgv2.PtrString(firstOrganizationID),
+	DisplayName: orgv2.PtrString(organizationNameUpdated),
+}
+
+func TestOrganizationTestSuite(t *testing.T) {
```
brianstrauch, line 60: Sorry you put all this work into this file... but we're trying to move away from this style of testing. We can accomplish everything in this file with our integration testing framework, with ~10x less code.

internal/cmd/organization/command_test.go
```diff
@@ -0,0 +1,155 @@
...
+	suite.Run(t, new(OrganizationTestSuite))
+}
+
+func (suite *OrganizationTestSuite) SetupTest() {
+	suite.conf = v1.AuthenticatedCloudConfigMock()
+	suite.accountClientMock = &ccloudv1mock.AccountInterface{
+		CreateFunc: func(arg0 context.Context, arg1 *ccloudv1.Account) (account *ccloudv1.Account, e error) {
```
MuweiHe, line 67: `func(_ context.Context, _ *ccloudv1.Account)`

internal/cmd/organization/command_test.go
```diff
@@ -0,0 +1,155 @@
...
+	orgClient := &orgv2.APIClient{
+		OrganizationsOrgV2Api: suite.V2ClientMock.orgClientMock,
+	}
+	resolverMock := &pcmd.FlagResolverImpl{
+		Out: os.Stdout,
+	}
+	prerunner := &cliMock.Commander{
+		FlagResolver: resolverMock,
+		Client:       client,
+		MDSClient:    nil,
```
MuweiHe, line 110:
```suggestion
```

internal/cmd/organization/command_update.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (c *command) newUpdateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update the current Confluent Cloud organization.",
+		Args:  cobra.NoArgs,
+		RunE:  c.update,
+	}
+
+	cmd.Flags().String("name", "", "New name for Confluent Cloud organization.")
```
brianstrauch, line 20:
```suggestion
	cmd.Flags().String("name", "", "Name of the Confluent Cloud organization.")
```

internal/cmd/organization/command_update.go
```diff
@@ -0,0 +1,42 @@
...
+		return err
+	}
+
+	updateOrganization := orgv2.OrgV2Organization{DisplayName: orgv2.PtrString(name)}
+	_, httpResp, err := c.V2Client.UpdateOrgOrganization(id, updateOrganization)
+	if err != nil {
+		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Organization, httpResp)
+	}
+
+	utils.ErrPrintf(cmd, errors.UpdateSuccessMsg, "name", "organization", id, name)
```
brianstrauch, line 40: Can we print the organization as a table, just like we do in `confluent org describe`?

internal/cmd/organization/command_update.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (c *command) update(cmd *cobra.Command, args []string) error {
+	id := c.Context.GetOrganization().GetResourceId()
+
+	name, err := cmd.Flags().GetString("name")
+	if err != nil {
+		return err
+	}
+
+	updateOrganization := orgv2.OrgV2Organization{DisplayName: orgv2.PtrString(name)}
```
MuweiHe, line 34: what's the behavior if no `--name` is set? Will it error?

test/test-server/org_handlers.go
```diff
@@ -57,17 +60,51 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
-	return envList
 }
+
+func getOrgResourcesList[OrgV2Resource orgv2.OrgV2Environment | orgv2.OrgV2Organization] (resources []*OrgV2Resource) []OrgV2Resource {
+	resList := []OrgV2Resource{}
+	for _, res := range resources {
+		resList = append(resList, *res)
+	}
+	return resList
+}
```
MuweiHe, line 110:
```suggestion
}

```

test/fixtures/output/organization/list.golden
```diff
@@ -0,0 +1,5 @@
+  Current |   ID    | Name
```
brianstrauch, line 1: Should we be sorting this list? What do we do for `environment list` or `kafka cluster list`?

brianstrauch (changes requested):

internal/cmd/organization/command_update.go
sgagniere: `name` is a required flag: `_ = cmd.MarkFlagRequired("name")`

internal/cmd/organization/command.go
sgagniere: Good point; changed it to `Manage your ...`

sgagniere: Changed it to `Manage your...`

internal/cmd/organization/command_describe.go
sgagniere: CDMUM recommended that we not do this without a compelling reason. The list endpoint is the only exception to the general rule of not allowing access to other orgs.

internal/cmd/organization/command_test.go
sgagniere: No worries; it wasn't all that much work for this part. I'll remove this in a later commit.

internal/pkg/errors/error_message.go
sgagniere: `[n]` is an argument index, so `%[1]s` means 'first argument formatted as a string'

I use this to repeat the argument twice instead of having to write the argument twice.

internal/pkg/errors/catcher.go
sgagniere: Tested in prod w/ `environment describe env-qwerty` and `organization describe`; both return 403 Forbidden (the latter gives an error detail explaining that the API is not available, which makes sense since it hasn't hit prod yet).

I'll make tickets (this also happens when describing non-existent Kafka clusters)

internal/cmd/organization/command_describe.go
brianstrauch: Maybe the idea is that in the future, the describe endpoint will return sensitive data that's not returned by the list endpoint?

If that's the case, the describe endpoint could just omit those sensitive fields if it's not the current org...

This seems really inconsistent & bad UX!

test/fixtures/output/organization/list.golden
sgagniere: `kafka cluster list` appears to do some type of sorting (I didn't investigate too much, but w/ 4 basic clusters in the same cloud and region it sorts by name).

Both `environment list` and `organization list` are sorted by creation date in the response (can be checked by looking at `GetMetadata()`), and after that it's just whatever sorting happens in `Print()`. Which, at least for `environment` and `organization`, looks to be a reversal of the response order but with the active resource always listed last.

sgagniere: For the integration tests here, I think the reason it looks unsorted is because of that last bit where the active one always gets moved to the bottom of the list.

internal/cmd/organization/command_describe.go
sgagniere: Hmm, checking the thread again this discussion was actually about the behavior of the describe endpoint; and how we did not want to design it to allow access for different orgs. As far as the CLI is concerned I think that might have been discussed in a meeting instead.

Well, I can ask them if they're ok with the CLI working around it via the list endpoint.

test/fixtures/output/organization/list.golden
brianstrauch: Ah, I think it sorts by default, but we should be ignoring the first column

internal/cmd/organization/command_list.go
```diff
@@ -0,0 +1,45 @@
+package organization
+
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+type listOut struct {
```
brianstrauch, line 10: I think it would be helpful if we also showed `IsCurrent` for describe and update. It's always going to be true, but just for the user to sanity check what they're doing, and to match other commands.

internal/pkg/errors/error_message.go
brianstrauch: Didn't know you could do that, thanks!

test/test-server/org_handlers.go
```diff
@@ -64,6 +64,47 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
+		err := json.NewEncoder(w).Encode(organization)
+		require.NoError(t, err)
+	}
+}
+
+// Handler for: "/org/v2/organizations"
+func handleOrgOrganizations(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		if r.Method == http.MethodGet {
```
brianstrauch, line 96:
```suggestion
		switch r.Method {
		case http.MethodGet:
```

test/test-server/org_handlers.go
```diff
@@ -64,6 +64,47 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
 	}
 }
 
+// Handler for: "/org/v2/organizations/{id}"
+func handleOrgOrganization(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		orgId := vars["id"]
```
brianstrauch, line 71: Probably a good idea to match the arg names, just like how we try to match flag names.
```suggestion
		id := vars["id"]
```

brianstrauch:
As a follow up to this PR, can you use the new organization API to replace equivalent ccloudv1 calls, like the one found in `PersistCCloudCredentialsToConfig()`?

internal/cmd/organization/command_describe.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (c *command) newDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe",
+		Short: "Describe the current Confluent Cloud organization.",
+		Args:  cobra.NoArgs,
+		RunE:  c.describe,
+	}
+
+	pcmd.AddOutputFlag(cmd)
```
brianstrauch, line 26: `--context`?

internal/cmd/organization/command_update.go
```diff
@@ -0,0 +1,50 @@
...
+		Use:   "update",
+		Short: "Update the current Confluent Cloud organization.",
+		Args:  cobra.NoArgs,
+		RunE:  c.update,
+	}
+
+	cmd.Flags().String("name", "", "Name of the Confluent Cloud organization.")
+	_ = cmd.MarkFlagRequired("name")
+
+	pcmd.AddOutputFlag(cmd)
```
brianstrauch, line 24: Do we want to add `--context`?
```suggestion
	cmd.Flags().String("name", "", "Name of the Confluent Cloud organization.")
	pcmd.AddOutputFlag(cmd)
	
	_ = cmd.MarkFlagRequired("name")
```

internal/cmd/organization/command_update.go
```diff
@@ -0,0 +1,50 @@
...
+
+	updateOrganization := orgv2.OrgV2Organization{DisplayName: orgv2.PtrString(name)}
+	organization, httpResp, err := c.V2Client.UpdateOrgOrganization(id, updateOrganization)
+	if err != nil {
+		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Organization, httpResp)
+	}
+
+	table := output.NewTable(cmd)
+	table.Add(&out{
+		IsCurrent: organization.GetId() == c.Context.GetOrganization().GetResourceId(),
```
brianstrauch, line 45:
```suggestion
		IsCurrent: organization.GetId() == id,
```

test/organization_test.go
```diff
@@ -0,0 +1,16 @@
+package test
+
+func (s *CLITestSuite) TestOrganization() {
+	tests := []CLITest{
+		{args: "organization describe", fixture: "organization/describe.golden"},
+		{args: "organization describe -o json", fixture: "organization/describe-json.golden"},
+		{args: "organization list", fixture: "organization/list.golden"},
+		{args: "organization list -o json", fixture: "organization/list-json.golden"},
+		{args: "organization update --name default-updated", fixture: "organization/update.golden"},
```
brianstrauch, line 9: `update -o`?

test/test-server/org_handlers.go
```diff
@@ -64,6 +64,48 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		id := vars["id"]
+		w.Header().Set("Content-Type", "application/json")
+
+		displayName := "default"
+		switch r.Method {
+		case http.MethodPatch:
+			req := orgv2.OrgV2Environment{}
+			err := json.NewDecoder(r.Body).Decode(&req)
```
brianstrauch, line 78: More idiomatic; if you were to pass this variable into 10 different functions, it would be easier not to have to add `&` every time.
```suggestion
			req := &orgv2.OrgV2Environment{}
			err := json.NewDecoder(r.Body).Decode(req)
```

test/test-server/org_handlers.go
```diff
@@ -64,6 +64,48 @@ func handleOrgEnvironments(t *testing.T) http.HandlerFunc {
...
+			require.NoError(t, err)
+			displayName = *req.DisplayName
```
brianstrauch, line 80:
```suggestion
			displayName = req.GetDisplayName()
```

internal/cmd/organization/command_describe.go
sgagniere: I'm on the fence. For describe it wouldn't matter too much, but I think it might be bad for update because the org itself is not part of the context, so a user could end up updating the name of the wrong org.

brianstrauch: > a user could end up updating the name of the wrong org

Organizations have unique resource IDs, right? I don't think that should be a problem. The `--context` flag would allow you to describe or update an organization without needing to log in to that org again, seems useful to me!

sgagniere: They have unique resource id's, but the resource id is not part of the context name. I just did a test using a pair of stag accounts; the first is in 2 orgs. I logged into the second and then used the first user's context for this flag and the org that got updated depended on whichever one the first user last logged in to.

Because it's not in the name, you can't choose and you have to call describe with that context or look in the config json file to see which org will be updated.

brianstrauch: > the org that got updated depended on whichever one the first user last logged in to

Ah, I see. Sounds like we don't want the `--context` flag for any of the `confluent organization` commands, then.

internal/cmd/organization/command_list.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (c *command) newListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List Confluent Cloud organizations.",
+		Args:  cobra.NoArgs,
+		RunE:  c.list,
+	}
+
+	pcmd.AddContextFlag(cmd, c.CLICommand)
```
brianstrauch, line 18:
```suggestion
```

internal/cmd/organization/command_describe.go
sgagniere: Yeah it's unfortunate. Is adding the org to the context something we could consider doing, maybe for v4? That would resolve this.

internal/cmd/organization/command_list.go
sgagniere: Ack, I didn't realize that was there.

internal/cmd/organization/command_describe.go
brianstrauch: Hmm, the organization should be able to be inferred from the context, right?

sgagniere: Yeah it's stored in the context.

brianstrauch: Yeah, can you make a ticket for that? It looks like right now if a user is in more than one organization, their old context will get overridden by a new one when the log in to a different organization... not good!

sgagniere: Will do, and that sounds like the behavior I was seeing when I was testing this.

brianstrauch (approved):

PR #1729, opened 2023-02-22
[CLI-1560-1] Add warning msg when command requires an env but no env is selected

MuweiHe:
Release Notes
-------------
New Features
- Display a warning message when the command potentially requires an active env, but no env is actually selected.

Checklist
---------
   * yes: ok
   
What
----

Some commands will require an active env. Before this change, even when no active env is selected or the former active env was deleted, command will still see the no-longer-exist env as the env for this command. Now add a warning when there's no active env, and reset env to nil and active env is deleted.

References
----------

https://confluentinc.atlassian.net/browse/CLI-1560

Test & Review
-------------

```
// a command that potentially gets the env for it
mhe@C02DV0KVMD6T cli % confluent env list
WARNING: This command requires an environment; no environment selected.

Suggestions:
    This issue may occur if this user has no valid role bindings. Contact an Organization Admin to create a role binding for this user.

  Current |    ID     | Name   
----------+-----------+--------
          | env-2mdy1 | newer  

// a command that doesn't require env:
mhe@C02DV0KVMD6T cli % confluent api-key list
  Current |       Key        | Description |  Owner   |   Owner Email    |  Resource Type  |  Resource   |       Created         
----------+------------------+-------------+----------+------------------+-----------------+-------------+-----------------------
          | 6VWFDUGHJVV45MCQ |             | u-e02dw5 | mhe@confluent.io | kafka           | lkc-3rddg0  | 2023-02-16T17:39:05Z  
```

Open Questions / Follow-ups
---------------------------


internal/pkg/errors/error_message.go
```diff
@@ -62,6 +62,7 @@ const (
 	EnvNotFoundErrorMsg    = `environment "%s" not found`
 	EnvNotFoundSuggestions = "List available environments with `confluent environment list`."
 	EnvSwitchErrorMsg      = "failed to switch environment: failed to save config"
+	UseExistingEnvMsg      = `Switch to an active environment with "confluent environment use"`
```
sgagniere, line 65: Suggestion: `Select a new active environment` instead of `Switch to an active environment`

brianstrauch:
Ticket for reference: https://confluentinc.atlassian.net/browse/CLI-1560

While this is a good thing to do, I don't think it closes this ticket since the reported error will still occur for `confluent kafka cluster list` if the user doesn't take our recommendation to run `confluent environment use`.

MuweiHe:
> Ticket for reference: https://confluentinc.atlassian.net/browse/CLI-1560
> 
> While this is a good thing to do, I don't think it closes this ticket since the reported error will still occur for `confluent kafka cluster list` if the user doesn't take our recommendation to run `confluent environment use`.

Yes, I think I'll fix it in v4, as it'll bring breaking changes now.

internal/cmd/environment/command_delete.go
```diff
@@ -45,5 +45,8 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 	}
 
 	utils.ErrPrintf(cmd, errors.DeletedResourceMsg, resource.Environment, id)
+	if id == c.EnvironmentId() {
```
sgagniere, line 48: We might want to remove stale env info from the context so that `c.EnvironmentId()` stops returning the deleted id as well in case someone doesn't immediately switch to a different one.

sgagniere (approved):

internal/pkg/errors/error_message.go
brianstrauch:
```suggestion
	UseExistingEnvMsg      = "Select a new environment with `confluent environment use`."
```

brianstrauch: I'd also drop the "active", the point of selecting an environment is to make it active.

internal/cmd/environment/command_delete.go
MuweiHe: i tried setting context auth account to nil, but it'll print out the `no org set` warning for every command you run until you set one. I think we didn't see this issue because we were always using the already deleted env id.

MuweiHe: Any idea how to fix this? I don't want to bring a warning for every following command, that might mess up users scripts

brianstrauch:
> Yes, I think I'll fix it in v4, as it'll bring breaking changes now.

@MuweiHe I don't think that fixing the error message is a breaking change. It's a bug fix since the error message is being improved.

internal/cmd/environment/command_delete.go
brianstrauch: > What should c.EnvironmentId() return if it shouldn't return the already deleted one?

It should definitely return `nil`!

> i tried setting context auth account to nil

This is the right way to do it, IMO. The warning (and possibly a few other things) just needs to be fixed.

test/fixtures/output/environment/11-prompt.golden
```diff
@@ -1,2 +1,3 @@
 Are you sure you want to delete environment "not-595"?
 To confirm, type "default". To cancel, press Ctrl-C: Deleted environment "not-595".
+Switch to an active environment with "confluent environment use"
```
brianstrauch, line 3: Any way we can put `Suggestions:` on top of this, in the same format we use for our errors? (We should convey to the user that this isn't mandatory)

MuweiHe: yes sure

MuweiHe:
> > Yes, I think I'll fix it in v4, as it'll bring breaking changes now.
> 
> @MuweiHe I don't think that fixing the error message is a breaking change. It's a bug fix since the error message is being improved.

I mean like Steven said, we should delete the deleted env from context and config. But what should we return instead of this deleted env id, to avoid printing out a `no env selected` warning for every command until you set an active env? That might be a breaking change

brianstrauch:
> to avoid printing out a no env selected

Why would we want to avoid printing this if there actually is no environment selected?

MuweiHe:
> > to avoid printing out a no env selected
> 
> Why would we want to avoid printing this if there actually is no environment selected?

Might be a breaking change right now? I'll come together with e.g. `api-key list` output. I'll think of a way to avoid printing that for every command but only when `c.EnvironmentId()` is actually called.

internal/cmd/environment/command_delete.go
```diff
@@ -45,5 +45,12 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 	}
 
 	utils.ErrPrintf(cmd, errors.DeletedResourceMsg, resource.Environment, id)
+	if id == c.EnvironmentId(cmd) {
+		c.Context.SetEnvironment(nil)
+
+		if err := c.Config.Save(); err != nil {
+			return errors.Wrap(err, errors.EnvSwitchErrorMsg)
+		}
+	}
```
MuweiHe, line 54: reset context env to nil when active env is deleted

internal/pkg/cmd/prerunner.go
```diff
@@ -322,11 +326,6 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(cmd *cobra
 			return setContextErr
 		}
 
-		if command.Context.GetEnvironment() == nil {
```
brianstrauch, line 325: What's the reason for moving the suggestion out of here?

brianstrauch: Are there unauthenticated commands that look up the environment?

MuweiHe: Remember your comments in steven's previous pr? adding this check here is like assuming all authenticatedCLICommand will require an active env, while that's not always true (for e.g. confluent api-key list).

brianstrauch: Got it, ok!

internal/pkg/cmd/prerunner.go
```diff
@@ -165,7 +165,11 @@ func (c *AuthenticatedCLICommand) AuthToken() string {
 	return c.Context.GetAuthToken()
 }
 
-func (c *AuthenticatedCLICommand) EnvironmentId() string {
+func (c *AuthenticatedCLICommand) EnvironmentId(command *cobra.Command) string {
```
brianstrauch, line 168: Instead of passing `cmd` everywhere (this makes functions significantly harder to unit test!) can we return the environment ID and the suggestion (which may be "") in a similar manner to how we return an error in Go?

MuweiHe: Then we'll need to add `if warning!="", printf warning` everywhere `c.EnvironmentId()` is called. That's less efficient and more duplicate code

sgagniere: We could go for the middle ground: `_, _ = fmt.Fprint(os.Stderr, errors.EnvNotSetErrorMsg+noEnvSuggestions+"\n")` so that we don't need to pass in command just for the `ErrPrint` and we don't need to check for the empty string at every call.

brianstrauch: Hmm, good call @sgagniere. I think we used to pass `cmd` everywhere so our unit tests (which were really integration tests) could record the output. We could make a PR to rewrite the `util.Print*()` functions so they don't need `cmd` (and during that process, rewrite the unit tests that break into integration tests).

brianstrauch: ^ If the above sounds good, let's merge this PR as-is.

sgagniere: Sounds good to me too.

sgagniere (approved):

internal/pkg/cmd/prerunner.go
brianstrauch: I'll take care of this today!

PR #1732, opened 2023-02-23
Fix panic in api-key list

sgagniere:
Release Notes
-------------
Bug Fixes
- Resolve a panic that could occur in ``confluent api-key list``

Checklist
---------
   * yes: ok

What
----
Change from direct access of User.ResourceId and User.Id to get functions.

References
----------
n/a

Test & Review
-------------
Ran unit tests

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
Are there any downsides to filling the map with `""` --> `""`?

PR #1742, opened 2023-02-27
Replace unit tests that mock cmd

brianstrauch:
Checklist
---------
   * yes: ok

What
----
As discussed in https://github.com/confluentinc/cli/pull/1729#discussion_r1116330443, we can avoid passing `cmd` throughout our code, since our integration testing framework doesn't require it.

As a first step, this PR changes the `util.Print*()` functions to print to stdout and stderr, and fix any unit tests that break (or remove, and validate that integration tests provide equal test coverage)

The next PR will remove the chain of `cmd` variables passed to the `util.Print*()` functions.

Test & Review
-------------
Verified that all tests still pass


internal/cmd/login/command_test.go
```diff
@@ -832,9 +832,8 @@ func TestLoginWithExistingContext(t *testing.T) {
 		loginCmd, cfg := newLoginCmd(auth, userInterface, s.isCloud, req, mockNetrcHandler, AuthTokenHandler, mockLoginCredentialsManager, LoginOrganizationManager)
 
 		// Login to the CLI control plane
-		output, err := pcmd.ExecuteCommand(loginCmd, s.args...)
+		_, err := pcmd.ExecuteCommand(loginCmd, s.args...)
 		req.NoError(err)
-		req.NotContains(output, fmt.Sprintf(errors.LoggedInAsMsg, promptUser))
```
sgagniere, line 837: Is this unnecessary because it's already caught by integration tests?

sgagniere (approved):
lgtm, thanks!

internal/cmd/login/command_test.go
brianstrauch: Yep!

PR #1743, opened 2023-02-28
[CLI-2101] Prevent panic when the parsing environment flag into the context

sgagniere:
Release Notes
-------------

Breaking Changes
- <PLACEHOLDER>

New Features
- <PLACEHOLDER>

Bug Fixes
- Prevent a panic that could occur when using commands with the ``--environment`` flag

Checklist
---------
   * yes: ok

What
----
If `State` or `Auth` are unexpectedly nil, trying to get `c.State.Auth.Accounts` will panic. This PR replaces the direct access with `Get` functions that check for both a non-nil state and auth.

References
----------
n/a

Test & Review
-------------
All tests pass.
Checked that commands using `--environment` flag still work correctly, and that everything works if you delete the the environment list from the config file (although I'll note that this is true even before this PR, so it's unclear why exactly this issue occurred in the first place).

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/config/v1/context.go
+++ b/internal/pkg/config/v1/context.go
@@ -214,6 +214,13 @@ func (c *Context) GetEnvironments() []*ccloudv1.Account {
 	return nil
 }
 
+func (c *Context) SetEnvironments(environments []*ccloudv1.Account) {
+	if c.GetAuth() == nil {
+		c.SetAuth(new(AuthConfig))
+	}
+	c.GetAuth().Accounts = environments
+}
+
 func (c *Context) GetAuthToken() string {
 	if state := c.GetState(); state != nil {
 		return state.AuthToken
--- a/internal/pkg/dynamic-config/dynamic_context.go
+++ b/internal/pkg/dynamic-config/dynamic_context.go
@@ -39,7 +39,7 @@ func (d *DynamicContext) ParseFlagsIntoContext(cmd *cobra.Command, client *cclou
 		}
 
 		// If environment ID is not found in config, make api call and check against those accounts
-		if !d.verifyEnvironmentId(environment, d.State.Auth.Accounts) {
+		if !d.verifyEnvironmentId(environment, d.GetEnvironments()) {
 			if client == nil {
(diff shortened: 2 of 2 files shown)


internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -39,7 +39,7 @@ func (d *DynamicContext) ParseFlagsIntoContext(cmd *cobra.Command, client *cclou
 		}
 
 		// If environment ID is not found in config, make api call and check against those accounts
-		if !d.verifyEnvironmentId(environment, d.State.Auth.Accounts) {
+		if !d.verifyEnvironmentId(environment, d.GetEnvironments()) {
```
brianstrauch, line 42: Do we still need to use the ccloudv1 client in this function? I thought we had a ccloudv2 function for getting org environments.

brianstrauch (approved):
Thanks!

internal/pkg/dynamic-config/dynamic_context.go
sgagniere: This doesn't actually use the v1 client, it's just grabbing the environment from the local context state.

internal/pkg/config/v1/context.go
```diff
@@ -214,6 +214,13 @@ func (c *Context) GetEnvironments() []*ccloudv1.Account {
 	return nil
 }
 
+func (c *Context) SetEnvironments(environments []*ccloudv1.Account) {
```
MuweiHe, line 217: Or `SetAccounts` cause that's the actual name of this field?

sgagniere: I think I still prefer `Environments`, because it matches the other `Get` and `Set` methods for accounts/environments in this file.

internal/pkg/dynamic-config/dynamic_context.go
brianstrauch: Oops, I meant the `getAllEnvironments()` function later down... On second thought, this is probably a breaking change, so we might want to save it for v4.

PR #1747, opened 2023-03-01
[CLI-2262] Check for empty flag values in `price list`

brianstrauch:
Release Notes
-------------
Bug Fixes
- Prevent empty values from being passed to flags in ``confluent price list``

Checklist
---------
   * yes: ok

What
----
`confluent price list --cloud "" --region ""` used to be accepted. Empty values should not be accepted, so now we error.

Used the shlex package to parse integration test commands in a similar manner to bash/zsh (now you can pass multi-word arguments in quotes)

References
----------
https://confluentinc.atlassian.net/browse/CLI-2262

Test & Review
-------------
Added an integration test and fixed other integration tests after switching to shlex.


sgagniere (approved):
+1

PR #1748, opened 2023-03-01
[CLI-962] add `--use` flag in `api-key create` to automatically use the created api-key for the provided resource

MuweiHe:
Release Notes
-------------
New Features
- Add the ``--use`` flag to ``confluent api-key create`` to select the API key after it has been created

Checklist
---------
   * yes: ok

What
----

new flag ``--use`` in ``confluent api-key create`` which will use the created API key for the provided resource after successfully creating the key.
If the resource provided is not a kafka cluster, print out error that ``use`` was not effective.

References
----------

https://confluentinc.atlassian.net/browse/CLI-962

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/api-key/command_create.go
+++ b/internal/cmd/api-key/command_create.go
@@ -50,6 +50,7 @@ func (c *command) newCreateCommand() *cobra.Command {
 
 	cmd.Flags().String(resourceFlagName, "", `The resource ID. Use "cloud" to create a Cloud API key.`)
 	cmd.Flags().String("description", "", "Description of API key.")
+	cmd.Flags().Bool("use", false, "automatically use this apikey for the provided resource.")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
@@ -131,6 +132,21 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
 		if err := c.keystore.StoreAPIKey(userKey, clusterId); err != nil {
 			return errors.Wrap(err, errors.UnableToStoreAPIKeyErrorMsg)
 		}
+	}
+
+	use, err := cmd.Flags().GetBool("use")
+	if err != nil {
+		return err
+	}
+	if use {
+		if resourceType != resource.KafkaCluster {
+			return errors.Wrap(errors.New(errors.NonKafkaNotImplementedErrorMsg), "--use flag ineffective")
+		}
+		err = c.Context.UseAPIKey(userKey.Key, clusterId)
+		if err != nil {
+			return errors.NewWrapErrorWithSuggestions(err, errors.APIKeyUseFailedErrorMsg, fmt.Sprintf(errors.APIKeyUseFailedSuggestions, userKey.Key))
+		}
+		utils.Printf(cmd, errors.UseAPIKeyMsg, userKey.Key, clusterId)
 	}
 
 	return nil
--- a/internal/cmd/api-key/command_create.go
+++ b/internal/cmd/api-key/command_create.go
@@ -50,7 +50,7 @@ func (c *command) newCreateCommand() *cobra.Command {
 
 	cmd.Flags().String(resourceFlagName, "", `The resource ID. Use "cloud" to create a Cloud API key.`)
 	cmd.Flags().String("description", "", "Description of API key.")
-	cmd.Flags().Bool("use", false, "automatically use this apikey for the provided resource.")
+	cmd.Flags().Bool("use", false, "Use the created apikey for the provided resource.")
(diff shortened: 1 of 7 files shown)


brianstrauch:
This goes against the principle of "each command does one thing" so I'm a bit weary of adding it, although it's definitely useful. Should we add this to `context create`, `environment create`, and `kafka cluster create` as well?

internal/cmd/api-key/command_create.go
```diff
@@ -50,6 +50,7 @@ func (c *command) newCreateCommand() *cobra.Command {
 
 	cmd.Flags().String(resourceFlagName, "", `The resource ID. Use "cloud" to create a Cloud API key.`)
 	cmd.Flags().String("description", "", "Description of API key.")
+	cmd.Flags().Bool("use", false, "Use the created apikey for the provided resource.")
```
sgagniere, line 53:
```suggestion
	cmd.Flags().Bool("use", false, "Use the created API key for the provided resource.")
```

sgagniere (approved):
lgtm!

internal/cmd/api-key/command_create.go
```diff
@@ -133,6 +134,21 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	use, err := cmd.Flags().GetBool("use")
+	if err != nil {
+		return err
+	}
+	if use {
+		if resourceType != resource.KafkaCluster {
+			return errors.Wrap(errors.New(errors.NonKafkaNotImplementedErrorMsg), `"--use" set but ineffective`)
```
brianstrauch, line 143:
```suggestion
			return errors.Wrap(errors.New(errors.NonKafkaNotImplementedErrorMsg), "`--use` set but ineffective")
```

MuweiHe:
> This goes against the principle of "each command does one thing" so I'm a bit weary of adding it, although it's definitely useful. Should we add this to `context create`, `environment create`, and `kafka cluster create` as well?

This one was specifically requested. let's see if people request the other ones and add them as needed.

internal/cmd/api-key/command_create.go
```diff
@@ -133,6 +134,21 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	use, err := cmd.Flags().GetBool("use")
+	if err != nil {
+		return err
+	}
+	if use {
```
brianstrauch, line 141: Can you write an integration test for this, or break it out into a unit test?

MuweiHe: added an int test

test/api_key_test.go
```diff
@@ -109,6 +109,9 @@ func (s *CLITestSuite) TestAPIKey() {
 		{args: "api-key create --description json-output --resource lkc-other1 -o json", fixture: "api-key/45.golden"},
 		{args: "api-key create --description yaml-output --resource lkc-other1 -o yaml", fixture: "api-key/46.golden"},
 
+		// create api-key and use for the resource
+		{args: "api-key create --description my-cool-app --resource lkc-cool1 --use", fixture: "api-key/60.golden"}, // MYKEY4
```
brianstrauch: Let's run `api-key describe` immediately after to prove that `--use` worked?

brianstrauch: We might also want to split up big functions like `TestAPIKey()`... it takes way longer to debug when there are so many lines.

brianstrauch:
Can you use the `--use` flag here? :smile: https://github.com/confluentinc/cli/blob/0a41d73b69996ed4e31364188387d06be5476180/test/cli_test.go#L163-L172

MuweiHe:
> Can you use the `--use` flag here? 😄
> 
> https://github.com/confluentinc/cli/blob/0a41d73b69996ed4e31364188387d06be5476180/test/cli_test.go#L163-L172

nice catch！

test/cli_test.go
```diff
@@ -159,13 +159,7 @@ func (s *CLITestSuite) runIntegrationTest(tt CLITest) {
...
 
 		if tt.authKafka != "" {
-			output := runCommand(t, testBin, []string{}, "api-key create --resource "+tt.useKafka, 0, "")
-			if *debug {
-				fmt.Println(output)
-			}
-			// HACK: we don't have scriptable output yet so we parse it from the table
-			key := strings.TrimSpace(strings.Split(strings.Split(output, "\n")[3], "|")[2])
-			output = runCommand(t, testBin, []string{}, fmt.Sprintf("api-key use %s --resource %s", key, tt.useKafka), 0, "")
+			output := runCommand(t, testBin, []string{}, "api-key create --use --resource "+tt.useKafka, 0, "")
```
brianstrauch, line 162:
```suggestion
			output := runCommand(t, testBin, []string{}, fmt.Sprintf("api-key create --resource %s --use", tt.useKafka), 0, "")
```

brianstrauch (approved):

PR #1752, opened 2023-03-01
Allow integration tests to be run from IDE

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Fix for a regression introduced by #1719. We should build the integration test CLI from cli_test.go instead of `make integration test` so the binary gets updated when tests are run from the IDE instead of the Makefile.

Also fixed a Makefile warning, also introduced by #1719.

Test & Review
-------------
Manual verification


MuweiHe (approved):
lgtm!

PR #1755, opened 2023-03-01
[CLI-454] Print pretty json schema in `confluent sr schema describe`

MuweiHe:
Release Notes
-------------
New Features
- Print pretty JSON output for ``confluent schema-registry schema describe``

Checklist
---------
   * yes: ok

What
----

protobuf schema:
```
mhe@C02DV0KVMD6T cli % confluent sr schema describe 100003                                          
Schema ID: 100003
Type: PROTOBUF
Schema:
syntax = "proto3";
package clickstream;

message events {
  string ip = 1;
  int32 userid = 2;
  string remote_user = 3;
}
```
avro and json schema, which used to be one line of json string.
```
mhe@C02DV0KVMD6T cli % confluent sr schema describe 100002
Schema ID: 100002
Type: JSON
Schema:
{
    "namespace": "io.confluent.examples.clients.basicavro",
    "name": "Payment",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "amount",
            "type": "double"
        }
    ]
}
```

References
----------

https://confluentinc.atlassian.net/browse/CLI-454

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/cmd/schema-registry/command_schema_describe.go
```diff
@@ -223,7 +224,18 @@ func printSchema(cmd *cobra.Command, schemaID int64, schema string, sType string
...
+	if sType == "PROTOBUF" {
+		utils.Println(cmd, "Schema:\n"+schema)
+	} else {
+		var jsonBuffer bytes.Buffer
+		if err := json.Indent(&jsonBuffer, []byte(schema), "", "    "); err != nil {
+			return err
+		}
+		schemaString := jsonBuffer.String()
+		utils.Println(cmd, "Schema:\n"+schemaString)
+	}
```
brianstrauch, line 237: Looks like we can extract the `utils.Println()` calls, since they're duplicated. It's also unclear what `sType` means; can we rename it?
```suggestion
	switch sType {
	case "PROTOBUF":
		utils.Println(cmd, "Schema:")
		utils.Println(cmd, schema)
	default:
		var jsonBuffer bytes.Buffer
		if err := json.Indent(&jsonBuffer, []byte(schema), "", "    "); err != nil {
			return err
		}
		schemaString := jsonBuffer.String()
		utils.Println(cmd, "Schema:")
		utils.Println(schemaString)
	}
```

internal/cmd/schema-registry/command_schema_describe.go
```diff
@@ -223,7 +224,21 @@ func printSchema(cmd *cobra.Command, schemaID int64, schema string, sType string
...
+		utils.Println(cmd, schema)
+	default:
+		var jsonBuffer bytes.Buffer
+		if err := json.Indent(&jsonBuffer, []byte(schema), "", "    "); err != nil {
+			return err
+		}
+		schemaString := jsonBuffer.String()
+		utils.Println(cmd, "Schema:")
+		utils.Println(schemaString)
+	}
```
brianstrauch, line 240: I'm assuming that "JSON" and "AVRO" are the only other types here. If there's ever a new type in the future we should not treat it as JSON by default. We can also get rid of some duplication here too.
```suggestion
	switch sType {
	case "JSON", "AVRO":
		var jsonBuffer bytes.Buffer
		if err := json.Indent(&jsonBuffer, []byte(schema), "", "    "); err != nil {
			return err
		}
		schema = jsonBuffer.String()
	}
	
	utils.Println(cmd, "Schema:")
	utils.Println(schemaString)
```

internal/cmd/schema-registry/command_schema_test.go
```diff
@@ -68,10 +68,10 @@ func (suite *SchemaTestSuite) SetupTest() {
 				return srsdk.RegisterSchemaResponse{Id: id}, nil, nil
 			},
 			GetSchemaFunc: func(_ context.Context, _ int32, _ *srsdk.GetSchemaOpts) (srsdk.SchemaString, *http.Response, error) {
-				return srsdk.SchemaString{Schema: "Potatoes"}, nil, nil
+				return srsdk.SchemaString{Schema: `{"Potatoes":1}`}, nil, nil
```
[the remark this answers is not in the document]
MuweiHe: yes

internal/cmd/schema-registry/command_schema_describe.go
```diff
@@ -223,7 +224,18 @@ func printSchema(cmd *cobra.Command, schemaID int64, schema string, sType string
 	if sType != "" {
 		utils.Println(cmd, "Type: "+sType)
 	}
-	utils.Println(cmd, "Schema: "+schema)
+
+	if sType != "PROTOBUF" {
```
MuweiHe, line 228: protobuf is formatted already. The reason for `!=` is that avro schemas created using web ui and sometimes using cli, they have no schema type set which is `""`.

internal/cmd/schema-registry/command_schema_describe.go
```diff
@@ -226,16 +226,16 @@ func printSchema(cmd *cobra.Command, schemaID int64, schema string, sType string
 	}
 
 	switch sType {
-	case "JSON", "AVRO":
+	case "JSON", "AVRO", "":
```
brianstrauch: Does the backend really return `""`?

brianstrauch (approved):

PR #1760, opened 2023-03-03
[CLI-2171] Always use latest Go version for glibc Linux builds

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Remove golang and goreleaser install from the base images.
The `build_linux_glibc.sh` script now reads the go version and passes it as an argument to the Dockerfiles.
The current goreleaser version is already installed by the `gorelease-linux-glibc` make targets, so there's no point installing it again in the images so I've just left that out.

I've also added the Dockerfiles used to build the base images. With this PR there really shouldn't be a need to update these as frequently as before, but I've added them in case we do need to update them again for whatever reason.

References
----------

Test & Review
-------------
Manual testing.

Open Questions / Follow-ups
---------------------------


dockerfiles/Dockerfile_linux_glibc_arm64_base
```diff
@@ -0,0 +1,9 @@
+FROM --platform=linux/arm64/v8 ubuntu:bionic-20220913
+
+RUN apt update
+
+RUN apt -y install make sudo
+
+RUN sudo apt -y install git wget build-essential rename
```
brianstrauch, line 7: Is it possible to pre-install this on the base image?

scripts/build_linux_glibc.sh
```diff
@@ -5,17 +5,17 @@ aws ecr get-login-password --region us-west-1 | docker login --username AWS --pa
...
+docker build . -f ./dockerfiles/Dockerfile_linux_glibc -t cli-linux-glibc-builder-image --build-arg GO_VERSION=$(cat .go-version)
 docker container create --name cli-linux-glibc-builder cli-linux-glibc-builder-image
 docker container cp cli-linux-glibc-builder:/go/src/github.com/confluentinc/cli/dist/. ./dist/
 docker container rm cli-linux-glibc-builder
 
 # Build linux/arm64
 architecture=$(uname -m)
 if [ "$architecture" == 'x86_64' ]; then
-  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64_from_amd64 -t cli-linux-glibc-arm64-builder-image
+  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64_from_amd64 -t cli-linux-glibc-arm64-builder-image  --build-arg GO_VERSION=$(cat .go-version)
```
brianstrauch, line 16:
```suggestion
  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64_from_amd64 -t cli-linux-glibc-arm64-builder-image --build-arg GO_VERSION=$(cat .go-version)
```

scripts/build_linux_glibc.sh
```diff
@@ -5,17 +5,17 @@ aws ecr get-login-password --region us-west-1 | docker login --username AWS --pa
...
 else
-  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64 -t cli-linux-glibc-arm64-builder-image
+  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64 -t cli-linux-glibc-arm64-builder-image  --build-arg GO_VERSION=$(cat .go-version)
```
brianstrauch, line 18:
```suggestion
  docker build . -f ./dockerfiles/Dockerfile_linux_glibc_arm64 -t cli-linux-glibc-arm64-builder-image --build-arg GO_VERSION=$(cat .go-version)
```

dockerfiles/Dockerfile_linux_glibc
```diff
@@ -1,5 +1,12 @@
 FROM --platform=linux/amd64 050879227952.dkr.ecr.us-west-1.amazonaws.com/confluentinc/cli-centos-base-amd64:latest
 
+ARG GO_VERSION
+
+RUN wget "https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz" && \
+    sudo tar -C /usr/local -xzf "go$GO_VERSION.linux-amd64.tar.gz"
+
+ENV PATH=${PATH}:/usr/local/go/bin:/root/go/bin
```
brianstrauch, line 8: If we `cd cli` first, then we shouldn't have to pass `$GO_VERSION`. We could just use `cat .go-version` instead.

brianstrauch: Might want to look into cloning `goenv` so the setup is the same as in the CLI?

sgagniere: I don't think we need the setup to be that similar, but that's a good point about the `.go-version`, I'll make that change.

dockerfiles/Dockerfile_linux_glibc_arm64_base
sgagniere: These are the instructions for the base image.

dockerfiles/Dockerfile_linux_glibc
```diff
@@ -2,6 +2,12 @@ FROM --platform=linux/amd64 050879227952.dkr.ecr.us-west-1.amazonaws.com/conflue
 
 COPY . /go/src/github.com/confluentinc/cli/
```
brianstrauch, line 3: What if we changed the path to just `cli/`? Not sure we need the whole thing.

dockerfiles/Dockerfile_linux_glibc_arm64_base
brianstrauch: Ah, we just hadn't uploaded them until now?

brianstrauch (approved):

dockerfiles/Dockerfile_linux_glibc_arm64_base
sgagniere: Yeah these were previously only in ECR as (built) images. I'm still going to build and upload these newer images; adding them to the repo is just a convenience so any of us can more easily update them again if that's ever necessary.

brianstrauch: Do we have a wiki page explaining how to update the images? Or better yet, should we write a make target that does it automatically?

sgagniere: I just added a script that generates them with the correct tags.

I didn't add the login commands (it errors out if you aren't logged in) and it doesn't automatically push them (but it prints out the exact commands to run to push them), because I don't want people to replace the base images on accident. These should not be updated all that frequently, so it's not much of a roadblock.

brianstrauch: How are we supposed to know how to log in, then? I feel like copying and pasting instructions could get old after doing it two or three times, so how about a prompt, like "Upload to ECR? (y/n):"

sgagniere: I was planning to add it to the release doc, but I just put in the login commands. Next commit will have the confirmation prompt.

PR #1761, opened 2023-03-03
AsyncAPI support for primitive AVRO schemas

pagrawal10:
Release Notes
-------------

Bug Fixes

Checklist
---------
   * yes: ok

What
----

- Handling the primitive schema case which was resulting in json.unmarshal error. 
- Added some null checks
- In case of being unable to unmarshal the schema, do not give an error. Instead produce a spec without the message payload.

References
----------

https://confluent.slack.com/archives/CM6PEBEGM/p1677858972060349

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/cmd/asyncapi/account_details.go
```diff
@@ -74,7 +74,8 @@ func (d *accountDetails) getSchemaDetails() error {
 	if err != nil {
 		return err
 	}
-	var unmarshalledSchema map[string]any
+	d.channelDetails.schema = &schema
+	unmarshalledSchema := make(map[string]any)
 	if schema.SchemaType == "" {
 		d.channelDetails.contentType = "application/avro"
```
MuweiHe, line 80: set `schema.SchemaType = AVRO` to be consistent with other places in cli code?

internal/cmd/asyncapi/account_details.go
```diff
@@ -87,11 +88,18 @@ func (d *accountDetails) getSchemaDetails() error {
 	// JSON or Avro Format
 	err = json.Unmarshal([]byte(schema.Schema), &unmarshalledSchema)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal schema: %v", err)
-
+		primitiveTypes := []string{"string", "null", "boolean", "int", "long", "float", "double", "bytes"}
+		for _, primitiveType := range primitiveTypes {
+			if schema.Schema == "\""+primitiveType+"\"" {
```
MuweiHe, line 93: use `sprintf` here

internal/cmd/asyncapi/account_details.go
```diff
@@ -74,24 +74,38 @@ func (d *accountDetails) getSchemaDetails() error {
...
+	case "JSON":
 		d.channelDetails.contentType = "application/json"
-	} else if schema.SchemaType == "PROTOBUF" {
+		break
+	case "AVRO":
+		d.channelDetails.contentType = "application/avro"
+		break
+	case "PROTOBUF":
 		log.CliLogger.Warn("Protobuf not supported.")
 		d.channelDetails.contentType = "PROTOBUF"
```
brianstrauch, line 91: Why do we set the `contentType` if it's not supported?

internal/cmd/asyncapi/account_details.go
```diff
@@ -133,7 +147,9 @@ func (d *accountDetails) buildMessageEntity() *spec.MessageEntity {
 	if d.channelDetails.bindings != nil {
 		(*spec.MessageEntity).WithBindings(entityProducer, d.channelDetails.bindings.messageBinding)
 	}
-	(*spec.MessageEntity).WithPayload(entityProducer, d.channelDetails.unmarshalledSchema)
+	if d.channelDetails.unmarshalledSchema != nil {
+		(*spec.MessageEntity).WithPayload(entityProducer, d.channelDetails.unmarshalledSchema)
```
brianstrauch, line 151: Why do we need to dereference?

brianstrauch (changes requested):
I don't have any context for what this PR is accomplishing. What do you mean by
> Handling the primitive schema case

internal/cmd/asyncapi/account_details.go
pagrawal10: It has no use right now.

pagrawal10:
> 

@brianstrauch , in case of primitive schemas(e.g. `string`) for avro, json.unmarshal() function fails and returns an error. This PR is used to track that. Please refer to https://confluent.slack.com/archives/CM6PEBEGM/p1677858972060349 for further details.

internal/cmd/asyncapi/account_details.go
```diff
@@ -74,24 +74,36 @@ func (d *accountDetails) getSchemaDetails() error {
...
+		break
+	case "PROTOBUF":
+		return fmt.Errorf("protobuf")
 	}
 	// JSON or Avro Format
 	err = json.Unmarshal([]byte(schema.Schema), &unmarshalledSchema)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal schema: %v", err)
-
+		primitiveTypes := []string{"string", "null", "boolean", "int", "long", "float", "double", "bytes"}
```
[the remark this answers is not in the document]
pagrawal10: Primitive schemas in Avro, like, `{"type":"string"}` are same as just "string" and when we call the DefaultApi.GetSchemaByVersion(), it returns `string` which cannot be parsed by json.unmarshal(). Thus, we need to handle those cases manually.

internal/cmd/asyncapi/command_export.go
```diff
@@ -155,7 +155,7 @@ func (c *command) export(cmd *cobra.Command, _ []string) (err error) {
 func (c *command) getChannelDetails(details *accountDetails, flags *flags) error {
 	output.Printf("Adding operation: %s\n", details.channelDetails.currentTopic.GetTopicName())
 	err := details.getSchemaDetails()
-	if details.channelDetails.contentType == "PROTOBUF" {
+	if err.Error() == "protobuf" {
 		log.CliLogger.Info("Protobuf is not supported.")
 		return fmt.Errorf("protobuf")
```
brianstrauch, line 160: Users typically don't run commands without the `-v` flags... we should return a more descriptive error than just "protobuf"
```suggestion
		return errors.New("protobuf is not supported")
```

internal/cmd/asyncapi/account_details.go
```diff
@@ -74,24 +74,36 @@ func (d *accountDetails) getSchemaDetails() error {
...
+		schema.SchemaType = "AVRO"
+	}
+	switch schema.SchemaType {
+	case "JSON":
 		d.channelDetails.contentType = "application/json"
-	} else if schema.SchemaType == "PROTOBUF" {
-		log.CliLogger.Warn("Protobuf not supported.")
-		d.channelDetails.contentType = "PROTOBUF"
-		return nil
+		break
```
brianstrauch, line 85: The `switch` statement doesn't need `break`s in Go
```suggestion
```

internal/cmd/asyncapi/account_details.go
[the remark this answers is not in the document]
brianstrauch: It also seems wise to move this block of code into a separate function and unit test it.

internal/cmd/asyncapi/account_details.go
```diff
@@ -74,24 +74,36 @@ func (d *accountDetails) getSchemaDetails() error {
...
+	case "AVRO":
+		d.channelDetails.contentType = "application/avro"
+		break
+	case "PROTOBUF":
+		return fmt.Errorf("protobuf")
```
brianstrauch, line 90: Let's come up with a more descriptive error here? If I saw this as a user, I would be really confused.

internal/cmd/asyncapi/account_details.go
```diff
@@ -74,24 +74,36 @@ func (d *accountDetails) getSchemaDetails() error {
...
 		return err
 	}
-	var unmarshalledSchema map[string]any
+	d.channelDetails.schema = &schema
+	unmarshalledSchema := make(map[string]any)
 	if schema.SchemaType == "" {
-		d.channelDetails.contentType = "application/avro"
-	} else if schema.SchemaType == "JSON" {
+		schema.SchemaType = "AVRO"
+	}
```
brianstrauch, line 81: Shouldn't we do this by default when we get the schema? (Do this inside of `GetSchemaByVersion()`)

brianstrauch (changes requested):

internal/cmd/asyncapi/account_details.go
brianstrauch: It's a better practice to propagate the user-facing error all the way up the stack!

internal/cmd/asyncapi/command_export.go
```diff
@@ -155,9 +155,9 @@ func (c *command) export(cmd *cobra.Command, _ []string) (err error) {
 func (c *command) getChannelDetails(details *accountDetails, flags *flags) error {
 	output.Printf("Adding operation: %s\n", details.channelDetails.currentTopic.GetTopicName())
 	err := details.getSchemaDetails()
-	if err.Error() == "protobuf" {
+	if err == errors.New("protobuf is not supported") {
 		log.CliLogger.Info("Protobuf is not supported.")
-		return fmt.Errorf("protobuf")
+		return errors.New("protobuf is not supported")
```
brianstrauch, line 160: Don't you think that logging the error and returning the error is redundant?
```suggestion
		return err
```

internal/cmd/asyncapi/command_export_test.go
```diff
@@ -302,8 +303,15 @@ func TestGetChannelDetails(t *testing.T) {
 	schema, _, _ = details.srClient.DefaultApi.GetSchemaByVersion(*new(context.Context), "subject2", "1", nil)
 	details.channelDetails.schema = &schema
 	err = c.getChannelDetails(details, flags)
-	require.Error(t, err, fmt.Errorf("protobuf"))
+	require.Error(t, err, errors.New("protobuf is not supported"))
+}
 
+func TestHandlePrimitiveSchemas(t *testing.T) {
+	unmarshalledSchema, err := handlePrimitiveSchemas(fmt.Sprintf("\"string\""), fmt.Errorf("unable to unmarshal schema"))
```
brianstrauch, line 310:
```suggestion
	unmarshalledSchema, err := handlePrimitiveSchemas(`"string"`, fmt.Errorf("unable to unmarshal schema"))
```

internal/cmd/asyncapi/account_details.go
pagrawal10: Refer to this thread: https://confluent.slack.com/archives/CM6PEBEGM/p1677785572528619?thread_ts=1677785042.749209&cid=CM6PEBEGM

brianstrauch (approved):
+1 pending CI

brianstrauch:
@pagrawal10 FYI, there's a new "Release Notes" section in the PR description template. I did this one for you, but please fill it out in the future if your PR has a user-facing change.

pagrawal10:
@brianstrauch , thanks for informing! Will keep it in mind going forward.

PR #1762, opened 2023-03-04
Use the same struct for `schema-registry` commands

brianstrauch:
Checklist
---------
   * yes: ok

What
----
I started working on adding autocompletion for SR flags, but quickly realized autocomplete functions can't be shared between subcommands unless we combine all of the SR commands under a single struct (this is what we do for most other commands).


internal/cmd/schema-registry/command_config.go
```diff
@@ -1,33 +1,23 @@
...
 	}
 
-	c := &configCommand{srClient: srClient}
-
 	if cfg.IsCloudLogin() {
-		c.AuthenticatedStateFlagCommand = pcmd.NewAuthenticatedStateFlagCommand(cmd, prerunner)
-		c.AddCommand(c.newDescribeCommand())
+		cmd.AddCommand(c.newConfigDescribeCommand())
 	} else {
-		c.AuthenticatedStateFlagCommand = pcmd.NewAuthenticatedWithMDSStateFlagCommand(cmd, prerunner)
```
MuweiHe, line 29: this no longer needed?

internal/cmd/schema-registry/command_config_describe_onprem.go
```diff
@@ -10,12 +10,12 @@ import (
...
 
-func (c *configCommand) newDescribeCommandOnPrem() *cobra.Command {
+func (c *command) newConfigDescribeCommandOnPrem() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:         "describe",
 		Short:       "Describe top-level or subject-level schema compatibility.",
-		Args:        cobra.MaximumNArgs(0),
-		RunE:        c.onPremDescribe,
+		Args:        cobra.NoArgs,
+		RunE:        c.configDescribeOnPrem,
```
MuweiHe, line 18: other commands like kafka are called `onpremOperation` as well. i think it's easier to read with `onprem` prefix.

internal/cmd/schema-registry/command_config.go
brianstrauch: It's only called once by the top-level `confluent schema-registry` command

internal/cmd/schema-registry/command_config_describe_onprem.go
brianstrauch: Can you give me an example? I think all of the function names should be suffixed with "OnPrem".

MuweiHe: https://github.com/confluentinc/cli/blob/ac394e7f40f93c2ee424d7e8acf26269d5e12d4b/internal/cmd/kafka/command_cluster_list_onprem.go#L31

brianstrauch: Thanks, I'll standardize these soon!

brianstrauch: (in a different PR)

MuweiHe (approved):

PR #1768, opened 2023-03-06
Revert cc-cli-service release process

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Condensed the cc-cli-service release process into `make update-db` and `make promote`. Kubernetes will occasionally re-deploy the service, so we need to make sure the service gets built with the latest migration files, and we can no longer use `[ci-skip]`.


sgagniere (approved):
😢

PR #1769, opened 2023-03-06
Disable plugins by default on Windows

brianstrauch:
Release Notes
-------------
Breaking Changes
- Disable plugins by default on Windows, but allow manual re-enablement in ~/.confluent/config.json

Checklist
---------
   * yes: ok

What
----
Plugin search can be prohibitively slow on Windows. Users with a long `$PATH` often report search times of around 20s, even when running `confluent version`. This is unacceptable UX, so plugins should be disabled by default on Windows. A user can enable them in ~/.confluent/config.json and they won't be disabled again in the lifetime of the configuration file.

Test & Review
-------------
Removed the `runtime.OS == "windows"` condition and tested manually.

### DIFF
--- a/internal/pkg/config/load/load_latest.go
+++ b/internal/pkg/config/load/load_latest.go
@@ -1,6 +1,8 @@
 package load
 
 import (
+	"runtime"
+
 	"github.com/confluentinc/cli/internal/pkg/config"
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 )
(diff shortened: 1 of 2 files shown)


internal/pkg/config/load/load_latest.go
```diff
@@ -35,6 +37,12 @@ func loadLatestNoErr(latestCfg *v1.Config, cfgIndex int) (config.Config, error)
 func migrateToLatest(cfg config.Config) *v1.Config {
 	switch cfg := cfg.(type) {
 	case *v1.Config:
+		// On Windows, plugin search is prohibitively slow for users with a long $PATH, so plugins should be disabled by default.
+		if runtime.GOOS == "windows" && !cfg.DisablePluginsOnce {
+			cfg.DisablePlugins = true
+			cfg.DisablePluginsOnce = true
```
sgagniere, line 43: The only thing I would suggest is renaming this one. If someone on Windows runs `plugin list`, sees the suggestion to `Set "disable_plugins": false in ~/.confluent/config.json.`, and then sees `DisablePlugins` and `DisablePluginsOnce`, they might get confused and set both to `false`. Then this block would put them right back where they started.

sgagniere (approved):

internal/pkg/config/load/load_latest.go
brianstrauch: Hmm, any suggestions for what to call it?

sgagniere: I can't think of anything good. Maybe as an alternative the suggestion message can be updated to explicitly instruct users not to change the value of the `DisablePluginsOnce` parameter.

PR #1771, opened 2023-03-07
Chown /etc so cli can read/write /etc/machine-id on Alpine for encryption

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix ``cipher: message authentication failed`` issue on Alpine image

Checklist
---------
   * yes: ok

What
----

got multiple reports saying alpine image is not working because of `cipher authentication error`. Because `/etc` was a read-only dir and cli generates/reads machine id on `/etc/machine-id`, encryption was failing because no actual machine id is being used. chown this dir to solve the read/write issue.

References
----------

Test & Review
-------------

Tested alpine image that commands and `--save` are working.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -26,6 +26,7 @@ RUN curl -L --http1.1 https://cnfl.io/cli | sh -s -- -b /bin
 
 RUN chmod +x /bin/confluent
 RUN chown $USER:$USER /bin/confluent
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
Make sure you write release notes!

PR #1773, opened 2023-03-08
Add more linters

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Added a few basic linters, `dupword,misspell,prealloc,usestdlibvars,whitespace`, and fixed all of their issues.

Test & Review
-------------
`make lint-go` passes


internal/cmd/price/command_list.go
```diff
@@ -255,20 +255,14 @@ func filterTable(table map[string]*ccloudv1.UnitPrices, filters []string, metric
 	return filteredTable, nil
 }
 
-func getKeys(m map[string]string) []string {
-	var slice []string
-	for key := range m {
-		slice = append(slice, key)
-	}
+func getSortedKeys(m map[string]string) []string {
```
sgagniere, line 258: I think we should put this in utils too even though it's currently only used here, because it may be useful elsewhere later on and someone could end up rewriting it after not finding it in utils.

internal/pkg/utils/slice.go
```diff
@@ -0,0 +1,27 @@
...
+	for i, x := range haystack {
+		if x == needle {
+			return append(haystack[:i], haystack[i+1:]...)
+		}
+	}
+	return haystack
+}
+
+func RemoveDuplicates(slice []string) []string {
+	m := make(map[string]bool)
```
sgagniere, line 22: Unfortunate that we can't use set here because of circular imports

sgagniere (approved):
Thanks +1

internal/pkg/utils/slice.go
brianstrauch: Yeah, I was thinking about this too... I'll make a generic `types` package!

brianstrauch: (To avoid the circular import error, which I think we would get)

PR #1774, opened 2023-03-09
Add more linters

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Clean up lots of dead code and improve tests!


sgagniere (approved):
Nice

PR #1777, opened 2023-03-10
Make SR ID and KSQL optional for pipeline create command

guruz09:
Release Notes
-------------
New Features
- Add ``--ksql-cluster`` and ``--update-schema-registry`` to ``confluent pipeline update``

Checklist
---------
   * No: DO NOT MERGE until the required features are live in prod  

What
----

Stream Designer team is making ksql cluster and sr cluster as optional while creating the pipeline. Hence we want to mark these are optional and let the user update these in `pipeline update` command.

References
----------

* [PR](https://github.com/confluentinc/api/pull/947) for API changes
* [PR](https://github.com/confluentinc/cc-pipeline-service/pull/1083) for SD changes

Test & Review
-------------
* got a green build and all the tests are passing. Will test the feature manually once the SD backend code is merged.


internal/cmd/pipeline/command_update.go
```diff
@@ -44,6 +53,7 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 			"If <secret-value> is empty, the named secret will be removed from Stream Designer.")
 	}
 	cmd.Flags().Bool("activation-privilege", true, "Grant or revoke the privilege to activate this pipeline.")
+	cmd.Flags().Bool("update-schema-registry", true, "Update the pipeline with the latest schema registry cluster.")
```
brianstrauch, line 56: Is the use case that customers want to be able to update their pipeline without updating their Schema Registry cluster? Not sure why we'd want that. Maybe you meant the default to be `false` here?

internal/cmd/pipeline/command_update.go
```diff
@@ -31,11 +31,20 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 				Text: `Revoke privilege to activate Stream Designer pipeline "pipe-12345".`,
 				Code: `confluent pipeline update pipe-12345 --activation-privilege false`,
 			},
+			examples.Example{
+				Text: `Request to update Stream Designer pipeline "pipe-12345", with new ksql cluster Id.`,
```
brianstrauch, line 35:
```suggestion
				Text: `Update Stream Designer pipeline "pipe-12345" with KSQL cluster ID "lksqlc-123456".`,
```

internal/cmd/pipeline/command_update.go
```diff
@@ -31,11 +31,20 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
...
+				Code: `confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456.`,
+			},
+			examples.Example{
+				Text: `Request to update Stream Designer pipeline "pipe-12345", with new schema registry cluster Id.`,
+				Code: `confluent pipeline update pipe-12345 --update-schema-registry`,
```
brianstrauch, line 40: Why are we passing the flag in this example if its default value is `true`? Passing this flag does nothing.

internal/cmd/pipeline/command_update.go
```diff
@@ -56,17 +66,18 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 
 	cluster, err := c.Context.GetKafkaClusterForCommand()
 	if err != nil {
 		return err
 	}
 
-	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && !cmd.Flags().Changed("activation-privilege") {
-		return fmt.Errorf("one of the update options must be provided: --name, --description, --sql-file, --secret, --activation-privilege")
+	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && !cmd.Flags().Changed("activation-privilege") && !cmd.Flags().Changed("update-schema-registry") {
+		return fmt.Errorf("one of the update options must be provided: --name, --description, --sql-file, --secret, --activation-privilege, --update-schema-registry")
```
brianstrauch, line 77: What about `--ksql-cluster`?

brianstrauch (changes requested):

internal/cmd/pipeline/command_create.go
```diff
@@ -65,9 +66,17 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
...
 	if err != nil {
-		return err
+		if !strings.Contains(err.Error(), "Schema Registry not enabled") {
+			// ignore if the SR is not enabled
+			return err
+		}
+	}
+
+	if srCluster != nil {
+		srId = srCluster.Id
```
brianstrauch, line 79: Is there a `srCluster.GetId()` function we can call here to handle the `nil` check for us?

internal/cmd/pipeline/command_update.go
brianstrauch: Also, can we put backticks around each of the flags?

guruz09 (commented):
Thank you @brianstrauch for the comments. I have addressed them and published a new revision. Please have another pass. Thanks.

internal/cmd/pipeline/command_create.go
```diff
@@ -68,14 +68,24 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
...
+			return err
+		}
 	}
 
 	// validate sr id
+	srId := ""
 	srCluster, err := c.Config.Context().SchemaRegistryCluster(cmd)
 	if err != nil {
-		return err
+		if !strings.Contains(err.Error(), "Schema Registry not enabled") {
```
MuweiHe, line 81: if sr is not enabled, it will continue to create?

internal/cmd/pipeline/command_update.go
```diff
@@ -60,17 +70,19 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
 	description, _ := cmd.Flags().GetString("description")
 	sqlFile, _ := cmd.Flags().GetString("sql-file")
 	secrets, _ := cmd.Flags().GetStringArray("secret")
+	ksqlCluster, _ := cmd.Flags().GetString("ksql-cluster")
```
MuweiHe, line 73: why are these err all being ignored？

internal/cmd/pipeline/command_update.go
```diff
@@ -32,12 +32,12 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
...
 			examples.Example{
-				Text: `Request to update Stream Designer pipeline "pipe-12345", with new ksql cluster Id.`,
+				Text: `Update Stream Designer pipeline "pipe-12345" with KSQL cluster ID "lksqlc-123456".`,
 				Code: `confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456.`,
 			},
 			examples.Example{
-				Text: `Request to update Stream Designer pipeline "pipe-12345", with new schema registry cluster Id.`,
-				Code: `confluent pipeline update pipe-12345 --update-schema-registry`,
+				Text: `Update Stream Designer pipeline "pipe-12345" with new schema registry cluster Id.`,
+				Code: `confluent pipeline update pipe-12345 --update-schema-registry true`,
```
brianstrauch, line 40: This is redundant. The existence of this flag implies `true`.
```suggestion
				Code: `confluent pipeline update pipe-12345 --update-schema-registry`,
```

test/fixtures/output/pipeline/update-help.golden
```diff
@@ -16,13 +16,13 @@ Revoke privilege to activate Stream Designer pipeline "pipe-12345".
 
   $ confluent pipeline update pipe-12345 --activation-privilege false
 
-Request to update Stream Designer pipeline "pipe-12345", with new ksql cluster Id.
+Update Stream Designer pipeline "pipe-12345" with KSQL cluster ID "lksqlc-123456".
 
   $ confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456.
 
-Request to update Stream Designer pipeline "pipe-12345", with new schema registry cluster Id.
+Update Stream Designer pipeline "pipe-12345" with new schema registry cluster Id.
```
brianstrauch, line 23:
```suggestion
Update Stream Designer pipeline "pipe-12345" with new Schema Registry cluster ID.
```

internal/cmd/pipeline/command_create.go
guruz09: Yes, thats the new feature.

internal/cmd/pipeline/command_update.go
guruz09: We have empty check later. What do you suggest here?

brianstrauch:
```suggestion
	ksqlCluster, err := cmd.Flags().GetString("ksql-cluster")
	if err != nil {
		return err
	}
```

guruz09 (commented):
Thanks for the comments. Updated the PR addressing your comments. Please have another pass. Thank you.

internal/cmd/pipeline/command_update.go
```diff
@@ -36,8 +37,8 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 				Code: `confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456.`,
 			},
 			examples.Example{
-				Text: `Update Stream Designer pipeline "pipe-12345" with new schema registry cluster Id.`,
-				Code: `confluent pipeline update pipe-12345 --update-schema-registry true`,
+				Text: `Update Stream Designer pipeline "pipe-12345" with new Schema Registry cluster ID.`,
```
brianstrauch, line 40:
```suggestion
				Text: `Update Stream Designer pipeline "pipe-12345" with a new Schema Registry cluster ID.`,
```

internal/cmd/pipeline/command_update.go
```diff
@@ -36,8 +37,8 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
...
+				Code: `confluent pipeline update pipe-12345 --update-schema-registry`,
```
brianstrauch, line 41:
```suggestion
				Code: "confluent pipeline update pipe-12345 --update-schema-registry",
```

internal/cmd/pipeline/command_create.go
```diff
@@ -68,14 +88,19 @@ func (c *command) create(cmd *cobra.Command, _ []string) error {
...
+		}
 	}
 
 	// validate sr id
-	srCluster, err := c.Config.Context().SchemaRegistryCluster(cmd)
+	srCluster, err := c.Context.FetchSchemaRegistryByEnvironmentId(context.Background(), environmentId)
 	if err != nil {
-		return err
+		if !strings.Contains(err.Error(), "Schema Registry not enabled") {
+			// ignore if the SR is not enabled
```
brianstrauch, line 101: This comment seems a bit misplaced since it's inside the `if`.

internal/cmd/pipeline/command_update.go
```diff
@@ -31,11 +32,20 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 				Text: `Revoke privilege to activate Stream Designer pipeline "pipe-12345".`,
 				Code: `confluent pipeline update pipe-12345 --activation-privilege false`,
 			},
+			examples.Example{
+				Text: `Update Stream Designer pipeline "pipe-12345" with KSQL cluster ID "lksqlc-123456".`,
+				Code: `confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456.`,
```
brianstrauch, line 37: Code snippets shouldn't end with periods
```suggestion
				Code: "confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456",
```

internal/cmd/pipeline/command_update.go
```diff
@@ -44,6 +54,7 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 			"If <secret-value> is empty, the named secret will be removed from Stream Designer.")
 	}
 	cmd.Flags().Bool("activation-privilege", true, "Grant or revoke the privilege to activate this pipeline.")
+	cmd.Flags().Bool("update-schema-registry", false, "Update the pipeline with the latest schema registry cluster.")
```
brianstrauch, line 57:
```suggestion
	cmd.Flags().Bool("update-schema-registry", false, "Update the pipeline with the latest Schema Registry cluster.")
```

internal/cmd/pipeline/command_update.go
brianstrauch: Yes please!

internal/cmd/pipeline/command_update.go
```diff
@@ -60,17 +71,19 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 	cluster, err := c.Context.GetKafkaClusterForCommand()
 	if err != nil {
 		return err
 	}
 
-	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && !cmd.Flags().Changed("activation-privilege") {
-		return fmt.Errorf("one of the update options must be provided: --name, --description, --sql-file, --secret, --activation-privilege")
+	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && !cmd.Flags().Changed("activation-privilege") &&
+		ksqlCluster == "" && !cmd.Flags().Changed("update-schema-registry") {
+		return fmt.Errorf("one of the update options must be provided: `--name`, `--description`, `--sql-file`, `--secret`, `--activation-privilege`, `--update-schema-registry`")
```
brianstrauch, line 83: Did you forget `--ksql-cluster` in this list?

brianstrauch: Maybe we can also sort this list alphabetically since it's getting really long.

internal/cmd/pipeline/command_update.go
```diff
@@ -103,6 +116,24 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 
+	if ksqlCluster != "" {
+		if _, err := c.V2Client.DescribeKsqlCluster(ksqlCluster, environmentId); err != nil {
+			return err
+		}
+		updatePipeline.Spec.SetKsqlCluster(streamdesignerv1.ObjectReference{Id: ksqlCluster})
+	}
+
+	if cmd.Flags().Changed("update-schema-registry") {
+		updateSRCluster, _ := cmd.Flags().GetBool("update-schema-registry")
```
brianstrauch, line 127: Please try to keep a 1-1 mapping between flag names and variables. It makes the code a bit easier to read.
```suggestion
		updateSchemaRegistry, err := cmd.Flags().GetBool("update-schema-registry")
		if err != nil {
			return err
		}
```

internal/pkg/ccloudv2/stream-designer.go
```diff
@@ -55,17 +55,23 @@ func (c *Client) executeListPipelines(envId, clusterId, pageToken string) (strea
 func (c *Client) CreatePipeline(envId, clusterId, name, description, sourceCode string, secretMappings *map[string]string, ksqlId, srClusterId string) (streamdesignerv1.SdV1Pipeline, error) {
```
brianstrauch, line 55: It might be a better idea to construct the pipeline object outside of this function and pass it in (instead of 8 arguments!)

test/fixtures/output/pipeline/create-help.golden
```diff
@@ -9,7 +9,7 @@ Create a new Stream Designer pipeline
   $ confluent pipeline create --name test-pipeline --ksql-cluster lksqlc-12345 --description "this is a test pipeline"
 
 Flags:
-      --ksql-cluster string   REQUIRED: KSQL cluster for the pipeline.
+      --ksql-cluster string   KSQL cluster for the pipeline.
```
brianstrauch, line 12: I think this belongs lower in the list.

brianstrauch (changes requested):

internal/cmd/pipeline/command_update.go
xiangxin72, line 57: since it's a bool flag, it makes it feel like we can do `--update-schema-registry=false` to `unset` the schema-registry, but actually it won't.

In general, since both ksql and sr are not optional at creation and can be updated later, I feel we should also consider the story of `unset`
* if we can unset ksql/sr, then this update command probably need new flag to indicate `--remove-ksql-cluster` and `--remove-schema-registry`
* if we do not want to allow unset after it's set once, then we may have some communication work to do (since feels a bit odd why such restriction exists)

guruz09: YEs, i had a chat about that with Randall on `unset`. I don't think our APIs can support that at this point of time. Maybe we can handle that separately?

brianstrauch: When the API supports it, we can check to see if the user has modified the `--update-schema-registry` flag, and set/unset depending on if the value is true/false.

brianstrauch: We may want to consider renaming to `--set-schema-registry`?

xiangxin72: before API supports it, then `--update-schema-registry=false` or `--set-schema-registry=false` will be no-op? How should we communicate that to make it clearer?

brianstrauch: I don't expect any customers to be passing `--set-schema-registry false` before this change is made, so I'd be ok with making the change I suggested above as a breaking change. (No need to communicate anything, but let's wait until the API change is made before making this change in the CLI?)

guruz09: Sorry, maybe my comment was confusing. We dont have a mechanism to unset any of the configs. Ex: once the user sets a ksql cluster for a pipeline they can only change it to a new ksql cluster but cant remove it. This is the behavior for all of the metadata configs. so when when user calls with  `--update-schema-registry=false`, we just don't update the pipeline with new/updated sr cluster.

guruz09 (commented):
Thanks for the comments @brianstrauch. Addressed it in the. next comit.

internal/cmd/pipeline/command_update.go
xiangxin72: Will we ever support it someday? It feels odd to me why we won't, so I assume someday we will want to either update/refresh ksql and SR, or disable/remove ksql and sr.

So we are just deciding what to start with for now that can be easily adapted later. If we can ignore users passing us `false` for now (which is no op today) and later support it (which will unset the SR). Then it feels more accurate to use `--enable-schema-registry` than `--update-schema-registry` or `--set-schema-registry`? The later doesn't feel like a bool flag. (With SR, users don't provide the `lsrc-id` so `update` or `set` doesn't feel very natual)

internal/cmd/pipeline/command_update.go
```diff
@@ -31,11 +32,20 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
 				Text: `Revoke privilege to activate Stream Designer pipeline "pipe-12345".`,
 				Code: `confluent pipeline update pipe-12345 --activation-privilege false`,
 			},
+			examples.Example{
+				Text: `Update Stream Designer pipeline "pipe-12345" with KSQL cluster ID "lksqlc-123456"`,
+				Code: `confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456`,
```
brianstrauch, line 37:
```suggestion
				Code: "confluent pipeline update pipe-12345 --ksql-cluster lksqlc-123456",
```

internal/cmd/pipeline/command_update.go
```diff
@@ -56,21 +67,50 @@ func (c *command) newUpdateCommand(enableSourceCode bool) *cobra.Command {
...
 	if err != nil {
 		return err
 	}
 
-	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && !cmd.Flags().Changed("activation-privilege") {
-		return fmt.Errorf("one of the update options must be provided: --name, --description, --sql-file, --secret, --activation-privilege")
+	if name == "" && description == "" && sqlFile == "" && len(secrets) == 0 && ksqlCluster == "" &&
+		!cmd.Flags().Changed("activation-privilege") && !cmd.Flags().Changed("update-schema-registry") {
+		return fmt.Errorf("one of the update options must be provided:" +
+			"` --name`," +
```
brianstrauch, line 103:
```suggestion
			" `--name`," +
```

internal/cmd/pipeline/command_update.go
guruz09: As discussed offline, We will add a new flag to unset the sr cluster when our backend API supports that and we can continue with the current approach to update the cluster with `--update-schema-registry` flag.

internal/cmd/pipeline/command_create.go
```diff
@@ -1,10 +1,13 @@
 package pipeline
 
 import (
+	"context"
 	"fmt"
+	streamdesignerv1 "github.com/confluentinc/ccloud-sdk-go-v2/stream-designer/v1"
```
brianstrauch, line 6: Please make sure your imports are sorted. `make lint` should catch this.

PR #1778, opened 2023-03-13
AsyncAPI fixing Protobuf Errors

pagrawal10:
Release Notes
-------------

Bug Fixes:
- Skip Protobuf schemas in `confluent asyncapi export` 

Checklist
---------
  Yes

What
----

Fixes the error message comparison in case of Protobuf schemas and produces an AsyncAPI spec with skipping all the protobuf topics. 

References
----------

Follow-up PR to https://github.com/confluentinc/cli/pull/1761

Test & Review
-------------

Tested locally against a Devel cluster

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -63,6 +63,7 @@ type flags struct {
 
 // messageOffset is 5, as the schema ID is stored at the [1:5] bytes of a message as meta info (when valid)
 const messageOffset int = 5
+const protobufErrorMessage string = "protobuf is not supported"
 
 func newExportCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


internal/cmd/asyncapi/command_export.go
```diff
@@ -117,11 +118,11 @@ func (c *command) export(cmd *cobra.Command, _ []string) (err error) {
 					currentSubject: subject,
 				}
 				err := c.getChannelDetails(accountDetails, flags)
-				if err != nil && err == errors.New("protobuf is not supported") {
-					log.CliLogger.Info("Protobuf is not supported.")
-					continue
-				}
 				if err != nil {
+					if err.Error() == protobufErrorMessage {
+						log.CliLogger.Info("Protobuf is not supported.")
```
brianstrauch, line 123: Why not log the error? It's the same text.

pagrawal10: Using log.CliLogger.log() ?

brianstrauch: Yeah: `log.CliLogger.Info(err.Error())`

brianstrauch (approved):

PR #1786, opened 2023-03-13
Use `cmd` where possible, remove unnecessary command structs

brianstrauch:
Checklist
---------
   * yes: ok

What
----
I realized there were a few structs that were not being used- removed those and switched to using `cmd` instead of `c` where possible (easier to detect these "dead" structs in the future)

Test & Review
-------------
Tests still pass

### DIFF
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -69,20 +69,23 @@ func newExportCommand(prerunner pcmd.PreRunner) *cobra.Command {
 		Use:   "export",
 		Short: "Create an AsyncAPI specification for a Kafka cluster.",
 	}
+
 	c := &command{AuthenticatedStateFlagCommand: pcmd.NewAuthenticatedStateFlagCommand(cmd, prerunner)}
-	c.RunE = c.export
-	c.Flags().String("file", "asyncapi-spec.yaml", "Output file name.")
-	c.Flags().String("group-id", "consumerApplication", "Consumer Group ID for getting messages.")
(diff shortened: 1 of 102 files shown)


internal/cmd/asyncapi/command_export.go
```diff
@@ -69,20 +69,23 @@ func newExportCommand(prerunner pcmd.PreRunner) *cobra.Command {
...
-	c.Flags().String("group-id", "consumerApplication", "Consumer Group ID for getting messages.")
-	c.Flags().Bool("consume-examples", false, "Consume messages from topics for populating examples.")
-	c.Flags().String("spec-version", "1.0.0", "Version number of the output file.")
-	c.Flags().String("kafka-api-key", "", "Kafka cluster API key.")
-	c.Flags().String("schema-registry-api-key", "", "API key for Schema Registry.")
-	c.Flags().String("schema-registry-api-secret", "", "API secret for Schema Registry.")
-	c.Flags().String("schema-context", "default", "Use a specific schema context.")
+	cmd.RunE = c.export
+
+	cmd.Flags().String("file", "asyncapi-spec.yaml", "Output file name.")
```
MuweiHe, line 76: are they different, setting flags to c or cmd?

MuweiHe (approved):

internal/cmd/asyncapi/command_export.go
brianstrauch: `cmd` is "inside of" `c`, so they're the same, but this way will make it easier to detect the "dead" structs I mentioned in the description.

PR #1787, opened 2023-03-14
Add `gci` linter

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Sort imports with [gci](https://github.com/daixiang0/gci). Order imports by:
- standard imports
- everything else
- imports in confluentinc
- imports in confluentinc/cli


sgagniere (approved):
💯

PR #1788, opened 2023-03-14
Add more linters

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Add linters, most of which are code formatters.


internal/pkg/secret/encryption_engine.go
```diff
@@ -103,20 +103,7 @@ func (c *EncryptEngineImpl) UnwrapDataKey(dataKey string, iv string, algo string
 	return base64.StdEncoding.DecodeString(dataKeyEnc)
 }
 
-func (c *EncryptEngineImpl) Encrypt(plainText string, key []byte) (data string, ivStr string, err error) {
-	defer func() {
```
sgagniere, line 107: This doesn't look like a linter change; was removing this block intended?

test/kafka_test.go
```diff
@@ -375,8 +375,10 @@ func (s *CLITestSuite) TestKafkaTopicCreate() {
 		{args: fmt.Sprintf("kafka topic create topic-X --url %s --config retention.ms=as --no-authentication", kafkaRestURL), contains: "Error: REST request failed: Invalid value as for configuration retention.ms: Not a number of type LONG (40002)\n", exitCode: 1, name: "creating topic with correct key incorrect config value should fail"},
 		// Success
 		{args: fmt.Sprintf("kafka topic create topic-X --url %s --no-authentication", kafkaRestURL), fixture: "kafka/topic/create-topic-success.golden", name: "correct URL with default params (part 6, repl 3, no configs) should create successfully"},
-		{args: fmt.Sprintf("kafka topic create topic-X --url %s --partitions 7 --replication-factor 2 --config retention.ms=100000,compression.type=gzip --no-authentication", kafkaRestURL),
-			fixture: "kafka/topic/create-topic-success.golden", name: "correct URL with valid optional params should create successfully"},
+		{
+			args:    fmt.Sprintf("kafka topic create topic-X --url %s --partitions 7 --replication-factor 2 --config retention.ms=100000,compression.type=gzip --no-authentication", kafkaRestURL),
```
sgagniere, line 379: Seems a bit awkward that this one isn't single line but the ones around it are.

sgagniere (commented):
Just one concern

internal/pkg/secret/encryption_engine.go
brianstrauch: If you think about it, it does nothing, right? I suppose it might catch a `panic()` but it doesn't look like there's anything that could `panic()`...

test/kafka_test.go
brianstrauch: Yeah agreed... it's really long though. Let me try changing it back.

test/kafka_test.go
```diff
@@ -375,8 +375,10 @@ func (s *CLITestSuite) TestKafkaTopicCreate() {
...
+			fixture: "kafka/topic/create-topic-success.golden", name: "correct URL with valid optional params should create successfully",
+		},
```
brianstrauch, line 381:
```suggestion
		{args: fmt.Sprintf("kafka topic create topic-X --url %s --partitions 7 --replication-factor 2 --config retention.ms=100000,compression.type=gzip --no-authentication", kafkaRestURL), fixture: "kafka/topic/create-topic-success.golden", name: "correct URL with valid optional params should create successfully"},
```

internal/pkg/secret/encryption_engine.go
brianstrauch: Hey @tadsul, looks like you wrote this code. Do you remember why you chose to use `recover()` here? Were we really seeing a `panic()`?

brianstrauch: Hmm... I wonder if since `err` is defined in the return type, it's being returned from inside the `defer()`...

brianstrauch: Nope, this doesn't work: https://play.golang.com/p/B0sKK_uwiWn

sgagniere: Yeah sounds good. I tried looking into the history to see if there was some motivation for it and there was a comment asking about tests for those errors but no response.

sgagniere (approved):
+1

PR #1789, opened 2023-03-14
[CLI-2065] Delete multiple resources with one command

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add support for multiple arguments to `delete` commands

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
`delete` commands taking an id argument can now take multiple arguments. The current behavior is the following:
- The command makes a GET request (calling either Describe or List depending on the # of args) to validate existence of the resource. By default, if not all ids are accessible for this user an error is returned.
- Resources are deleted in a loop, and errors from this step are collected and reported at the end. In other words, the CLI will attempt to delete each resource even if a previous deletion failed.

References
----------

Test & Review
-------------
Manual testing
Added multiple integration tests

Open Questions / Follow-ups
---------------------------


internal/cmd/api-key/command_delete.go
```diff
@@ -10,13 +10,15 @@ import (
...
+	"github.com/confluentinc/cli/internal/pkg/types"
+	"github.com/confluentinc/cli/internal/pkg/utils"
 )
 
 func (c *command) newDeleteCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "delete <api-key>",
-		Short:             "Delete an API key.",
-		Args:              cobra.ExactArgs(1),
+		Use:               "delete <api-key-1> [api-key-2] ... [api-key-N]",
```
brianstrauch, line 19: Nit: We use lower-kebab-case for all of our commands and flags, so we should probably do the same for args.
```suggestion
		Use:               "delete <api-key-1> [api-key-2] ... [api-key-n]",
```

internal/cmd/api-key/command_delete.go
```diff
@@ -28,18 +30,67 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+	if ok, err := form.ConfirmDeletionYesNo(cmd, resource.ApiKey, args); err != nil || !ok {
+		return err
+	}
+
+	var errs error
+	for _, id := range args {
+		if httpResp, err := c.V2Client.DeleteApiKey(id); err != nil {
+			errs = errors.Join(errs, errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, id, httpResp))
+		} else {
+			output.Printf(errors.DeletedResourceMsg, resource.ApiKey, id)
```
brianstrauch, line 47: If the user deletes 10 resources, instead of having 10 separate lines, how about:
```
Deleted API keys "12345", "67890", ... and "13579".
```

internal/cmd/api-key/command_delete.go
```diff
@@ -28,18 +30,67 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+	// Single
+	if len(args) == 1 {
+		if _, _, err := c.V2Client.GetApiKey(args[0]); err != nil {
+			return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.NotFoundErrorMsg, resource.ApiKey, args[0]), fmt.Sprintf(errors.DeleteNotFoundSuggestions, resource.ApiKey))
+		}
+		return nil
+	}
+
+	// Multiple
+	apiKeys, err := c.V2Client.ListApiKeys("", "")
```
brianstrauch, line 70: What if the user deletes 2 API keys, but they have 1000? This will take a while...

internal/cmd/api-key/command_delete.go
```diff
@@ -28,18 +30,67 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+	set := types.NewSet()
+	for _, apiKey := range apiKeys {
+		set.Add(apiKey.GetId())
+	}
+
+	validArgs, invalidArgs := set.IntersectionAndDifference(args)
+	if force, err := cmd.Flags().GetBool("force"); err != nil {
+		return err
+	} else if force && len(invalidArgs) > 0 {
+		args = validArgs
```
brianstrauch, line 84: I'm not 100% convinced about this... `--force` is used to circumvent the prompt, and in my opinion that's what it should be used for here to. We're using it to circumvent invalid flags here, so IMO that should be a separate flag if that's something we want to do.

internal/cmd/api-key/command_delete.go
brianstrauch: (it would use the same `plural()` logic as the "not found" error message)

internal/cmd/api-key/command_delete.go
```diff
@@ -28,18 +30,67 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+		return nil
+	}
+
+	invalidArgsStr := utils.ArrayToCommaDelimitedStringWithAnd(invalidArgs)
+	if len(invalidArgs) == 1 {
+		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.NotFoundErrorMsg, resource.ApiKey, invalidArgsStr), fmt.Sprintf(errors.DeleteNotFoundSuggestions, resource.ApiKey))
+	} else if len(invalidArgs) > 1 {
+		return errors.NewErrorWithSuggestions(fmt.Sprintf(errors.NotFoundErrorMsg, utils.Plural(resource.ApiKey), invalidArgsStr), fmt.Sprintf(errors.DeleteNotFoundSuggestions, resource.ApiKey))
```
brianstrauch, line 92: Can we move the `Plural()` function into the `resource` package? We might have a few edge cases (-es suffix instead of -s suffix) and it would be nice if the resource names and their plurals lived in the same file.

internal/pkg/errors/errors.go
```diff
@@ -1,21 +1,27 @@
...
+	return perrors.Wrapf(err, fmt, args...)
 }
 
 func Errorf(fmt string, args ...any) error {
-	return errors.Errorf(fmt, args...)
+	return perrors.Errorf(fmt, args...)
+}
+
+func Join(errs ...error) error {
+	return errors.Join(errs...)
```
brianstrauch, line 26: So `errors.Join()` is not part of `github.com/pkg/errors`?

internal/pkg/errors/strings.go
```diff
@@ -121,14 +121,15 @@ const (
...
-	DeleteACLsConfirmMsg          = "Are you sure you want to delete the ACLs corresponding to these parameters?"
-	RequestedDeleteResourceMsg    = "Requested to delete %s \"%s\".\n"
-	UpdatedResourceMsg            = "Updated %s \"%s\".\n"
+	CreatedResourceMsg             = "Created %s \"%s\".\n"
+	DeletedResourceMsg             = "Deleted %s \"%s\".\n"
+	DeleteResourceConfirmMsg       = "Are you sure you want to delete %s \"%s\"?\nTo confirm, type \"%s\". To cancel, press Ctrl-C"
+	DeleteResourceConfirmYesNoMsg  = `Are you sure you want to delete %s "%s"?`
+	DeleteResourcesConfirmYesNoMsg = "Are you sure you want to delete %ss %s?"
+	DeleteACLConfirmMsg            = "Are you sure you want to delete the ACL corresponding to these parameters?"
+	DeleteACLsConfirmMsg           = "Are you sure you want to delete the ACLs corresponding to these parameters?"
```
brianstrauch, line 130: Unrelated: It would be best if this was moved to internal/cmd/acl

internal/pkg/utils/utils.go
```diff
@@ -140,6 +140,30 @@ func ArrayToCommaDelimitedString(arr []string) string {
 	return delimitedStr.String()
 }
 
+func ArrayToCommaDelimitedStringWithAnd(arr []string) string {
```
brianstrauch, line 143: How about passing "and" or "or" into this function?
```suggestion
func SliceToCommaSeparatedString(arr []string, conjunction string) string {
```

test/api_key_test.go
```diff
@@ -104,6 +104,10 @@ func (s *CLITestSuite) TestAPIKey() {
 		{args: "api-key create --resource lkc-cool1 --service-account sa-1337 --description auditlog-key", fixture: "api-key/42.golden", disableAuditLog: true}, // MYKEY11
 		{args: "api-key list", fixture: "api-key/43.golden", disableAuditLog: true},
 
+		// delete multiple API keys
+		{args: "api-key delete MYKEY6 MYKEY7 MYKEY13 MYKEY14", fixture: "api-key/61.golden", exitCode: 1},
+		{args: "api-key delete MYKEY6 MYKEY7 MYKEY8", input: "y\n", fixture: "api-key/62.golden"},
```
brianstrauch, line 109: Can we move all the deletion tests into a separate function? This one is way too long?

brianstrauch (changes requested):

internal/cmd/api-key/command_delete.go
sgagniere: I can go ahead and switch to making `n` calls to the describe endpoint instead

internal/pkg/errors/errors.go
sgagniere: Nope. It's part of the standard library like `fmt` or `os`.

internal/pkg/utils/delete_utils.go
```diff
@@ -27,13 +27,21 @@ func ValidateArgsForDeletion(cmd *cobra.Command, args []string, resourceType str
...
+			output.ErrPrintln(invalidArgsErrMsg)
+			return validArgs, nil
+		}
+
+		if len(validArgs) == 1 {
+			return nil, errors.NewErrorWithSuggestions(invalidArgsErrMsg, fmt.Sprintf(errors.DeleteNotFoundSuggestions, resourceType))
+		} else if len(validArgs) > 1 {
+			return nil, errors.NewErrorWithSuggestions(invalidArgsErrMsg, fmt.Sprintf(errors.DeleteNotFoundSuggestions, resource.Plural(resourceType)))
+		} else {
+			return nil, errors.New(invalidArgsErrMsg)
```
brianstrauch: If we short circuit on this error first, we'll be able to remove the duplication above.

internal/cmd/api-key/command_delete.go
```diff
@@ -21,6 +21,7 @@ func (c *command) newDeleteCommand() *cobra.Command {
 	}
 
 	pcmd.AddForceFlag(cmd)
+	pcmd.AddWarnFlag(cmd)
```
brianstrauch, line 25: Not a huge fan of `--warn`... We have `--if-exists` elsewhere in the CLI, so that might make more sense? i.e. `confluent api-key delete A B C --if-exists`

brianstrauch: Maybe `--skip-invalid`?

brianstrauch: Also, what's the argument for adding this flag in the first place? Do we really need it? Why can't the user delete the invalid entries?

brianstrauch (changes requested):

internal/cmd/api-key/command_delete.go
```diff
@@ -6,39 +6,71 @@ import (
...
+func (c *command) validateArgs(cmd *cobra.Command, args []string) ([]string, error) {
+	describeFunc := func(id string) error {
+		if _, _, err := c.V2Client.GetApiKey(id); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	validArgs, err := deletion.ValidateArgsForDeletion(cmd, args, resource.ApiKey, describeFunc)
+	err = errors.NewWrapAdditionalSuggestions(err, fmt.Sprintf(errors.ListResourceSuggestions, resource.ApiKey, "api-key"))
```
brianstrauch, line 73: You should be able to get "api-key" from `cmd`. Something like `cmd.Parent().Name()`

internal/pkg/errors/error_message.go
```diff
@@ -495,6 +497,7 @@ const (
 	// General
 	DeleteResourceErrorMsg        = `failed to delete %s "%s": %v`
 	DeleteResourceConfirmErrorMsg = `input does not match "%s"`
+	ListResourceSuggestions       = "List available %ss with `confluent %s list`."
```
brianstrauch, line 500: We should probably call the `plural()` function, since there might be edge cases.
```suggestion
	ListResourceSuggestions       = "List available %s with `confluent %s list`."
```

internal/pkg/resource/resource.go
```diff
@@ -68,3 +83,42 @@ func LookupType(resourceId string) string {
...
+	}
+
+	if last := string(resource[len(resource)-1]); last == "s" || last == "x" || last == "z" {
+		return resource + "es"
+	}
+	if len(resource) > 1 {
+		if lastTwo := string(resource[len(resource)-2:]); lastTwo == "ch" || lastTwo == "sh" {
+			return resource + "es"
+		}
+	}
```
brianstrauch, line 121: If you use `strings.HasSuffix()`, you won't need to do the length checks, and you can use a `map[string]string`!

test/api_key_test.go
```diff
@@ -167,6 +166,24 @@ func (s *CLITestSuite) TestApiKeyDescribe() {
...
+		{args: "api-key delete MYKEY6 MYKEY17 MYKEY18", fixture: "api-key/delete/1.golden", exitCode: 1},
+		{args: "api-key delete MYKEY6 MYKEY17 MYKEY18 --skip-invalid --force", input: "y\n", fixture: "api-key/delete/2.golden"},
+		{args: "api-key delete MYKEY7 MYKEY8 MYKEY19", input: "y\n", fixture: "api-key/delete/3.golden", exitCode: 1},
+		{args: "api-key delete MYKEY7 MYKEY8", input: "y\n", fixture: "api-key/delete/4.golden"},
+		{args: "api-key delete MYKEY17 MYKEY18", fixture: "api-key/delete/5.golden", exitCode: 1},
+	}
+
+	resetConfiguration(s.T(), false)
+
+	for _, tt := range tests {
```
brianstrauch, line 181: I've started writing "test" since that's less confusing than "tt"
```suggestion
	for _, test := range tests {
```

test/byok_test.go
```diff
@@ -17,6 +17,8 @@ func (s *CLITestSuite) TestBYOK() {
 		// delete tests
 		{args: "byok delete cck-001", input: "y\n", fixture: "byok/delete_1.golden"},
 		{args: "byok delete cck-404", fixture: "byok/delete_2.golden", exitCode: 1},
+		{args: "byok delete cck-002 cck-006 cck-007 cck-100", fixture: "byok/delete_3.golden", exitCode: 1},
```
brianstrauch, line 20: I've never been a huge fan of numbering integration tests... they always get out of order, and they're not descriptive.
```suggestion
		{args: "byok delete cck-002 cck-006 cck-007 cck-100", fixture: "byok/delete-multiple.golden", exitCode: 1},
```

brianstrauch: Is this testing that one of these IDs doesn't exist?

test/fixtures/output/api-key/47.golden
```diff
@@ -1,4 +1,4 @@
-Error: error getting API key: resource not found: unknown API key UNKNOWN
+Error: error getting API key UNKNOWN: resource not found: unknown API key UNKNOWN
```
brianstrauch, line 1: It would be great to put quotes around all resources!
```suggestion
Error: error getting API key "UNKNOWN": resource not found: unknown API key "UNKNOWN"
```

internal/pkg/errors/error_message.go
```diff
@@ -471,6 +470,9 @@ const (
 		"To create an API key, use `confluent api-key create --resource %s`.\n" +
 		"To store an existing API key, use `confluent api-key store --resource %s`."
 	FailedToReadInputErrorMsg = "failed to read input"
+	NotFoundErrorMsg          = "%s not found: %s"
```
brianstrauch, line 473:
```suggestion
	NotFoundErrorMsg          = `%s "%s" not found`
```

internal/pkg/resource/resource.go
```diff
@@ -68,3 +83,42 @@ func LookupType(resourceId string) string {
...
+	for _, resourceId := range args {
+		if LookupType(resourceId) != resourceType {
+			malformed = append(malformed, resourceId)
+		}
+	}
+
+	if len(malformed) == 1 {
+		return errors.Errorf(`failed parsing resource ID %s: missing prefix "%s-" is required`, malformed[0], prefix)
+	} else if len(malformed) > 1 {
+		return errors.Errorf(`failed parsing resource IDs %s: missing prefix "%s-" is required`, utils.ArrayToCommaDelimitedString(malformed, "and"), prefix)
```
brianstrauch, line 103: Wording is redundant. Either "missing prefix x" or "prefix x is required"

internal/cmd/api-key/command_delete.go
```diff
@@ -6,39 +6,71 @@ import (
...
+		return err
+	}
+
+	var errs error
+	var deleted []string
+	for _, id := range args {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			errs = errors.Join(errs, errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, id, r))
+		} else {
+			deleted = append(deleted, id)
```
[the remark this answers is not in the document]
sgagniere: Nope. This gets turned into a message about successful deletion, so I'm only adding it in if the err was `nil`.

internal/cmd/byok/command_delete.go
```diff
@@ -6,40 +6,67 @@ import (
...
+		Use:               "delete <id-1> [id-2] ... [id-n]",
+		Short:             "Delete self-managed keys.",
+		Long:              "Delete self-managed keys from Confluent Cloud.",
+		Args:              cobra.MinimumNArgs(1),
 		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
 		RunE:              c.delete,
 	}
 
 	pcmd.AddForceFlag(cmd)
+	pcmd.AddSkipInvalidFlag(cmd)
```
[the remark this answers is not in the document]
sgagniere: I'll get back to you on this tomorrow!

test/byok_test.go
sgagniere: > Is this testing that one of these IDs doesn't exist?
Yup. I'll do a pass at the end and rename all the number-name files I introduced. I agree that that naming scheme is not good.

test/fixtures/output/stream-share/delete-consumer-share-multiple-fail.golden
```diff
@@ -0,0 +1,5 @@
+Error: consumer share not found: "ss-12346"
+
+Suggestions:
+    Use the `--skip-invalid` flag to suppress this error and delete the remaining consumer share entered.
```
brianstrauch, line 4: Should this be plural? Also are we sure we want this flag in the first place?
```suggestion
    Use the `--skip-invalid` flag to suppress this error and delete the remaining consumer shares entered.
```

test/test-server/cloud_kafka_handlers.go
```diff
@@ -47,8 +46,8 @@ var KafkaQuotas = []*kafkaquotasv1.KafkaQuotasV1ClientQuota{
 func handleKafkaClientQuota(t *testing.T) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		quotaId := mux.Vars(r)["id"]
-		if i := slices.IndexFunc(KafkaQuotas, func(quota *kafkaquotasv1.KafkaQuotasV1ClientQuota) bool { return quota.GetId() == quotaId }); i != -1 {
-			quota := KafkaQuotas[i]
+		if i := getV2Index(kafkaQuotas, quotaId); i != -1 {
```
brianstrauch, line 49: What does `getV2Index()` mean?

test/test-server/cloud_kafka_handlers.go
```diff
@@ -57,14 +56,18 @@ func handleKafkaClientQuota(t *testing.T) http.HandlerFunc {
 				_, err := io.WriteString(w, "")
 				require.NoError(t, err)
 			case http.MethodPatch:
+				quotaPatch := &kafkaquotasv1.KafkaQuotasV1ClientQuota{ // make a deep copy so changes don't reflect in subsequent tests
```
brianstrauch, line 59: Can we just change the type from a slice of pointers to a regular slice, and do a normal copy?

test/test-server/quotas_handler.go
```diff
@@ -19,45 +19,45 @@ func handleAppliedQuotas(t *testing.T) http.HandlerFunc {
...
-			AppliedLimit: int32ToPtr(15),
+			Usage:        ptr(int32(10)),
+			AppliedLimit: ptr(int32(15)),
 		}
 
 		quota2 := servicequotav1.ServiceQuotaV1AppliedQuota{
-			Id:           stringToPtr("quota_a"),
-			Scope:        stringToPtr("kafka_cluster"),
-			DisplayName:  stringToPtr("Qutoa A"),
+			Id:           ptr("quota_a"),
```
brianstrauch, line 33: Which one should we prefer? (I assume the following function exists)
```suggestion
			Id:           servicequotav1.PtrString("quota_a"),
```

test/test-server/cloud_kafka_handlers.go
sgagniere: I just couldn't think of a better name. It's to get the index of the entry in the array that matches the id entered. I wrote it specifically for things in ccloud-sdk-go-v2 which all have a `GetId()` method, hence why I slapped v2 in the name.

brianstrauch: Is your goal with this to add state to the mock backend? How about we make a generic `type mockDatabase []T` that has functions like `Add()`, `Lookup()`, and `Delete()`?

test/test-server/quotas_handler.go
sgagniere: I definitely prefer to use the PtrString method from the sdk. I did a global find and replace for our in-house string to pointer methods which is probably why I didn't catch this.

test/test-server/cloud_kafka_handlers.go
sgagniere: Kind of. I initially added state before I realized that this was incredibly annoying: having state in our test backend means that running a specific test with `INTEGRATION_TEST_ARGS` can fail if updates, posts, and deletes that normally run beforehand don't get run. It also obfuscates what's actually changing when the output has `updated-name` but the test only updated a different field...

So what I'm trying to do here is adding a consistent initial state, so that describe/list/delete are all consistent. And none of the delete/patch/post actually change anything, so you always have the same set of resources to work with when you add additional integration tests, and you don't have to update unrelated golden files when you re-arrange or remove integration tests.

Side note: I didn't bother trying to do this for api-key because we have a state for that one with the keystore...

test/fixtures/output/stream-share/delete-consumer-share-multiple-fail.golden
sgagniere: No, we don't need this flag. So I'll remove this. Kind of moot now, but to answer the first question: the message was set up to be singular or plural based on the count of remaining valid entries.

internal/cmd/iam/command_acl_delete.go
```diff
@@ -3,8 +3,8 @@ package iam
 import (
 	"github.com/spf13/cobra"
 
+	aclutil "github.com/confluentinc/cli/internal/pkg/acl"
```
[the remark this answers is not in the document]
sgagniere: There's an `acl` variable so I'll rename this to `pacl` to keep with our convention.

internal/pkg/errors/error_message.go
```diff
@@ -471,6 +470,9 @@ const (
...
+	AccessForbiddenErrorMsg   = "%s(s) not found or access forbidden: %s"
```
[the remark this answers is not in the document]
sgagniere: Removed it; it ended up unused.

test/fixtures/output/api-key/47.golden
sgagniere: Agreed. But I think I'll leave this one alone, because half of this is the backend error that doesn't have quotation marks. I checked that the real backend returns the same thing.

I think we can probably toss the backend error, but I'm not sure if there are other (meaningful) messages that it might return aside from `unknown API key %s`

internal/cmd/kafka/command_quota_delete.go
```diff
@@ -53,11 +54,15 @@ func (c *quotaCommand) delete(cmd *cobra.Command, args []string) error {
...
-		_, err := c.V2Client.DescribeKafkaQuota(id)
+		quota, err := c.V2Client.DescribeKafkaQuota(id)
+		if err == nil && displayName == "" { // store the first valid user name
+			displayName = quota.Spec.GetDisplayName()
+		}
 		return err
 	}
 
-	return deletion.ValidateArgsForDeletion(cmd, args, resource.ClientQuota, describeFunc)
+	return displayName, deletion.ValidateArgsForDeletion(cmd, args, resource.ClientQuota, describeFunc)
```
brianstrauch, line 67: Based on the function name, it seems odd that this returns a string.

sgagniere: It is odd, I agree. The problem is that I don't want this clogging up the delete function, but at the same time I don't want to make an extra call just to get the name.

How about this? I rename these functions to `validateArgsAndConfirmDeletion` or something similar and roll the confirmation check into these `validateArgs` functions.

internal/cmd/api-key/command_delete.go
```diff
@@ -56,11 +52,19 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func (c *command) validateArgs(cmd *cobra.Command, args []string) error {
+func (c *command) confirmDeletion(cmd *cobra.Command, args []string) error {
 	describeFunc := func(id string) error {
 		_, _, err := c.V2Client.GetApiKey(id)
```
brianstrauch, line 57: Why are we ignoring the `detail` field of the HTTP response (second argument) here again?

internal/cmd/connect/command_cluster_delete.go
```diff
@@ -88,5 +77,19 @@ func (c *clusterCommand) validateArgs(cmd *cobra.Command, environmentId, kafkaCl
 		return err
 	}
 
-	return connectorIdToName, deletion.ValidateArgsForDeletion(cmd, args, resource.Connector, describeFunc)
+	if err := deletion.ValidateArgsForDeletion(cmd, args, resource.Connector, describeFunc); err != nil {
+		return err
+	}
+
+	if len(args) == 1 {
+		if err := form.ConfirmDeletionWithString(cmd, resource.Connector, args[0], connectorIdToName[args[0]]); err != nil {
```
brianstrauch, line 85: How about getting the name from the describe API call instead of passing this `map` in?

internal/cmd/api-key/command_delete.go
sgagniere: I'm only looking to see whether there was an error or not. All id's that return an error here are reported together in a different error message along with a suggestion to use the list command to view valid ids.

internal/cmd/connect/command_cluster_delete.go
sgagniere: I need these names in the main `delete` function since `V2Client.DeleteConnector` takes the name instead of the id, so I pass in the map to avoid making two of every describe call.

I'm just recycling the map for this line.

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +26,45 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
 	}
-	httpResp, err := c.V2Client.DeleteApiKey(apiKey)
-	if err != nil {
-		return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, httpResp)
+
+	var errs error
+	var deleted []string
+	for _, id := range args {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			errs = errors.Join(errs, errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r))
```
MuweiHe, line 38: what does a joined error look like?

internal/pkg/deletion/delete_utils.go
```diff
@@ -0,0 +1,41 @@
...
+
+	return nil
+}
+
+func PrintSuccessfulDeletionMsg(successful []string, resourceType string) {
+	if len(successful) == 1 {
+		output.Printf(errors.DeletedResourceMsg, resourceType, successful[0])
+	} else if len(successful) > 1 {
+		output.Printf("Deleted %s %s.\n", resource.Plural(resourceType), utils.ArrayToCommaDelimitedString(successful, "and"))
+	}
```
MuweiHe, line 40: add a msg like `None found`: like `None was deleted`? output something so the user knows cli did the work?

internal/cmd/api-key/command_delete.go
brianstrauch: +1, I don't see any examples in the golden files, so let's make sure we have at least one

internal/cmd/api-key/command_delete.go
sgagniere: Yeah I'm not using the errors from the describe calls; I'm only looking at whether or not an error happened. What I meant about combined error message was this:
```
Error: connectors "lcc-456", "lcc-789", and "lcc-101112" not found

Suggestions:
    List available connectors with `confluent connect cluster list`.
```

internal/cmd/api-key/command_delete.go
sgagniere: I only made one: `test/fixtures/output/context/delete/check-success-operation-fail.golden`

The format is:
```
Error: error_1
error_2
error_3
...
```

internal/pkg/deletion/delete_utils.go
sgagniere: It always outputs something: each resource in the list either reports an error in the multi-error or it adds to the success message at the end.

internal/cmd/environment/command_delete.go
```diff
@@ -28,33 +28,60 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
 
-	if err := c.V2Client.DeleteOrgEnvironment(id); err != nil {
+	return nil
+}
+
+func (c *command) confirmDeletion(cmd *cobra.Command, args []string) error {
+	var displayName string
+	describeFunc := func(id string) error {
+		environment, err := c.V2Client.GetOrgEnvironment(id)
+		if err == nil && id == args[0] {
```
MuweiHe, line 66: why is it always args[0]?

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +26,45 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+		} else {
+			deleted = append(deleted, id)
+			if err := c.keystore.DeleteAPIKey(id); err != nil {
+				errs = errors.Join(errs, err)
+			}
+		}
+	}
+	deletion.PrintSuccessfulDeletionMsg(deleted, resource.ApiKey)
```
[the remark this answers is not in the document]
sgagniere: We could do it that way. It's the difference between reporting the successes first or the failures first. Which do we prefer?

internal/cmd/connect/command_cluster_delete.go
```diff
@@ -41,31 +39,57 @@ func (c *clusterCommand) newDeleteCommand() *cobra.Command {
...
+	if err := c.confirmDeletion(cmd, environmentId, kafkaCluster.ID, args, connectorIdToName); err != nil {
 		return err
 	}
 
-	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmMsg, resource.Connector, clusterId, connector.Info.GetName())
-	if _, err := form.ConfirmDeletion(cmd, promptMsg, connector.Info.GetName()); err != nil {
+	var errs error
+	var deleted []string
+	for _, id := range args {
+		if _, err := c.V2Client.DeleteConnector(connectorIdToName[id], environmentId, kafkaCluster.ID); err != nil {
```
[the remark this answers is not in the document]
sgagniere: Agreed :(

internal/cmd/environment/command_delete.go
```diff
@@ -28,33 +28,60 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+			deleted = append(deleted, id)
+			if id == c.Context.GetCurrentEnvironment() {
+				c.Context.SetCurrentEnvironment("")
+
+				if err := c.Config.Save(); err != nil {
+					errs = errors.Join(errs, errors.Wrap(err, errors.EnvSwitchErrorMsg))
+				}
+			}
+			c.Context.DeleteEnvironment(id)
+			_ = c.Config.Save()
```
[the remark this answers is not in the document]
sgagniere: Sure. There are a handful that have extra handling, so I can throw those into a function for post-delete operations.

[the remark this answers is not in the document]
sgagniere: Sure. There are a handful that have extra handling, so I can throw those into a function for post-delete operations.

internal/cmd/iam/command_acl_delete.go
```diff
@@ -46,11 +46,11 @@ func (c *aclCommand) delete(cmd *cobra.Command, _ []string) error {
...
 	}
 
-	promptMsg := errors.DeleteACLConfirmMsg
+	promptMsg := pacl.DeleteACLConfirmMsg
 	if len(bindings) > 1 {
-		promptMsg = errors.DeleteACLsConfirmMsg
+		promptMsg = pacl.DeleteACLsConfirmMsg
 	}
-	if ok, err := form.ConfirmDeletion(cmd, promptMsg, ""); err != nil || !ok {
+	if ok, err := form.ConfirmDeletionYesNoCustomPrompt(cmd, promptMsg); err != nil || !ok {
```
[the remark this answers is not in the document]
sgagniere: The user input is `y/n` and the function takes in a prompt for the user instead of using the boilerplate prompt.

internal/cmd/kafka/command_broker_delete.go
```diff
@@ -47,17 +43,52 @@ func (c *brokerCommand) delete(cmd *cobra.Command, args []string) error {
...
+			deleted = append(deleted, id)
+		}
+	}
+	if len(deleted) == 1 {
+		output.Printf("Started deletion of broker %[1]s. To monitor the remove-broker task run `confluent kafka broker get-tasks %[1]s --task-type remove-broker`.\n", deleted[0])
+	} else if len(deleted) > 1 {
+		output.Printf("Started deletion of brokers %s. To monitor a remove-broker task run `confluent kafka broker get-tasks <id> --task-type remove-broker`.\n", utils.ArrayToCommaDelimitedString(deleted, "and"))
+	}
+
+	return nil
```
MuweiHe, line 67: refactor this like other delete's, to be consistent?

internal/cmd/kafka/command_broker_delete.go
```diff
@@ -47,17 +43,52 @@ func (c *brokerCommand) delete(cmd *cobra.Command, args []string) error {
...
+			errs = errors.Join(errs, kafkarest.NewError(restClient.GetConfig().BasePath, err, resp))
+		} else {
+			deleted = append(deleted, id)
+		}
+	}
+	if len(deleted) == 1 {
+		output.Printf("Started deletion of broker %[1]s. To monitor the remove-broker task run `confluent kafka broker get-tasks %[1]s --task-type remove-broker`.\n", deleted[0])
+	} else if len(deleted) > 1 {
+		output.Printf("Started deletion of brokers %s. To monitor a remove-broker task run `confluent kafka broker get-tasks <id> --task-type remove-broker`.\n", utils.ArrayToCommaDelimitedString(deleted, "and"))
+	}
```
MuweiHe, line 65: refactor this like other delete's, to be consistent?

internal/pkg/cmd/cobra.go
```diff
@@ -45,3 +45,18 @@ func RegisterFlagCompletionFunc(cmd *cobra.Command, flag string, f func(*cobra.C
...
+
+func FullParentName(c *cobra.Command) string {
+	var name string
+	if c.HasParent() {
+		name = c.Parent().Name()
+		c = c.Parent()
+	}
+
+	for c.HasParent() {
+		name = c.Parent().Name() + " " + name
```
MuweiHe, line 57: use `fmt`?

internal/pkg/errors/error_message.go
```diff
@@ -495,6 +492,7 @@ const (
 	// General
 	DeleteResourceErrorMsg        = `failed to delete %s "%s": %v`
 	DeleteResourceConfirmErrorMsg = `input does not match "%s"`
+	ListResourceSuggestions       = "List available %ss with `%s list`."
```
MuweiHe, line 495:
```suggestion
	ListResourceSuggestions       = "List available %ss with `confluent %s list`."
```

internal/pkg/errors/strings.go
```diff
@@ -118,14 +118,10 @@ const (
...
-	DeletedResourceMsg            = "Deleted %s \"%s\".\n"
-	DeleteResourceConfirmMsg      = "Are you sure you want to delete %s \"%s\"?\nTo confirm, type \"%s\". To cancel, press Ctrl-C"
-	DeleteResourceConfirmYesNoMsg = `Are you sure you want to delete %s "%s"?`
-	DeleteACLConfirmMsg           = "Are you sure you want to delete the ACL corresponding to these parameters?"
-	DeleteACLsConfirmMsg          = "Are you sure you want to delete the ACLs corresponding to these parameters?"
-	RequestedDeleteResourceMsg    = "Requested to delete %s \"%s\".\n"
-	UpdatedResourceMsg            = "Updated %s \"%s\".\n"
+	CreatedResourceMsg       = "Created %s \"%s\".\n"
+	DeletedResourceMsg       = "Deleted %s \"%s\".\n"
+	DeleteResourceConfirmMsg = "Are you sure you want to delete %s \"%s\"?\nTo confirm, type \"%s\". To cancel, press Ctrl-C"
```
MuweiHe, line 123: i remember some resources are like api-key `id`, some are cluster `name`. has it changed?

internal/cmd/environment/command_delete.go
sgagniere: Currently, some delete operations ask a user to type a string. The new version will still do that, but only if the user is deleting a single resource (otherwise it just asks for a `y/n` response). So I only need to keep one string.

The reason it's always the first is that this check requires that all id's belong to valid resources. So the only way to reach the string typing prompt is if there's only one resource, and it exists, so the 0 index id.

internal/pkg/errors/error_message.go
sgagniere: The function that returns the full parent command includes `confluent`, so I left it off of this string.

internal/pkg/errors/strings.go
sgagniere: They still work that way for single deletion, because I still want it to match the UI flow. But for multiple ids, all of them switch to a `y/n` prompt.

internal/cmd/iam/command_acl_delete.go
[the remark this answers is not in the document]
brianstrauch: Since we're passing `promptMsg` here, and it's clear we're passing a custom prompt message, maybe we can just do this:
```suggestion
	if ok, err := form.ConfirmDeletionYesNo(cmd, promptMsg); err != nil || !ok {
```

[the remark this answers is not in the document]
sgagniere: That's what I'm calling the boilerplate version `ConfirmDeletionYesNo`. I'm open to naming suggestions, or other suggestions.

I guess I could just have one version that takes in a prompt, but then I would be passing in the same prompt for the majority of these calls but that just increases the amount of boilerplate code in the delete files.

Incidentally, this is why I wish golang had overloading.

internal/cmd/kafka/command_broker_delete.go
sgagniere: I dunno... I kept it this way because this one had a special output to begin with, and I don't want to modify the common delete message function for one special case.

internal/cmd/api-key/command_delete.go
brianstrauch: I feel like I like the format of [multierror](https://github.com/hashicorp/go-multierror) better...

sgagniere: I also like their format, but the default includes 2 extra newlines at the end that I found very annoying, so I used the built in `Join` instead.

But, looking at that repo, it looks like you can pass in a custom format. I think I'll switch to multierror and just write a format func that doesn't have the extraneous new lines.

internal/cmd/api-key/command_delete.go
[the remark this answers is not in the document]
brianstrauch: Oh, now I see what's going on here. We should print successes first.

internal/cmd/api-key/command_delete.go
```diff
@@ -32,24 +31,19 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	errs := &multierror.Error{ErrorFormat: errors.CustomMultierrorList}
-	var deleted []string
-	for _, id := range args {
+	deleted, err := deletion.DeleteResources(args, func(id string) error {
```
brianstrauch, line 34: Would it make more sense to put this in the existing `resource` package? 
```suggestion
	deleted, err := resource.Delete(args, func(id string) error {
```

internal/cmd/api-key/command_delete.go
```diff
@@ -32,24 +31,19 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
...
-		} else {
-			deleted = append(deleted, id)
-			if err := c.deletePostProcess(id); err != nil {
-				errs = multierror.Append(errs, err)
-			}
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
 		}
-	}
+		return nil
+	}, func(id string) error {
```
brianstrauch, line 39: Can we define these functions outside of the function call? This syntax is messy, IMO.

internal/cmd/api-key/command_delete.go
sgagniere: Yeah. I went ahead and just moved everything out of that package and into other packages. `resource` already had a `ValidatePrefixes` function so I went ahead and put `ValidateArgs` there as well. `DefaultPostProcess` is only meant to be used with (what is now) `Delete`, so it also went there.

I put the default prompt string functions into form, next to the actual prompting functions themselves.

internal/pkg/resource/resource.go
```diff
@@ -75,6 +76,9 @@ var resourceToPrefix = map[string]string{
 	User:                  UserPrefix,
 }
 
+// Singular words ending w/ these suffixes generally add an extra -es syllable in their plural forms
+var pluralExtraSyllableSuffix = types.NewSet("s", "x", "z", "ch", "sh")
```
brianstrauch, line 80:
```suggestion
var pluralSuffixes = types.NewSet("s", "x", "z", "ch", "sh")
```

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,42 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
-	apiKey := args[0]
 
-	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmYesNoMsg, resource.ApiKey, apiKey)
-	if ok, err := form.ConfirmDeletion(cmd, promptMsg, ""); err != nil || !ok {
+	if err := c.confirmDeletion(cmd, args); err != nil {
 		return err
 	}
-	httpResp, err := c.V2Client.DeleteApiKey(apiKey)
+
+	deleted, err := resource.Delete(args, func(id string) error {
```
brianstrauch, line 33: Can we define this function outside of the function call? I find this really difficult to read.

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,42 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
+		}
+		return nil
+	}, c.postProcess)
```
brianstrauch, line 38: Why do we need to pass this function in? Can't we just call it after `resource.Delete()`?

internal/pkg/resource/resource.go
```diff
@@ -70,3 +93,88 @@ func LookupType(resourceId string) string {
...
+	for suffix := range pluralExtraSyllableSuffix {
+		if strings.HasSuffix(resource, suffix) {
+			return resource + "es"
+		}
+	}
+
+	return resource + "s"
+}
+
+func DefaultPostProcess(_ string) error {
```
brianstrauch, line 153: Can we just pass `nil` instead of this function? It's unclear at first glance what `DefaultPostProcess` means/does.

internal/cmd/api-key/command_delete.go
sgagniere: This is to avoid looping through all the id's twice, once to delete and once to perform additional tasks.

internal/pkg/resource/resource.go
sgagniere: Oh, good point.

internal/cmd/api-key/command_delete.go
```diff
@@ -1,23 +1,20 @@
...
 )
 
 func (c *command) newDeleteCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "delete <api-key>",
-		Short:             "Delete an API key.",
-		Args:              cobra.ExactArgs(1),
-		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		Use:               "delete <api-key-1> [api-key-2] ... [api-key-n]",
+		Short:             "Delete API keys.",
```
brianstrauch, line 15: Currently it sounds like you can only delete multiple API keys.
```suggestion
		Short:             "Delete one or more API keys.",
```

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,44 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+
+	deleteFunc := func(id string) error {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
+		}
+		return nil
+	}
+
+	deleted, err := resource.Delete(args, deleteFunc, c.postProcess)
+	resource.PrintDeleteSuccessMsg(deleted, resource.ApiKey)
```
brianstrauch, line 41: Why are we printing a success message before we check for an error?

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,44 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
 
-	output.Printf(errors.DeletedResourceMsg, resource.ApiKey, apiKey)
-	return c.keystore.DeleteAPIKey(apiKey)
+	if err := resource.ValidateArgs(pcmd.FullParentName(cmd), args, resource.ApiKey, describeFunc); err != nil {
+		return err
+	}
+
+	if ok, err := form.ConfirmDeletionYesNo(cmd, form.DefaultYesNoPromptString(resource.ApiKey, args)); err != nil || !ok {
+		return err
+	}
```
brianstrauch, line 61: So if the user types "y", `ok` should be `true`, and we return `nil`, and if the user types "n", `ok` should be `false` and if `err == nil`, we should also return `nil`? I don't see what the difference is.

internal/cmd/connect/command_cluster_delete.go
```diff
@@ -41,31 +37,57 @@ func (c *clusterCommand) newDeleteCommand() *cobra.Command {
...
+	return err
+}
+
+func (c *clusterCommand) confirmDeletion(cmd *cobra.Command, environmentId, kafkaClusterId string, args []string, connectorIdToName map[string]string) error {
+	describeFunc := func(id string) error {
+		connector, err := c.V2Client.GetConnectorExpansionById(id, environmentId, kafkaClusterId)
+		if err == nil {
+			connectorIdToName[id] = connector.Info.GetName()
+		}
 		return err
```
brianstrauch, line 74: A bit more verbose, but this is the more idiomatic way to write this.
```suggestion
		if err != nil {
			return err
		}
		connectorIdToName[id] = connector.Info.GetName()
		
		return nil
```

brianstrauch: (In the current approach, the fallthrough combined with the assumption that `err == nil` is confusing)

internal/cmd/context/command_delete.go
```diff
@@ -27,20 +23,36 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
-	if err != nil {
+	if err := c.confirmDeletion(cmd, args); err != nil {
+		return err
+	}
+
+	deleteFunc := func(id string) error {
+		if err := c.Config.DeleteContext(id); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 34:
```suggestion
		return c.Config.DeleteContext(id)
```

internal/cmd/environment/command_delete.go
```diff
@@ -28,33 +27,65 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
-	id := args[0]
+	if err := c.confirmDeletion(cmd, args); err != nil {
+		return err
+	}
+
+	deleteFunc := func(id string) error {
+		if err := c.V2Client.DeleteOrgEnvironment(id); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 38:
```suggestion
		return c.V2Client.DeleteOrgEnvironment(id)
```

internal/cmd/flink/command_statement_delete.go
```diff
@@ -40,19 +37,36 @@ func (c *command) statementDelete(cmd *cobra.Command, args []string) error {
...
-	if _, err := client.GetStatement(environmentId, args[0], c.Context.LastOrgId); err != nil {
+	if err := c.confirmDeletionStatement(cmd, client, environmentId, args); err != nil {
+		return err
+	}
+
+	deleteFunc := func(id string) error {
+		if err := client.DeleteStatement(environmentId, id, c.Context.LastOrgId); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 48:
```suggestion
		return client.DeleteStatement(environmentId, id, c.Context.LastOrgId)
```

internal/cmd/iam/command_acl_delete.go
```diff
@@ -46,11 +49,11 @@ func (c *aclCommand) delete(cmd *cobra.Command, _ []string) error {
 		return c.handleACLError(cmd, err, response)
 	}
 
-	promptMsg := errors.DeleteACLConfirmMsg
+	promptMsg := fmt.Sprintf(pacl.DeleteACLConfirmMsg, "ACL")
 	if len(bindings) > 1 {
-		promptMsg = errors.DeleteACLsConfirmMsg
+		promptMsg = fmt.Sprintf(pacl.DeleteACLConfirmMsg, resource.Plural("ACL"))
```
brianstrauch, line 54: Can we add "ACL" to the list of resources? It seems weird to be calling `resource.Plural()` on a constant string.

internal/cmd/iam/command_pool_delete.go
```diff
@@ -42,20 +38,46 @@ func (c *identityPoolCommand) delete(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
 
-	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmMsg, resource.IdentityPool, args[0], pool.GetDisplayName())
-	if _, err := form.ConfirmDeletion(cmd, promptMsg, pool.GetDisplayName()); err != nil {
+	deleteFunc := func(id string) error {
+		if err := c.V2Client.DeleteIdentityPool(id, provider); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 49:
```suggestion
		return c.V2Client.DeleteIdentityPool(id, provider)
```

internal/cmd/iam/command_pool_delete.go
```diff
@@ -42,20 +38,46 @@ func (c *identityPoolCommand) delete(cmd *cobra.Command, args []string) error {
...
+	resource.PrintDeleteSuccessMsg(deleted, resource.IdentityPool)
+
+	return err
+}
+
+func (c *identityPoolCommand) confirmDeletion(cmd *cobra.Command, provider string, args []string) error {
+	var displayName string
+	describeFunc := func(id string) error {
+		pool, err := c.V2Client.GetIdentityPool(id, provider)
+		if err == nil && id == args[0] {
```
brianstrauch, line 62: Please handle the error first!

internal/cmd/iam/command_provider_delete.go
```diff
@@ -34,20 +30,46 @@ func (c *identityProviderCommand) newDeleteCommand() *cobra.Command {
...
 		return err
 	}
 
-	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmMsg, resource.IdentityProvider, args[0], provider.GetDisplayName())
-	if _, err := form.ConfirmDeletion(cmd, promptMsg, provider.GetDisplayName()); err != nil {
+	deleteFunc := func(id string) error {
+		if err := c.V2Client.DeleteIdentityProvider(id); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 41:
```suggestion
		return c.V2Client.DeleteIdentityProvider(id)
```

brianstrauch (changes requested):

internal/cmd/api-key/command_delete.go
sgagniere: `deleted` is a slice of resources which were successfully deleted. For each resource, `resource.Delete` either adds it's ID to `deleted` if there were no errors, or else adds the error to a multierror.

The idea here is that I print out the ones which succeeded, and then the error will tell the user which ones didn't delete.

internal/cmd/api-key/command_delete.go
sgagniere: This is an artifact leftover from a previous refactor. I'll change it.

internal/cmd/iam/command_provider_delete.go
sgagniere: I have to say, this is getting progressively more embarrassing with each one 😭

internal/cmd/api-key/command_delete.go
brianstrauch: This seems a bit weird. How about we call `PrintDeleteSuccessMsg()` from inside `resource.Delete()`?

internal/cmd/api-key/command_delete.go
sgagniere: It was also a bug that deleted if users said `n`, so I'm going to add more integration tests to prevent it in the next commit.

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,42 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
-	if ok, err := form.ConfirmDeletion(cmd, promptMsg, ""); err != nil || !ok {
+	if confirm, err := c.confirmDeletion(cmd, args); err != nil {
 		return err
+	} else if !confirm {
+		return nil
+	}
+
+	deleteFunc := func(id string) error {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
```
brianstrauch, line 37: Don't we only care about the existence of an error here? Why are we wrapping it?

internal/cmd/api-key/command_delete.go
```diff
@@ -28,17 +25,42 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+		}
+		return nil
 	}
-	httpResp, err := c.V2Client.DeleteApiKey(apiKey)
+
+	deleted, err := resource.Delete(args, deleteFunc, c.postProcess)
```
brianstrauch, line 42: Why do we need to pass a post-process function? Can't we just call it afterwards?

internal/cmd/context/command_delete.go
```diff
@@ -27,20 +23,31 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+
+func (c *command) confirmDeletion(cmd *cobra.Command, args []string) (bool, error) {
+	describeFunc := func(id string) error {
+		_, err := c.Config.FindContext(id)
 		return err
 	}
 
-	output.Printf(errors.DeletedResourceMsg, resource.Context, ctx.Name)
-	return nil
+	if err := resource.ValidateArgs(pcmd.FullParentName(cmd), args, resource.Context, describeFunc); err != nil {
```
brianstrauch, line 48: Let's just pass `cmd` directly? We're calling `pcmd.FullParentName()` 24 times

internal/cmd/api-key/command_delete.go
sgagniere: This is the func for deleting, where we do want the error. The one for describing (existence check) is where we don't need the error.

internal/cmd/api-key/command_delete.go
sgagniere: I passed it in so that I wouldn't have to loop over the id's a second time.

If it improves readability, I can make it take the list of successfully deleted resources as an argument and call it outside.

brianstrauch: Yeah, I think that would help. We should 1) avoid passing functions into functions where possible and 2) avoid non-descriptive terminology like "postProcess"

internal/cmd/api-key/command_delete.go
```diff
@@ -39,9 +40,12 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	deleted, err := resource.Delete(args, deleteFunc, c.postProcess)
-	resource.PrintDeleteSuccessMsg(deleted, resource.ApiKey)
+	deletedIDs, err := resource.Delete(args, deleteFunc)
```
brianstrauch, line 43:
```suggestion
	deletedIds, err := resource.Delete(args, deleteFunc)
```

brianstrauch: It feels weird to be waiting to handle this error.

sgagniere: You left feedback last time (that I accidentally marked resolved) about rolling the error message into the `Delete` function (and I think Muwei asked for that too).

After I do that, this error will be handled immediately in most calls. The only ones where it won't be will be for resources like api-key where I need to do further tasks, like deleting the keys from the key store.

Since some deletes can succeed and some can fail, I can't return the error until I've removed the correct keys from the key store and added any potential new errors to the multierror.

internal/cmd/context/command_delete.go
sgagniere: I tried this and remembered why I did it this way: I get an import cycle error because the `cmd` pkg depends on `resource`. I think the `cmd` package is probably too large, which makes import cycles easy to run into.

I guess the easy thing to do is to just move `FullParentName` to `resource`, but it's called in one other place and `cmd` feels like the natural place for this function.

Does that seem like an alright solution for now?

sgagniere:
@brianstrauch I did a refactor on this that reduces the amount of duplicated code, reduces the amount of complexity inside the actual command implementation files, and should address most if not all previous concerns.

CONTRIBUTING.md
```diff
@@ -266,6 +266,10 @@ func (c *command) newDescribeCommand() *cobra.Command {
 
 See the [Autocompletion](internal/pkg/cmd/AUTOCOMPLETION.md) resource for implementation details.
 
+#### Adding a New Delete Command
+
+For most resource types, a `delete` command should support multiple arguments. The exceptions are resources which do not have an ID (e.g. ACLs, role bindings) or unique resources (e.g. the schema registry cluster).
```
brianstrauch, line 271:
```suggestion
For most resource types, a `delete` command should support multiple arguments. The exceptions are resources which do not have an ID (e.g. ACLs, role bindings) or unique resources (e.g. the Schema Registry cluster).
```

CONTRIBUTING.md
```diff
@@ -266,6 +266,10 @@ func (c *command) newDescribeCommand() *cobra.Command {
...
+See [Supporting Multiple Deletion](pkg/deletion/README.md) for instructions on how to write such commands.
```
brianstrauch, line 272:
```suggestion
See [Supporting Multiple Deletion](pkg/deletion/README.md) for instructions on how to write such commands.

```

internal/api-key/command_delete.go
```diff
@@ -28,17 +26,38 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
-		return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, httpResp)
+
+	deleteFunc := func(id string) error {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
+		}
+		return nil
+	}
+
+	deletedIDs, err := deletion.Delete(args, deleteFunc, resource.ApiKey)
```
brianstrauch, line 46:
```suggestion
	deletedIds, err := deletion.Delete(args, deleteFunc, resource.ApiKey)
```

internal/api-key/command_delete.go
```diff
@@ -28,17 +26,38 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+
+	if err := deletion.ValidateAndConfirmDeletionYesNo(cmd, args, existenceFunc, resource.ApiKey); err != nil {
 		return err
 	}
-	httpResp, err := c.V2Client.DeleteApiKey(apiKey)
-	if err != nil {
-		return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, httpResp)
+
+	deleteFunc := func(id string) error {
+		if r, err := c.V2Client.DeleteApiKey(id); err != nil {
```
brianstrauch, line 40: I think we've standardized around this
```suggestion
		if httpResp, err := c.V2Client.DeleteApiKey(id); err != nil {
```

internal/api-key/command_delete.go
```diff
@@ -28,17 +26,38 @@ func (c *command) newDeleteCommand() *cobra.Command {
...
+			return errors.CatchApiKeyForbiddenAccessError(err, deleteOperation, r)
+		}
+		return nil
+	}
+
+	deletedIDs, err := deletion.Delete(args, deleteFunc, resource.ApiKey)
+
+	errs := multierror.Append(err, c.deleteKeysFromKeyStore(deletedIDs))
```
brianstrauch, line 48: I'm not sure we have coverage for the multiple-error case?

sgagniere: I just added a multierror integration test for `environment delete`.

For the specific scenario here (appending more errors after `Delete`, it's not really something that's easily testable in integration tests because as far as I can tell, all instances where we do post-delete tasks only return errors if the user isn't logged in or if `Save()` had an error.

pkg/resource/resource.go
```diff
@@ -74,3 +98,76 @@ func LookupType(resourceId string) string {
...
+
+	return errors.NewErrorWithSuggestions(invalidArgsErrMsg, invalidResourceSuggestion)
+}
+
+func Plural(resource string) string {
+	if resource == "" {
+		return ""
+	}
+
+	for suffix := range pluralExtraSyllableSuffix {
```
brianstrauch, line 166: Should this slice be defined in this function?

pkg/resource/resource.go
```diff
@@ -74,3 +98,76 @@ func LookupType(resourceId string) string {
...
+
+	if len(invalidArgs) != 0 {
+		return ResourcesNotFoundError(c, resourceType, invalidArgs...)
+	}
+
+	return nil
+}
+
+func ResourcesNotFoundError(c *cobra.Command, resourceType string, invalidArgs ...string) error {
+	NotFoundErrorMsg := `%s %s not found`
```
brianstrauch, line 140:
```suggestion
	notFoundErrorMsg := `%s %s not found`
```

brianstrauch (approved):

internal/pipeline/command_delete.go
```diff
@@ -48,18 +48,29 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
...
-	output.Printf(errors.RequestedDeleteResourceMsg, resource.Pipeline, args[0])
-	return nil
+	deleteFunc := func(id string) error {
+		return c.V2Client.DeleteSdPipeline(environmentId, cluster.ID, id)
+	}
+
+	deletedIds, err := deletion.DeleteWithoutMessage(args, deleteFunc)
+	deleteMsg := "Requested to delete %s %s.\n"
+	if len(deletedIds) == 1 {
+		output.Printf(deleteMsg, resource.Pipeline, fmt.Sprintf("\"%s\"", deletedIds[0]))
```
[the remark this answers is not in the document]
sgagniere: Hmm, I did this because the plural version calls `ArrayToCommaDelimitedString`, which puts quotation marks around everything, and I wanted it to be consistent.

Perhaps I should remove the quotations from that function in a followup PR, because it's not actually what we would expect to happen just based on the name...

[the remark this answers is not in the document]
sgagniere: Lol oops, I completely misunderstood.

PR #1793, opened 2023-03-16
[CLI-2137] remove prefix check for `anonymous-id`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Allow ``confluent kafka topic produce`` to run when using a context created with ``confluent context create``

Checklist
---------
   * yes: ok

What
----
The `anonymous-id` Kafka cluster id should not be checked for the `lkc-` prefix.

References
----------

Test & Review
-------------
Manual testing using the steps outlined in the ticket.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -834,6 +834,12 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
 		var clusterId string
 		switch ctx.Credential.CredentialType {
 		case v1.APIKey:
+			if cmd.Flags().Changed("cluster") {
+				output.ErrPrintln("WARNING: The `--cluster` flag is not parsed when using API key credentials.")
+			}
+			if cmd.Flags().Changed("environment") {
+				output.ErrPrintln("WARNING: The `--environment` flag is not parsed when using API key credentials.")
+			}
 			clusterId = r.getClusterIdForAPIKeyCredential(ctx)
 		case v1.Username:
 			unsafeTrace, err := cmd.Flags().GetBool("unsafe-trace")
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -838,7 +838,7 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
 				output.ErrPrintln("WARNING: The `--cluster` flag is ignored when using API key credentials.")
 			}
 			if cmd.Flags().Changed("environment") {
-				output.ErrPrintln("WARNING: The `--environment` flag is not parsed when using API key credentials.")
+				output.ErrPrintln("WARNING: The `--environment` flag is ignored when using API key credentials.")
 			}
 			clusterId = r.getClusterIdForAPIKeyCredential(ctx)
 		case v1.Username:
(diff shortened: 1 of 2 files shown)


MuweiHe (approved):

brianstrauch (changes requested):
I don't think this will fix the customer's issue. Can you explain?

sgagniere:
@brianstrauch I elaborated on this in the thread.

internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -108,7 +108,7 @@ func (d *DynamicContext) GetKafkaClusterForCommand() (*v1.KafkaClusterConfig, er
 	}
 
 	cluster, err := d.FindKafkaCluster(clusterId)
-	if presource.LookupType(clusterId) != presource.KafkaCluster {
+	if presource.LookupType(clusterId) != presource.KafkaCluster && clusterId != "anonymous-id" {
```
brianstrauch, line 111: Taking this check away will show an error that says something along the lines of "missing lkc- prefix", even if the user passes `--cluster lkc-12345`, right? Sounds like we're just replacing one bad error with another.

sgagniere: I do not see any more errors in `kafka topic produce` using a bootstrapped context with this change.

On the `--cluster` flag: it doesn't actually do anything here because `produce` is a `HasAPIKeyCLICommand` which doesn't parse `--environment` and `--cluster` flags when the credential is an API key instead of a username. Running `produce` from this context is using an API key credential.

I might put in a warning message about this in the APIKey credential case if these flags are supplied because it was pretty confusing to me as well for a bit why the flag had no effect.

brianstrauch: So there's no way to select a different Kafka cluster?

sgagniere: Nope. Bootstrapped contexts are tied to a specific cluster.

brianstrauch: Alright, let's add a warning message that says that!

internal/pkg/cmd/prerunner.go
```diff
@@ -834,6 +834,12 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
 		var clusterId string
 		switch ctx.Credential.CredentialType {
 		case v1.APIKey:
+			if cmd.Flags().Changed("cluster") {
+				output.ErrPrintln("WARNING: The `--cluster` flag is not parsed when using API key credentials.")
```
sgagniere, line 838:
```suggestion
				output.ErrPrintln("WARNING: The `--cluster` flag is ignored when using API key credentials.")
```

internal/pkg/cmd/prerunner.go
```diff
@@ -834,6 +834,12 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
 		var clusterId string
 		switch ctx.Credential.CredentialType {
 		case v1.APIKey:
+			if cmd.Flags().Changed("cluster") {
+				output.ErrPrintln("WARNING: The `--cluster` flag is ignored when using API key credentials.")
+			}
+			if cmd.Flags().Changed("environment") {
+				output.ErrPrintln("WARNING: The `--environment` flag is not parsed when using API key credentials.")
```
sgagniere, line 841:
```suggestion
				output.ErrPrintln("WARNING: The `--environment` flag is ignored when using API key credentials.")
```

PR #1795, opened 2023-03-16
Fix incompatible `--nodebuginfo` flag in debian packaging

brianstrauch:
Checklist
---------
   * yes: ok

What
----
If EL7, use `--define "debug_package %{nil}"`, else (EL8) use `--nodebuginfo`.
https://stackoverflow.com/questions/36983051/rpmbuild-how-to-skip-generation-of-debuginfo-packages-without-change-spec-fi

Test & Review
-------------
Not sure how to test, need help from DevProd team.


DABH (approved):
Do you need to run `make update-debian-patch` or whatever?

brianstrauch:
> Do you need to run make update-debian-patch or whatever?

*sigh*... Yes, I always forget. Thanks.

PR #1796, opened 2023-03-16
[CLI-2356] Fix inability to hard delete schema versions

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix an issue where ``confluent schema-registry schema delete --permanent`` prevents permanent deletion of schema versions

Checklist
---------
   * yes: ok

What
----
Existence check fails for soft deleted schema versions without the `Deleted` optional argument in the `GET` request.

Unfortunately, using this optional parameter will still result in successful calls for non deleted schema versions in addition to soft deleted ones. So for the purposes of existence checking, we proceed when the `--permanent` flag is used if:

`GetSchemaByVersion` with `Deleted` returns no error and `GetSchemaByVersion` without `Deleted` returns an error.

I also updated the confirmation message to say "permanently delete" when that flag is supplied.

References
----------

Test & Review
-------------
Manual testing

Attempt to permanently delete a version that hasn't been soft deleted (expect failure):
```
confluent sr schema delete --subject test1 --version 1 --permanent
Error: you must first soft delete a schema version before you can hard delete it
```
Soft delete it (expect success):
```
confluent sr schema delete --subject test1 --version 1            
Are you sure you want to delete schema "test1 (version 1)"?
To confirm, type "test1". To cancel, press Ctrl-C: test1
Successfully soft deleted version "1" for subject "test1".
  Version  
-----------
        1  
```
Try to soft delete it again (expect failure):
```
confluent sr schema delete --subject test1 --version 1
Error: Schema Registry subject or version not found

Suggestions:
    List available subjects with `confluent schema-registry subject list`.
    List available versions with `confluent schema-registry subject describe`.
```
Now  hard delete it (expect success):
```
confluent sr schema delete --subject test1 --version 1 --permanent
Are you sure you want to permanently delete schema "test1 (version 1)"?
To confirm, type "test1". To cancel, press Ctrl-C: test1
Successfully hard deleted version "1" for subject "test1".
  Version  
-----------
        1  
```
Try to hard delete it again (expect failure):
```
confluent sr schema delete --subject test1 --version 1 --permanent
Error: Schema Registry subject or version not found

Suggestions:
    List available subjects with `confluent schema-registry subject list`.
    List available versions with `confluent schema-registry subject describe`.
```
Try to hard delete a version that doesn't exist (expect failure):
```
confluent sr schema delete --subject test1 --version 3 --permanent
Error: Schema Registry subject or version not found

Suggestions:
    List available subjects with `confluent schema-registry subject list`.
    List available versions with `confluent schema-registry subject describe`.
```


internal/cmd/schema-registry/command_schema_delete.go
```diff
@@ -67,24 +67,33 @@ func deleteSchema(cmd *cobra.Command, srClient *srsdk.APIClient, ctx context.Con
...
+
 	checkVersion := version
 	if version == "all" {
 		// check that at least one version for the input subject exists
 		checkVersion = "latest"
 	}
-	_, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil)
-	if err != nil {
+	if permanent {
+		getOpts := &srsdk.GetSchemaByVersionOpts{Deleted: optional.NewBool(true)}
```
brianstrauch, line 81:
```suggestion
		opts := &srsdk.GetSchemaByVersionOpts{Deleted: optional.NewBool(true)}
```

internal/cmd/schema-registry/command_schema_delete.go
```diff
@@ -67,24 +67,33 @@ func deleteSchema(cmd *cobra.Command, srClient *srsdk.APIClient, ctx context.Con
...
+		if _, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, getOpts); err != nil {
+			return errors.CatchSchemaNotFoundError(err, httpResp)
+		} else if _, _, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil); err == nil {
+			return errors.New("you must first soft delete a schema version before you can hard delete it")
```
brianstrauch, line 85: Wait really? That seems like a silly requirement.

brianstrauch: Does the backend not return an equivalent error message?

brianstrauch: Nit: Match the wording with the flag name.
```suggestion
			return errors.New("you must first soft delete a schema version before you can permanently delete it")
```

sgagniere: > Wait really? That seems like a silly requirement.

Yeah. If you try to delete one you get the following backend error: `"Subject 'test3' Version 1 was not deleted first before being permanently deleted`.

> Does the backend not return an equivalent error message?

The headache here is that I can't get that error without calling the delete endpoint, and for the purposes of checking existence beforehand the get endpoint returns no error. So I have to infer whether or not it was soft deleted with this unfortunate second call if I know that it wasn't hard deleted after the first call.

internal/cmd/schema-registry/command_schema_delete.go
sgagniere: Thanks! This was an artifact from an earlier iteration where I didn't want it conflict with `opts` defined further down and I forgot to switch it back after I moved it inside an if block.

internal/cmd/schema-registry/command_schema_delete.go
brianstrauch: Can we let the backend team know about this? Maybe they can help fix some error messages?

sgagniere: I'll make a ticket asking if they can update the sdk, but otherwise I think this is working as intended. The `deleted` option for GET requests seems to be just "return the version even if it was soft-deleted" and not "return the version only if it was soft-deleted"

brianstrauch (approved):
+1 provided tests pass... speaking of which, we should definitely have some integration tests here.

PR #1798, opened 2023-03-17
Fix regex in `make promote`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The regular expression was treating "." as a wildcard, not a period. They needed to be escaped.

Test & Review
-------------
Tested on `dry-run` branch!


sgagniere (approved):
Just tried it out and it works, thanks

PR #1801, opened 2023-03-17
Prevent panic in `api-key list` and `describe`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Prevent a panic which could occur in various ``confluent api-key`` commands

Checklist
---------
   * yes: ok

What
----
If `State.Auth.User` is `nil` for whatever reason, a panic will result in `getAllUsers` when that function attempts to access the `User.Id` field.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------
Followup: investigate the cause of incorrectly populated contexts.

### DIFF
--- a/internal/cmd/api-key/command_list.go
+++ b/internal/cmd/api-key/command_list.go
@@ -153,10 +153,9 @@ func getServiceAccountsMap(serviceAccounts []iamv2.IamV2ServiceAccount) map[stri
 func getUsersMap(users []*ccloudv1.User) map[int32]*ccloudv1.User {
 	userMap := make(map[int32]*ccloudv1.User)
 	for _, user := range users {
-		if user == nil {
-			continue
+		if user != nil {
+			userMap[user.GetId()] = user
 		}
-		userMap[user.GetId()] = user
 	}
 	return userMap
 }
--- a/internal/cmd/api-key/command_list.go
+++ b/internal/cmd/api-key/command_list.go
@@ -153,7 +153,10 @@ func getServiceAccountsMap(serviceAccounts []iamv2.IamV2ServiceAccount) map[stri
 func getUsersMap(users []*ccloudv1.User) map[int32]*ccloudv1.User {
 	userMap := make(map[int32]*ccloudv1.User)
 	for _, user := range users {
-		userMap[user.Id] = user
+		if user == nil {
+			continue
+		}
+		userMap[user.GetId()] = user
 	}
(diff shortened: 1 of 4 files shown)


internal/pkg/config/v1/context_state.go
```diff
@@ -19,6 +21,20 @@ type ContextState struct {
 	Nonce            []byte      `json:"nonce,omitempty"`
 }
 
+func (c *ContextState) GetAuth() *AuthConfig {
```
brianstrauch, line 24: This function already exists for `Context`!

internal/pkg/config/v1/context_state.go
```diff
@@ -19,6 +21,20 @@ type ContextState struct {
...
+	if c != nil {
+		return c.Auth
+	}
+	return nil
+}
+
+func (c *ContextState) GetUser() *ccloudv1.User {
```
brianstrauch, line 31: Same with this one!

internal/cmd/api-key/command_list.go
```diff
@@ -153,7 +153,9 @@ func getServiceAccountsMap(serviceAccounts []iamv2.IamV2ServiceAccount) map[stri
 func getUsersMap(users []*ccloudv1.User) map[int32]*ccloudv1.User {
 	userMap := make(map[int32]*ccloudv1.User)
 	for _, user := range users {
-		userMap[user.Id] = user
+		if user != nil {
```
brianstrauch, line 156: This check seems redundant given the other check made in this PR.

brianstrauch: I'd rather use the getter syntax than adding defensive `if` statements everywhere.

internal/pkg/config/v1/context_state.go
sgagniere: Oh right, I'll switch it around.

internal/cmd/api-key/command_list.go
sgagniere: It's mostly redundant, but it's to preempt another edge case: if the backend returns a `nil` user as part of the list.

sgagniere: Ok I removed the if. It still has the getter function.

brianstrauch (approved):

PR #1808, opened 2023-03-20
Check salt length for old contexts as well to avoid incorrect salt length error in decryption

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fixed a decryption error when running ``confluent context list`` with old contexts from `~/.confluent/config.json`

Checklist
---------
   * yes: ok

What
----

when logged in with older than 3.2.0, then you run newer 3.2.0, new salt and nonce will only be generated for current context that you're logged into. When running `context list`, other old contexts are only being read and i.e. some random mock contexts with mock tokens (not jwt format) would be decrypted, while there's no salt or nonce used to generate them in the first place. Add a check when decrypting tokens, only decrypt when: it is a valid format token, it's not nil, and it already has salt/nonce that were used to generate the token.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/config/v1/context_state.go
+++ b/internal/pkg/config/v1/context_state.go
@@ -21,7 +21,7 @@ type ContextState struct {
 
 func (c *ContextState) DecryptContextStateAuthToken(ctxName string) error {
 	reg := regexp.MustCompile(authTokenRegex)
-	if !reg.MatchString(c.AuthToken) && c.AuthToken != "" {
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
Has this been tested?

MuweiHe:
> Has this been tested?

yes i reproduced the issue using our mock contexts left in config.json after running integration tests.. and this solves that issue.

PR #1809, opened 2023-03-21
Adjust Makefiles to use AWS SSO instead of gimme-aws-creds

lyoung-confluent:
https://confluent.slack.com/archives/C038ZJ00P/p1679368991765529

### DIFF
--- a/mk-files/release.mk
+++ b/mk-files/release.mk
@@ -3,7 +3,7 @@ ARCHIVE_TYPES=darwin_amd64.tar.gz darwin_arm64.tar.gz linux_amd64.tar.gz linux_a
 # If you set up your laptop following https://github.com/confluentinc/cc-documentation/blob/master/Operations/Laptop%20Setup.md
 # then assuming caas.sh lives here should be fine
 define aws-authenticate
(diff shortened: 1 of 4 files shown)


mk-files/release.mk
```diff
@@ -3,7 +3,7 @@ ARCHIVE_TYPES=darwin_amd64.tar.gz darwin_arm64.tar.gz linux_amd64.tar.gz linux_a
 # If you set up your laptop following https://github.com/confluentinc/cc-documentation/blob/master/Operations/Laptop%20Setup.md
 # then assuming caas.sh lives here should be fine
 define aws-authenticate
-	source ~/git/go/src/github.com/confluentinc/cc-dotfiles/caas.sh && if ! aws sts get-caller-identity; then eval $$(gimme-aws-creds --output-format export --roles "arn:aws:iam::050879227952:role/administrator"); fi
+	source ~/git/go/src/github.com/confluentinc/cc-dotfiles/caas.sh && AWS_PROFILE=cc-production-1/prod-administrator aws sts get-caller-identity >/dev/null
```
brianstrauch, line 6: @lyoung-confluent I assume we don't need to run `aws sts get-caller-identity` here anymore? From what I've heard, `AWS_PROFILE=cc-production-1/prod-administrator` should be all we need.

lyoung-confluent: Yes it's _probably_ not needed, though this has the benefit of causing the auth to trigger/occur right here instead of at the first `aws` command that is run. Up to you

brianstrauch (approved):

PR #1810, opened 2023-03-22
`DRY_RUN=true` mode for releases

brianstrauch:
What
----
Prefix any release make target with `DRY_RUN=true` to avoid making external changes (opening PRs, pushing tags, etc). Makes testing less stressful, and should improve iteration speed.

Note: `make gorelease`, and by extension `make release-to-stag` and `make release` do not work due to https://github.com/goreleaser/goreleaser/issues/3883 (although I'd still like to merge this since running `DRY_RUN=true make gorelease` will fail without uploading anything)

Test & Review
-------------
Ran all make targets with `DRY_RUN=true` and validated that they didn't make any non-local changes.


mk-files/release-test.mk
```diff
@@ -18,13 +18,13 @@ verify-archive-installer:
 # if ARCHIVES_VERSION is empty, latest folder will be tested
 .PHONY: test-installer
 test-installer:
-	@echo Running packaging/installer tests
-	@bash test-installer.sh $(ARCHIVES_VERSION)
+	$(call dry-run,bash test-installer.sh $(ARCHIVES_VERSION))
```
MuweiHe, line 21: How is this done not downloading from prod/stag?

brianstrauch: Not sure what you mean, but I chose to gate this with the `DRY_RUN` flag because since nothing has been uploaded, we can't test downloading.

brianstrauch: This is something that we should wait to do a release to test. We can't really test it without first uploading the new binary.

Of course, we could always test it using the old binary, but that seems counterintuitive to me...

PR #1811, opened 2023-03-22
Prevent panics in api key and kafka commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Prevent panics in ``confluent kafka cluster use`` and ``confluent api-key`` commands for users without an environment

Checklist
---------
   * yes: ok

What
----
`Auth.Account` is nil for users without an environment (such as any user without a role-binding). Hence, commands which call `FindKafkaCluster` result in panic for these users.

References
----------

Test & Review
-------------
Added a new integration test
Manual verification by deleting all role-binds for one of my users and checking that the panics are replaced with errors.


internal/pkg/cmd/prerunner.go
```diff
@@ -168,8 +168,7 @@ func (c *AuthenticatedCLICommand) AuthToken() string {
 
 func (c *AuthenticatedCLICommand) EnvironmentId() string {
 	if c.Context.GetEnvironment() == nil {
-		noEnvSuggestions := errors.ComposeSuggestionsMessage("This issue may occur if this user has no valid role bindings. Contact an Organization Admin to create a role binding for this user.")
-		output.ErrPrintln(errors.EnvNotSetErrorMsg + noEnvSuggestions)
+		output.ErrPrintln(errors.EnvNotSetErrorMsg + errors.ComposeSuggestionsMessage(errors.EnvNotSetSuggestions))
```
brianstrauch, line 171: Let's prefix this with `WARNING:` for now, and then in the future when we display warnings by default, we can convert this to a real warning.

internal/pkg/errors/error_message.go
```diff
@@ -66,7 +66,8 @@ const (
 	EnvNotFoundErrorMsg            = `environment "%s" not found`
 	OrgResourceNotFoundSuggestions = "List available %[1]ss with `confluent %[1]s list`."
 	EnvSwitchErrorMsg              = "failed to switch environment: failed to save config"
-	EnvNotSetErrorMsg              = "this command requires an environment; no environments found"
+	EnvNotSetErrorMsg              = "an environment is required; no environments found"
```
brianstrauch, line 69: I feel like erroring with "no environment found" implies that an environment was required.
```suggestion
	EnvNotSetErrorMsg              = "no environment found"
```

test/test-server/ccloud_handlers.go
```diff
@@ -98,14 +98,18 @@ func (c *CloudRouter) HandleMe(t *testing.T, isAuditLogEnabled bool) http.Handle
 			org.Marketplace = &ccloudv1.Marketplace{Partner: ccloudv1.MarketplacePartner_AWS}
 		}
 
+		accounts := environments
+		if os.Getenv("CONFLUENT_CLOUD_EMAIL") == "new@user.com" {
```
brianstrauch, line 102: How about something more helpful?
```suggestion
		if os.Getenv("CONFLUENT_CLOUD_EMAIL") == "no-environment-user@example.com" {
```

test/test-server/ccloud_handlers.go
```diff
@@ -98,14 +98,18 @@ func (c *CloudRouter) HandleMe(t *testing.T, isAuditLogEnabled bool) http.Handle
...
+			accounts = []*ccloudv1.Account{}
```
brianstrauch, line 103: We're trying to move away from this old nomenclature.
```suggestion
			environments = []*ccloudv1.Account{}
```

brianstrauch (changes requested):

test/test-server/ccloud_handlers.go
sgagniere: My rational for this was that new users are naturally in this state where they have no environments.

But I'll make this change.

internal/pkg/errors/error_message.go
```diff
@@ -66,7 +66,7 @@ const (
 	EnvNotFoundErrorMsg            = `environment "%s" not found`
 	OrgResourceNotFoundSuggestions = "List available %[1]ss with `confluent %[1]s list`."
 	EnvSwitchErrorMsg              = "failed to switch environment: failed to save config"
-	EnvNotSetErrorMsg              = "an environment is required; no environments found"
+	EnvNotSetErrorMsg              = "no environment found"
```
brianstrauch, line 69:
```suggestion
	NoEnvironmentFoundErrorMsg              = "no environment found"
```

internal/pkg/cmd/prerunner.go
sgagniere: We could, but it would end up being more verbose I think since we're not actually returning an error here. So we would have to print the error message and then also call `DisplaySuggestionsMessage`.

MuweiHe:
does it show the warning for those commands that don't actually require an active environment?

sgagniere:
@MuweiHe It shouldn't, because the warning/error are generated inside `EnvironmentId()` and `AuthenticatedEnvId()`. Any command which ends up calling these functions should be one which has a need for an environment.

internal/pkg/cmd/prerunner.go
brianstrauch: Can't we do something like the following?
1. `err := NewErrorWithSuggestions()`
2. `output.Println(err)`

sgagniere: Just tried
```
err := errors.NewErrorWithSuggestions(errors.NoEnvironmentFoundErrorMsg, errors.NoEnvironmentFoundSuggestions)
output.ErrPrintln(err)
```
and the suggestion doesn't print:
```
no environment found
```
Same result if I do `output.ErrPrintln(err.Error())`.

Since the suggestion is stored in the `ErrorWithSuggestionImpl` struct I also tried casting to (a pointer to) this struct but it still only prints the error: `output.ErrPrintln(err.(*errors.ErrorWithSuggestionsImpl))`.

test/fixtures/output/environment/11-prompt.golden
```diff
@@ -1,6 +1,6 @@
 Are you sure you want to delete environment "not-595"?
 To confirm, type "default". To cancel, press Ctrl-C: Deleted environment "not-595".
-this command requires an environment; no environments found
+WARNING: no environment found
```
brianstrauch, line 3: Shouldn't this be an error and not a warning? Doesn't the operation fail?

sgagniere: The function w/ this ultimately gets called in dozens of files, so I'm not completely sure if there are any which can continue or if all of them invariably fail without this id.

This function returns only a string and no error, though, so I can start working on a followup PR to convert this to an error and add error handling to its calls but we should probably merge without it and this can go in next week's release.

brianstrauch: I'm guessing that 99% of the time this should be an error and not a warning. Returning a `(string, error)` is definitely the way to go here (but it'll be a decent amount of work)!

brianstrauch (approved):

PR #1813, opened 2023-03-24
Release improvements

brianstrauch:
Checklist
---------
   * yes: ok

What
----
* Remove bump.txt (this can be inferred, for instance a patch version doesn't end with ".0")
* Fix GitHub release notes header, remove "=======" above title
* Combine release notes and command reference into one PR
* Update install script automatically
* Replace release.svg with [Release](https://github.com/confluentinc/cli/releases/latest), remove need to commit before every release
* Make underlines the same length as headers
* Standardize newlines between docs and S3 release notes, combine into the same function
* Automatically convert single backticks to double backticks for docs

Test & Review
-------------
Tested all modified targets with `DRY_RUN=true` enabled


mk-files/release-notes.mk
```diff
@@ -5,23 +5,17 @@ release-notes:
...
 	cd - && \
 	cp release-notes/release-notes.rst $(DOCS_CONFLUENT_CLI) && \
 	cd $(DOCS_CONFLUENT_CLI) && \
 	git commit -am "New release notes for v$${version}" && \
-	$(call dry-run,git push -u origin cli-v$${version}-release-notes) && \
-	base="master" && \
-	if [ "$${bump}" = "patch" ]; then \
-		base=$$(echo $${version} | sed $(STAGING_BRANCH_REGEX)); \
-	fi && \
-	$(call dry-run,gh pr create -B $${base} --title "New release notes for v$${version}" --body "")
```
sgagniere, line 24: So this target will now update a branch, and it's the target at the end of `make release` that opens the combined PR?

sgagniere (approved):
+1

mk-files/release-notes.mk
brianstrauch: Yep, exactly!

PR #1814, opened 2023-03-24
Change missing environment warning to an error

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Followup to a previous PR fixing panics.

Instead of displaying a warning about missing environments and letting the CLI fail further down the line, this PR just returns that error directly.

There are a few places where the error is ignored:
- in autocompletion code, if there is an error it just returns nil instead of the error.
- in `confluent environment` functions, the calls to `EnvironmentId()` happen after successful http requests and is used to indicate if the environment in question is the current one; it would be confusing to return an error here so the error is just ignored.

References
----------

Test & Review
-------------
Updated one golden file and ran tests

Open Questions / Follow-ups
---------------------------


internal/pkg/cmd/flags.go
```diff
@@ -25,15 +25,15 @@ func AddApiKeyFlag(cmd *cobra.Command, c *AuthenticatedCLICommand) {
...
-		return AutocompleteApiKeys(c.EnvironmentId(), c.V2Client)
+		return AutocompleteApiKeys(c.V2Client)
 	})
 }
 
 func AddApiSecretFlag(cmd *cobra.Command) {
 	cmd.Flags().String("api-secret", "", "API key secret.")
 }
 
-func AutocompleteApiKeys(environment string, client *ccloudv2.Client) []string {
```
brianstrauch, line 36: :+1:

brianstrauch (approved):
I thought there would be more usages of this function. Can you check if there is any code that _should_ be using this function?

PR #1815, opened 2023-03-24
Update CLI from GitHub

brianstrauch:
Checklist
---------
   * yes: ok

What
----
* On `confluent update`, and every 24h, fetch CLI releases from GitHub instead of S3
* Use https://github.com/inconshreveable/go-update to update binaries in-place instead of hand-written solution
* Retire `~/.confluent/update_check` in favor of a `last_update_check` configuration field

Test & Review
-------------
Unit tests + manual testing


internal/cmd/update/command.go
```diff
@@ -89,89 +65,197 @@ func (c *command) update(cmd *cobra.Command, _ []string) error {
...
+	for i, release := range releases {
+		version, err := getReleaseVersion(release)
 		if err != nil {
-			errMsg = fmt.Sprintf(errors.ReleaseNotesVersionCheckErrorMsg, err)
+			continue
 		}
-		if !isSameVersion {
-			errMsg = fmt.Sprintf(errors.ReleaseNotesVersionMismatchErrorMsg, latestBinaryVersion, latestReleaseNotesVersion)
+
+		if !found && (major || version.Major == current.Major) {
```
sgagniere, line 196: Is `releases` initially newest-first? If so, I'm wondering what would happen if the user didn't choose `--major` and the 30th release satisfies `version.Major == current.Major`; it seems like we would get `lo == hi` and end up returning nothing.

If it's initially sorted oldest-first then ignore this comment.

sgagniere (approved):

internal/cmd/update/command.go
brianstrauch: It's newest first.

> I'm wondering what would happen if the user didn't choose --major and the 30th release satisfies version.Major == current.Major

Yeah, we'd have to iterate through all 30 in that case. Not sure if there's any way we can only fetch minor versions... let me look into it, but I don't think it's possible.

brianstrauch:
This turned out to be extremely variable… sometimes a download would take 2s, other times it would take 2m. This is unacceptable for production. Maybe we can come back to this when GitHub download speeds improve.

PR #1816, opened 2023-03-24
[CIAM-2686] Add identity namespace behind an LD flag

lucy-fan:
Release Notes
-------------

Breaking Changes
- N/A

New Features
- N/A (adds a new namespace but is hidden behind LD flag so not user-facing)

Bug Fixes
- N/A

Checklist
---------
   * yes: ok

What
----
We are adding a new namespace `identity` behind an LD flag so that during EA we can introduce customers to the new role, `IdentityAdmin`,  in this namespace.

This PR adds the `identity` namespace to the `role describe` and `role list`.

References
----------

Test & Review
-------------
Tested manually in stag.
```
confluent iam rbac role describe IdentityAdmin               
+---------------+-----------------------------------------------------------------+
| Name          | IdentityAdmin                                                   |
| Access Policy | [                                                               |
|               |   {                                                             |
|               |     "bindingScope": "organization",                             |
|               |     "bindWithResource": false,                                  |
|               |     "allowedOperations": [                                      |
|               |       {                                                         |
|               |         "resourceType": "CloudApiKey",                          |
|               |         "operations": ["Describe", "Alter", "Delete", "Create"] |
|               |       },                                                        |
|               |       {                                                         |
|               |         "resourceType": "User",                                 |
|               |         "operations": ["Describe", "Create", "Invite"]          |
|               |       },                                                        |
|               |       {                                                         |
|               |         "resourceType": "Organization",                         |
|               |         "operations": ["Describe"]                              |
|               |       },                                                        |
|               |       {                                                         |
|               |         "resourceType": "ServiceAccount",                       |
|               |         "operations": ["Describe", "Alter", "Create"]           |
|               |       }                                                         |
|               |     ]                                                           |
|               |   }                                                             |
|               | ]                                                               |
|               |                                                                 |
+---------------+-----------------------------------------------------------------+

```

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/iam/command_rbac_role_describe.go
+++ b/internal/cmd/iam/command_rbac_role_describe.go
@@ -11,7 +11,9 @@ import (
 	"github.com/confluentinc/mds-sdk-go-public/mdsv2alpha1"
 
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/featureflags"
 	"github.com/confluentinc/cli/internal/pkg/output"
 )
 
@@ -49,6 +51,12 @@ func (c *roleCommand) ccloudDescribe(cmd *cobra.Command, role string) error {
 		ksqlNamespace.Value(),
 		publicNamespace.Value(),
 		streamCatalogNamespace.Value(),
+	}
+
+	// check if IdentityAdmin is enabled
+	ldClient := v1.GetCcloudLaunchDarklyClient(c.Context.PlatformName)
+	if featureflags.Manager.BoolVariation("auth.rbac.identity_admin.enable", c.Context, ldClient, true, false) {
+		namespacesList = append(namespacesList, identityNamespace.Value())
 	}
 
 	namespaces := optional.NewString(strings.Join(namespacesList, ","))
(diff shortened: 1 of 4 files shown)


internal/cmd/iam/command_rbac_role_describe.go
```diff
@@ -51,6 +53,12 @@ func (c *roleCommand) ccloudDescribe(cmd *cobra.Command, role string) error {
 		streamCatalogNamespace.Value(),
 	}
 
+	// check if IdentityAdmin is enabled
+	ldClient := v1.GetCcloudLaunchDarklyClient(c.Context.PlatformName)
+	if featureflags.Manager.BoolVariation("auth.rbac.identity_admin.enable", c.Context, ldClient, true, false) {
+		namespacesList = append(namespacesList, identityNamespace.Value())
```
brianstrauch, line 59: Would it be possible to create an API that returns the list of namespaces, so we don't have to keep making trivial PRs like this one?

brianstrauch (approved):

brianstrauch:
Nit: Can we get rid of the newline at the end of the "Access Policy"?

internal/cmd/iam/command_rbac_role_describe.go
lucy-fan: We use new namespaces to "hide" new roles until they are ready for GA, or if we intend to use it for internal roles only. You can see all the actual namespaces here: https://github.com/confluentinc/ce-kafka/blob/master/ce-rbac/src/main/java/io/confluent/security/rbac/Role.java

Specifically for this role it is planned on being moved to the `public` namespace as a post-GA task so we would just have to delete these lines and update tests.

PR #1817, opened 2023-03-24
Install from GitHub instead of S3

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Now that `confluent` v1 has been deprecated, and all versions after `v2.0.0` are downloadable from GitHub, we can change the installer script to use GitHub instead of S3.

This will allow us to stop uploading to S3 in the future (once the last version of the CLI that gets updates from S3 is deprecated, which is probably not for a while)

Test & Review
-------------
`bash test-installer.sh` (CI will also run this from Linux)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,20 +5,9 @@ set -e
 # Manually modified to download from S3 for confluentinc/cli goreleaser config
 #
 # The major modifications include:
-# * reworked `github_release` into `s3_release` function
-# * updated `TARBALL_URL` and `CHECKSUM_URL` to point to S3 instead of GitHub API
-# * added a new `-l` flag to list versions from S3, since we can't link to our (private) GitHub repo
+# * added a new `-l` flag to list versions
 # * extracted a `BINARY` variable instead of having binary names hardcoded in `execute`
-# * updated version/tag handling of the `v` prefix; it's expected in GitHub and inconsistently used in S3
 # * updated the usage message, logging, and file comments a bit
-
-S3_URL=https://s3-us-west-2.amazonaws.com/confluent.cloud
-S3_CONTENT_CHECK_URL="${S3_URL}?prefix="
-if [ -n "$OVERRIDE_S3_FOLDER" ]
-then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -133,22 +114,11 @@ adjust_os() {
...
-}
-s3_release() {
-  version=$1
-  test -z "$version" && version="latest"
-  s3url="${S3_CONTENT_CHECK_URL}${PROJECT_NAME}/archives/${version#v}/&delimiter=/"
-  xml=$(http_copy "$s3url")
-  exists=$(echo "$xml" | grep "<Key>") || return 1
-  test -z "$version" && return 1
-  echo "$version"
+releases() {
```
sgagniere, line 117: I ran `./install.sh -l` because I figured it's easier to parse the regex when I know what the match looks like, but I see an empty list. Running `curl -s https://api.github.com/repos/confluentinc/cli/releases` on it's own I see a rate limit error message.

install.sh
```diff
@@ -406,23 +376,20 @@ main() {
...
   adjust_os
 
-  if [ ${OS} = "windows" ]; then
-    FORMAT=zip ;
-  fi
-
-  log_info "found version: ${VERSION} for ${TAG}/${OS}/${ARCH}"
-
   # If < v3, archive version is prefixed with "v"
+  VERSION=${TAG}
```
sgagniere, line 380: This doesn't seem to work if the tag is `latest` because then it tries to download `TARBALL=${BINARY}_latest_${OS}_${ARCH}.${FORMAT}`

install.sh
brianstrauch: How many times did you run this? I must've ran it over 20 times, and I never got a rate limit error.

sgagniere: Around 5 times.

sgagniere: I just waited an hour and it works.

brianstrauch:
This turned out to be extremely variable… sometimes a download would take 2s, other times it would take 2m. This is unacceptable for production. Maybe we can come back to this when GitHub download speeds improve.

PR #1818, opened 2023-03-27
WIP: Add initial implementation of cli script.

conradwt:
This PR supports the following features:

- add initial implementation cli install

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+
+uname_arch() {
+  arch=$(uname -m)
+
+  case $arch in
+    x86_64) arch="amd64" ;;
+    x86) arch="386" ;;
+    i686) arch="386" ;;
+    i386) arch="386" ;;
+    aarch64) arch="arm64" ;;
+    armv*) arch="armv6" ;;
+    armv*) arch="armv6" ;;
+    armv*) arch="armv6" ;;
+  esac
+
+  echo ${arch}
+}
+
+uname_os() {
+  os=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+  echo "$os"
+}
+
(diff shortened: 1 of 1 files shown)


brianstrauch:
Thanks for the PR! We already have an installation script (see install.sh) that installs the CLI from S3. We're transitioning to GitHub [here](https://github.com/confluentinc/cli/pull/1817). Let me know if there's anything we're missing!

conradwt:
@brianstrauch This is great to hear.  First, the install docs on GitHub are vague about how to actually install the binary.  I have been downloading and performing the following actions:

```zsh
chmod +x confluent_3.6.0_darwin_arm64 && sudo mv confluent_3.6.0_darwin_arm64 /usr/local/bin/confluent
```

However, if you're installing the binary for the first time, you should use the following from the [docs](https://docs.confluent.io/confluent-cli/current/install.html#install-confluent-cli):

```zsh
curl -sL --http1.1 https://cnfl.io/cli | sh -s -- latest
```

Note:  This isn't a very good default because it will install it in $HOME/.bin, Thus, I recommend the following instead for most Unix system users:

```zsh
curl -sL --http1.1 https://cnfl.io/cli | sh -s -- -b /usr/local/bin
```

Finally, if one needs to update an existing installation of the binary, I would definitely recommend the following instead of performing the above command:

```zsh
confluent update
```

This information should find its way into the GitHub docs and should be easier to find in the actual docs.

brianstrauch:
Thanks for the input @conradwt! I'll update the README with a few of these changes today.

> I would prefer something as easy as the following from ddosify
Can you explain why you'd prefer this over the existing 1-line install command?

brianstrauch:
Applied these suggestions in https://github.com/confluentinc/cli/pull/1827. Thanks again!

PR #1824, opened 2023-03-28
Add suggestion to check access to /etc when cipher authentication fails

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add suggestions to check CLI's access to ``/etc`` when encryption fails

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok  

What
----

Add suggestion to check access to /etc when cipher authentication fails
```
mhe@C02DV0KVMD6T cli % confluent kafka topic list
Error: cipher: message authentication failed. Check cli read/write permission to /etc.
```

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/secret/encryption_other.go
+++ b/internal/pkg/secret/encryption_other.go
@@ -92,7 +92,7 @@ func Decrypt(username, encrypted string, salt, nonce []byte) (string, error) {
 	}
 	decryptedPassword, err := aesgcm.Open(nil, nonce, cipherText, []byte(username))
 	if err != nil {
-		return "", errors.New(fmt.Sprintf("%s. %s", err.Error(), "Check cli read/write permission to /etc."))
+		return "", errors.New(fmt.Sprintf("%s. %s", err.Error(), "Check cli read/write permission to /etc/machine-id."))
 	}
 
 	return string(decryptedPassword), nil
--- a/internal/pkg/secret/encryption_other.go
+++ b/internal/pkg/secret/encryption_other.go
@@ -91,7 +91,7 @@ func Decrypt(username, encrypted string, salt, nonce []byte) (string, error) {
 	}
 	decryptedPassword, err := aesgcm.Open(nil, nonce, cipherText, []byte(username))
 	if err != nil {
-		return "", err
+		return "", errors.New(err.Error() + ". Check cli read/write permission to /etc.")
 	}
 
 	return string(decryptedPassword), nil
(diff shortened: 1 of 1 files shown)


internal/pkg/secret/encryption_other.go
```diff
@@ -91,7 +92,7 @@ func Decrypt(username, encrypted string, salt, nonce []byte) (string, error) {
 	}
 	decryptedPassword, err := aesgcm.Open(nil, nonce, cipherText, []byte(username))
 	if err != nil {
-		return "", err
+		return "", errors.New(fmt.Sprintf("%s. %s", err.Error(), "Check cli read/write permission to /etc/machine-id."))
```
brianstrauch, line 95:
```suggestion
		return "", fmt.Errorf("CLI does not have write permission for `/etc/machine-id`: %w", err)
```

brianstrauch (changes requested):

internal/pkg/secret/encryption_other.go
brianstrauch: I've used a few command line tools that will prompt the user for their system password when they run into permission issues like this one. Do you think we could try something similar here?

MuweiHe: I'm not sure this is the only root cause... it could be just that `open` failed

internal/pkg/secret/encryption_other.go
```diff
@@ -91,7 +92,7 @@ func Decrypt(username, encrypted string, salt, nonce []byte) (string, error) {
 	}
 	decryptedPassword, err := aesgcm.Open(nil, nonce, cipherText, []byte(username))
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("CLI does not have write permission for `/etc/machine-id`, or `~/.confluent/config.json` was corrupted: %w", err)
```
sgagniere, line 95:
```suggestion
		return "", fmt.Errorf("CLI does not have write permission for `/etc/machine-id`, or `~/.confluent/config.json` is corrupted: %w", err)
```

sgagniere: Nitpick: match the tense of the first and second half of this message

internal/pkg/secret/encryption_other.go
brianstrauch: Gotcha, let's just suggest that they manually add write permissions to the file.

brianstrauch: I was hoping there might be a solution that doesn't require user intervention, since we've already gotten multiple customer complaints.

brianstrauch (approved):

PR #1833, opened 2023-03-31
On `confluent update`, no longer remove CLI if path is different

brianstrauch:
Release Notes
-------------
Bug Fixes
- Fix `confluent update` for Windows

Checklist
---------
   * yes: ok

What
----
When updating from `ccloud` to `confluent`, the path was different, and we deleted `ccloud`. On Windows, the path can be different due to the difference between `confluent` and `confluent.exe`. This PR deletes that check since it's no longer relevant.

Test & Review
-------------
Waiting for validation from @sgagniere that this fixes the issue on Windows.


sgagniere (changes requested):
We no longer get the error, but there's still an issue: the new binary is missing the `.exe` extension, and running without that gives `confluent is not recognized as an internal or external command, operable program or batch file.`

PR #1835, opened 2023-03-31
Standardize commands and fix typos

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Fixes for a few annoyances and typos.


sgagniere (approved):
👍

PR #1840, opened 2023-04-04
Store current environment at top level of context

brianstrauch:
Release Notes
-------------
New Features
- Add `--context` flag to all `environment` subcommands

Checklist
---------
   * yes: ok

What
----
This is a pre-step for `use` commands that we plan to add in the future. This fixes a few problems:
* Storing the entire environment struct stores unnecessary info and bloats the configuration file
* Storing the entire environment struct can result in data getting out-of-sync with the backend if a user interacts with the UI
* Storing at the top level will allow us to store a hierarchy of the "current" resources that have been marked with the relevant `use` commands

Test & Review
-------------
Updated all relevant unit and integration tests, additional manual testing

### DIFF
--- a/internal/cmd/environment/command_update.go
+++ b/internal/cmd/environment/command_update.go
@@ -21,6 +21,8 @@ func (c *command) newUpdateCommand() *cobra.Command {
 	}
 
 	cmd.Flags().String("name", "", "New name for Confluent Cloud environment.")
+	pcmd.AddContextFlag(cmd, c.CLICommand)
+	pcmd.AddOutputFlag(cmd)
 
 	cobra.CheckErr(cmd.MarkFlagRequired("name"))
 
@@ -36,11 +38,16 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
 	}
 
 	updateEnvironment := orgv2.OrgV2Environment{DisplayName: orgv2.PtrString(name)}
-	_, httpResp, err := c.V2Client.UpdateOrgEnvironment(id, updateEnvironment)
+	environment, httpResp, err := c.V2Client.UpdateOrgEnvironment(id, updateEnvironment)
 	if err != nil {
 		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Environment, httpResp)
 	}
 
-	output.ErrPrintf(errors.UpdateSuccessMsg, "name", "environment", id, name)
-	return nil
+	table := output.NewTable(cmd)
+	table.Add(&out{
+		IsCurrent: environment.GetId() == c.Context.GetCurrentEnvironment(),
+		Id:        environment.GetId(),
+		Name:      environment.GetDisplayName(),
+	})
+	return table.Print()
 }
--- a/internal/cmd/environment/command_update.go
+++ b/internal/cmd/environment/command_update.go
@@ -6,9 +6,7 @@ import (
 	orgv2 "github.com/confluentinc/ccloud-sdk-go-v2/org/v2"
 
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
-	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/output"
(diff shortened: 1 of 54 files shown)


internal/pkg/config/v1/context.go
```diff
@@ -191,32 +198,33 @@ func (c *Context) GetSuspensionStatus() *ccloudv1.SuspensionStatus {
 	return c.GetOrganization().GetSuspensionStatus()
 }
 
-func (c *Context) GetEnvironment() *ccloudv1.Account {
-	if auth := c.GetAuth(); auth != nil {
-		return auth.Account
+func (c *Context) GetCurrentEnvironment() string {
+	if c.CurrentEnvironment != "" {
```
sgagniere, line 202:
```suggestion
	if c != nil && c.CurrentEnvironment != "" {
```

sgagniere: This shouldn't happen, but just in case we should guard against panics.

internal/pkg/config/v1/context.go
```diff
@@ -191,32 +198,33 @@ func (c *Context) GetSuspensionStatus() *ccloudv1.SuspensionStatus {
...
+	if auth := c.GetAuth(); auth != nil {
+		return auth.Account.GetId()
 	}
-	c.GetAuth().Account = environment
+
+	return ""
 }
 
-func (c *Context) GetEnvironments() []*ccloudv1.Account {
+func (c *Context) SetCurrentEnvironment(id string) {
```
sgagniere, line 213: Same comment as above. Maybe we could just return at the top if `c == nil`

internal/pkg/config/v1/context.go
```diff
@@ -191,32 +198,33 @@ func (c *Context) GetSuspensionStatus() *ccloudv1.SuspensionStatus {
...
 
-func (c *Context) SetEnvironments(environments []*ccloudv1.Account) {
-	if c.GetAuth() == nil {
-		c.SetAuth(new(AuthConfig))
+	if id != "" {
+		if _, ok := c.Environments[id]; !ok {
+			c.Environments[id] = &EnvironmentContext{}
+		}
+	} else {
+		delete(c.Environments, id)
```
sgagniere, line 226: I'm confused about this line. Do we expect/suspect that `c.Environments` has an empty key if this function is called with an empty `id`?

internal/pkg/config/v1/context.go
brianstrauch: If we're unable to set this, there's a bigger authentication problem that we should probably know about, so I'd opt to leaving this as is.

sgagniere: I agree that we want to know about it. In that case I still think we should return an error instead of (potentially) a panic.

internal/pkg/config/v1/context.go
brianstrauch: Oops, this should be `c.CurrentEnvironment`, not `id`.

brianstrauch: Ah, good point. I don't think this is right.

sgagniere: Ah, that makes sense!

internal/cmd/environment/command_update.go
```diff
@@ -36,11 +38,16 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 	updateEnvironment := orgv2.OrgV2Environment{DisplayName: orgv2.PtrString(name)}
-	_, httpResp, err := c.V2Client.UpdateOrgEnvironment(id, updateEnvironment)
+	environment, httpResp, err := c.V2Client.UpdateOrgEnvironment(id, updateEnvironment)
 	if err != nil {
 		return errors.CatchOrgV2ResourceNotFoundError(err, resource.Environment, httpResp)
 	}
 
-	output.ErrPrintf(errors.UpdateSuccessMsg, "name", "environment", id, name)
-	return nil
+	table := output.NewTable(cmd)
```
sgagniere, line 46: 👍

sgagniere (approved):

PR #1841, opened 2023-04-07
Update autocompletion for `api-key describe` and `connect cluster pause/resume`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add argument autocompletion for `confluent api-key describe`
- Add support for autocompleting multiple arguments for `confluent connect cluster pause` and `confluent connect cluster resume`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
The `api-key describe` change is self explanatory.

For `pause` and `resume`, I split the old `validArgs` function into two functions `validArgs` and `validArgsMultiple`, where only `validArgs` checks the length of args.

References
----------
n/a

Test & Review
-------------
Manual testing. For connect, I checked that I can autocomplete multiple arguments for `pause` and `resume` but not for `describe`, which only accepts 1 argument.

Open Questions / Follow-ups
---------------------------
I'm going to make these changes in the CLI-2065 PR to support multiple argument autocompletion for delete commands.

### DIFF
--- a/internal/cmd/connect/command_cluster.go
+++ b/internal/cmd/connect/command_cluster.go
@@ -48,6 +48,10 @@ func (c *clusterCommand) validArgs(cmd *cobra.Command, args []string) []string {
 		return nil
 	}
 
+	return c.validArgsMultiple(cmd, args)
+}
+
+func (c *clusterCommand) validArgsMultiple(cmd *cobra.Command, args []string) []string {
 	if err := c.PersistentPreRunE(cmd, args); err != nil {
 		return nil
 	}
--- a/internal/cmd/api-key/command_describe.go
+++ b/internal/cmd/api-key/command_describe.go
@@ -27,10 +27,11 @@ type out struct {
 
 func (c *command) newDescribeCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "describe <id>",
-		Short: "Describe an API key.",
-		Args:  cobra.ExactArgs(1),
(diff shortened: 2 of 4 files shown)


internal/cmd/connect/command_cluster.go
```diff
@@ -48,6 +48,10 @@ func (c *clusterCommand) validArgs(cmd *cobra.Command, args []string) []string {
 		return nil
 	}
 
+	return c.validArgsMultiple(cmd, args)
```
brianstrauch, line 51: Doesn't the check on line 47 guarantee that `len(args) == 0`?

brianstrauch (changes requested):

internal/cmd/connect/command_cluster.go
sgagniere: Yeah. The only difference between the autocompletion code for commands which take one or multiple arguments is the check on line 47. If it takes one arg, then it only autocompletes if there are no args. Otherwise, it autocompletes regardless.

So I'm saving some lines by having one call the other. I did realize that the naming is unfortunate, though.

What about `validArg` and `validArgs`, with `validArg` having the length check and calling `validArgs`?

brianstrauch: Oh, I see. This took me a while to understand, but I think the current naming makes sense.

brianstrauch (approved):

PR #1852, opened 2023-04-11
Consolidate command types

brianstrauch:
Checklist
---------
   * yes: ok

What
----
While investigating a separate issue, I realized that most commands are "StateFlag" commands, and the few commands that aren't don't have the same flags, i.e. `--environment` or `--cluster` (and if they did, they should also "StateFlag" commands). This PR makes all authenticated commands "StateFlag" commands by default.

Test & Review
-------------
Verified tests still pass.

### DIFF
--- a/internal/cmd/kafka/command_replica.go
+++ b/internal/cmd/kafka/command_replica.go
@@ -7,7 +7,7 @@ import (
 )
 
 type replicaCommand struct {
-	*pcmd.AuthenticatedStateFlagCommand
+	*pcmd.AuthenticatedCLICommand
 }
 
 func newReplicaCommand(prerunner pcmd.PreRunner) *cobra.Command {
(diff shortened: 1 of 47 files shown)


internal/cmd/kafka/command_replica.go
```diff
@@ -17,7 +17,7 @@ func newReplicaCommand(prerunner pcmd.PreRunner) *cobra.Command {
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
 	}
 
-	c := &replicaCommand{pcmd.NewAuthenticatedStateFlagCommand(cmd, prerunner)}
+	c := &replicaCommand{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
```
sgagniere, line 20: Here too.

sgagniere (changes requested):
Just one concern:

sgagniere (approved):
+1

Thanks!

PR #1853, opened 2023-04-11
Remove hard failure from `kafka cluster describe` if topic count can't be retrieved

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Display a warning instead of returning an error if the topic count can not be found in `confluent kafka cluster describe`

Checklist
---------
   * yes: ok

What
----
Customers using private link clusters cannot reach the endpoint used to get the topic list, and hence cannot use `kafka cluster describe`. This PR replaces the error with a warning and returns 0 for the topic count, which has been switched to an `omitempty` field.

References
----------

Test & Review
-------------
Added two new integration tests

Open Questions / Follow-ups
---------------------------


internal/cmd/kafka/command_cluster_describe.go
```diff
@@ -101,8 +100,8 @@ func (c *clusterCommand) outputKafkaClusterDescription(cmd *cobra.Command, clust
 
 	if getTopicCount {
 		topicCount, err := c.getTopicCountForKafkaCluster(cluster)
-		if err != nil {
-			return err
+		if err != nil { // topicCount is 0 when err != nil, and will be omitted by `omitempty`
+			output.ErrPrintln("WARNING: The CLI was unable to retrieve Kafka topics for this cluster. The topic count will be omitted.")
```
brianstrauch, line 104: How about we log this to DEBUG? I don't think we need to print this to stdout.

internal/cmd/kafka/command_cluster_describe.go
```diff
@@ -178,9 +177,9 @@ func (c *clusterCommand) getTopicCountForKafkaCluster(cluster *cmkv2.CmkV2Cluste
 		return 0, err
 	}
 
-	topics, httpResp, err := kafkaREST.CloudClient.ListKafkaTopics(cluster.GetId())
+	topics, _, err := kafkaREST.CloudClient.ListKafkaTopics(cluster.GetId())
 	if err != nil {
-		return 0, kafkarest.NewError(kafkaREST.CloudClient.GetUrl(), err, httpResp)
+		return 0, err
```
brianstrauch, line 182: Doesn't this change make the error less detailed?

internal/cmd/kafka/command_cluster_describe.go
sgagniere: I like it; I'll make that change.

brianstrauch: Actually, INFO would make more sense since it's underneath WARN

internal/cmd/kafka/command_cluster_describe.go
brianstrauch: Doesn't that mean you're missing the http in the URL?

sgagniere: Yeah, but it's a red herring. The url itself is apparently empty for these private link setups, so it's sending requests to an empty url with the topics endpoint tacked on at the end, which is why it reports an empty protocol scheme.

brianstrauch: huh, maybe we should add some logic where it doesn't send the request if the URL is `""`?

test/kafka_test.go
```diff
@@ -81,7 +83,7 @@ func (s *CLITestSuite) TestKafka() {
 		{args: "kafka cluster describe lkc-describe-infinite -o yaml", fixture: "kafka/43.golden"},
 
 		{args: "kafka cluster describe lkc-describe-dedicated-provisioning", fixture: "kafka/cluster-describe-dedicated-provisioning.golden"},
-		{args: "kafka cluster describe lkc-describe-topic-error", fixture: "kafka/cluster-describe-no-topic-count.golden"},
+		{args: "kafka cluster describe lkc-describe-topic-error -vv", contains: errors.OmitTopicCountMsg},
```
brianstrauch: Not a huge fan of `contains`... can we just print the complete output?

internal/pkg/errors/strings.go
```diff
@@ -45,6 +45,7 @@ const (
 	StartingConsumerMsg      = "Starting Kafka Consumer. Use Ctrl-C to exit."
 	UpdateTopicConfigMsg     = "Updated the following configuration values for topic \"%s\":\n"
 	UpdateTopicConfigRestMsg = "Updated the following configuration values for topic \"%s\" (read-only configs were not updated):\n"
+	OmitTopicCountMsg        = "The topic count will be omitted as Kafka topics for this cluster could not be retrieved: %s"
```
brianstrauch, line 48: IMO, no need to make a variable in a separate package since this string is only used once. Also, it's more idiomatic to use `%v` here (that way we can pass `err` instead of `err.Error()`)

test/kafka_test.go
sgagniere: I switched to `contains` because `-vv` introduces a timestamp into the output, so I can't easily put that in a golden file, which is annoying.

internal/pkg/errors/strings.go
sgagniere: Oh nice, yeah I'll switch to %v. As for why I put it here: it's because I'm checking for this string in the integration test, and I don't want to hardcode it twice.

test/kafka_test.go
brianstrauch: Oh, I didn't see the `-vv`. In general, testing the verbose output doesn't seem super necessary so we've never done it.

internal/cmd/kafka/command_cluster_describe.go
```diff
@@ -101,8 +102,8 @@ func (c *clusterCommand) outputKafkaClusterDescription(cmd *cobra.Command, clust
 
 	if getTopicCount {
 		topicCount, err := c.getTopicCountForKafkaCluster(cluster)
-		if err != nil {
-			return err
+		if err != nil { // topicCount is 0 when err != nil, and will be omitted by `omitempty`
```
brianstrauch, line 105: It's a better practice to put comments above the line they refer too, this prevents lines from getting too long.
```suggestion
		// topicCount is 0 when err != nil, and will be omitted by `omitempty`
		if err != nil {
```

internal/cmd/kafka/command_cluster_describe.go
```diff
@@ -36,7 +37,7 @@ type describeStruct struct {
 	ByokKeyId          string `human:"BYOK Key ID" serialized:"byok_key_id"`
 	EncryptionKeyId    string `human:"Encryption Key ID" serialized:"encryption_key_id"`
 	RestEndpoint       string `human:"REST Endpoint" serialized:"rest_endpoint"`
-	TopicCount         int    `human:"Topic Count" serialized:"topic_count"`
+	TopicCount         int    `human:"Topic Count,omitempty" serialized:"topic_count,omitempty"`
```
brianstrauch, line 40: This would technically be a breaking change, so I'd rather not apply it to the serialized output. Can you make a V4 ticket for this?
```suggestion
	TopicCount         int    `human:"Topic Count,omitempty" serialized:"topic_count"`
```

sgagniere: Will do. In this case, the serialized output should just report 0, right?

Or perhaps -1?

test/test-server/kafka_rest_router.go
```diff
@@ -151,7 +151,7 @@ func handleKafkaRPTopics(t *testing.T) http.HandlerFunc {
 				"kind": "KafkaTopicList",
 				"metadata": {"self": "http://localhost:9391/v3/clusters/cluster-1/topics","next": null},
 				"data": [
-				  {
+					{
```
sgagniere, line 154: Well this is annoying, but I'll leave it in because this is the go fmt'd output I got after I manually undid my changes here.

internal/cmd/kafka/command_cluster_describe.go
sgagniere: Ok I think we should just proceed for now to get the functionality in, and if we can get confirmation that it's just a missing hostname we optimize later.

test/test-server/kafka_rest_router.go
brianstrauch: I'll make a note to JSON-encode this in a separate PR

brianstrauch (approved):

PR #1856, opened 2023-04-13
Replace our pointer conversion functions with SDK equivalents

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
I'm splitting off some of the tangential changes from the CLI-2065 into separate PRs since that one is getting large.

The SDKs we're using already come with functions to convert strings and int32s to pointers so there's no need to write our own functions to do it. We also had two identical versions of the string to pointer function, which are both removed here.

References
----------

Test & Review
-------------
Tests still pass.

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
Thanks!

PR #1858, opened 2023-04-13
Use sorted keys for context autocompletion

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Context autocompletion extracts the key slice from a map, which is unordered. This means that the list order is not deterministic, which is causing a flaky autocompletion test.

This PR is a one line change to return a sorted key slice instead.

References
----------

Test & Review
-------------
Tests pass

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
🤌

PR #1847, opened 2023-04-08
Update autocompletion for various commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add argument autocompletion for `confluent iam`, `confluent kafka`, and `confluent pipeline` commands

Bug Fixes
- PLACEHOLDER

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok

What
----
Found some more missing autocompletion support while extending autocompletion support for CLI-2065. I also replaced some direct access with get function calls in multiple autocompletion functions.

On zsh, this is what `user` autocompletion looks like (it shows the ID and the full name):
```
confluent iam user describe u-
u-abc123  -- Steven
u-def456  -- Steven G
u-ghi789  -- aa
```

Marked as draft because I'm still going through looking for any more commands that don't have auto-completion.

References
----------
n/a

Test & Review
-------------
Manual testing of service account, user, and api-key.

Will test provider and pool before merging.

Open Questions / Follow-ups
---------------------------
Add autocompletion to `confluent schema-registry` commands. This requires slightly more work: all subcommands are defined with methods on a single `command` struct, so `validArgs` can only be defined once without adding prefixes/suffixes that break the pattern with the rest of the CLI. So the followup will slightly refactor schema-registry to have structs `clusterCommand`, `exporterCommand`, etc.

Waiting on https://github.com/confluentinc/cli/pull/1828 to avoid introducing merge conflicts into that PR.

PR #1860, opened 2023-04-14
Help output and integration test naming changes

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Followup to comments from autocompletion PR. Two things here:
- Change argument labels on the `Use` lines to match the field label (as shown in describe/list). The only exception is `ksql cluster configure-acls`, where topic arguments aren't shown in any output so I just renamed it to `topic-name` for clarity.
- Rename test suite functions to have the form `Test{Command}_`, e.g. `TestEnvironment_Describe`. I also changed command acronyms to be lowercase after the first letter.

References
----------
Followup to the comments here: https://github.com/confluentinc/cli/pull/1847

Test & Review
-------------
Ran tests

Open Questions / Follow-ups
---------------------------


internal/cmd/connect/command_plugin_describe.go
```diff
@@ -19,7 +19,7 @@ type pluginDescribeOut struct {
 
 func (c *pluginCommand) newDescribeCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "describe <plugin>",
+		Use:               "describe <plugin-name>",
```
brianstrauch, line 22: Since this command has "plugin" in its name, this seems redundant.
```suggestion
		Use:               "describe <name>",
```

internal/cmd/kafka/command_broker_describe.go
```diff
@@ -45,7 +45,7 @@ func (c *brokerCommand) newDescribeCommand() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("all", false, "Get cluster-wide broker configurations (non-default values only).")
-	cmd.Flags().String("config-name", "", "Get a specific configuration value (pair with --all to see a cluster-wide config.")
+	cmd.Flags().String("config-name", "", "Get a specific configuration value (pair with --all to see a cluster-wide config).")
```
brianstrauch, line 48: Until https://github.com/spf13/pflag/issues/373 is fixed...
```suggestion
	cmd.Flags().String("config-name", "", `Get a specific configuration value (pair with "--all" to see a cluster-wide config).`)
```

internal/cmd/kafka/command_broker_delete.go
```diff
@@ -17,7 +17,7 @@ import (
 
 func (c *brokerCommand) newDeleteCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "delete <id>",
+		Use:   "delete <broker-id>",
```
brianstrauch, line 20: The existing text is less redundant, what do you think?

test/asyncapi_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	testserver "github.com/confluentinc/cli/test/test-server"
 )
 
-func (s *CLITestSuite) TestAsyncApiExport() {
+func (s *CLITestSuite) TestAsyncApi_Export() {
```
brianstrauch, line 12: The existing way is preferred, since the command is named `confluent async-api export`. If we wanted to add more context to this test name, it might look like `TestAsyncApiExport_MoreContext()`

test/asyncapi_test.go
```diff
@@ -55,7 +55,7 @@ func (s *CLITestSuite) TestAsyncApiImport() {
 	}
 }
 
-func (s *CLITestSuite) TestAsyncApiImportWithWorkflow() {
+func (s *CLITestSuite) TestAsyncApi_ImportWithWorkflow() {
```
brianstrauch, line 58:
```suggestion
func (s *CLITestSuite) TestAsyncApiImport_WithWorkflow() {
```

brianstrauch (changes requested):

test/asyncapi_test.go
brianstrauch: The reasoning for this is you might want to run all of the `confluent asyncapi export` commands, so `make integration-test INTEGRATION_TEST_ARGS="-run TestCLI/TestAsyncapiExport" would run any of the tests with the prefix "TestAsyncapiExport".

brianstrauch:
```suggestion
func (s *CLITestSuite) TestAsyncapiExport() {
```

internal/cmd/connect/command_plugin_describe.go
sgagniere: I did it for the sake of consistency because that's how it's displayed in `pluginListOut`.

How about keep it as `plugin-name` here, and change both to `name` in v4 (because you're right that it's redundant)?

internal/cmd/kafka/command_broker_delete.go
sgagniere: Same as above: I'm matching the list output.

internal/cmd/connect/command_plugin_describe.go
brianstrauch: This isn't a breaking change, so we change both to "name" now!

test/asyncapi_test.go
sgagniere: Oh, ok. I misunderstood the discussion. I'll do another pass and tidy this up.

internal/cmd/kafka/command_broker_delete.go
brianstrauch: > I'm matching the list output.

What do you mean? The column name of `confluent kafka broker list`?

test/asyncapi_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	testserver "github.com/confluentinc/cli/test/test-server"
 )
 
-func (s *CLITestSuite) TestAsyncApiExport() {
+func (s *CLITestSuite) TestAsyncapiExport() {
```
sgagniere, line 12:
```suggestion
func (s *CLITestSuite) TestAsyncApiExport() {
```

internal/cmd/kafka/command_broker_delete.go
sgagniere: Yeah.

internal/cmd/connect/command_plugin_describe.go
sgagniere: That would make the human and seralized versions out of sync because
`Class string `human:"Plugin Name" serialized:"plugin_name"`
changing `plugin_name` to `name` would be a breaking change, right?

test/price_test.go
```diff
@@ -7,7 +7,7 @@ const (
 	exampleRegion = "us-east-1"
 )
 
-func (s *CLITestSuite) TestPriceList() {
+func (s *CLITestSuite) TestPrice() {
```
sgagniere, line 10:
```suggestion
func (s *CLITestSuite) TestPriceList() {
```

internal/cmd/kafka/command_broker_delete.go
brianstrauch: Seems a bit unrelated. The use case here is a user wants to delete a broker, so they type `confluent kafka broker delete -h`. Just a simple `<id>` seems sufficient since they've already typed "broker" and they know what resource they're about to delete.

sgagniere: Alright. I'll make the change, but I think we should change the field to ID in the get outputs in v4 for the same reason.

internal/cmd/connect/command_plugin_describe.go
brianstrauch: I thought we were just changing the usage strings in the PR, not the struct fields!

internal/cmd/kafka/command_broker_delete.go
brianstrauch: Feel free to make a ticket!

internal/cmd/connect/command_plugin_describe.go
sgagniere: Right. What I mean is that I'm matching the argument label in the usage string to the corresponding field output in the struct for consistency.

But like broker, I'll change this to `name` in the usage string and make a v4 ticket to change redundant output labels in the structs.

internal/cmd/kafka/command_broker_delete.go
sgagniere: Made this change for `partition-id` as well.

internal/cmd/kafka/command_mirror_describe.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func (c *mirrorCommand) newDescribeCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "describe <destination-topic-name>",
+		Use:               "describe <mirror-topic-name>",
```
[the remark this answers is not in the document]
sgagniere: Made the change.

PR #1861, opened 2023-04-15
Remove `api-key` unit tests

brianstrauch:
Checklist
---------
   * yes: ok

What
----
These unit tests are better suited as integration tests, since they mock the entire command and backend.

Test & Review
-------------
Verified that these tests are covered by the integration tests (and updated the integration tests accordingly)


internal/cmd/api-key/command_store.go
```diff
@@ -36,21 +36,13 @@ func (c *command) newStoreCommand() *cobra.Command {
...
 			examples.Example{
-				Text: "Pipe the API secret",
-				Code: "confluent api-key store my-key -",
-			},
-			examples.Example{
-				Text: "Provide the API secret in a file",
-				Code: "confluent api-key store my-key @my-secret.txt",
+				Text: "Get prompted for both the API key and secret",
+				Code: "confluent api-key store",
 			},
```
brianstrauch, line 46: This style of input differs from the rest of our commands, we should not recommend this. Users can use `cat my-secret.txt | confluent api-key store my-key` instead.

internal/pkg/cmd/flag_resolver.go
```diff
@@ -31,11 +31,6 @@ func (r *FlagResolverImpl) ValueFrom(source string, prompt string, secure bool)
 		if prompt == "" {
 			return "", ErrNoValueSpecified
 		}
-		if yes, err := r.Prompt.IsPipe(); err != nil {
```
brianstrauch, line 34: This seemed like an unecessary restriction. It forces users to type "-" if they want to pipe input. As far as I know, this isn't a standard in most CLIs, so we should remove this as not to confuse users.

test/api_key_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 )
 
-func (s *CLITestSuite) TestAPIKey() {
+func (s *CLITestSuite) TestApiKey() {
```
sgagniere, line 14: My other PR will do this renaming, so you can leave it off of this one. Same with the other two further down.

sgagniere (approved):

PR #1862, opened 2023-04-15
Use organization ID in `Context` instead of `AuthContext`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The organization ID is stored in three places:
* `context.last_org_id`
* `context.auth.account.organization.id`
* `context.auth.organization.id`

To move away from using Confluent Cloud v1 structs, make code more readable, and to help with the PKCE implementation, only the first is used. Since information internal to an organization is subject to change, this will also fix bugs where a user makes a change in the UI (for instance, disabling their organization's audit log) and those changes are not propagated to the CLI. Unfortunately, this requires making a few extra API calls for a few select commands, but ultimately it's the correct thing to do.

Note: the name "last_org_id" is confusing and should be renamed to "current_org_id" in v4

References
----------
Merge #1861 first

Test & Review
-------------
Shouldn't be any issues since `last_org_id` has been used for login for a while. Even so, made sure all tests pass and amended out-of-date tests.


internal/cmd/stream-share/command_provider_invite_create.go
```diff
@@ -75,11 +75,11 @@ func (c *command) createEmailInvite(cmd *cobra.Command, _ []string) error {
 	deliveryMethod := "Email"
 	resources := []string{
 		fmt.Sprintf("crn://confluent.cloud/organization=%s/environment=%s/schema-registry=%s/kafka=%s/topic=%s",
-			c.Config.GetLastUsedOrgId(), environment, srCluster.Id, cluster, topic),
+			c.Context.GetOrganization().GetResourceId(), environment, srCluster.Id, cluster, topic),
```
sgagniere, line 78: Should this be `c.Context.GetCurrentOrganization()`?

sgagniere (approved):
+1

PR #1863, opened 2023-04-15
Update test handlers

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
The goal of this PR is:
- Make sure the test backend supports more than 1 valid resource
- Return errors for get and delete requests if the resource id doesn't match the valid ones
- Make get/delete succeed or fail on the same resource ids

Which I need for multiple deletion.

Also fixed a few typos.

References
----------

Test & Review
-------------
Ran/updated tests

Open Questions / Follow-ups
---------------------------


internal/cmd/connect/command_cluster.go
```diff
@@ -65,10 +66,12 @@ func (c *clusterCommand) autocompleteConnectors() []string {
 		return nil
 	}
 
+	keys := types.GetSortedKeys(connectors)
```
sgagniere, line 69: This is to fix a flaky autocomplete test because the connectors are in an unordered map.

I couldn't use `GetSortedValues` here because the values don't satisfy `constraints.Ordered`.

test/fixtures/output/iam/identity-pool/update.golden
```diff
@@ -1,5 +1,5 @@
 +----------------+-------------------------------------------------+
-| ID             | op-55555                                        |
+| ID             | pool-12345                                      |
```
sgagniere, line 2: Typo fix: pool commands were getting provider format ids.

test/fixtures/output/kafka/quota/delete-prompt.golden
```diff
@@ -1,2 +1,2 @@
-Are you sure you want to delete client quota "cq-123"?
-To confirm, type "cq-123". To cancel, press Ctrl-C: Deleted client quota "cq-123".
+Are you sure you want to delete client quota "cq-1234"?
```
sgagniere, line 1: This was a mismatch between the handler (`cq-1234`) and the tests (`cq-123`), so I just picked one.

test/kafka_test.go
```diff
@@ -480,11 +480,11 @@ func (s *CLITestSuite) TestKafkaClientQuotas() {
 		{args: "kafka quota list --cluster lkc-1234", fixture: "kafka/quota/list.golden"},
 		{args: "kafka quota list --cluster lkc-1234 --principal sa-5678 -o json", fixture: "kafka/quota/list-json.golden"},
 		{args: "kafka quota list --cluster lkc-1234 -o yaml", fixture: "kafka/quota/list-yaml.golden"},
-		{args: "kafka quota describe cq-123 --cluster lkc-1234", fixture: "kafka/quota/describe.golden"},
-		{args: "kafka quota describe cq-123 --cluster lkc-1234 -o json", fixture: "kafka/quota/describe-json.golden"},
-		{args: "kafka quota delete cq-123 --force", fixture: "kafka/quota/delete.golden"},
-		{args: "kafka quota delete cq-123", input: "cq-123\n", fixture: "kafka/quota/delete-prompt.golden"},
-		{args: "kafka quota update cq-123 --ingress 100 --egress 100 --add-principals sa-4321 --remove-principals sa-1234 --name newName", fixture: "kafka/quota/update.golden"},
+		{args: "kafka quota describe cq-1234 --cluster lkc-1234", fixture: "kafka/quota/describe.golden"},
```
brianstrauch, line 483: Nit: They typically have len=5
```suggestion
		{args: "kafka quota describe cq-12345 --cluster lkc-12345", fixture: "kafka/quota/describe.golden"},
```

test/test-server/cloud_kafka_handlers.go
```diff
@@ -40,8 +46,7 @@ func handleKafkaClientQuota(t *testing.T) http.HandlerFunc {
 			err = json.NewEncoder(w).Encode(req)
 			require.NoError(t, err)
 		case http.MethodDelete:
-			idStr := mux.Vars(r)["id"]
-			require.Equal(t, "cq-123", idStr)
```
brianstrauch, line 44: Thanks. Using `require.Equal()` in the handlers has always seemed weird to me.

test/test-server/iam_handlers.go
```diff
@@ -211,26 +210,25 @@ func handleIamUsers(t *testing.T) http.HandlerFunc {
...
+				Description: iamv2.PtrString("at your service."),
 			}
+			err := json.NewEncoder(w).Encode(serviceAccount)
+			require.NoError(t, err)
 		case http.MethodPatch:
 			var req iamv2.IamV2ServiceAccount
 			err := json.NewDecoder(r.Body).Decode(&req)
 			require.NoError(t, err)
-			res := &iamv2.IamV2ServiceAccount{Id: iamv2.PtrString(id), Description: req.Description}
+			res := &iamv2.IamV2ServiceAccount{Id: req.Id, Description: req.Description}
```
brianstrauch, line 231:
```suggestion
			res := &iamv2.IamV2ServiceAccount{Id: req.GetId(), Description: req.Description}
```

test/test-server/byok_handlers.go
```diff
@@ -37,9 +37,14 @@ func handleByokKeyGet(t *testing.T, keyStr string, byokStoreV1 map[string]*byokv
 			require.NoError(t, err)
 			return
 		}
-		byokKey := byokStoreV1[keyStr]
-		err := json.NewEncoder(w).Encode(byokKey)
-		require.NoError(t, err)
+		if byokKey, ok := byokStoreV1[keyStr]; !ok {
+			err := writeResourceNotFoundError(w)
+			require.NoError(t, err)
+			return
```
MuweiHe, line 43: redundant `return` if there's nothing after if-else?

test/test-server/connect_handler.go
```diff
@@ -71,7 +71,26 @@ func handleConnectors(t *testing.T) http.HandlerFunc {
...
 				},
 			}
-			err := json.NewEncoder(w).Encode(map[string]connectv1.ConnectV1ConnectorExpansion{"az-connector": connectorExpansion})
+			secondConnectorExpansion := connectv1.ConnectV1ConnectorExpansion{
+				Id: &connectv1.ConnectV1ConnectorExpansionId{Id: connectv1.PtrString("lcc-111")},
+				Status: &connectv1.ConnectV1ConnectorExpansionStatus{
+					Name: "az-connector-2",
+					Connector: connectv1.ConnectV1ConnectorExpansionStatusConnector{
+						State: "RUNNING",
+						Trace: connectv1.PtrString(""),
```
MuweiHe, line 80: if it's empty can it be omitted?

test/test-server/ccloud_handlers.go
```diff
@@ -61,7 +61,6 @@ const (
 	serviceAccountID           = int32(12345)
 	serviceAccountResourceID   = "sa-12345"
 	identityProviderResourceID = "op-12345"
-	identityPoolResourceID     = "pool-12345"
```
MuweiHe, line 64: i see it's still being used in multiple places. what was the intention here to delete it？

sgagniere: I removed it because the variable itself was only used in once place after I replaced another instance of it.

But yeah, I just did a search and the string `"pool-12345"` looks like it appears in more places, so I can restore this and switch the other uses of this string to this variable.

test/test-server/connect_handler.go
sgagniere: Yup! I wenet ahead and removed it from the block above, too.

test/test-server/iam_handlers.go
sgagniere: Reverted this because it needs a pointer to a string instead of a string. I could do `iamv2.PtrString(req.GetId())`, but this is simpler.

internal/cmd/connect/command_cluster.go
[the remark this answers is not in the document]
sgagniere: Changed to `names`.

test/kafka_test.go
sgagniere: I started changing this and found that there's a whole lot of id's with length 4 after the `-`, so this can go in its own cleanup PR I think.

MuweiHe (approved):

test/test-server/iam_handlers.go
brianstrauch: `iamv2.PtrString(req.GetId())` is definitely more explicit. You can tell it's a pointer that way. `req.Id` may or may not be a pointer.

sgagniere: Made the change!

PR #1875, opened 2023-04-25
Format release notes from JSON

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Take the JSON representation of the release notes from confluentinc/cli-release, and format for:
* GitHub
* S3
* Docs
* More formats to come, such as Homebrew, APT, etc.

Test & Review
-------------
Unit tested, ran make targets with `DRY_RUN=true`


cmd/releasenotes/main.go
```diff
@@ -1,20 +1,124 @@
...
+		fmt.Print(docs(r))
+	case "github":
+		fmt.Print(github(r))
+	case "s3":
+		fmt.Print(s3(r))
+	}
+}
+
+func docs(r *releaseNotes) string {
+	title := fmt.Sprintf("[%s] Confluent CLI v%s Release Notes", time.Now().Format("1/2/2006"), "3.10.0")
```
sgagniere, line 56:
```suggestion
	title := fmt.Sprintf("[%s] Confluent CLI v%s Release Notes", time.Now().Format("1/2/2006"), r.version)
```

cmd/releasenotes/main.go
```diff
@@ -1,20 +1,124 @@
...
+			out += bullet(line) + "\n"
+		}
+		out += "\n"
+	}
+
+	return out
+}
+
+func s3(r *releaseNotes) string {
+	title := fmt.Sprintf("[%s] Confluent CLI v%s Release Notes", time.Now().Format("1/2/2006"), "3.10.0")
```
sgagniere, line 96:
```suggestion
	title := fmt.Sprintf("[%s] Confluent CLI v%s Release Notes", time.Now().Format("1/2/2006"), r.version)
```

sgagniere (changes requested):

cmd/releasenotes/main.go
```diff
@@ -1,20 +1,124 @@
...
+	file, err := os.Open(path)
+	if err != nil {
 		panic(err)
 	}
 
-	if err := r.Write(releaseNotesPath); err != nil {
+	version := strings.TrimSuffix(filepath.Base(path), filepath.Ext(path))
+
+	r := &releaseNotes{
+		date:     time.Now(),
```
sgagniere, line 36: This field doesn't seem like it's used. In the `docs` and `s3` functions you're calling `time.Now()`, so I think either this field should be removed, or those functions should use this.

sgagniere (approved):

cmd/releasenotes/main.go
```diff
@@ -1,20 +1,124 @@
...
+	"Bug Fixes",
+}
 
+// Usage: go run main.go 3.10.0.json docs
 func main() {
-	r := releasenotes.New()
+	path, format := os.Args[1], os.Args[2]
 
-	if err := r.ReadFromGithub(); err != nil {
+	file, err := os.Open(path)
```
MuweiHe, line 28: does it need a `defer close()`

brianstrauch: Doesn't hurt, thanks!

PR #1871, opened 2023-04-20
[FEDRAMP-629, IDENTITY-2574] Okta SSO login

brianstrauch:
Release Notes
-------------
New Features
- PLACEHOLDER

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * no: DO NOT MERGE until the required features are live in prod  

What
----
Only supports login for `confluent login --url https://infra.confluentgov-internal.com`

Test & Review
-------------
Mostly manual testing for now.

PR #1876, opened 2023-04-27
Stop refreshing tokens in the anonymous prerunner

brianstrauch:
Checklist
---------
   * yes: ok

What
----
It's a bit disorienting to run an unauthenticated command such as `confluent version` and see a message such as "token expired"... ideally these functions should not perform any authentication-related logic.

Refactored the `ValidateToken()` and `updateToken()` so they do what their names suggest, and prevent tokens from being updated/deleted by the Anonymous prerunner.

While testing #1871, I would occasionally see two calls to `/api/sessions` if the token was expired and the user ran `confluent login`, which inspired this PR.

Test & Review
-------------
Manual testing, patched unit/integration tests

Open Questions / Follow-ups
---------------------------
There is still an issue with logging in to SSO accounts in different environments (i.e. prod and then fedramp) which will also need to be fixed.

### DIFF
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -172,18 +172,6 @@ func (r *PreRun) Anonymous(command *CLICommand, willAuthenticate bool) func(*cob
 		r.notifyIfUpdateAvailable(cmd, command.Version.Version)
 		r.warnIfConfluentLocal(cmd)
 
-		if r.Config != nil {
-			ctx := command.Config.Context()
-			err := r.ValidateToken(command.Config, unsafeTrace)
-			switch err.(type) {
-			case *ccloudv1.ExpiredTokenError:
-				if err := ctx.DeleteUserAuth(); err != nil {
-					return err
-				}
-				output.ErrPrintln(errors.TokenExpiredMsg)
-			}
-		}
-
 		LabelRequiredFlags(cmd)
 
 		return nil
@@ -281,8 +269,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 			return err
 		}
 
-		if err := r.ValidateToken(command.Config, unsafeTrace); err != nil {
-			return err
+		if err := r.ValidateToken(command.Config); err != nil {
+			if err := r.updateToken(err, command.Config.Context(), unsafeTrace); err != nil {
+				return err
+			}
 		}
 
 		if err := r.setV2Clients(command); err != nil {
@@ -559,7 +549,13 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 			return err
 		}
 
-		return r.ValidateToken(command.Config, unsafeTrace)
+		if err := r.ValidateToken(command.Config); err != nil {
+			if err := r.updateToken(err, command.Config.Context(), unsafeTrace); err != nil {
+				return err
+			}
+		}
+
+		return nil
 	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/pkg/cmd/prerunner_test.go
```diff
@@ -184,15 +184,13 @@ func TestPreRun_TokenExpires(t *testing.T) {
...
-	}
-	rootCmd := pcmd.NewAnonymousCLICommand(root, r)
+	root := &cobra.Command{Run: func(cmd *cobra.Command, args []string) {}}
+	rootCmd := pcmd.NewAuthenticatedCLICommand(root, r)
 	root.Flags().CountP("verbose", "v", "Increase verbosity")
 	root.Flags().Bool("unsafe-trace", false, "")
 
 	_, err := pcmd.ExecuteCommand(rootCmd.Command)
-	require.NoError(t, err)
+	require.Error(t, err)
```
MuweiHe, line 193: 😨

internal/pkg/cmd/prerunner.go
```diff
@@ -863,47 +861,33 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
...
 func (r *PreRun) updateToken(tokenError error, ctx *dynamicconfig.DynamicContext, unsafeTrace bool) error {
-	if ctx == nil {
-		log.CliLogger.Debug("Dynamic context is nil. Cannot attempt to update auth token.")
-		return tokenError
-	}
 	log.CliLogger.Debug("Updating auth tokens")
 	token, refreshToken, err := r.getUpdatedAuthToken(ctx, unsafeTrace)
 	if err != nil || token == "" {
 		log.CliLogger.Debug("Failed to update auth tokens")
+		_ = ctx.DeleteUserAuth()
```
MuweiHe, line 880: 👍

internal/pkg/cmd/prerunner.go
```diff
@@ -863,47 +861,33 @@ func (r *PreRun) HasAPIKey(command *HasAPIKeyCLICommand) func(*cobra.Command, []
...
-		return r.updateToken(new(ccloudv1.InvalidTokenError), ctx, unsafeTrace)
-	case *ccloudv1.ExpiredTokenError:
-		return r.updateToken(new(ccloudv1.ExpiredTokenError), ctx, unsafeTrace)
-	}
-	if err.Error() == errors.MalformedJWTNoExprErrorMsg {
-		return r.updateToken(errors.New(errors.MalformedJWTNoExprErrorMsg), ctx, unsafeTrace)
-	} else {
-		return r.updateToken(err, ctx, unsafeTrace)
-	}
+	return r.JWTValidator.Validate(ctx.Context)
```
MuweiHe, line 872: how does this handle invalid token / expired token?

internal/pkg/cmd/prerunner_test.go
brianstrauch: This is intended! We have to force an error to happen (corresponds to the backend returning `""` for the auth token) because if the auth token exists, it'll try to send it to the backend which we don't have mocks for in our unit tests.

internal/pkg/cmd/prerunner.go
brianstrauch: You can take a look at the implementation! It checks for the existence of the `exp` field and if it's expired.

internal/pkg/cmd/prerunner.go
```diff
@@ -281,8 +269,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 			return err
 		}
 
-		if err := r.ValidateToken(command.Config, unsafeTrace); err != nil {
-			return err
+		if err := r.ValidateToken(command.Config); err != nil {
+			if err := r.updateToken(err, command.Config.Context(), unsafeTrace); err != nil {
```
sgagniere, line 273: It's slightly awkward to read this line w/ two different `err` variables on the same line; how about using `err2` for the inner one?

MuweiHe (approved):

PR #1877, opened 2023-04-27
Resolve bug preventing permanent deletion of schemas

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add a new `--deleted` flag to `confluent schema-registry schema list` to include soft-deleted schemas in the output

Bug Fixes
- Schemas can now be permanently deleted with `confluent schema-registry schema delete` when using `--version all`

Checklist
---------
   * yes: ok

What
----
`GetSchemaByVersion` from `schema-registry-sdk-go` doesn't currently work with the `Deleted` optional argument when the version is `latest` or `all`, so until that issue is resolved I'm disabling the existence check.

I also added a flag to `sr schema list` to show soft deleted schemas in addition to everything else.

References
----------
n/a

Test & Review
-------------
Ran all tests
Made sure that all manual test cases from https://github.com/confluentinc/cli/pull/1796 still work as expected

Additionally tested the following case (payment is a soft-deleted subject):
```
confluent sr schema delete --subject payment --version all --permanent
Are you sure you want to permanently delete schema "payment (version all)"?
To confirm, type "payment". To cancel, press Ctrl-C: payment
Successfully hard deleted all versions for subject "payment"
  Version  
-----------
        1  
```

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/schema-registry/command_schema_delete.go
+++ b/internal/cmd/schema-registry/command_schema_delete.go
@@ -78,11 +78,13 @@ func deleteSchema(cmd *cobra.Command, srClient *srsdk.APIClient, ctx context.Con
 		checkVersion = "latest"
 	}
 	if permanent {
-		opts := &srsdk.GetSchemaByVersionOpts{Deleted: optional.NewBool(true)}
-		if _, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, opts); err != nil {
-			return errors.CatchSchemaNotFoundError(err, httpResp)
-		} else if _, _, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil); err == nil {
-			return errors.New("you must first soft delete a schema version before you can permanently delete it")
+		if checkVersion != "latest" {
+			opts := &srsdk.GetSchemaByVersionOpts{Deleted: optional.NewBool(true)}
+			if _, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, opts); err != nil {
+				return errors.CatchSchemaNotFoundError(err, httpResp)
+			} else if _, _, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil); err == nil {
+				return errors.New("you must first soft delete a schema version before you can permanently delete it")
+			}
 		}
 	} else if _, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil); err != nil {
 		return errors.CatchSchemaNotFoundError(err, httpResp)
--- a/internal/cmd/schema-registry/command_schema_list.go
+++ b/internal/cmd/schema-registry/command_schema_list.go
@@ -43,6 +43,7 @@ func (c *command) newSchemaListCommand() *cobra.Command {
 	}
 
 	cmd.Flags().String("subject-prefix", "", "List schemas for subjects with a given prefix.")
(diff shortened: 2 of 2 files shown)


internal/cmd/schema-registry/command_schema_list.go
```diff
@@ -43,6 +43,7 @@ func (c *command) newSchemaListCommand() *cobra.Command {
 	}
 
 	cmd.Flags().String("subject-prefix", "", "List schemas for subjects with a given prefix.")
+	cmd.Flags().Bool("deleted", false, "Include soft-deleted schemas.")
```
sgagniere, line 46: This is kind of misleading since this shows both non-deleted and soft-deleted schemas.

I chose this to match the flag from `confluent sr subject list --deleted`, which also shows non-deleted and soft-deleted subjects

brianstrauch: Several other commands have an `--all` flag. Should we do the same here or would that be too vague?

internal/cmd/schema-registry/command_schema_delete.go
```diff
@@ -78,11 +78,13 @@ func deleteSchema(cmd *cobra.Command, srClient *srsdk.APIClient, ctx context.Con
 		checkVersion = "latest"
 	}
 	if permanent {
-		opts := &srsdk.GetSchemaByVersionOpts{Deleted: optional.NewBool(true)}
-		if _, httpResp, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, opts); err != nil {
-			return errors.CatchSchemaNotFoundError(err, httpResp)
-		} else if _, _, err := srClient.DefaultApi.GetSchemaByVersion(ctx, subject, checkVersion, nil); err == nil {
-			return errors.New("you must first soft delete a schema version before you can permanently delete it")
+		if checkVersion != "latest" {
```
brianstrauch, line 81: What happens when both `--permanent` and `--version latest` are passed?

MuweiHe: `permanent` works on already soft-deleted schemas, and because of the bug, the call will return error for `latest` that's why we're bypassing it

sgagniere: Actually, it looks like `--permanent` and `--version latest` is an invalid combination for the delete endpoint itself.

I'll check w/ the backend team to see if this is another bug like the one @MuweiHe mentioned for the get endpoint, or if this is expected behavior.

MuweiHe: oh i was trying to say already soft-deleted so `opts: deleted=true`. we on the same page?

sgagniere: Yup. I was just noticing that this applies to the delete endpoint as well.

internal/cmd/schema-registry/command_schema_list.go
brianstrauch: Perfect!

internal/cmd/schema-registry/command_schema_list.go
```diff
@@ -68,7 +68,7 @@ func (c *command) listSchemas(cmd *cobra.Command, srClient *srsdk.APIClient, ctx
 		return err
 	}
 
-	showDeleted, err := cmd.Flags().GetBool("deleted")
+	showDeleted, err := cmd.Flags().GetBool("all")
```
brianstrauch, line 71:
```suggestion
	all, err := cmd.Flags().GetBool("all")
```

PR #1828, opened 2023-03-30
[WIP] DGS-5661: moving schema-registry cluster commands to srcm/v2

Anshul-Goyal:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok
   * no: DO NOT MERGE until the required features are live in prod  

What
----

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

PR #1878, opened 2023-04-27
Migrate to SRCM v2 for schema-registry cluster delete, upgrade, describe commands

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Delete schema registry URL of an environment when the schema registry cluster is deleted

Checklist
---------
   * yes: ok

What
----

Migrate to srcm v2. 
`enable` command is still wip and will be in a follow-up pr.

References
----------

https://github.com/confluentinc/cli/pull/1828/files

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


test/test-server/srcm_handlers.go
```diff
@@ -18,7 +18,7 @@ const (
 	idRegion1          = "sgreg-1"
 	packageTypeRegion1 = "advanced"
 	cloudRegion1       = "aws"
-	nameRegion1        = "us-east-2"
+	nameRegion1        = "us-west-2"
```
sgagniere, line 21: I think we should either keep this as `east` to match `displayNameRegion1 = "Ohio (us-east-2)"`, or change the display name.

internal/cmd/schema-registry/command_cluster_delete.go
```diff
@@ -40,24 +39,22 @@ func (c *command) newClusterDeleteCommand() *cobra.Command {
...
 func (c *command) clusterDelete(cmd *cobra.Command, _ []string) error {
-	ctx := context.Background()
-
 	environmentId, err := c.Context.EnvironmentId()
 	if err != nil {
 		return err
 	}
 
-	cluster, err := c.Context.FetchSchemaRegistryByEnvironmentId(ctx, environmentId)
+	clusters, err := c.V2Client.GetSchemaRegistryClustersByEnvironment(environmentId)
```
brianstrauch, line 47: Can an environment have more than one SR cluster now?

internal/cmd/schema-registry/command_cluster_describe.go
```diff
@@ -68,10 +70,20 @@ func (c *command) clusterDescribe(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	cluster, err := c.Context.FetchSchemaRegistryByEnvironmentId(ctx, environmentId)
+	clusters, err := c.V2Client.GetSchemaRegistryClustersByEnvironment(environmentId)
 	if err != nil {
 		return err
 	}
+
+	cluster := clusters[0]
```
brianstrauch, line 78: Should we check the length to prevent a panic?

internal/cmd/schema-registry/command_cluster_describe.go
```diff
@@ -104,22 +116,23 @@ func (c *command) clusterDescribe(cmd *cobra.Command, _ []string) error {
...
+	freeSchemasLimit := defaultSchemaLimitEssentials
+	if strings.ToLower(clusterSpec.GetPackage()) == essentialsPackage {
 		user, err := c.Client.Auth.User(context.Background())
 		if err != nil {
 			return err
 		}
 		prices, err := c.Client.Billing.GetPriceTable(context.Background(), user.GetOrganization(), streamGovernancePriceTableProductName)
 		if err == nil {
-			priceKey := getMaxSchemaLimitPriceKey(cluster.Package, cluster.ServiceProvider, cluster.ServiceProviderRegion)
+			internalPackageName, _ := getPackageInternalName(clusterSpec.GetPackage())
```
brianstrauch, line 134: Why are we ignoring the error?

internal/cmd/schema-registry/command_cluster_test.go
```diff
@@ -93,14 +98,34 @@ func (suite *ClusterTestSuite) SetupTest() {
 			return resp, nil, nil
 		},
 	}
+	suite.srcmClusterMock = &srcmv2mock.ClustersSrcmV2Api{
```
brianstrauch, line 101: It would be amazing if these unit tests could be converted to integration tests in this PR!

internal/pkg/ccloudv2/schema_registry.go
```diff
@@ -51,3 +64,51 @@ func (c *Client) ListSchemaRegistryRegions(cloud, packageType string) ([]srcm.Sr
...
+
+		pageToken, done, err = extractNextPageToken(page.GetMetadata().Next)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if len(list) == 0 {
+		return nil, errors.NewSRNotEnabledError()
+	}
```
brianstrauch, line 103: This is a bit different than our normal style. Can we move this outside of this function?

internal/pkg/dynamic-config/client.go
```diff
@@ -53,31 +53,13 @@ func (d *DynamicContext) FetchAPIKeyError(apiKey string, clusterID string) error
 	return &errors.UnconfiguredAPISecretError{APIKey: apiKey, ClusterID: clusterID}
 }
 
-func (d *DynamicContext) FetchSchemaRegistryByEnvironmentId(context context.Context, accountId string) (*ccloudv1.SchemaRegistryCluster, error) {
-	existingClusters, err := d.Client.SchemaRegistry.GetSchemaRegistryClusters(context, &ccloudv1.SchemaRegistryCluster{
-		AccountId: accountId,
-		Name:      "account schema-registry",
-	})
+func (d *DynamicContext) FetchSchemaRegistryByEnvironmentId(context context.Context, accountId string) (*srcmv2.SrcmV2Cluster, error) {
```
brianstrauch, line 56: No need for this to be a pointer!
```suggestion
func (d *DynamicContext) FetchSchemaRegistryByEnvironmentId(context context.Context, accountId string) (srcmv2.SrcmV2Cluster, error) {
```

internal/cmd/schema-registry/command_cluster_describe.go
MuweiHe: length is being checked within the package before returning the list

brianstrauch: What do you think about moving that logic out here? It would be more explicit

internal/cmd/schema-registry/command_cluster_describe.go
MuweiHe: cc @Anshul-Goyal

internal/cmd/schema-registry/command_cluster_delete.go
MuweiHe: cc @Anshul-Goyal

Anshul-Goyal: No, we can only have one SR per environment.

brianstrauch: Why does this API return a list of SR clusters, then? Is that something we plan to support in the future?

brianstrauch: It also looks like this API is paginated too? (A bit weird, considering it will only ever return 0 or 1 results)

internal/cmd/schema-registry/command_cluster_describe.go
MuweiHe: ok!

internal/cmd/schema-registry/command_cluster_test.go
MuweiHe: they're already covered in inte-tests! Do you wish to remove them unit tests?

brianstrauch: As long as you're certain we're not losing any test coverage, yes! This would make maintenance easier in the long-term.

internal/cmd/schema-registry/command_cluster_delete.go
```diff
@@ -63,6 +63,12 @@ func (c *command) clusterDelete(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	ctx := c.Config.Context()
+	ctx.SchemaRegistryClusters[environmentId] = nil
+	if err := ctx.Save(); err != nil {
```
brianstrauch, line 68: If the cluster is deleted but the configuration file is not saved, we should still print the message below. Can we ignore this error?

brianstrauch (changes requested):

internal/cmd/schema-registry/command_cluster_delete.go
MuweiHe: good point we should still print the `deleted` msg. But idk about ignoring the `save config` error. Is there any other places that we're ignoring it?

brianstrauch: Should be fine! Searching through the repo I see 4-5 other places where we ignore the error!

internal/cmd/schema-registry/command_cluster_upgrade.go
```diff
@@ -35,36 +39,57 @@ func (c *command) newClusterUpgradeCommand() *cobra.Command {
...
 	}
 
-	cluster, err := c.Context.FetchSchemaRegistryByEnvironmentId(environmentId)
+	_, err = getPackageInternalName(packageDisplayName)
 	if err != nil {
 		return err
 	}
 
-	if packageInternalName == cluster.Package {
+	if strings.EqualFold(clusterSpec.GetPackage(), packageDisplayName) {
```
brianstrauch, line 69: Should we upper case these strings, and then do a regular `==` comparison? I don't think many people know what `EqualFold()` does (I had to look it up)

internal/cmd/schema-registry/command_cluster_upgrade.go
```diff
@@ -35,36 +39,57 @@ func (c *command) newClusterUpgradeCommand() *cobra.Command {
...
+	_, err = c.V2Client.UpgradeSchemaRegistryCluster(*clusterUpdateRequest, cluster.GetId())
+	if err != nil {
 		return err
 	}
 
 	output.Printf(errors.SchemaRegistryClusterUpgradedMsg, environmentId, packageDisplayName)
 	return nil
 }
+
+func createClusterUpdateRequest(packageType, environmentId string) *srcmv2.SrcmV2ClusterUpdate {
```
brianstrauch, line 84: Can we create this struct manually instead of calling these functions? It's not clear what the "defaults" are, and we don't use this approach anywhere else.

internal/pkg/ccloudv2/schema_registry.go
```diff
@@ -4,27 +4,39 @@ import (
...
-	var list []srcm.SrcmV2Region
+func (c *Client) GetStreamGovernanceRegionById(regionId string) (srcmv2.SrcmV2Region, error) {
+	req := c.SchemaRegistryClient.RegionsSrcmV2Api.GetSrcmV2Region(c.SchemaRegistryApiContext(), regionId)
+	region, httpResp, err := c.SchemaRegistryClient.RegionsSrcmV2Api.GetSrcmV2RegionExecute(req)
+	return region, errors.CatchCCloudV2Error(err, httpResp)
+}
+
+func (c *Client) CreateSchemaRegistryCluster(srCluster srcmv2.SrcmV2Cluster) (srcmv2.SrcmV2Cluster, error) {
+	req := c.SchemaRegistryClient.ClustersSrcmV2Api.CreateSrcmV2Cluster(c.SchemaRegistryApiContext()).SrcmV2Cluster(srCluster)
+	createdCluster, httpResp, err := c.SchemaRegistryClient.ClustersSrcmV2Api.CreateSrcmV2ClusterExecute(req)
```
brianstrauch, line 34: Call `.Execute()` instead?

internal/cmd/schema-registry/command_cluster_delete.go
```diff
@@ -44,21 +44,29 @@ func (c *command) clusterDelete(cmd *cobra.Command, _ []string) error {
...
 	}
 
-	err = c.Client.SchemaRegistry.DeleteSchemaRegistryCluster(cluster)
+	err = c.V2Client.DeleteSchemaRegistryCluster(clusters[0].GetId(), environmentId)
 	if err != nil {
 		return err
 	}
 
+	ctx := c.Config.Context()
+	ctx.SchemaRegistryClusters[environmentId] = nil
```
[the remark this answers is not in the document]
Anshul-Goyal: @brianstrauch - AFAIK, neither the HTTP endpoint URL nor the cluster ID ever changes during the lifetime of a SR cluster.

Just for context, are we talking about deleting and recreating the clusters in the same environment here?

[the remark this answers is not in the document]
brianstrauch: > Just for context, are we talking about deleting and recreating the clusters in the same environment here?

Yes, we don't want to cache the Schema Registry cluster URL in the CLI if it's possible for a customer to delete and recreate it from the UI (which would change the URL)

internal/cmd/schema-registry/command_cluster_upgrade.go
```diff
@@ -66,12 +66,20 @@ func (c *command) clusterUpgrade(cmd *cobra.Command, _ []string) error {
...
 		return nil
 	}
 
-	clusterUpdateRequest := createClusterUpdateRequest(packageDisplayName, environmentId)
+	clusterUpdateRequest := &srcmv2.SrcmV2ClusterUpdate{
+		Spec: &srcmv2.SrcmV2ClusterSpecUpdate{
+			Package: srcmv2.PtrString(packageDisplayName),
+			Environment: &srcmv2.GlobalObjectReference{
+				Id: environmentId,
+			},
```
brianstrauch, line 79:
```suggestion
			Environment: &srcmv2.GlobalObjectReference{Id: environmentId},
```

internal/cmd/schema-registry/command_cluster_delete.go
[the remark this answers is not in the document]
Anshul-Goyal: Makes sense, because that is possible. A customer could delete their cluster via the UI and recreate a cluster in the same environment in another region or a cluster with different physical cluster in the same region, changing the SR cluster url for that env.

PR #1887, opened 2023-05-01
Set up GoReleaser Pro

brianstrauch:
What
----
Add prebuilt Linux binaries to the GoReleaser pipeline. This is a first step before we can easily release to various package managers.

References
----------
https://goreleaser.com/customization/builds/#import-pre-built-binaries

Test & Review
-------------
Tested with `DRY_RUN=true`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,8 +1,5 @@
 project_name: confluent
 
-partial:
-  by: target
-
 builds:
   - binary: confluent
     id: confluent-alpine-amd64
@@ -115,6 +112,18 @@ builds:
       post:
         - cmd: scripts/gon_filepath_editor.sh {{ .Path }} arm64
         - cmd: gon dist/gon_confluent_arm64.hcl
+  - builder: prebuilt
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    goamd64:
+      - v1
+    
+    prebuilt:
+      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_v1{{ end }}/confluent
+
 
 release:
   disable: "{{.Env.DRY_RUN}}"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,8 @@
 project_name: confluent
 
+partial:
+  by: target
+
 builds:
   - binary: confluent
     id: confluent-alpine-amd64
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -120,10 +120,8 @@ builds:
       - arm64
     goamd64:
       - v1
-    
     prebuilt:
       path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_v1{{ end }}/confluent
-
 
 release:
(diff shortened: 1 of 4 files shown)


.goreleaser.yml
```diff
@@ -112,6 +112,16 @@ builds:
...
+  - builder: prebuilt
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    goamd64:
+      - v1
+    prebuilt:
+      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_v1{{ end }}/confluent
```
sgagniere, line 124: Something to keep in mind if we upgrade the Goreleaser version in the future: directory names inside `dist/` aren't guaranteed to be the same between goreleaser versions according to the docs.

It might be safer to parse the path from the `dist/artifacts.json` files of the linux builds; the docs do say that we should always be able to get the correct path from there.

mk-files/release.mk
```diff
@@ -75,25 +75,11 @@ endif
...
-	$(call dry-run,aws s3 cp dist/confluent_linux_amd64_v1/confluent $(S3_STAG_PATH)/confluent-cli/binaries/$(VERSION_NO_V)/confluent_$(VERSION_NO_V)_linux_amd64) && \
-	$(call dry-run,aws s3 cp dist/confluent_linux_arm64/confluent $(S3_STAG_PATH)/confluent-cli/binaries/$(VERSION_NO_V)/confluent_$(VERSION_NO_V)_linux_arm64) && \
-	cat dist/confluent_$(VERSION_NO_V)_checksums_linux.txt >> dist/confluent_$(VERSION_NO_V)_checksums.txt && \
-	cat dist/confluent_$(VERSION_NO_V)_checksums_linux_arm64.txt >> dist/confluent_$(VERSION_NO_V)_checksums.txt && \
-	$(call dry-run,aws s3 cp dist/confluent_$(VERSION_NO_V)_checksums.txt $(S3_STAG_PATH)/confluent-cli/archives/$(VERSION_NO_V)/confluent_$(VERSION_NO_V)_checksums.txt) && \
-	$(call dry-run,aws s3 cp dist/confluent_$(VERSION_NO_V)_checksums.txt $(S3_STAG_PATH)/confluent-cli/binaries/$(VERSION_NO_V)/confluent_$(VERSION_NO_V)_checksums.txt) && \
-	echo "UPLOADING LINUX BUILDS TO GITHUB" && \
-	make upload-linux-build-to-github
-	
+	GORELEASER_KEY=$(GORELEASER_KEY) GITHUB_TOKEN=$(token) S3FOLDER=$(S3_STAG_FOLDER_NAME)/confluent-cli DRY_RUN=$(DRY_RUN) goreleaser continue
```
sgagniere, line 82: Much cleaner!

sgagniere (approved):

.goreleaser.yml
brianstrauch: > It might be safer to parse the path from the dist/artifacts.json files of the linux builds

Since the path in the dist/artifacts.json file is based on the path in the Docker image, I'll make a note about this and come back to this in the future (since we may or may not still be using Docker)!

PR #1888, opened 2023-05-03
[CLI-916] Add `confluent connect plugin install` command

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `confluent connect plugin install` command to install Connect plugins for Confluent Platform

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Add `confluent-hub` functionality into the CLI as the On Prem command `confluent connect plugin install`. It takes one argument, which can be either a path to a local zip file downloaded from Confluent Hub, or a plugin id in the form `<owner/name:version>`, e.g. `confluentinc/kafka-connect-datagen:latest`.

There are 4 supported flags. Flags `plugin-dir` and `force` correspond to `component-dir` and `no-prompt` from the original tool.

This version has a slightly modified flow compared to the original tool. The new flow is as follows:
1. Validate the plugin argument by extracting the plugin manifest from the local archive or Confluent Hub.
2. Detect Confluent Platform installations and prompt the user to select one if `--plugin-dir` is not specified; then prompt the user if they would like to use the default plugin directory for that Confluent Platform installation. If they refuse, ask them to provide a directory path.
3. Check if any version of this plugin is already installed at the selected location, and if so, prompt the user to uninstall it to continue.
4. Ask the user to accept all licenses listed in the manifest.
5. Install the plugin into the selected directory.
6. Detect worker config files if the user did not specify `--worker-configs`. The detection is based on the chosen installation from step 2; if the user provided `--plugin-dir` then they will be prompted to select a Confluent Platform installation here instead. Next, ask the user if they want to update all detected files. If not, ask them if they want to update each one individually.

For `dry-run` mode, steps 3, 5, and 6 skip filesystem modifications.
For `force` mode, step 2 will use the default plugin directory of the first Confluent Platform installation found (usually CONFLUENT_PLATFORM) if `--plugin-dir` is not specified, step 3 will auto-delete the existing plugin installation, step 4 will inform the user that they have agreed to all licenses, and step 6 will update all detected worker config files if `--worker-configs` is not specified.

References
----------
Closes #1199

Test & Review
-------------
Manual testing in progress

Open Questions / Follow-ups
---------------------------


brianstrauch:
Release notes should mention connect. We might also want to consider deprecating the "Confluent Hub" terminology, but you should check with the connect team. How about:

> Add `confluent connect plugin install` to install connect plugins for Confluent Platform

internal/cmd/connect/command_plugin_install.go
```diff
@@ -68,10 +63,10 @@ func (c *pluginCommand) newInstallCommand() *cobra.Command {
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
 	}
 
-	cmd.Flags().String("component-dir", "", "The plugin installation directory. Defaults to $share/confluent-hub-components for archive deployment, and to /usr/share/confluent-hub-components for deb/rpm deployment.")
-	cmd.Flags().StringSlice("worker-configs", []string{}, "Comma-delineated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the component directory in addition to any preexisting directories.")
+	cmd.Flags().String("plugin-dir", "", "The plugin installation directory. Defaults to $share/confluent-hub-components for archive deployment, and to /usr/share/confluent-hub-components for deb/rpm deployment.")
```
brianstrauch:
```suggestion
	cmd.Flags().String("plugin-dir", "", "The plugin installation directory. Defaults to $SHARE/confluent-hub-components for archive deployment, and to /usr/share/confluent-hub-components for DEB/RPM deployment.")
```

brianstrauch: Is there any way we can infer what type of deployment this is before the command is run, and set the default value depending on the deployment type?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -68,10 +63,10 @@ func (c *pluginCommand) newInstallCommand() *cobra.Command {
...
+	cmd.Flags().StringSlice("worker-configs", []string{}, "Comma-delineated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any preexisting directories.")
```
brianstrauch:
```suggestion
	cmd.Flags().StringSlice("worker-configs", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any pre-existing directories.")
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -100,43 +95,43 @@ func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
 		pluginManifest = remoteManifest
 	}
 
-	// Select component-dir
+	// Select plugin-dir
 	var ins *installation
```
brianstrauch, line 99: Let's define this variable right before we need to use it.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -146,14 +141,14 @@ func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
 	}
 
 	if dryRun {
-		output.Println("Skipping installation of connector onto worker as part of dry run mode.")
+		output.Println("Dry run: skipping installation of plugin.")
```
brianstrauch, line 144:
```suggestion
		output.Println("Dry run: skipping plugin installation.")
```

brianstrauch: I think our other dry run text is formatted as "[DRY RUN] Message", let's double-check and stay consistent with that!

internal/cmd/connect/command_plugin_install.go
```diff
@@ -205,20 +200,20 @@ func parsePluginId(plugin string) (string, string, string, error) {
...
 		if err != nil {
-			return nil, errors.Errorf("failed to examine file %s inside local archive file %s: %v", zipFile.Name, archivePath, err)
+			return nil, errors.Wrapf(err, "failed to examine file %s inside local archive file %s", zipFile.Name, archivePath)
 		}
 
 		if isManifest {
 			manifestFile, err := zipFile.Open()
 			if err != nil {
-				return nil, errors.Errorf("failed to open manifest file %s inside local archive file %s: %v", zipFile.Name, archivePath, err)
+				return nil, errors.Wrapf(err, "failed to open manifest file %s inside local archive file %s", zipFile.Name, archivePath)
```
brianstrauch, line 218: We may want to do this across the whole CLI?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -235,12 +230,12 @@ func getLocalManifest(archivePath string) (*manifest, error) {
...
 	}
-	
-	return nil, errors.Errorf("failed to find manifest file inside local archive file %s", archivePath)
+
+	return nil, errors.Errorf(`failed to find manifest file inside local archive file "%s"`, archivePath)
 }
 
 func getRemoteManifest(owner, name, version string) (*manifest, error) {
-	manifestUrl := fmt.Sprintf("%s/%s/%s", hubPluginsUrl, owner, name)
+	manifestUrl := fmt.Sprintf("%s/%s/%s", "https://api.hub.confluent.io/api/plugins", owner, name)
```
brianstrauch, line 240:
```suggestion
	manifestUrl := fmt.Sprintf("https://api.hub.confluent.io/api/plugins/%s/%s", owner, name)
```

internal/cmd/connect/install_utils.go
```diff
@@ -36,15 +36,16 @@ type listOut struct {
...
+func getInstallation(cmd *cobra.Command, force bool) (*installation, error) {
 	installations, err := findInstallationDirectories()
 	if err != nil {
 		return nil, err
 	}
 
 	if len(installations) == 0 {
-		return nil, errors.NewErrorWithSuggestions("unable to detect Confluent Platform installation", "Pass the component directory and worker configuration files to the `--component-dir` and `--worker-configs` flags.")
-	} else if noPrompt {
+		return nil, errors.NewErrorWithSuggestions("unable to detect Confluent Platform installation", "Pass the plugin directory and worker configuration files to the `--plugin-dir` and `--worker-configs` flags.")
```
brianstrauch:
```suggestion
		return nil, errors.NewErrorWithSuggestions("unable to detect a Confluent Platform installation", "Pass the plugin directory and worker configuration files with `--plugin-dir` and `--worker-configs`.")
```

internal/cmd/connect/install_utils.go
```diff
@@ -227,7 +229,7 @@ func standardWorkerConfigLocations(ins *installation) ([]WorkerConfig, error) {
 		if utils.DoesPathExist(confluentCurrentFile) {
 			confluentCurrentContent, err := os.ReadFile(confluentCurrentFile)
 			if err != nil {
-				return nil, errors.Errorf("failed to read possible $CONFLUENT_CURRENT file %s: %v", confluentCurrentFile, err)
+				return nil, errors.Wrapf(err, `failed to read possible $CONFLUENT_CURRENT file "%s"`, confluentCurrentFile)
```
brianstrauch, line 232: This seems like an artifact of the old `confluent local` commands. Not sure we need it anymore, but I'd double check.

internal/cmd/connect/install_utils.go
```diff
@@ -309,25 +311,30 @@ func chooseWorkerConfigs(cmd *cobra.Command, ins *installation, noPrompt bool) (
 	}
 
 	var filteredWorkerConfigs []WorkerConfig
-	if noPrompt || len(workerConfigs) == 0 {
+	if len(workerConfigs) == 0 {
+		output.Println("No worker config files found.")
```
brianstrauch, line 315:
```suggestion
		output.Println("No worker configuration files found.")
```

internal/cmd/connect/install_utils.go
```diff
@@ -309,25 +311,30 @@ func chooseWorkerConfigs(cmd *cobra.Command, ins *installation, noPrompt bool) (
...
+		return filteredWorkerConfigs, nil
+	}
+
+	output.Println("Detected the following worker config files:")
```
brianstrauch, line 319:
```suggestion
	output.Println("Detected the following worker configuration files:")
```

internal/cmd/connect/command_plugin_install.go
brianstrauch: Can you explain what this is, or give an example of where you download one, or how you write one?

internal/cmd/connect/command_plugin_install.go
sgagniere: I just removed this part of the string altogether; I may add back a reworded version later if we want it.

brianstrauch: The comment still applies though, it might be nice to decide the default beforehand (this way it shows up for the user if they run `--help`)

brianstrauch: Although on second thought, I'm not sure how we'd choose which default to show in our autogenerated docs... (maybe there's a way we can hide it?)

sgagniere: I'm inclined to just write a general message about how the command will select a default; I'll think about the wording.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -119,15 +126,10 @@ func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
...
-	dryRun, err := cmd.Flags().GetBool("dry-run")
-	if err != nil {
-		return err
-	}
-
 	if previousInstallations, err := existingPluginInstallation(pluginDir, pluginManifest); err != nil {
 		return err
 	} else if len(previousInstallations) > 0 {
-		output.Println("A version of this plugin is already installed.")
+		output.Printf("\nA version of this plugin is already installed.\n")
```
brianstrauch, line 132:
```suggestion
		output.Print("\nA version of this plugin is already installed.\n")
```

internal/cmd/connect/command_plugin_install.go
sgagniere: Should be a quick followup. Took a quick look; there's only a handful of spots where `Errorf` is used to format an existing error like this.

brianstrauch: Don't forget `fmt.Errorf()`!

internal/cmd/connect/command_plugin_install.go
sgagniere: They're Java `.properties` files packaged with Confluent Platform. I'm not sure how they're used so I don't know if users can write their own.

I can add an example using the `plugin-directory` and `worker-configs` flags. I'll think about how to word that.

internal/cmd/connect/install_utils.go
```diff
@@ -164,35 +165,16 @@ func hasArchiveInstallation(dir string) bool {
...
-		if j == i {
-			continue
-		}
-		if equalInstallations(i, j) {
-			installations[i].Use = fmt.Sprintf("%s, %s", installations[i].Use, ins.Use)
-		} else {
-			uniqueInstallations = append(uniqueInstallations, installations[i])
-			i = j
+
+	s := types.NewSet()
```
brianstrauch, line 170:
```suggestion
	set := types.NewSet()
```

internal/cmd/connect/install_utils.go
```diff
@@ -164,35 +165,16 @@ func hasArchiveInstallation(dir string) bool {
...
+	for _, ins := range installations {
+		typePathStr := fmt.Sprintf("type=%s,path=%s", ins.Type, ins.Path)
```
brianstrauch, line 172: Is there anything we can do to make `set` generic for any struct, so we don't have to do this?

sgagniere: I'm pretty sure we can do something with generics. Should be a simple followup.

test/connect_test.go
```diff
@@ -87,3 +120,29 @@ func (s *CLITestSuite) TestConnect_Autocomplete() {
 		s.runIntegrationTest(test)
 	}
 }
+
+func (s *CLITestSuite) zipManifest() {
```
sgagniere, line 124: I did this instead of committing the zip file itself so that any changes to the manifest are easier to parse in the diffs.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/examples"
+	"github.com/confluentinc/cli/internal/pkg/form"
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/utils"
+)
+
+type manifest struct {
+	Owner struct {
+		Username string `json:"username"`
```
brianstrauch, line 28: Do we want to support any other formats? YAML? (or is JSON required here?)

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/examples"
+	"github.com/confluentinc/cli/internal/pkg/form"
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/utils"
+)
+
+type manifest struct {
+	Owner struct {
```
brianstrauch, line 27: Let's split this into individual structs? They might be more convenient to work with in the future if they're split up, and these nested structs have a tendency to get a bit out of hand if the schema ever evolves.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	} `json:"license"`
+}
+
+const (
+	archiveInstallation = "ARCHIVE"
+	packageInstallation = "PACKAGE"
+
+	invalidDirectoryErrorMsg       = `plugin directory "%s" does not exist`
+	unexpectedInstallationErrorMsg = "unexpected installation type: %s"
+)
```
brianstrauch, line 51: Use a new `const` block when grouping related constants!
```suggestion
const (
	archiveInstallation = "ARCHIVE"
	packageInstallation = "PACKAGE"
)

const (
	invalidDirectoryErrorMsg       = `plugin directory "%s" does not exist`
	unexpectedInstallationErrorMsg = "unexpected installation type: %s"
)
```

brianstrauch: (Do we really need the first two, though?)

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "Install the latest release of the Datagen connector onto your local Confluent Platform environment.",
+				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest",
+			},
+		),
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
+	}
+
+	cmd.Flags().String("plugin-directory", "", "The plugin installation directory.")
```
brianstrauch, line 68: Let's add `cmd.MarkFlagDirname()` at the bottom of this function?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	cmd.Flags().StringSlice("worker-configs", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any preexisting directories.")
+	cmd.Flags().Bool("dry-run", false, "Simulate an operation without making any changes.")
```
brianstrauch, line 70: Let's make all of our `--dry-run` flags have the same wording?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+		}
+
+		remoteManifest, err := getRemoteManifest(owner, name, version)
+		if err != nil {
+			return err
+		}
+		pluginManifest = remoteManifest
+	}
+
+	// Select plugin-directory
```
brianstrauch, line 107: This comment doesn't seem to match the code underneath it.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	dryRun, err := cmd.Flags().GetBool("dry-run")
+	if err != nil {
+		return err
+	}
+	if dryRun {
+		output.Println("[DRY RUN] Performing a dry run of this command.")
+	}
+	// Verify that the argument corresponds to a valid plugin
+	var pluginManifest *manifest
+	if utils.DoesPathExist(args[0]) {
```
brianstrauch, line 86: Should we put this `if-else` block in a function that returns the manifest?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	if err != nil {
+		return err
+	}
+
+	pluginDir, err := getPluginDirFromFlag(cmd)
+	if err != nil {
+		return err
+	}
+
+	var ins *installation
```
brianstrauch, line 118: Let's find a better name for this since it has a really long scope, and `ins` isn't super descriptive.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	for _, workerConfig := range workerConfigs {
+		if err := updateWorkerConfig(pluginDir, workerConfig, dryRun); err != nil {
+			return err
+		}
+	}
+
+	if dryRun {
+		output.Println("\n[DRY RUN] Dry run successfully completed. All requirements are met; you may proceed with the installation.")
```
brianstrauch, line 196:
```suggestion
		output.Println("\n[DRY RUN] Success. All requirements are met and you may proceed with the installation.")
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	return nil
+}
+
+func parsePluginId(plugin string) (string, string, string, error) {
+	parsePluginErrorMsg := "plugin not found"
+	parsePluginSuggestions := "Provide either a path to a local file or a plugin id from Confluent Hub with the format: `<owner>/<name>:<version>`."
```
brianstrauch, line 204:
```suggestion
	parsePluginSuggestions := "Provide a path to a local file or provide a plugin ID from Confluent Hub with the format: `<owner>/<name>:<version>`."
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+
+	ownerNameSplit := strings.Split(plugin, "/")
```
brianstrauch, line 206: You might like this format better:
```suggestion
	x := strings.SplitN(plugin, "/", 2)
	owner, name := x[0], x[1]
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+
+			return pluginManifest, nil
+		}
+	}
+
+	return nil, errors.Errorf(`failed to find manifest file inside local archive file "%s"`, archivePath)
+}
+
+func getRemoteManifest(owner, name, version string) (*manifest, error) {
+	manifestUrl := fmt.Sprintf("https://api.hub.confluent.io/api/plugins/%s/%s", owner, name)
```
brianstrauch, line 256: Is there a Confluent Hub SDK? Would it be overkill to create one? It might be useful if we add more functionality in the future (i.e. searching for plugins). At the very least, this should probably be moved to a separate file or package.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+		manifestUrl = fmt.Sprintf("%s/versions/%s", manifestUrl, version)
+	}
+
+	r, err := http.Get(manifestUrl)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return nil, errors.New("failed to read manifest file from Confuent Hub")
```
brianstrauch, line 267: Why not surface the backend error?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+	pluginManifest := new(manifest)
+	if err := json.Unmarshal(body, &pluginManifest); err != nil {
+		return nil, err
+	}
+
+	return pluginManifest, nil
+}
+
+func getPluginDirFromFlag(cmd *cobra.Command) (string, error) {
+	if cmd.Flags().Changed("plugin-directory") {
```
brianstrauch, line 285:
```suggestion
	if !cmd.Flags().Changed("plugin-directory") {
		return "", nil
	}
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+
+func existingPluginInstallation(pluginDir string, pluginManifest *manifest) ([]string, error) {
+	// Bundled installations
+	if utils.DoesPathExist(filepath.Join(pluginDir, pluginManifest.Name)) {
+		return nil, errors.New("unable to install plugin because it is already bundled")
+	}
+
+	// Other previous installations
+	immediateDirectory := filepath.Join(pluginDir, fmt.Sprintf("%s-%s", pluginManifest.Owner.Username, pluginManifest.Name))
+	uberJar := filepath.Join(pluginDir, fmt.Sprintf("%s-%s.jar", pluginManifest.Name, pluginManifest.Version))
```
brianstrauch, line 313: Uber? What does that mean?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+		installations = append(installations, immediateDirectory)
+	}
+	if utils.DoesPathExist(uberJar) {
+		installations = append(installations, uberJar)
+	}
+
+	return installations, nil
+}
+
+func uninstall(pathToPlugin string, dryRun, force bool) error {
```
brianstrauch, line 326: Should we add `confluent connect plugin uninstall`?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+		})
+		if err := f.Prompt(form.NewPrompt(os.Stdin)); err != nil {
+			return err
+		}
+		if !f.Responses["confirm"].(bool) {
+			return errors.New("previous versions must be uninstalled to continue")
+		}
+	}
+	if dryRun {
+		output.Println("[DRY RUN] Skipping uninstallation of the existing version.")
```
brianstrauch, line 343:
```suggestion
		output.Println("[DRY RUN] Success.")
```

brianstrauch, line 343: How about we always write: "Uninstalled package XXX" but for dry-run mode it's prefixed with `DRY RUN`?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+}
+
+func checkLicenseAcceptance(pluginManifest *manifest, force bool) error {
+	for _, license := range pluginManifest.License {
+		if force {
+			output.Printf("\nImplicitly agreeing to the following license:\n%s\n%s\n", license.Name, license.Url)
+		} else {
+			f := form.New(form.Field{
+				ID:        "confirm",
+				Prompt:    fmt.Sprintf("\nLicense:\n%s\n%s\nI agree to this software license agreement. ", license.Name, license.Url),
```
brianstrauch, line 443: Do we need the trailing space here?

brianstrauch (commented):
This code could really benefit from making functions smaller / more modular and adding unit tests!

internal/cmd/connect/command_plugin_install.go
sgagniere: The manifests are JSON only, I believe.

internal/cmd/connect/command_plugin_install.go
sgagniere: Nah not really, I just replaced references to these consts with the actual string

internal/cmd/connect/command_plugin_install.go
sgagniere: I have no idea. That's what they're called in the original code so I just carried it over.

internal/cmd/connect/command_plugin_install.go
sgagniere: Good idea.

A dedicated uninstall command probably needs a bit more work, though. This function only removes the plugin, but doesn't update the worker files. I'll need to think about what the right way to do that is (since it adds the parent directory and not the plugin directory itself), and probably add this in when I add the on-prem version of `list`.

internal/cmd/connect/command_plugin_install.go
sgagniere: We can do that. Do we want to use this wording here?

internal/cmd/connect/command_plugin_install.go
sgagniere: Oops, nope.

internal/cmd/connect/command_plugin_install.go
brianstrauch: Did you reach out to the Connect team yet? Maybe someone there can clarify this for you? It would be helpful to write down what this means in a comment.

internal/cmd/connect/command_plugin_install.go
brianstrauch: Wait, how does this function get called currently if there's no `uninstall` command?

sgagniere: It's called on line 118. The command first checks if this plugin already exists in the plugin directory, and then calls this function if so to prompt the user to confirm removal of the existing version.

internal/cmd/connect/command_plugin_install.go
sgagniere: I think I prefer this. There shouldn't be any difference between the function calls themselves when N=2, and I can't assign `x[1]` until I've checked that it exists (which it generally won't with a badly formed id).

internal/cmd/connect/command_plugin_install.go
sgagniere: I don't think there is. I wasn't even able to find documentation on this API at all. I'll move it somewhere though.

internal/cmd/connect/command_plugin_install.go
sgagniere: I reached out again a few minutes ago.

internal/cmd/connect/command_plugin_install.go
brianstrauch: Oh, it replaces it if it already exists.

sgagniere: I'll rename this function to `replaceInstallation` until I get to writing the `uninstall` command.

internal/cmd/connect/command_plugin_install.go
sgagniere: What about one struct but flattened instead?

Splitting into multiple structs probably means passing at least 2 structs around in most places.

brianstrauch: You can still nest them, but when you define them you can define them separately.

sgagniere: Ok. Does splitting out License and Archive sounds good?

brianstrauch: I was thinking owner, license, and archive should be defined separately since they're all structs

internal/cmd/connect/command_plugin_install.go
```diff
@@ -24,18 +24,27 @@ import (
 )
 
 type manifest struct {
+	Name    string `json:"name"`
+	Title   string `json:"title"`
+	Version string `json:"version"`
+}
```
brianstrauch: This is what I was thinking of!
```suggestion
type manifest struct {
	Owner   owner   `json:"owner"`
	Name    string  `json:"name"`
	Title   string  `json:"title"`
	Version string  `json:"version"`
	Archive archive `json:"archive"
	License license `json:"license"`
}
```

brianstrauch: This is what I was thinking of! This way all of our functions still only need to return one variable.
```suggestion
type manifest struct {
	Owner   owner   `json:"owner"`
	Name    string  `json:"name"`
	Title   string  `json:"title"`
	Version string  `json:"version"`
	Archive archive `json:"archive"
	License license `json:"license"`
}
```

internal/cmd/connect/install_utils.go
sgagniere: This ended up being pretty simple and didn't require too many changes elsewhere in the code, so I'll roll it into this PR.

Since I'm only comparing two of the fields of this struct, I modified the struct as well.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -24,31 +24,28 @@ import (
...
 type manifest struct {
-	Name    string `json:"name"`
-	Title   string `json:"title"`
-	Version string `json:"version"`
+	Name    string    `json:"name"`
+	Title   string    `json:"title"`
+	Version string    `json:"version"`
+	Owner   owner     `json:"owner"`
+	Archive archive   `json:"archive"`
+	License []license `json:"license"`
```
brianstrauch, line 31: Plural since it's a list?
```suggestion
	Licenses []license `json:"licenses"`
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -206,120 +203,98 @@ func parsePluginId(plugin string) (string, string, string, error) {
...
-func getManifest(id string) (*manifest, *manifestOwner, *manifestArchive, *manifestLicense, error) {
+func getManifest(id string) (*manifest, error) {
 	if utils.DoesPathExist(id) {
 		// if installing plugin from local archive
-		return getLocalManifest(id)
+		localManifest, err := getLocalManifest(id)
+		if err != nil {
+			return nil, err
+		}
+		return localManifest, nil
```
brianstrauch, line 216:
```suggestion
		return getLocalManifest(id)
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -206,120 +203,98 @@ func parsePluginId(plugin string) (string, string, string, error) {
...
-			return nil, nil, nil, nil, err
+			return nil, err
 		}
 
-		return getRemoteManifest(owner, name, version)
+		remoteManifest, err := getRemoteManifest(owner, name, version)
+		if err != nil {
+			return nil, err
+		}
+		return remoteManifest, nil
```
brianstrauch, line 228:
```suggestion
		return getRemoteManifest(owner, name, version)
```

venkatteki (approved):
LGTM

MuweiHe (commented):
how is manual testing going?

internal/cmd/connect/hub_api_utils.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return nil, errors.New("failed to retrieve archive from Confuent Hub")
+	}
+	defer r.Body.Close()
+
+	return io.ReadAll(r.Body)
```
MuweiHe, line 57:
```suggestion

	defer r.Body.Close()
	return io.ReadAll(r.Body)
```

internal/cmd/connect/install_utils.go
```diff
@@ -0,0 +1,439 @@
...
+	Path string
+}
+
+type WorkerConfig struct {
+	Path string
+	Use  string
+}
+
+type listOut struct {
+	Number      string `human:""`
```
MuweiHe, line 39: `Number` is omitted?

internal/cmd/connect/install_utils.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+		choice, err := strconv.Atoi(f.Responses["installation"].(string))
+		if err != nil || choice < 1 || choice > len(installations) {
+			return nil, errors.Errorf("your choice must be in the range %d to %d (inclusive)", 1, len(installations))
+		}
+		return &installations[choice-1], nil
+	}
+}
+
+func findInstallationDirectories() ([]platformInstallation, error) {
```
MuweiHe, line 90: does these commands require more read/write access than our plugin commands before? like to certain paths?

internal/cmd/connect/install_utils.go
```diff
@@ -0,0 +1,439 @@
...
+	if filepath.ToSlash(cliDirectory) == "/usr/bin" && hasPackageInstallation {
+		installation := platformInstallation{
+			Location: platformLocation{
+				Type: "PACKAGE",
+				Path: filepath.FromSlash("/"),
+			},
+			Use: cliUse,
+		}
+		result = append(result, installation)
+	} else if filepath.Base(cliDirectory) == "bin" && hasArchiveInstallation(filepath.Dir(cliDirectory)) {
```
MuweiHe, line 157: how is these 2 conditions (`usr/bin` and `bin`) determined? Could there be any chance that neither of these 2 conditions are true?

internal/cmd/connect/install_utils.go
sgagniere: Yeah I wanted it to show up with an empty header. The number is only there for the interactive prompt which asks the user to pick a number.

internal/cmd/connect/install_utils.go
sgagniere: `/usr/bin` is where the CLI is installed for package manager installations, and `bin` is where it's installed in archive installations.

The second one is also local, so if the CLI determines that it's in side a folder called `bin` it will check the parent directory of `bin` to see if it's a Confluent Platform installation.

Unless the user relocates `confluent` for whatever reason, these should not change.

internal/cmd/connect/install_utils.go
sgagniere: I'm not sure what read/write permissions the plugin commands need, but this command needs read access to for the various paths checked here (it looks for the existence of specific sub-directories to detect Confluent Platform installations), and it needs read/write permissions for installing the plugin and updating the worker config files.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,458 @@
...
+		return err
+	}
+
+	workerConfigs, err := getWorkerConfigsFromFlag(cmd)
+	if err != nil {
+		return err
+	}
+
+	var installation *platformInstallation
+	prompt := form.NewPrompt(os.Stdin)
```
sgagniere, line 117: Turns out the reason I couldn't provide multiple inputs was because I was creating a new prompt every time. Creating this one at the start and passing it around is the only way I found to allow integration testing.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -100,19 +102,25 @@ func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Select plugin-directory
+	// Parse plugin & worker flags
```
brianstrauch, line 105: The function names below convey the same meaning as this comment
```suggestion
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -279,6 +282,30 @@ func getPluginDirFromFlag(cmd *cobra.Command) (string, error) {
 	return pluginDir, nil
 }
 
+func getWorkerConfigsFromFlag(cmd *cobra.Command) ([]string, error) {
+	if !cmd.Flags().Changed("worker-configs") {
+		return nil, nil
+	}
```
brianstrauch, line 288: This doesn't look like it's needed. If the flag isn't passed, we'll default to an empty slice, and nothing will happen below.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -408,7 +435,7 @@ func unzipPlugin(pluginManifest *manifest, zipFiles []*zip.File, pluginDir strin
 	return nil
 }
 
-func checkLicenseAcceptance(pluginManifest *manifest, force bool) error {
+func checkLicenseAcceptance(pluginManifest *manifest, prompt *form.RealPrompt, force bool) error {
```
brianstrauch, line 438: I think RealPrompt implements Prompt? It seems like the intention is to use Prompt here, so we can pass in a mock prompt during unit tests (which you should write)

test/test-server/hub_handlers.go
```diff
@@ -0,0 +1,128 @@
...
+		switch r.Method {
+		case http.MethodGet:
+			responseManifest := newManifest()
+			if utils.DoesPathExist(archivePath) {
+				archive, err := os.ReadFile(archivePath)
+				require.NoError(t, err)
+	
+				responseManifest.Archive = ccstructs.Archive{
+					Url:  fmt.Sprintf("%s/api/plugins/confluentinc/integration-test-plugin/versions/0.1.0/confluentinc-integration-test-plugin.zip", TestHubUrl.String()),
+					Md5:  fmt.Sprintf("%x", md5.Sum(archive)),
```
sgagniere, line 59: I'm calculating the md5 sum instead of hardcoding to future proof a bit. Also, a zip can sometimes give a different md5 because of metadata.

test/test-server/hub_handlers.go
```diff
@@ -0,0 +1,128 @@
...
+			w.WriteHeader(http.StatusNoContent)
+		}
+	}
+}
+
+// Handler for: "/api/plugins/{owner}/{id}/versions/{version}"
+func handleHubPluginVersion(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		if vars["id"] == "dne-connector" || vars["owner"] != "confluentinc" || vars["version"] != "0.0.5" {
```
sgagniere, line 81: Using `0.0.5` here instead of `0.1.0` just to make sure that the integration test is reaching the correct endpoint (basically just a test of `GetRemoteManifest`).

internal/pkg/hub/manifest.go
```diff
@@ -7,36 +7,12 @@ import (
 	"net/http"
 	"net/http/httputil"
 
+	"github.com/confluentinc/cli/internal/pkg/ccstructs"
 	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/log"
 )
 
-type Manifest struct {
```
sgagniere: I got an import cycle error since this package calls `testserver` for the test URL, but I wanted to use this struct in the hub handler file. Couldn't put this struct in the `connect` package for the same reason (that package imports this one).

So I dumped this struct into our internal `ccstructs` package.

internal/pkg/ccstructs/hub_structs.go
```diff
@@ -0,0 +1,26 @@
+package ccstructs
```
brianstrauch, line 1: Are these really part of ccstructs? I thought Confluent Hub was for CP only.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -89,9 +65,11 @@ func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
 	}
 	if dryRun {
 		output.Println("[DRY RUN] Performing a dry run of this command.")
+	} else {
+		output.Println("Beginning installation.")
 	}
```
brianstrauch: How about: `[DRY RUN] Beginning installation` so they match?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -362,13 +340,13 @@ func installFromLocal(pluginManifest *manifest, archivePath, pluginDir string) e
...
-func installFromRemote(pluginManifest *manifest, pluginDir string) error {
-	archive, err := getRemoteArchive(pluginManifest)
+func (c *pluginCommand) installFromRemote(pluginManifest *ccstructs.Manifest, pluginDir string) error {
+	archive, err := c.HubClient.GetRemoteArchive(pluginManifest)
 	if err != nil {
 		return err
 	}
 
-	checksumErrorMsg := "%s checksum for downloaded archive (%s) does not match checksum in manifest (%s) for plugin %s"
+	checksumErrorMsg := "%s checksum for downloaded archive (%s) does not match checksum in manifest (%s) for plugin \"%s\""
```
brianstrauch, line 349:
```suggestion
	checksumErrorMsg := `%s checksum for downloaded archive (%s) does not match checksum in manifest (%s) for plugin "%s"`
```

internal/pkg/cmd/prerunner.go
```diff
@@ -70,6 +71,7 @@ type AuthenticatedCLICommand struct {
 	V2Client          *ccloudv2.Client
 	MDSClient         *mds.APIClient
 	MDSv2Client       *mdsv2alpha1.APIClient
+	HubClient         *hub.Client
```
brianstrauch, line 74: We have a lot of custom clients here that shouldn't be initialized for every authenticated command: Kafka REST, Schema Registry, Metrics, Flink (soon), and now Confluent Hub. I think Kafka REST is the only one to do this, but they should only be initialized before we use each of them (as a singleton). We should also try to converge on some shared code pattern for all of these.

internal/pkg/cmd/prerunner.go
```diff
@@ -555,6 +567,10 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 			}
 		}
 
+		if err := r.setHubClient(command); err != nil {
```
brianstrauch, line 570: As per my comment above, this will get initialized even for commands that don't need it, and as we add clients over time, this function will grow linearly. We should copy the style of Kafka REST and write a `GetConfluentHubClient()` (or similarly named) function that instantiates it right before it's used.

internal/pkg/hub/client.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// Client represents a Confluent Cloud Client as defined by ccloud-sdk-go-v2
+type Client struct {
+	URL    string
+	Debug  bool
+	Client *http.Client
+}
+
+func NewClient(isTest bool, unsafeTrace bool) *Client {
```
brianstrauch, line 17:
```suggestion
func NewClient(isTest, unsafeTrace bool) *Client {
```

internal/pkg/hub/client.go
```diff
@@ -0,0 +1,30 @@
...
+	if isTest {
+		url = testserver.TestHubUrl.String()
+	}
+
+	return &Client{
+		URL:   url,
+		Debug: unsafeTrace,
+		Client: &http.Client{
+			Timeout: 5 * time.Second,
+		},
```
brianstrauch, line 28:
```suggestion
		Client: &http.Client{Timeout: 5 * time.Second},
```

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/http"
+	"net/http/httputil"
+
+	"github.com/confluentinc/cli/internal/pkg/ccstructs"
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/log"
+)
+
+func (c *Client) GetRemoteManifest(owner, name, version string) (*ccstructs.Manifest, error) {
+	manifestUrl := fmt.Sprintf("%s/api/plugins/%s/%s", c.URL, owner, name)
```
brianstrauch, line 16: Have you talked to the Connect team about making an API spec / generated SDK for these endpoints?

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if c.Debug {
+		dump, err := httputil.DumpRequestOut(req, true)
+		if err != nil {
+			return nil, err
+		}
+		log.CliLogger.Tracef("\n%s\n", string(dump))
```
brianstrauch, line 31: We should make an effort to standardize the logging messages- typically log files don't contain whitespace gaps.
```suggestion
		log.CliLogger.Trace(string(dump))
```

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+	defer r.Body.Close()
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode != http.StatusOK {
+		response := make(map[string]interface{})
+		_ = json.Unmarshal(body, &response)
+		if errorMessage, ok := response["message"]; ok {
```
brianstrauch, line 56: Don't we have a general-purpose function that catches HTTP error details?

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+		log.CliLogger.Tracef("\n%s\n", string(dump))
+	}
+
+	r, err := c.Client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return nil, errors.New("failed to retrieve archive from Confuent Hub")
```
brianstrauch, line 90: Since we're naming this command `confluent connect plugin install`, I think we should fully move away from using the name "Confluent Hub".

```suggestion
		return nil, errors.New("failed to retrieve plugin archive")
```

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	defer r.Body.Close()
```
brianstrauch, line 93: The point of the `defer` keyword is to put it right after the resource is created. I think this would be best suited for after line 87.

test/fixtures/output/connect/plugin/install/platform-not-found.golden
```diff
@@ -1,3 +1,4 @@
+Beginning installation.
```
brianstrauch, line 1: How long does installation take? If < 1s, I think we should remove this message. Our other commands don't do this.

Of course, you can always log it with DEBUG!

brianstrauch, line 1: How long does installation take? If < 1s, I think we should remove this message. Our other commands don't do this.

Of course, you can always log it with DEBUG!

test/fixtures/output/connect/plugin/install/remote-bad-md5.golden
```diff
@@ -0,0 +1,10 @@
+Beginning installation.
+
+Using the only available Confluent Platform installation at "test/fixtures/input/connect/confluent-7.3.3".
+Do you want to install this plugin into "test/fixtures/input/connect/confluent-7.3.3/share/confluent-hub-components"? (y/n):
```
brianstrauch, line 4: We should test what happens when a user types "n". In this case, I would expect the program to exit and give a suggestion on how they can change the install directory.

test/fixtures/output/connect/plugin/install/remote-bad-md5.golden
```diff
@@ -0,0 +1,10 @@
+Beginning installation.
+
+Using the only available Confluent Platform installation at "test/fixtures/input/connect/confluent-7.3.3".
+Do you want to install this plugin into "test/fixtures/input/connect/confluent-7.3.3/share/confluent-hub-components"? (y/n): 
+License:
+Example License 1.0
+url-to-license-goes-here
```
brianstrauch, line 7: Can we fit this on one line? Also, it would be great to use an actual URL so we can better visualize what this will look like in practice.
```suggestion
License: Example License 1.0 (url-to-license-goes-here)
```

test/fixtures/output/connect/plugin/install/remote-bad-md5.golden
```diff
@@ -0,0 +1,10 @@
...
+I agree to this software license agreement. (y/n): 
+Installing Integration Test Plugin 0.1.0, provided by Confluent, Inc.
+Error: md5 checksum for downloaded archive (7a920f4187137251171abcd7c7dd9844) does not match checksum in manifest (12345) for plugin "integration-test-plugin"
```
brianstrauch, line 10:
```suggestion
Error: MD5 checksum for downloaded archive (7a920f4187137251171abcd7c7dd9844) does not match checksum in manifest for plugin "integration-test-plugin" (12345)
```

test/fixtures/output/connect/plugin/install/remote-bad-sha1.golden
```diff
@@ -0,0 +1,10 @@
+Beginning installation.
+
+Using the only available Confluent Platform installation at "test/fixtures/input/connect/confluent-7.3.3".
```
brianstrauch, line 3: What if the user has two versions of CP installed, but this command only detects one of them? I think we should always prompt, unless it's explicitly passed with a flag.

test/fixtures/output/connect/plugin/install/remote-specific-version.golden
```diff
@@ -0,0 +1,24 @@
...
+Detected the following worker configuration files:
+    |                                         Path                                         | Description  
+----+--------------------------------------------------------------------------------------+--------------
+  1 | test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-distributed.properties | Standard     
+  2 | test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-standalone.properties  | Standard     
+Do you want to update all detected config files? (y/n): 
+Adding plugin installation directory to the plugin path in the following files:
+	* test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-distributed.properties
+	* test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-standalone.properties
+[DRY RUN] Skipping update of worker configuration file "test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-distributed.properties".
```
brianstrauch, line 21: Instead of printing "skipping update", I think we should print the same message as we would normally print, but prefixed with "[DRY-RUN]"

test/fixtures/output/connect/plugin/install/remote-specific-version.golden
```diff
@@ -0,0 +1,24 @@
...
+[DRY RUN] Skipping update of worker configuration file "test/fixtures/input/connect/confluent-7.3.3/etc/kafka/connect-standalone.properties".
+
+[DRY RUN] Success. All requirements are met and you may proceed with the installation.
```
brianstrauch, line 24: How about "Installed plugin XXX." (matching the style of our other commands?)

brianstrauch: Or if there is more useful data to show, a table?

test/test-server/hub_router.go
```diff
@@ -0,0 +1,18 @@
...
+	"testing"
+
+	"github.com/gorilla/mux"
+)
+
+func NewHubRouter(t *testing.T) *mux.Router {
+	router := mux.NewRouter()
+	router.Use(defaultHeaderMiddleware)
+
+	router.HandleFunc("/api/plugins/{owner}/{id}", handleHubPlugin(t))
```
brianstrauch, line 13: Can we match the style of our other tests? We're using a list or map for all of our routes (I forget which).

brianstrauch (changes requested):

internal/pkg/ccstructs/hub_structs.go
sgagniere: It's not, but I don't have any better place to put it.

internal/pkg/hub/manifest.go
sgagniere: Nope.

internal/pkg/hub/manifest.go
sgagniere: I did this to match the style of the logging in the ccloud-sdk-go-v2 callAPI functions: `log.Printf("\n%s\n", string(dump))`.

internal/pkg/hub/manifest.go
sgagniere: It's still called Confluent Hub in the UI, so for now I think we should keep references to Confluent Hub. At the very least, there's another reference to Confluent Hub in a suggestion string explicitly telling the user to get the id from Confluent Hub that we should keep if nothing else.

internal/pkg/hub/manifest.go
sgagniere: We have some v2 error catchers, but I'm not sure beyond that.

test/fixtures/output/connect/plugin/install/remote-bad-md5.golden
sgagniere: I already do:
https://github.com/confluentinc/cli/pull/1888/files#diff-ae2a012856605648e0401a8e6d309dfda044924f7a0eb76c52d78a6f08f1d7fd
https://github.com/confluentinc/cli/pull/1888/files#diff-db1052b63cac15718484af3b558f1d9584af0a4ceb57b2f960101ec650ac762c
(`interactive-select-plugin-directory.golden` and `interactive-select-plugin-directory-fail.golden`)

When a user types `n`, it prompts the user to specify a directory. This is the original tool's behavior.

test/fixtures/output/connect/plugin/install/remote-bad-md5.golden
sgagniere: I made these uppercase, but it feels awkward to me to relocate the second checksum to the end, since both checksums are supposed to be for the plugin's archive.

internal/pkg/ccstructs/hub_structs.go
sgagniere: Moved it to a new pkg called `cpstructs`

test/fixtures/output/connect/plugin/install/remote-bad-sha1.golden
sgagniere: The CP installation is only used to find the standard installation directory if they didn't supply `--plugin-directory` or the worker config files if they didn't supply `--worker-configs`.

So in this case, the user could just decline on the next prompt and rerun with both flags.

If it's better UX to add a flag where users can supply a CP installation, I can do that.

test/fixtures/output/connect/plugin/install/platform-not-found.golden
sgagniere: I frankly just did it so that the spacing would work out across all the different logic branches, but I just removed it and wrestled with the spacing for a bit.

internal/cmd/connect/command_plugin.go
```diff
@@ -6,23 +6,31 @@ import (
...
 
-	c := &pluginCommand{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+	c := &pluginCommand{}
 
-	cmd.AddCommand(c.newDescribeCommand())
-	cmd.AddCommand(c.newListCommand())
+	if cfg.IsCloudLogin() {
+		c.AuthenticatedCLICommand = pcmd.NewAuthenticatedCLICommand(cmd, prerunner)
+
+		cmd.AddCommand(c.newDescribeCommand())
```
brianstrauch, line 27: We want users to get a message along the lines of "Please log in to Confluent Cloud" if they're logged in to Confluent Platform but they run `confluent connect plugin describe`. Can you move this line outside of the `if`, and make sure it has the correct annotation? I think this is what we want:
```
Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin}
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (c *pluginCommand) newInstallCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install <plugin>",
+		Short: "Install a Connect plugin.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.install,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "Install the latest release of the Datagen connector onto your local Confluent Platform environment.",
```
brianstrauch, line 40:
```suggestion
				Text: "Install the latest version of the Datagen connector into your local Confluent Platform environment.",
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+			examples.Example{
+				Text: "Install the latest release of the Datagen connector in a user specified directory and update one worker-config file.",
+				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configs $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
+			},
+		),
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
+	}
+
+	cmd.Flags().String("plugin-directory", "", "The plugin installation directory. If not specified, a default will be selected based on your Confluent Platform installation.")
+	cmd.Flags().StringSlice("worker-configs", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any preexisting directories.")
```
brianstrauch, line 52:
```suggestion
	cmd.Flags().StringSlice("worker-configs", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any pre-existing directories.")
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+	cmd.Flags().String("confluent-platform", "", "The path to a Confluent Platform archive installation. If not specified, this command will search for Confluent Platform installations in common locations.")
```
brianstrauch, line 53: > archive installation

Like a ZIP file? That doesn't sound right.

brianstrauch:
```suggestion
	cmd.Flags().String("confluent-platform", "", "The path to a Confluent Platform archive installation. By default, this command will search for Confluent Platform installations in common locations.")
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+	return cmd
+}
+
+func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
+	dryRun, err := cmd.Flags().GetBool("dry-run")
+	if err != nil {
+		return err
+	}
+
+	if err := c.InitializeHubClient(); err != nil {
```
brianstrauch, line 68: Should we follow the approach of the Flink Gateway and make this function return the HTTP client?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+		return err
+	}
+
+	force, err := cmd.Flags().GetBool("force")
+	if err != nil {
+		return err
+	}
+
+	if cmd.Flags().Changed("plugin-directory") && cmd.Flags().Changed("worker-configs") && cmd.Flags().Changed("confluent-platform") {
+		return errors.New("at most two of `--plugin-directory`, `--worker-configs`, and `--confluent-platform` may be set")
```
brianstrauch, line 83: This seems like a weird requirement. So any of the three combinations of two flags can be passed here?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+	cmd.Flags().String("confluent-platform", "", "The path to a Confluent Platform archive installation. If not specified, this command will search for Confluent Platform installations in common locations.")
+	pcmd.AddDryRunFlag(cmd)
+	cmd.Flags().Bool("force", false, "Proceed without user input.")
+
+	cobra.CheckErr(cmd.MarkFlagDirname("plugin-directory"))
+
+	return cmd
+}
+
+func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
```
brianstrauch, line 62: Any way we can reduce the size of this function? Are there any blocks of code we can pull out so it's easier to unit test?

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+			}
+			defer manifestFile.Close()
+
+			jsonByteArr, err := io.ReadAll(manifestFile)
+			if err != nil {
+				return nil, err
+			}
+
+			pluginManifest := new(cpstructs.Manifest)
+			if err := json.Unmarshal(jsonByteArr, &pluginManifest); err != nil {
```
brianstrauch, line 239: I think `json.NewDecoder().Decode()` is a cleaner interface!

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,445 @@
...
+			}
+		}
+
+		zipFileReader, err := zipFile.Open()
+		if err != nil {
+			return errors.Wrapf(err, "failed to read file %s from archive", zipFile.Name)
+		}
+		defer zipFileReader.Close()
+
+		destFile, err := os.OpenFile(destFilePath, os.O_CREATE|os.O_RDWR, 0644)
```
brianstrauch, line 406: Why do we need to set the file permissions here?

internal/pkg/hub/client.go
```diff
@@ -0,0 +1,28 @@
+package hub
+
+import (
+	"net/http"
+	"time"
+
+	testserver "github.com/confluentinc/cli/test/test-server"
+)
+
+// Client represents a Confluent Cloud Client as defined by ccloud-sdk-go-v2
```
brianstrauch, line 10: This seems like it was copied over.

internal/pkg/hub/client.go
```diff
@@ -0,0 +1,28 @@
...
+func NewClient(isTest, unsafeTrace bool) *Client {
+	url := "https://api.hub.confluent.io"
+	if isTest {
+		url = testserver.TestHubUrl.String()
+	}
+
+	return &Client{
+		URL:    url,
+		Debug:  unsafeTrace,
+		Client: &http.Client{Timeout: 5 * time.Second},
```
brianstrauch, line 26: Most of our HTTP clients use a retryable client... how come we're not using that here?

internal/pkg/hub/manifest.go
```diff
@@ -0,0 +1,105 @@
...
+	"io"
+	"net/http"
+	"net/http/httputil"
+
+	"github.com/confluentinc/cli/internal/pkg/cpstructs"
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/log"
+)
+
+const clientNotInitializedErrorMsg = "Hub client not initialized"
```
brianstrauch, line 15: Do we still want to refer to this as "Confluent Hub" or are we trying to deprecate that terminology?

test/fixtures/output/connect/plugin/install/both-flags-force.golden
```diff
@@ -0,0 +1,10 @@
+
```
brianstrauch, line 1: Can this newline be removed?

internal/cmd/connect/command_plugin_install.go
sgagniere: Archive installation just means it was installed by the user from a zip or tar file, as opposed to using a package manager.

internal/cmd/connect/command_plugin_install.go
sgagniere: The `--confluent-platform` flag gets ignored completely if both of the other flags are specified, which is why I error out if all three are there. The platform installation is only used as a helper to suggest values to the user for the other two if the user didn't provide them.

The user can provide zero, one, or two of the flags.

internal/pkg/hub/manifest.go
sgagniere: The website for plugins still calls it Confluent Hub, so I want to be consistent with it.

internal/cmd/connect/command_plugin_install.go
sgagniere: It's a required argument for this function, so I just went with Chris' value. This one gives rdwr permissions to the owner and rd permission to everyone else, which seems ok.

internal/cmd/connect/command_plugin_install.go
sgagniere: I still think we should find a solution where we aren't initializing clients we don't need but also don't have to explicitly initialize clients right before we use them, but for now I switched this to the flink gateway style.

test/fixtures/output/connect/plugin/install/both-flags-force.golden
sgagniere: Yup 👍

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,461 @@
...
+		return getLocalManifest(id)
+	} else {
+		// if installing plugin from Confluent Hub
+		owner, name, version, err := parsePluginId(id)
+		if err != nil {
+			return nil, err
+		}
+
+		return client.GetRemoteManifest(owner, name, version)
+	}
```
brianstrauch, line 186: The first branch always returns so we can save some indentation.
```suggestion
	}
	
	// if installing plugin from Confluent Hub
	owner, name, version, err := parsePluginId(id)
	if err != nil {
		return nil, err
	}
	
	return client.GetRemoteManifest(owner, name, version)
```

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,461 @@
...
+				return errors.New("you must accept all license agreements to install this plugin")
+			}
+		}
+	}
+	output.Print("\n")
+
+	return nil
+}
+
+func addDryRunPrefix(msg string) string {
```
brianstrauch, line 439: I wonder if other commands could reuse this?

internal/pkg/types/set.go
```diff
@@ -1,23 +1,24 @@
 package types
 
-type Set map[string]bool
+type Set[V comparable] map[V]bool
```
brianstrauch, line 3: Nice!

test/test-server/hub_handlers.go
```diff
@@ -0,0 +1,129 @@
...
+		Licenses: []cpstructs.License{
+			{
+				Name: "Apache License 2.0",
+				Url:  "https://www.apache.org/licenses/LICENSE-2.0",
+			},
+		},
+	}
+}
+
+// Handler for: "/api/plugins/{owner}/{id}"
```
brianstrauch, line 39: Has this team written an OpenAPI spec, or have you asked them to? It would be awesome to have an SDK for these endpoints. As of right now, I don't think `--unsafe-trace` will work.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,461 @@
...
+	"github.com/confluentinc/cli/internal/pkg/form"
+	"github.com/confluentinc/cli/internal/pkg/hub"
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/utils"
+)
+
+const (
+	invalidDirectoryErrorMsg       = `plugin directory "%s" does not exist`
+	unexpectedInstallationErrorMsg = "unexpected installation type: %s"
+	workerProcessRegexStr          = `org\.apache\.kafka\.connect\.cli\.Connect(Distributed|Standalone)`
```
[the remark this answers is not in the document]
sgagniere: I use it in two functions.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,461 @@
...
+			},
+			examples.Example{
+				Text: "Install the latest version of the Datagen connector in a user specified directory and update one worker-config file.",
+				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configs $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
+			},
+		),
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
+	}
+
+	cmd.Flags().String("plugin-directory", "", "The plugin installation directory. If not specified, a default will be selected based on your Confluent Platform installation.")
```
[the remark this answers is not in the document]
sgagniere: It does, but it changes depending on whether or not the user installed CP using a package manager.

internal/cmd/connect/command_plugin_install.go
sgagniere: Yup. I moved this to utils and changed the two other instances of dry run messages to call this.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,461 @@
...
+		Short: "Install a Connect plugin.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.install,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "Install the latest version of the Datagen connector into your local Confluent Platform environment.",
+				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest",
+			},
+			examples.Example{
+				Text: "Install the latest version of the Datagen connector in a user specified directory and update one worker-config file.",
```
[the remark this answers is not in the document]
sgagniere: Went ahead and made this change with the hyphen between 'user' and 'specified' instead of 'specified' and 'directory'.

test/test-server/hub_handlers.go
sgagniere: I'll ask.

Unsafe trace works; I put support for it in the functions in `internal/pkg/hub/manifest.go`. Here's an example (I cut out the response body since it's pretty large):
```
connect plugin install confluentinc/kafka-connect-datagen:latest --dry-run --unsafe-trace
2023-06-06T12:10:46.073+0800 [TRACE] 
GET /api/plugins/confluentinc/kafka-connect-datagen HTTP/1.1
Host: api.hub.confluent.io
User-Agent: Confluent-CLI/v3.16.0-SNAPSHOT-cdc7f720e (https://confluent.io; support@confluent.io)
Accept-Encoding: gzip



2023-06-06T12:10:46.073+0800 [DEBUG] performing request%!(EXTRA string=method, string=GET, string=url, *url.URL=https://api.hub.confluent.io/api/plugins/confluentinc/kafka-connect-datagen)
2023-06-06T12:10:47.028+0800 [TRACE] 
HTTP/1.1 200 OK
Content-Length: 2762
Access-Control-Allow-Origin: null
Cache-Control: no-transform, s-maxage=26
Connection: keep-alive
Content-Type: application/vnd.confluent.v1+json
Date: Tue, 06 Jun 2023 04:10:47 GMT
Etag: "9c90bbe00bf0b8bf2cba64d8c5e8bd7ffa4a9e6369a9869fc3679017a3279480"
Server: nginx
```

I did notice while testing this though that I didn't set the user agent (it's fixed in this output); I'll push that momentarily.

sgagniere: There doesn't appear to be an sdk but there's an openapi doc, I'll link it to you in slack if you're interested.

internal/cmd/connect/command_plugin_install.go
```diff
@@ -0,0 +1,456 @@
...
+			examples.Example{
+				Text: "Install the latest version of the Datagen connector in a user-specified directory and update a worker-config file.",
+				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configs $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
+			},
+		),
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
+	}
+
+	cmd.Flags().String("plugin-directory", "", "The plugin installation directory. If not specified, a default will be selected based on your Confluent Platform installation.")
+	cmd.Flags().StringSlice("worker-configs", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any prior directories.")
```
brianstrauch, line 54:
```suggestion
	cmd.Flags().StringSlice("worker-configurations", []string{}, "A comma-separated list of paths to one or more Kafka Connect worker configuration files. Each worker file will be updated to load plugins from the plugin directory in addition to any prior directories.")
```

internal/cmd/connect/command_plugin_install_test.go
```diff
@@ -0,0 +1,200 @@
...
+	archivePath := fmt.Sprintf("%s/test-plugin.zip", tempDir)
+	err = newTestZip(archivePath)
+	require.NoError(t, err)
+
+	manifest := &cpstructs.Manifest{
+		Name:    "unit-test-plugin",
+		Version: "0.0.0",
+		Owner: cpstructs.Owner{
+			Username: "confluentinc",
+		},
```
brianstrauch, line 158:
```suggestion
		Owner: cpstructs.Owner{Username: "confluentinc"},
```

internal/pkg/exec/exec.go
```diff
@@ -0,0 +1,21 @@
+package exec
```
brianstrauch, line 1: Is this file being used? Can't tell because the PR is so big and these names are very generic.

brianstrauch (approved):

internal/pkg/exec/exec.go
sgagniere: Yeah it's used in `internal/cmd/connect/install_utils.go`. I did this so that I can mock `os/exec` in the unit tests.

PR #1890, opened 2023-05-04
Fix GoReleaser Pro

brianstrauch:
What
----
No need to break the GoReleaser run into two separate commands. We can build the Linux binaries first and add them as `prebuilt`. Also fixed some small bugs.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -121,7 +121,7 @@ builds:
     goamd64:
       - v1
     prebuilt:
-      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_v1{{ end }}/confluent
+      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_{{ . }}{{ end }}/confluent
 
 release:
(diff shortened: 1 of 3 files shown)


.goreleaser.yml
```diff
@@ -121,7 +121,7 @@ builds:
     goamd64:
       - v1
     prebuilt:
-      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_v1{{ end }}/confluent
+      path: dist/confluent_{{ .Os }}_{{ .Arch }}_{{ with .Amd64 }}_{{ . }}{{ end }}/confluent
```
sgagniere, line 124: Is the `.` a wildcard?

sgagniere (approved):

.goreleaser.yml
brianstrauch: I think if `.Amd64` is defined, the `.` gets replaced with its value. I'm copying from the documentation here: https://goreleaser.com/customization/builds/#import-pre-built-binaries (Although I couldn't find any documentation that explains how it works... so I'm going in blind)

PR #1891, opened 2023-05-05
Fix dist folder management for `make gorelease`

brianstrauch:
What
----
The `dist/` folder should only be deleted if it exists, and it shouldn't be cleaned at the start of `goreleaser release` since it contains the Linux binaries.

### DIFF
--- a/mk-files/release.mk
+++ b/mk-files/release.mk
@@ -77,10 +77,10 @@ gorelease:
 	$(eval token := $(shell (grep github.com ~/.netrc -A 2 | grep password || grep github.com ~/.netrc -A 2 | grep login) | head -1 | awk -F' ' '{ print $$2 }'))
 	
 	$(aws-authenticate) && \
-	rm -r dist/ && \
+	if [ -d dist/ ]; rm -r dist/; fi && \
 	mkdir dist/ && \
 	scripts/build_linux_glibc.sh && \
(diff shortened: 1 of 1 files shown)


mk-files/release.mk
```diff
@@ -77,10 +77,10 @@ gorelease:
 	$(eval token := $(shell (grep github.com ~/.netrc -A 2 | grep password || grep github.com ~/.netrc -A 2 | grep login) | head -1 | awk -F' ' '{ print $$2 }'))
 	
 	$(aws-authenticate) && \
-	rm -r dist/ && \
+	if [ -d dist/ ]; rm -r dist/; fi && \
```
sgagniere, line 80: `rm -rf dist` should do this without the need for `if`. The `f` flag suppresses the `no file or directory` error.

brianstrauch:
```suggestion
	rm -rf dist && \
```

sgagniere (approved):

mk-files/release.mk
```diff
@@ -77,10 +77,10 @@ gorelease:
 	$(eval token := $(shell (grep github.com ~/.netrc -A 2 | grep password || grep github.com ~/.netrc -A 2 | grep login) | head -1 | awk -F' ' '{ print $$2 }'))
 	
 	$(aws-authenticate) && \
-	rm -r dist/ && \
+	rm -rf dist && \
```
brianstrauch, line 80:
```suggestion
	rm -rf dist/ && \
```

PR #1900, opened 2023-05-09
Fix FEDRAMP SSO

brianstrauch:
What
----
We need to call the correct realm endpoint to determine the IDP (connection) ID for the user's org.

Test & Review
-------------
Manual testing, still running into complications

### DIFF
--- a/internal/pkg/auth/auth_token_handler.go
+++ b/internal/pkg/auth/auth_token_handler.go
@@ -80,20 +80,16 @@ func (a *AuthTokenHandlerImpl) GetCCloudTokens(clientFactory CCloudClientFactory
 }
 
 func (a *AuthTokenHandlerImpl) getCCloudSSOToken(client *ccloudv1.Client, noBrowser bool, email, orgResourceId string) (string, string, error) {
-	var auth0ConnectionName string
-	if !isOkta(client) {
-		userSSO, err := a.getCCloudUserSSO(client, email, orgResourceId)
-		if err != nil {
-			log.CliLogger.Debugf("unable to obtain user SSO info: %v", err)
-			return "", "", errors.Errorf(errors.FailedToObtainedUserSSOErrorMsg, email)
-		}
-		if userSSO == "" {
-			return "", "", errors.Errorf(errors.NonSSOUserErrorMsg, email)
-		}
-		auth0ConnectionName = userSSO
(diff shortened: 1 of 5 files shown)


internal/pkg/auth/auth_token_handler.go
```diff
@@ -165,13 +160,9 @@ func (a *AuthTokenHandlerImpl) checkSSOEmailMatchesLogin(client *ccloudv1.Client
 }
 
 func login(client *ccloudv1.Client, req *ccloudv1.AuthenticateRequest) (*ccloudv1.AuthenticateReply, error) {
-	if isOkta(client) {
+	if sso.IsOkta(client.BaseURL) {
```
DABH, line 163: GetIsSso vs. IsOkta -- naming conventions / consistency, wdyt?

internal/pkg/auth/sso/auth_state.go
```diff
@@ -226,7 +222,7 @@ func (s *authState) getOAuthTokenResponse(payload *strings.Reader) (map[string]a
 	return data, nil
 }
 
-func (s *authState) getAuthorizationCodeUrl(ssoProviderConnectionName string) string {
+func (s *authState) getAuthorizationCodeUrl(ssoProviderConnectionName string, isOkta bool) string {
```
DABH, line 225: We wanted to avoid hard-coding a vendor name when originally writing the SSO code -- can we avoid that here?  What if in the future there are more IDPs we support?

brianstrauch: Due to lots of variability in how IDPs implement the OAuth "standard", the best approach might be to define an interface that each IDP can conform to. I can do this in a follow-up PR!

internal/pkg/auth/auth_token_handler.go
brianstrauch: `GetIsSso()` is autogenerated by the SDK, so we don't have much control over it!

brianstrauch: It's a panic-safe function for the `IsSso` field.

internal/pkg/auth/sso/auth_state.go
DABH: Sounds good, it would be good if you could create a ticket to track that follow-up work so we don't forget about it.  Thanks!

PR #1903, opened 2023-05-10
Add billing Cost CLI

Jayxuchen:
Release Notes
-------------

New Features
- Add `confluent billing cost list`

Checklist
---------
   * yes: ok

What
----

Adds new billing subcommand and cost subcommand. Allows CLI to call the [Cost API](https://confluentinc.atlassian.net/wiki/spaces/CS/pages/2930843726/WIP+1-Pager+Billing+Cost+Public+API).

References
----------

Test & Review
-------------

Local build and test against personal org.

Open Questions / Follow-ups
---------------------------


brianstrauch:
Thanks for the PR @Jayxuchen! Do you think it would make logical sense to move the `confluent admin payment` and `confluent admin promo` commands under `confluent billing` as a breaking change in the next major version? https://docs.confluent.io/confluent-cli/current/command-reference/admin/index.html#confluent-admin

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,107 @@
...
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+type costOut struct {
```
brianstrauch, line 17: There are a lot of fields here... it seems likely that this will cause line wrapping on small terminals which could be a minor annoyance. Could any of these fields be removed or are they all important to the end user?

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,107 @@
...
+	return cmd
+}
+
+func (c *command) list(cmd *cobra.Command, args []string) error {
+	startDate := args[0]
+	endDate := args[1]
+
+	_, err := time.Parse("2006-01-02", startDate)
+	if err != nil {
+		return fmt.Errorf("invalid start date format, expected format should look like: 2022-01-01")
```
brianstrauch, line 55: Can you move this block to a function so we don't need to repeat the error?

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,107 @@
...
+			StartDate:         cost.GetStartDate(),
+			EndDate:           cost.GetEndDate(),
+			Granularity:       cost.GetGranularity(),
+			LineType:          cost.GetLineType(),
+			Product:           cost.GetProduct(),
+			NetworkAccessType: cost.GetNetworkAccessType(),
+			Unit:              cost.GetUnit(),
+		}
+
+		// These fields may be empty depending on the line type, so casting floats as strings as to avoid zero-value
```
brianstrauch, line 80: The downside of doing this is these values will have quotes around them for JSON/YAML output. Can you copy what we did in `confluent price list`? (We have separate `humanOut` and `serializedOut` structs)

internal/pkg/ccloudv2/billing.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return list, nil
+}
+
+func (c *Client) executeListCosts(startDate, endDate, pageToken string) (billing.BillingV1CostList, *http.Response, error) {
+	req := c.BillingClient.CostsBillingV1Api.ListBillingV1Costs(c.billingApiContext()).
+		PageSize(10000).
```
brianstrauch, line 46: Are you sure you want this to be so big?

Jayxuchen: yes, larger page size would be better here since our largest customers have over 10000 line items in a month.

internal/cmd/billing/command_cost_list.go
Jayxuchen: @brianstrauch  is there a way to separate critical fields from verbose fields? I believe all these fields would be important to customers.

internal/cmd/billing/command_cost_list.go
Jayxuchen: For serializedOut if we keep using a `float64`, would the problem still exist where the price is 0? My initial thought was that if the field was a pointer `*float64` and it was nil, it would be either excluded from JSON/YAML or empty in human form, but it seems like using pointer will just show the pointer value.

internal/cmd/billing/command_cost.go
```diff
@@ -0,0 +1,23 @@
...
+	*pcmd.AuthenticatedCLICommand
+}
+
+func (c *commandCost) newCostCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "cost",
+		Short: "Manage Confluent Cloud Billing Costs.",
+	}
+
+	cl := &commandList{c.AuthenticatedCLICommand}
```
brianstrauch, line 19: No need to define a `commandList` struct, we can use `command` for everything within `confluent billing`.

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+type commandList struct {
+	*pcmd.AuthenticatedCLICommand
+}
```
brianstrauch, line 15:
```suggestion
```

internal/cmd/billing/command_cost_list.go
brianstrauch: We don't have anything like that. Typically the most important fields are ordered first. I'll suggest some renamings to make the field names shorter, which should save some space.

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+
+type costOut struct {
+	StartDate           string `human:"Start Date" serialized:"start_date"`
+	EndDate             string `human:"End Date" serialized:"end_date"`
+	Granularity         string `human:"Granularity" serialized:"granularity"`
```
brianstrauch, line 20: Do you mind writing some integration tests? I'd love to see an example of what this output looks like. Take a look at the `test/` directory!

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+	LineType            string `human:"Line Type" serialized:"line_type"`
+	Product             string `human:"Product" serialized:"product"`
+	ResourceId          string `human:"Resource ID" serialized:"resource_id"`
+	ResourceDisplayName string `human:"Resource Display Name" serialized:"resource_display_name"`
```
brianstrauch, line 24:
```suggestion
	ResourceId          string `human:"Resource ID" serialized:"resource_id"`
	ResourceName string `human:"Resource Name" serialized:"resource_name"`
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+	EnvironmentId       string `human:"Environment ID" serialized:"environment_id"`
+	NetworkAccessType   string `human:"Network Access Type" serialized:"network_access_type"`
+	Price               string `human:"Price" serialized:"price"`
+	Unit                string `human:"Unit" serialized:"unit"`
```
brianstrauch, line 28: We can combine these in the human output (should save us a column of space), and keep them separate in the serialized output.

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+	OriginalAmount      string `human:"Original Amount" serialized:"original_amount"`
+	DiscountAmount      string `human:"Discount Amount" serialized:"discount_amount"`
+	Amount              string `human:"Amount" serialized:"amount"`
+}
+
+func (c *commandList) newCostListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "list <start-date> <end-date>",
```
brianstrauch, line 36: I'm thinking start date and end date should be required flags instead of arguments (for clarity).

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+	EnvironmentId       string `human:"Environment ID" serialized:"environment_id"`
+	NetworkAccessType   string `human:"Network Access Type" serialized:"network_access_type"`
+	Price               string `human:"Price" serialized:"price"`
+	Unit                string `human:"Unit" serialized:"unit"`
+	OriginalAmount      string `human:"Original Amount" serialized:"original_amount"`
+	DiscountAmount      string `human:"Discount Amount" serialized:"discount_amount"`
+	Amount              string `human:"Amount" serialized:"amount"`
+}
+
+func (c *commandList) newCostListCommand() *cobra.Command {
```
brianstrauch, line 34: Make sure to add examples!

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	pcmd.AddOutputFlag(cmd)
+
+	return cmd
+}
+
+func (c *commandList) checkDateFormat(date string) error {
+	_, err := time.Parse("2006-01-02", date)
+	if err != nil {
```
brianstrauch, line 51: Doesn't matter too much here, but this syntax will help prevent bugs in the long run by limiting the scope of `err`.
```suggestion
	if _, err := time.Parse("2006-01-02", date); err != nil {
```

internal/cmd/billing/command_cost_list.go
brianstrauch: If the price is 0, is it not visible when it should be?

internal/cmd/billing/command_cost_test.go
```diff
@@ -0,0 +1,102 @@
...
+			NetworkAccessType: billingv1.PtrString("INTERNET"),
+			Price:             billingv1.PtrFloat64(0.123),
+			Unit:              billingv1.PtrString("GB"),
+			Amount:            billingv1.PtrFloat64(50.0),
+			OriginalAmount:    billingv1.PtrFloat64(50.0),
+		},
+	},
+}
+
+func (suite *CostTestSuite) SetupTest() {
```
brianstrauch, line 44: Ah, sorry to have you delete your code, but we're switching to a new (easier to write) form of integration tests, which you can find in the `test/` directory.

internal/cmd/command.go
```diff
@@ -126,6 +127,7 @@ func NewConfluentCommand(cfg *v1.Config) *cobra.Command {
 	cmd.AddCommand(shell.New(cmd, func() *cobra.Command { return NewConfluentCommand(cfg) }))
 	cmd.AddCommand(update.New(prerunner, cfg.Version, updateClient))
 	cmd.AddCommand(version.New(prerunner, cfg.Version))
+	cmd.AddCommand(billing.New(prerunner))
```
brianstrauch, line 130: Please insert in alphabetical order!

internal/pkg/ccloudv2/billing.go
brianstrauch: Oh wow, does it take a while to fetch when there are that many records?

internal/cmd/billing/command_cost_list.go
Jayxuchen: it's not visible for line types that price doesn't make sense for. For example Promo code amounts do not have unit prices, similarly Support costs do not have unit price since the amount is usage depedendent.

internal/pkg/ccloudv2/billing.go
Jayxuchen: We've seen the largest orgs (1 or 2 which are larger than everyone else) may take some time in the seconds, but on average it's still sub-second.
https://ccloud-production.datadoghq.com/apm/resource/cc-billing-service/opencensus/890b3a4ef53050d6?query=%40_top_level%3A1%20env%3Aprod%20service%3Acc-billing-service%20operation_name%3Aopencensus%20resource_name%3A%22GET%20%2Fbilling%2Fv1%2Fcosts%22&env=prod&spanType=service-entry&topGraphs=latency%3Alatency%2CbreakdownAs%3Apercentage%2Cerrors%3Acount%2Chits%3Acount&traces=qson%3A%28data%3A%28%29%2Cversion%3A%210%29&start=1683218550911&end=1683823350911&paused=false

internal/cmd/billing/command_cost_test.go
Jayxuchen: thanks @brianstrauch  i should've read the contribution guide :)

internal/cmd/billing/command_cost_list.go
brianstrauch: Makes sense, I don't think we've ever run in to this use case before. A temporary solution might be to create an `OptionalFloat64` type that overrides the `String()` function... let me think about this today and see if I can come up with a better solution.

internal/pkg/ccloudv2/billing.go
brianstrauch: Ok, this will be highest output-producing command, so we should test that printing 10k+ rows doesn't cause the terminal to stall.

internal/cmd/billing/command_cost_test.go
brianstrauch: No worries! Hopefully the other tests will be significantly easier to write.

internal/pkg/ccloudv2/billing.go
Jayxuchen: @brianstrauch  I can do this in the integration tests.

test/fixtures/output/billing/list.golden
```diff
@@ -0,0 +1,4 @@
+      Start Date      |      End Date       | Granularity |  Line Type  | Product | Resource ID | Resource Display Name | Environment ID | Network Access Type |   Price    | Unit | Original Amount | Discount Amount |   Amount     
+----------------------+---------------------+-------------+-------------+---------+-------------+-----------------------+----------------+---------------------+------------+------+-----------------+-----------------+--------------
+  2023-01-01 00:00:00 | 2023-01-02 00:00:00 | DAILY       | LINE_TYPE_1 | KAFKA   | lkc-123     | kafka_1               | env-123        | INTERNET            | 0.12300000 | GB   |     70.00000000 |     20.00000000 | 50.00000000
```
brianstrauch, line 3: Can we try to format the amounts for the human (non-serialized) output? There should be a helper function inside the `confluent price list` command.

Jayxuchen: I've intentionally kept the floating points to 8 decimal places since these are aggregated by daily amounts. If we truncate the decimal less then we've found that `sum(amounts)` would not match the month-end invoice due to rounding error.

internal/pkg/ccloudv2/billing.go
brianstrauch: I was more interested in seeing if large amounts of data would cause the terminal to freeze or crash (in practice, as opposed to integration tests)

test/fixtures/output/billing/list.golden
Jayxuchen: thanks, i've move the the function a separate util file under `pkg/utils` so both commands can use the same. It also helps us remove the `Unit` column.

internal/pkg/ccloudv2/billing.go
MuweiHe: +1 on this one

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,110 @@
...
+	LineType            string `human:"Line Type" serialized:"line_type"`
+	Product             string `human:"Product" serialized:"product"`
+	ResourceId          string `human:"Resource ID" serialized:"resource_id"`
+	ResourceDisplayName string `human:"Resource Display Name" serialized:"resource_display_name"`
+	EnvironmentId       string `human:"Environment ID" serialized:"environment_id"`
+	NetworkAccessType   string `human:"Network Access Type" serialized:"network_access_type"`
+	Price               string `human:"Price" serialized:"price"`
+	OriginalAmount      string `human:"Original Amount" serialized:"original_amount"`
+	DiscountAmount      string `human:"Discount Amount" serialized:"discount_amount"`
+	Amount              string `human:"Amount" serialized:"amount"`
```
MuweiHe, line 27: i assume none of these need omit-empty?

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,110 @@
...
+			NetworkAccessType: cost.GetNetworkAccessType(),
+		}
+
+		// These fields may be empty depending on the line type, so casting floats as strings as to avoid zero-value
+		if price, ok := cost.GetPriceOk(); ok {
+			out.Price = utils.FormatPrice(*price, cost.GetUnit())
+		}
+
+		if originalAmount, ok := cost.GetOriginalAmountOk(); ok {
+
```
MuweiHe, line 89: extra new line

internal/pkg/utils/bililng_utils.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func FormatPrice(price float64, unit string) string {
+	priceStr := fmt.Sprint(price)
+
+	// Require >= 2 digits after the decimal
+	if strings.Contains(priceStr, ".") {
+		// Extend the remainder if needed
+		r := strings.Split(priceStr, ".")
+		for len(r[1]) < 2 {
```
MuweiHe, line 15: can we add a length check here to avoid potential panic? e.g. `invalid price format`

internal/cmd/billing/command_cost_list.go
brianstrauch: (and they should be `float64`)

test/fixtures/output/billing/list.golden
brianstrauch: Can you move it under `pkg/billing` instead?

internal/pkg/ccloudv2/billing.go
sgagniere: Another thing we might want to consider here is scrollback limits on user terminals. 10,000 lines should be within the default for most users at this point (default should be based on available memory), but it would still be annoying if a user had to change their terminal settings to see this output.

User-side pagination (let user tab through pages of results) is on our radar, and it might alleviate both mine and Brian's concerns.

internal/cmd/billing/command_cost_list.go
Jayxuchen: @brianstrauch  would output support `*float64`? That way we could differentiate between 0 value and empty field.

internal/pkg/utils/bililng_utils.go
Jayxuchen: Looks like go always adds decimal to floats, we have a test where the value is just integer
https://github.com/confluentinc/cli/blob/239787f52228072beffa13388feb54c5f86c28be/internal/cmd/price/command_list_test.go#L10

internal/cmd/billing/command_cost_list.go
```diff
@@ -82,19 +82,18 @@ func (c *command) list(cmd *cobra.Command, args []string) error {
...
 		if price, ok := cost.GetPriceOk(); ok {
-			out.Price = utils.FormatPrice(*price, cost.GetUnit())
+			out.Price = billing.FormatPrice(*price, cost.GetUnit())
 		}
 
 		if originalAmount, ok := cost.GetOriginalAmountOk(); ok {
-
-			out.OriginalAmount = fmt.Sprintf("%.8f", *originalAmount)
+			out.OriginalAmount = *originalAmount
 		}
```
brianstrauch: Is `out.OriginalAmount = cost.GetOriginalAmount()` equivalent?

test/fixtures/output/help/cloud.golden
```diff
@@ -8,6 +8,7 @@ Available Commands:
   api-key         Manage API keys.
   asyncapi        Manage AsyncAPI document tooling.
   audit-log       Manage audit log configuration.
+  billing         Manage Confluent Cloud Billing.
```
brianstrauch, line 11:
```suggestion
  billing         Manage Confluent Cloud billing.
```

test/test-server/billing_handler.go
```diff
@@ -16,9 +16,9 @@ func handleBillingCosts(t *testing.T) http.HandlerFunc {
 		case http.MethodGet:
 			var costList billingv1.BillingV1CostList
 
-			// test over 10000 entries to ensure command does not stall terminal
+			// test 20
 			var data []billingv1.BillingV1Cost
-			for i := 0; i < 10100; i++ {
+			for i := 0; i < 20; i++ {
```
brianstrauch, line 21: Do we need all 20 iterations, or is a smaller number like 2 enough?

internal/cmd/billing/command_cost_list.go
Jayxuchen: @brianstrauch i noticed this issue, it prints out the pointer value instead of the actual value:
image

internal/cmd/billing/command_cost_list.go
Jayxuchen: ah yes i originally did this hoping that i could use pointer values to determine nil vs non-nil value but I dont think that's supported as i mentioned [here](https://github.com/confluentinc/cli/pull/1903#discussion_r1218794121)

internal/cmd/billing/command_cost_list.go
Jayxuchen: I dont think the `omitempty` works either for floats, as i still see 0 in the test.

go.mod
```diff
@@ -104,6 +104,7 @@ require (
 	github.com/bufbuild/protocompile v0.4.0 // indirect
 	github.com/c-bata/go-prompt v0.2.6 // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
+	github.com/confluentinc/ccloud-sdk-go-v2/billing v0.3.0
```
MuweiHe, line 107: did you run `go mod tidy`? This should be grouped together with other packages in ccloud-sdk-go-v2

internal/cmd/billing/command.go
```diff
@@ -0,0 +1,23 @@
...
+
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "billing",
+		Short:       "Manage Confluent Cloud billing.",
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
```
brianstrauch, line 17: We haven't implemented API key support for endpoints like yours yet. Will do soon!
```suggestion
		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonApiKeyCloudLogin},
```

internal/cmd/billing/command.go
```diff
@@ -0,0 +1,23 @@
...
+	}
```
brianstrauch, line 18:
```suggestion
	}
	
```

internal/cmd/billing/command.go
```diff
@@ -0,0 +1,23 @@
...
+	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+	c.Command.AddCommand(c.newCostCommand())
+
+	return c.Command
```
brianstrauch, line 22:
```suggestion
	cmd.AddCommand(c.newCostCommand())

	return cmd
```

internal/cmd/billing/command_cost_list.go
brianstrauch: Thoughts on this @Jayxuchen? IMO, B is a bit more clear than A:

A:
```
confluent billing cost list 01-02-2003 01-03-2002
```

B:
```
confluent billing cost list --start-date 01-02-2003 --end-date 01-03-2002
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	pcmd.AddOutputFlag(cmd)
+
+	return cmd
+}
+
+func (c *command) checkDateFormat(date string) error {
+	if _, err := time.Parse("2006-01-02", date); err != nil {
+		return fmt.Errorf("expected format should look like: 2022-01-01")
```
brianstrauch, line 47:
```suggestion
		return fmt.Errorf("must be formatted as: YYYY-MM-DD")
```

brianstrauch: It's unclear if month or day goes first if they're both "01"!

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return nil
+}
+
+func (c *command) list(cmd *cobra.Command, args []string) error {
+	startDate := args[0]
+	endDate := args[1]
+
+	err := c.checkDateFormat(startDate)
+	if err != nil {
```
brianstrauch, line 58:
```suggestion
	if err := c.checkDateFormat(startDate); err != nil {
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("invalid start date: %s", err.Error())
```
brianstrauch, line 59:
```suggestion
		return fmt.Errorf("invalid start date: %v", err)
```

internal/pkg/billing/billing_utils_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFormatPrice(t *testing.T) {
+	assert.Equal(t, "$1.00 USD/GB", FormatPrice(1, "GB"))
+	assert.Equal(t, "$1.20 USD/GB", FormatPrice(1.2, "GB"))
+	assert.Equal(t, "$1.23 USD/GB", FormatPrice(1.23, "GB"))
+	assert.Equal(t, "$1.23456 USD/GB", FormatPrice(1.23456, "GB"))
```
brianstrauch, line 13: We really only need one of these assertions, since these cases are covered in the test underneath now!

internal/pkg/ccloudv2/billing.go
```diff
@@ -0,0 +1,55 @@
...
+		StartDate(startDate).
+		EndDate(endDate)
```
brianstrauch, line 50:
```suggestion
	req := c.BillingClient.CostsBillingV1Api.ListBillingV1Costs(c.billingApiContext()).PageSize(10000).StartDate(startDate).EndDate(endDate)
```

internal/pkg/ccloudv2/billing.go
```diff
@@ -0,0 +1,55 @@
...
+	if pageToken != "" {
+		req = req.PageToken(pageToken)
+	}
+	return c.BillingClient.CostsBillingV1Api.ListBillingV1CostsExecute(req)
```
brianstrauch, line 54:
```suggestion
	return req.Execute()
```

test/fixtures/output/billing/list.golden
```diff
@@ -0,0 +1,5 @@
+      Start Date      |      End Date       | Granularity |  Line Type  |     Product     | Resource ID | Resource Display Name | Environment ID | Network Access Type |     Price     | Original Amount | Discount Amount | Amount  
+----------------------+---------------------+-------------+-------------+-----------------+-------------+-----------------------+----------------+---------------------+---------------+-----------------+-----------------+---------
+  2023-01-01 00:00:00 | 2023-01-02 00:00:00 | DAILY       | LINE_TYPE_1 | KAFKA           | lkc-0       | kafka_1               | env-123        | INTERNET            | $0.123 USD/GB | $70.00          | $20.00          | $50.00
```
brianstrauch, line 3: Do we need to print the start time here too? It looks like it's always 00:00:00... will that ever change?

test/fixtures/output/billing/list_invalid_format.golden
```diff
@@ -0,0 +1,15 @@
+Error: accepts 2 arg(s), received 1
```
brianstrauch, line 1: This looks more like a test for a missing argument than an invalid format as its name suggests.

test/fixtures/output/billing/list_missing_arg.golden
```diff
@@ -0,0 +1,15 @@
+Error: accepts 2 arg(s), received 1
```
brianstrauch, line 1: We typically don't include tests like this one, up to you though!

test/test-server/billing_handler.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// Handler for: "/billing/v1/costs"
+func handleBillingCosts(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case http.MethodGet:
+			var costList billingv1.BillingV1CostList
+
+			// test 20
```
brianstrauch, line 21:
```suggestion
```

test/test-server/billing_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stretchr/testify/require"
+
+	billingv1 "github.com/confluentinc/ccloud-sdk-go-v2/billing/v1"
+)
+
+// Handler for: "/billing/v1/costs"
+func handleBillingCosts(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case http.MethodGet:
```
brianstrauch, line 18: Looks like there's only one case in this `switch`, so we can probably just omit it entirely.

go.mod
Jayxuchen: oh i just noticed there's two separate require sections, which is why go mod tidy didn't re-order

Jayxuchen:
@brianstrauch updated to use `confluent billing cost list --start-date 01-02-2003 --end-date 01-03-2002`

test/fixtures/output/billing/list.golden
Jayxuchen: yes, so if we query for multiple days of costs, it will aggregate based on the granularity (currently only daily). So one row will have `2023-01-01 00:00:00 - 2023-01-02 00:00:00` and the next row could have `2023-01-02 00:00:00 - 2023-01-03 00:00:00`

test/fixtures/output/billing/list_invalid_format.golden
Jayxuchen: my mistake right, updating the test to actually test format error.

test/fixtures/output/billing/list.golden
brianstrauch: So granularity is going to be able to be changed in the future? Can we remove the timestamp if the granularity is a multiple of a day?

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,119 @@
...
+	Price               string `human:"Price,omitempty" serialized:"price,omitempty"`
+	OriginalAmount      string `human:"Original Amount" serialized:"original_amount"`
+	DiscountAmount      string `human:"Discount Amount,omitempty" serialized:"discount_amount,omitempty"`
+	Amount              string `human:"Amount,omitempty" serialized:"amount,omitempty"`
+}
+
+func (c *command) newCostListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "list --start-date <start-date> --end-date <end-date>",
+		Example: "list  --start-date 2023-01-01  --end-date 2023-01-10",
```
brianstrauch, line 33: Can you use `examples.BuildExampleString()` and define this at the bottom of the struct? Take a look at other commands for examples!

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,119 @@
...
+		Short:   "List Confluent Cloud billing costs.",
+		Long:    "List Confluent Cloud daily aggregated costs for a specific range of dates.",
+		Args:    cobra.NoArgs,
+		RunE:    c.list,
+	}
+
+	cmd.Flags().String("start-date", "", "Start Date.")
+	cmd.Flags().String("end-date", "", "End Date.")
+
```
brianstrauch, line 42:
```suggestion
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,119 @@
...
+	pcmd.AddOutputFlag(cmd)
+	cobra.CheckErr(cmd.MarkFlagRequired("start-date"))
+	cobra.CheckErr(cmd.MarkFlagRequired("end-date"))
```
brianstrauch, line 45:
```suggestion

	cobra.CheckErr(cmd.MarkFlagRequired("start-date"))
	cobra.CheckErr(cmd.MarkFlagRequired("end-date"))
	
```

brianstrauch (changes requested):

test/fixtures/output/billing/list.golden
Jayxuchen: yeah i think that's a good idea

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,129 @@
...
+	Price               string `human:"Price,omitempty" serialized:"price,omitempty"`
+	OriginalAmount      string `human:"Original Amount" serialized:"original_amount"`
+	DiscountAmount      string `human:"Discount Amount,omitempty" serialized:"discount_amount,omitempty"`
+	Amount              string `human:"Amount,omitempty" serialized:"amount,omitempty"`
+}
+
+func (c *command) newCostListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "list --start-date <start-date> --end-date <end-date>",
+		//Example: "list  --start-date 2023-01-01  --end-date 2023-01-10",
```
brianstrauch, line 35:
```suggestion
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,129 @@
...
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `List Billing costs from 2023-01-01 to 2023-01-10:`,
+				Code: "confluent billing list --start-date 2023-01-01 --end-date 2023-01-10",
+			}),
```
brianstrauch, line 40: Can you move this to the bottom of the struct and follow the same format as other commands?

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+
+	endDate, err := cmd.Flags().GetString("end-date")
+	if err != nil {
+		return err
+	}
+
+	err = c.checkDateFormat(startDate)
+	if err != nil {
```
brianstrauch, line 77:
```suggestion
	if err := c.checkDateFormat(startDate); err != nil {
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -0,0 +1,129 @@
...
+	list := output.NewList(cmd)
+	for _, cost := range costs {
+		out := &costOut{
+			Granularity:       cost.GetGranularity(),
+			LineType:          cost.GetLineType(),
+			Product:           cost.GetProduct(),
+			NetworkAccessType: cost.GetNetworkAccessType(),
+		}
+		if cost.GetGranularity() == "DAILY" {
+			out.StartDate = strings.Split(cost.GetStartDate(), " ")[0]
```
brianstrauch, line 100: Since we know what the suffix will look like:
```suggestion
			out.StartDate = strings.TrimSuffix(cost.GetStartDate(), " 00:00:00")
```

test/billing_test.go
```diff
@@ -0,0 +1,16 @@
+package test
+
+func (s *CLITestSuite) TestCostList() {
+	tests := []CLITest{
+		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02", fixture: "billing/list.golden"},
+		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o json", fixture: "billing/list_json.golden"},
```
brianstrauch, line 6: Match the style of our other tests!
```suggestion
		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o json", fixture: "billing/cost/list-json.golden"},
```

test/fixtures/output/billing/list_missing_arg.golden
```diff
@@ -0,0 +1,19 @@
+Error: required flag(s) "end-date" not set
```
brianstrauch, line 1: Rename to "missing flag"?

test/test-server/billing_handler.go
```diff
@@ -0,0 +1,61 @@
...
+
+		for i := 0; i < 2; i++ {
+			item := billingv1.BillingV1Cost{
+				StartDate:   billingv1.PtrString("2023-01-01 00:00:00"),
+				EndDate:     billingv1.PtrString("2023-01-02 00:00:00"),
+				Granularity: billingv1.PtrString("DAILY"),
+				LineType:    billingv1.PtrString("LINE_TYPE_1"),
+				Product:     billingv1.PtrString("KAFKA"),
+				Resource: &billingv1.BillingV1Resource{
+					Id:          billingv1.PtrString(fmt.Sprintf("lkc-%d", i)),
```
brianstrauch, line 28:
```suggestion
					Id:          billingv1.PtrString(fmt.Sprintf("lkc-0000%d", i)),
```

internal/cmd/billing/command_cost_list.go
```diff
@@ -31,17 +31,16 @@ type costOut struct {
...
+		Use:   "list --start-date <start-date> --end-date <end-date>",
+		Short: "List Confluent Cloud billing costs.",
+		Long:  "List Confluent Cloud daily aggregated costs for a specific range of dates.",
+		Args:  cobra.NoArgs,
+		RunE:  c.list,
 		Example: examples.BuildExampleString(
 			examples.Example{
 				Text: `List Billing costs from 2023-01-01 to 2023-01-10:`,
 				Code: "confluent billing list --start-date 2023-01-01 --end-date 2023-01-10",
 			}),
```
brianstrauch, line 43:
```suggestion
			},
		),
```

test/billing_test.go
```diff
@@ -2,11 +2,11 @@ package test
...
-		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o json", fixture: "billing/list_json.golden"},
-		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o yaml", fixture: "billing/list_yaml.golden"},
-		{args: "billing cost list --start-date 2023-01-01", fixture: "billing/list_missing_arg.golden", exitCode: 1},
-		{args: "billing cost list --start-date 01-02-2023 --end-date 01-01-2023", fixture: "billing/list_invalid_format.golden", exitCode: 1},
+		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02", fixture: "billing/cost/list.golden"},
+		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o json", fixture: "billing/cost/list_json.golden"},
+		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o yaml", fixture: "billing/cost/list_yaml.golden"},
+		{args: "billing cost list --start-date 2023-01-01", fixture: "billing/cost/list_missing_flag.golden", exitCode: 1},
+		{args: "billing cost list --start-date 01-02-2023 --end-date 01-01-2023", fixture: "billing/cost/list_invalid_format.golden", exitCode: 1},
```
brianstrauch, line 9:
```suggestion
		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o json", fixture: "billing/cost/list-json.golden"},
		{args: "billing cost list --start-date 2023-01-01 --end-date 2023-01-02 -o yaml", fixture: "billing/cost/list-yaml.golden"},
		{args: "billing cost list --start-date 2023-01-01", fixture: "billing/cost/list-missing-flag.golden", exitCode: 1},
		{args: "billing cost list --start-date 01-02-2023 --end-date 01-01-2023", fixture: "billing/cost/list-invalid-format.golden", exitCode: 1},
```

brianstrauch (approved):

PR #1909, opened 2023-05-12
Fix release process

brianstrauch:
What
----
* Now that the first two steps of the release have moved to Semaphore, the tag-release target needs to refer to the correct repository.
* Fix Docker build/manifest
* Add release notes in `make publish-docs` step

Test & Review
-------------
Tested during the v3.13.0 release


sgagniere (approved):
lgtm. Publishing the dockerfile for arm64 is new, right? Neat.

mk-files/semver.mk
```diff
@@ -19,10 +19,14 @@ split_version := $(subst .,$(_space),$(CLEAN_VERSION))
...
-	git tag -d v$$(cat release-notes/version.txt) || true
-	$(call dry-run,git push -d origin v$$(cat release-notes/version.txt)) || true
+	$(eval DIR=$(shell mktemp -d))
+	$(eval CLI_RELEASE=$(DIR)/cli-release)
+	
+	git clone git@github.com:confluentinc/cli-release.git $(CLI_RELEASE) && \
+	version=$$(ls $(CLI_RELEASE)/release-notes | sed -e s/.json$$// | sort --version-sort | tail -1) && \
+	git tag -d v$${version} || true && \
+	$(call dry-run,git push -d origin v$${version}) || true && \
+	git tag v$${version} && \
```
MuweiHe, line 29: double checking here that the `version` doesn't have `v` in it?

PR #1910, opened 2023-05-12
[RCCA-12451] Delay decrypting context from Load() to prerunner

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix decryption error when running non-authenticated commands

Checklist
---------
   * yes: ok

What
----

customer reported getting decryption (cipher) error when running `confluent update`, which is not an authenticated command. Loading config is being called at the very beginning of executing a command, and decrypting tokens there is unnecessary for anonymous commands like `update`. Delayed into prerunners to avoid potential issues.

References
----------

https://confluentinc.atlassian.net/browse/RCCA-12541

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/pkg/config/v1/mock.go
```diff
@@ -114,7 +114,7 @@ func APICredentialConfigMock() *Config {
 
 func UnauthenticatedCloudConfigMock() *Config {
 	c := AuthenticatedCloudConfigMock()
-	c.Contexts = nil
+	c.CurrentContext = ""
```
MuweiHe, line 117: logging out doesn't delete all contexts. setting it to nil will cause contexts mapping error.

internal/pkg/cmd/prerunner.go
```diff
@@ -233,6 +232,9 @@ func IsFlagRequired(flag *pflag.Flag) bool {
 // Authenticated provides PreRun operations for commands that require a logged-in Confluent Cloud user.
 func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		if err := r.Config.DecryptContextStates(); err != nil {
```
MuweiHe, line 235: added this step for other prerunner operations except anonymous ones.

internal/pkg/config/v1/config.go
```diff
@@ -133,6 +133,24 @@ func New() *Config {
 	}
 }
 
+func (c *Config) DecryptContextStates() error {
+	for _, context := range c.Contexts {
```
brianstrauch, line 137: Don't we only want to decrypt the tokens for the current state? There should be no need to decrypt tokens for a state we're not using.

internal/pkg/config/v1/config.go
```diff
@@ -259,14 +267,18 @@ func (c *Config) encryptContextStateTokens(tempAuthToken, tempAuthRefreshToken s
 		c.Context().GetState().Salt = salt
 		c.Context().GetState().Nonce = nonce
 	}
-	if tempAuthToken != "" {
+
+	reg1 := regexp.MustCompile(authTokenRegex)
+	if tempAuthToken != "" && reg1.MatchString(tempAuthToken) {
```
MuweiHe, line 272: avoid encrypting already encrypted tokens

internal/pkg/config/v1/config.go
```diff
@@ -134,7 +135,8 @@ func New() *Config {
 }
 
 func (c *Config) DecryptContextStates() error {
-	for _, context := range c.Contexts {
+	if c.CurrentContext != "" {
+		context := c.Contexts[c.CurrentContext]
```
brianstrauch, line 139: This should be equivalent!
```suggestion
		context := c.Context()
```

internal/pkg/config/v1/config.go
MuweiHe: you're right. Changed to only decrypt for current context!

internal/pkg/cmd/prerunner.go
```diff
@@ -289,6 +291,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 
 func (r *PreRun) ParseFlagsIntoContext(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		if err := r.Config.DecryptContextStates(); err != nil {
```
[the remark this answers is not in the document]
MuweiHe: good catch. I'll remove those redundant calls!

internal/pkg/cmd/prerunner.go
brianstrauch: Does this have to go before `Anonymous()`? It's a bit easier to reason about things when `Anonymous()` is called first.

internal/pkg/config/v1/config.go
```diff
@@ -133,6 +134,25 @@ func New() *Config {
 	}
 }
 
+func (c *Config) DecryptContextStates() error {
+	if c.CurrentContext != "" {
+		context := c.Context()
```
brianstrauch, line 139: This might be preferred?
```suggestion
	if context := c.Context(); context != nil {
```

internal/pkg/config/v1/config.go
```diff
@@ -133,6 +134,25 @@ func New() *Config {
...
+		state := c.ContextStates[context.Name]
```
brianstrauch, line 140: We can't use `context.State` here?

internal/pkg/config/v1/config.go
brianstrauch, line 272: If we can't come up with a better name than `reg1`, let's just do this:
```suggestion
	if tempAuthToken != "" && regexp.MustCompile(authTokenRegex).MatchString(tempAuthToken) {
```

internal/pkg/config/v1/config.go
MuweiHe: This was copied from the original code:
`context.State = c.ContextStates[context.Name]` so i think it's different

internal/pkg/config/v1/config.go
```diff
@@ -259,14 +266,16 @@ func (c *Config) encryptContextStateTokens(tempAuthToken, tempAuthRefreshToken s
...
+	if tempAuthToken != "" && regexp.MustCompile(authTokenRegex).MatchString(tempAuthToken) {
 		encryptedAuthToken, err := secret.Encrypt(c.Context().Name, tempAuthToken, c.Context().GetState().Salt, c.Context().GetState().Nonce)
 		if err != nil {
 			return err
 		}
 		c.Context().GetState().AuthToken = encryptedAuthToken
 	}
-	if tempAuthRefreshToken != "" {
+
+	if tempAuthRefreshToken != "" && regexp.MustCompile(authRefreshTokenRegex).MatchString(tempAuthRefreshToken) {
```
brianstrauch, line 278: This regular expression does not match the empty string, so this is redundant.
```suggestion
	if regexp.MustCompile(authRefreshTokenRegex).MatchString(tempAuthRefreshToken) {
```

internal/pkg/config/v1/config.go
```diff
@@ -259,14 +266,16 @@ func (c *Config) encryptContextStateTokens(tempAuthToken, tempAuthRefreshToken s
 		c.Context().GetState().Salt = salt
 		c.Context().GetState().Nonce = nonce
 	}
-	if tempAuthToken != "" {
+
+	if tempAuthToken != "" && regexp.MustCompile(authTokenRegex).MatchString(tempAuthToken) {
```
brianstrauch, line 270:
```suggestion
	if regexp.MustCompile(authTokenRegex).MatchString(tempAuthToken) {
```

brianstrauch (approved):

PR #1916, opened 2023-05-16
Release to Homebrew tap

brianstrauch:
Release Notes
-------------
New Features
- Publish with Homebrew

What
----
* Release duplicate versions of darwin and linux binaries with updates permanently disabled
* Upload these archived binaries to a new subdirectory in S3 called `homebrew`
* Open a PR in Homebrew tap with formula cli.rb
* Install command will look like `brew install confluentinc/tap/cli`


.goreleaser.yml
```diff
@@ -63,6 +63,26 @@ builds:
...
+      - all=-trimpath={{.Env.HOME}}/git
+    asmflags:
+      - all=-trimpath={{.Env.HOME}}/git
+    goos:
+      - darwin
+    goarch:
+      - amd64
+    hooks:
+      post:
+        - cmd: make download-licenses
```
sgagniere, line 83:
```suggestion
```

`confluent-darwin-amd64` post hook already runs this.

cmd/confluent/main.go
```diff
@@ -14,19 +14,24 @@ import (
...
+	date           = ""
+	disableUpdates = "false"
+	isTest         = "false"
 )
 
 func main() {
 	cfg, err := load.LoadAndMigrate(v1.New())
 	cobra.CheckErr(err)
 
+	disableUpdates, err := strconv.ParseBool(disableUpdates)
```
sgagniere, line 28: Just to double check my reading of these changes: users can't configure disable updates in their configs; updating is always enabled for archive/binary downloads and always disabled for package manager installs to not interfere with the package manager update feature.

.goreleaser.yml
brianstrauch: Good catch, forgot we only need to run this once.

cmd/confluent/main.go
brianstrauch: Yeah, on second thought, we should remove this from the configuration file altogether (instead of omitempty, use `json:"-"`)

brianstrauch: If a Homebrew user set `"updates_disabled": false`, it would reset to true the next time they ran any command, which could be disorienting.

brianstrauch: Never mind, that would be a breaking change.

.goreleaser.yml
```diff
@@ -220,4 +262,53 @@ docker_manifests:
...
       - "confluentinc/confluent-cli:latest-amd64"
       - "confluentinc/confluent-cli:latest-arm64"
-    skip_push: "{{ .Env.DRY_RUN }}"
\ No newline at end of file
+    skip_push: "{{ .Env.DRY_RUN }}"
+
+release:
+  ids:
+    - archive
+    - archive-alpine
```
MuweiHe, line 270: except alpine, all other linux builds are in archive now 💛

brianstrauch: Yep! I tried to get alpine in there too, but it looks like you can't have a duplicate linux/amd64 or linux/arm64.

scripts/gon_filepath_editor.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
```
sgagniere, line 1: Instead of deleting this file and hardcoding 4 source paths in 4 gon files, imo we should just make a second script that handles the homebrew gon files to fix the issue that @MuweiHe pointed out.

Something like this:
```
#!/bin/bash

# run sed inside this script to get around difficulties in running it directly inside a post hook
sed "s|path/to/file|$1|" scripts/gon_confluent_$2.hcl > dist/gon_confluent_$2_homebrew.hcl
```

The reason is because our build pipeline already broke once with hardcoded paths after updating goreleaser; their docs say that there is no guarantee that the paths will stay the same after updates.

sgagniere (changes requested):

sgagniere (approved):
+1

Remember to check for the (now 4) emails confirming mac signing, since this modifies the workflow slightly!

brianstrauch:
@sgagniere I added `output: true` so now we get visual confirmation in the terminal that signing was successful!

PR #1917, opened 2023-05-17
Clean up GoReleaser build script and release Dockerfiles

brianstrauch:
What
----
* Accept defaults during GoReleaser build for easier debugging
* Remove "glibc" from build script / dockerfile names since it's the default- this better matches the naming conventions in GoReleaser and is ultimately less confusing (up for debate if you disagree!)
* Fix `IsReleased()`

Test & Review
-------------
Tests still pass, `make build` and `make cross-build` work as intended.


Makefile
```diff
@@ -1,33 +1,35 @@
 GORELEASER_VERSION := v1.17.2
 
-.PHONY: build # compile natively based on the system
+# Compile natively based on the current system
+.PHONY: build 
 build:
-ifneq "" "$(findstring NT,$(shell uname))" # build for Windows
+ifneq "" "$(findstring NT,$(shell uname))" # windows
```
sgagniere, line 6: As a side note: this line actually doesn't work at all on Windows, since uname isn't a valid command (at least in current Windows 10 Home)... we should fix this at some point 😅

internal/pkg/version/version.go
```diff
@@ -37,7 +37,7 @@ func NewVersion(version, commit, buildDate string) *Version {
 }
 
 func (v *Version) IsReleased() bool {
-	return v.Version != "0.0.0" && !strings.Contains(v.Version, "dirty") && !strings.Contains(v.Version, "-g")
+	return v.Version != "0.0.0" && !strings.Contains(v.Version, "dirty") && !strings.Contains(v.Version, "-g") && !strings.Contains(v.Version, "SNAPSHOT")
```
sgagniere, line 40: 👍

sgagniere (approved):

Issue #1896, opened 2023-05-07
Error: update client failure: error updating CLI binary


conradwt:
I'm seeing the following error when trying to update confluent locally on Apple M1 using macOS 13.3.1 (a) (Ventura):

```zsh
➜ confluent update
Checking for updates...
New version of confluent is available
Current Version: v3.11.0
Latest Version:  v3.12.0
[5/4/2023] Confluent CLI v3.12.0 Release Notes
==============================================

New Features
------------
  - Add `--topics` flag to `confluent asyncapi export` to only export specified topics



Do you want to download and install this update? (y/n): y
Downloading confluent version 3.12.0...
Done. Downloaded 43.31 MB in 1 seconds. (32.41 MB/s)
Error: update client failure: error updating CLI binary: unable to copy /var/folders/0_/nkh8mcg92pq1fft8mrzcg98w0000gn/T/confluent3475877262/confluent-v3.12.0-darwin-arm64 to /usr/local/bin/confluent: remove /usr/local/bin/confluent: permission denied

Suggestions:
    Please submit a support ticket.
    In the meantime, see link for other ways to download the latest CLI version:
    https://docs.confluent.io/current/cli/installing.html
```

However, I'm not seeing the issue Intel macOS 11.7.6 (Big Sur):

```zsh
➜ confluent update
Checking for updates...
New version of confluent is available
Current Version: v3.11.0
Latest Version:  v3.12.0
[5/4/2023] Confluent CLI v3.12.0 Release Notes
==============================================

New Features
------------
  - Add `--topics` flag to `confluent asyncapi export` to only export specified topics



Do you want to download and install this update? (y/n): y
Downloading confluent version 3.12.0...
Done. Downloaded 45.67 MB in 3 seconds. (14.25 MB/s)
```

In short, one needs do `sudo confluent update` instead `confluent update`.  Thus, I believe that the `confluent` binary executable should invoke `sudo` in such situations.

brianstrauch:
Hi @conradwt, I haven't been able to replicate this even though I'm also using macOS 13.3.1 (a). Do you mind showing me the output of `ls -l /usr/local/bin/confluent`?

MuweiHe:
@conradwt we added a fix a while ago. can you confirm if you still see the issue? The latest cli version is v3.25.0

conradwt:
@MuweiHe I'm still seeing the following issue:

```zsh
➜ confluent update                    
Checking for updates...
New version of confluent is available
[...]

PR #1918, opened 2023-05-17
Use go-update package

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Tried to replicate https://github.com/confluentinc/cli/issues/1896 but couldn't... I'm hoping that switching to this package will fix the issue since they use a different strategy, but even if it doesn't this is a good thing to do.

Test & Review
-------------
Unit tests, manually tested


internal/pkg/update/client.go
```diff
@@ -246,62 +206,54 @@ func (c *client) UpdateBinary(cliName, version, path string, noVerify bool) erro
...
+		name += ".exe"
 	}
+	return name
+}
 
-	return nil
+func findChecksum(content, binary string) (string, error) {
+	for _, line := range strings.Split(content, "\n") {
+		x := strings.Split(line, "  ")
+		checksum, name := x[0], x[1]
```
MuweiHe, line 251: a length check to avoid panic?

internal/pkg/update/client.go
```diff
@@ -246,62 +206,54 @@ func (c *client) UpdateBinary(cliName, version, path string, noVerify bool) erro
...
 	if err != nil {
 		return errors.Wrapf(err, errors.DownloadVersionErrorMsg, cliName, version, downloadDir)
 	}
 
-	mb := float64(bytes) / 1024.0 / 1024.0
+	mb := float64(len(payload)) / 1024.0 / 1024.0
 	timeSpent := c.clock.Now().Sub(startTime).Seconds()
 	fmt.Fprintf(c.Out, "Done. Downloaded %.2f MB in %.0f seconds. (%.2f MB/s)\n", mb, timeSpent, mb/timeSpent)
 
+	opts := update.Options{}
```
MuweiHe, line 218: put it after a few lines to where it's initialized?

internal/pkg/update/client.go
```diff
@@ -246,62 +206,54 @@ func (c *client) UpdateBinary(cliName, version, path string, noVerify bool) erro
...
-			}
-			return errors.Wrapf(err, errors.CopyErrorMsg, newBin, newPath)
+			return err
 		}
-	} else {
-		err = c.copyFile(newBin, newPath)
-		if err != nil {
-			return errors.Wrapf(err, errors.CopyErrorMsg, newBin, newPath)
+
+		opts.Checksum = make([]byte, len(checksum)/2)
```
MuweiHe, line 231: continue: declare here

internal/pkg/update/repository.go
```diff
@@ -13,7 +13,7 @@ type Repository interface {
 	GetAvailableReleaseNotesVersions(name string) (version.Collection, error)
 	// Downloads the versioned package to download dir to downloadDir.
 	// Returns the full path to the downloaded package, the download size in bytes, or an error if one occurred.
-	DownloadVersion(name, version, downloadDir string) (downloadPath string, downloadedBytes int64, err error)
```
MuweiHe, line 16: is this assuming that path will always be `/bin`?

brianstrauch: The new package updates the binary in-place, so it will use whatever directory the binary is currently in.

internal/pkg/update/client.go
brianstrauch: What do you mean?

internal/pkg/update/client.go
brianstrauch: Good catch!

MuweiHe (approved):
lgtm once ci addressed

PR #1920, opened 2023-05-19
[CLI-2496] Resolve issue causing users to log into the wrong SSO org

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Prevent `confluent login --organization-id` from logging users into the wrong SSO-enabled organization

Checklist
---------
   * yes: ok

What
----
After logging in with SSO and obtaining the idp id, the organization resource id was not being sent in the final login request. Without the organization resource id, the backend interprets this as a request to log into the default organization. If that organization also uses the same idp, that request will succeed (*single* sign on), and inadvertently ignore the `organization-id` flag.

References
----------

Test & Review
-------------
Manually tested w/ @ivanhyf 's help (successfully logged into the correct organization using this fix).

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/auth/auth_token_handler.go
+++ b/internal/pkg/auth/auth_token_handler.go
@@ -94,7 +94,10 @@ func (a *AuthTokenHandlerImpl) getCCloudSSOToken(client *ccloudv1.Client, noBrow
 		return "", "", err
 	}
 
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
Do we have integration tests for this?

PR #1925, opened 2023-05-22
DGS-7258 Async API Ability to only export topics with prefix

kcorman0:
Release Notes
-------------
New Features
- Support topic prefixes ending with a wildcard (*) in `confluent asyncapi export --topics`

Checklist
---------
   * yes

What
----

If `--topics` specified, split into prefixes and regular topics. Then check that a topic is either present in the list of specified topics or has a prefix given by the user, otherwise don't export it.

For added tests, we only care about the CLI output (are topics correct?) so reusing files. File check is not useful here

Test & Review
-------------

Tested on prod confluent cloud, create topics and schemas (with TopicNamingStrategy) and export a subset of them with `--topics topic1,topic2,prefix*...`

### DIFF
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -20,6 +20,7 @@ import (
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
 	v1 "github.com/confluentinc/cli/internal/pkg/config/v1"
 	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/examples"
 	"github.com/confluentinc/cli/internal/pkg/kafkarest"
 	"github.com/confluentinc/cli/internal/pkg/log"
 	"github.com/confluentinc/cli/internal/pkg/output"
@@ -64,6 +65,12 @@ func newExportCommand(prerunner pcmd.PreRunner) *cobra.Command {
 		Short: "Export an AsyncAPI specification.",
 		Long:  "Export an AsyncAPI specification for a Kafka cluster and Schema Registry.",
 		Args:  cobra.NoArgs,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Export an AsyncAPI specification with topic "my-topic" and all topics starting with "prefix-".`,
+				Code: `confluent asyncapi export --topics "my-topic,prefix-*"`,
+			},
+		),
 	}
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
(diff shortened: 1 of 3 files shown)


internal/cmd/asyncapi/command_export.go
```diff
@@ -108,12 +108,22 @@ func (c *command) export(cmd *cobra.Command, _ []string) error {
 	}
 	channelCount := 0
 
-	topicsSpecified := types.NewSet(flags.topics...)
+	topicsSpecified := types.NewSet()
+	var prefixesSpecified []string
+	// Split user-specified topics into topics/topic prefixes
+	for _, userTopic := range flags.topics {
+		if userTopic[len(userTopic)-1:] == "*" {
```
MuweiHe, line 115: can we add a length check here to avoid panic? in case user input "" by mistake

internal/cmd/asyncapi/command_export.go
```diff
@@ -78,7 +78,7 @@ func newExportCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd.Flags().String("schema-registry-api-key", "", "API key for Schema Registry.")
 	cmd.Flags().String("schema-registry-api-secret", "", "API secret for Schema Registry.")
 	cmd.Flags().String("schema-context", "default", "Use a specific schema context.")
-	cmd.Flags().StringSlice("topics", nil, "A comma-separated list of topics to export.")
+	cmd.Flags().StringSlice("topics", nil, "A comma-separated list of topics to export. Supports prefixes ending in wildcard (*).")
```
brianstrauch, line 81:
```suggestion
	cmd.Flags().StringSlice("topics", nil, "A comma-separated list of topics to export. Supports prefixes ending with a wildcard (*).")
```

internal/cmd/asyncapi/command_export.go
brianstrauch: https://pkg.go.dev/strings#HasSuffix has this check built-in!

internal/cmd/asyncapi/command_export.go
```diff
@@ -575,3 +588,21 @@ func createConsumer(broker string, clusterCreds *v1.APIKeyPair, groupId string)
 	}
 	return consumer, nil
 }
+
+// Check if topic matches user-specified topics/prefixes. True if user didn't specify
+func topicMatch(topic string, topics types.Set, prefixes []string) bool {
```
brianstrauch, line 593: We can keep one set/slice for all of our topics / prefixed topics, and check for the trailing "*" in this function.

brianstrauch (changes requested):

brianstrauch (approved):
Unit tests for `topicMatch()` would be appreciated!

internal/cmd/asyncapi/command_export.go
```diff
@@ -575,3 +573,17 @@ func createConsumer(broker string, clusterCreds *v1.APIKeyPair, groupId string)
...
 }
+
+// Check if topic matches user-specified topics/prefixes. True if user didn't specify
+func topicMatch(topic string, userTopics []string) bool {
+	if len(userTopics) == 0 {
+		return true
+	}
+
+	for _, userTopic := range userTopics {
+		if strings.HasSuffix(userTopic, "*") && strings.HasPrefix(topic, userTopic[:len(userTopic)-1]) || userTopic == topic {
```
brianstrauch, line 584: https://pkg.go.dev/strings#TrimSuffix might be more readable here.

PR #1929, opened 2023-05-24
[RCCA-12774] Only call `POST /api/access_tokens` when necessary

brianstrauch:
Checklist
---------
   * yes: ok

What
----
`POST /api/access_tokens` was being called for every authenticated command, even though it is only needed for two commands. This PR matches the style of Kafka REST, and only sets up the appropriate HTTP client when it's needed. Other standalone HTTP clients should match this pattern in a follow-up PR, but this fix is most important since it makes a wasteful HTTP call. I expect that this will improve the speed of all of our authenticated commands by a significant amount.

Deleted a few unit tests that are already covered by integration tests, and cleaned up some related dead code.

Test & Review
-------------
Integration tests still pass & manually tested.


internal/pkg/cmd/prerunner.go
```diff
@@ -118,6 +120,35 @@ func (c *AuthenticatedCLICommand) GetKafkaREST() (*KafkaREST, error) {
 	return (*c.KafkaRESTProvider)()
 }
 
+func (c *AuthenticatedCLICommand) GetMetricsClient() (*ccloudv2.MetricsClient, error) {
```
sgagniere, line 123: The KafkaREST style is kind of tedious, I think.

The on-prem Kafka commands use the following line: `c.PersistentPreRunE = prerunner.InitializeOnPremKafkaRest(c.AuthenticatedCLICommand)` (although you currently still need to get the client before you use it...)

It seems cleaner to do something along these lines: in commands which need a particular client, use `c.PersistenPreRunE` to call a prerunner function to set that client's field in `AuthenticatedCLICommand`. Then access the client in the style that we currently use for all non-KafkaREST clients (`c.V2Client.Blah()`).

This way, we don't initialize clients that aren't use, and we also don't need to call `GetClient` functions potentially multiple times for the same command.

sgagniere (commented):
`I expect that this will improve the speed of all of our authenticated commands by a significant amount.`
👍

internal/pkg/cmd/prerunner.go
brianstrauch: Yeah, agreed. Let's save this for a separate PR though, since it's a substantial rewrite. Since we completely overwrite the `PersistentPreRunE` field, we should try to chain functions to the existing prerun function. I started writing support for that in the past, so there should be some helper functions already set up for that!

internal/pkg/ccloudv2/metrics.go
```diff
@@ -22,35 +19,30 @@ type responseDataPoint struct {
 	Value     float32   `json:"value"`
 }
 
-func newMetricsClient(baseUrl, userAgent string, unsafeTrace, isTest bool) *metricsv2.APIClient {
+type MetricsClient struct {
+	*metricsv2.APIClient
+	authToken string
+}
+
+func NewMetricsClient(url, userAgent string, unsafeTrace bool, authToken string) *MetricsClient {
```
MuweiHe, line 27:
```suggestion
func NewMetricsClient(url, userAgent, authToken string, unsafeTrace bool) *MetricsClient {
```

brianstrauch: I'm trying to match the order that other similar functions use!

sgagniere (approved):

PR #1930, opened 2023-05-24
[Draft] Various formatting updates

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
We keep modifying the following patterns as we come across them, so I figured I would just get the majority of them in one go. I've included the regexes I've used to find them:

1. Multiline err blocks that could be single line `^[\t ,_]*err :?= [\w\.\(\)," -:\[\];]+\n[\t ]*if err !=`
Includes any number of underscores before `err`.
```
err := Foo()
if err != nil {
```
to
```
if err := Foo(); err != nil {
```

2. Multiline ok blocks that could be single line `^[\t ,_]*ok :?= [\w\.\(\)," -:\[\];]+\n[\t ]*if !?ok`
Includes both `ok` and `!ok`.
```
_, ok := Map[key]
if ok {
```
to
```
if _, ok := Map[key]; ok {
```

3. Declarations of a variable followed immediately by a return of that variable `^[\t ]*([\w]+) := [\w\.\(\)," -:\[\];\{\}]+\s*return (\1|nil|\1, |nil )*\n`
Includes `nil`s with the returned variable.
```
result := Bar()
return result, nil
```
to
```
return Bar(), nil
```

4. Multiline struct literals that only set one field `[\w]+\{\n[\t ]*[\w]+:[\t ]*[\w]+,\n[\t ]*\},*\n`
```
foo := Bar{
  a: b,
}
```
to
```
foo := Bar{a: b}
```
References
----------

Test & Review
-------------
All tests pass, but I want to carefully double check that I didn't introduce any subtle bugs, hence why I marked this as a draft PR.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -125,8 +125,7 @@ func (c *command) export(cmd *cobra.Command, _ []string) error {
 					currentTopic:   topic,
 					currentSubject: subject,
 				}
-				err := c.getChannelDetails(accountDetails, flags)
-				if err != nil {
+				if err := c.getChannelDetails(accountDetails, flags); err != nil {
 					if err.Error() == protobufErrorMessage {
 						log.CliLogger.Info(err.Error())
 						continue
@@ -163,8 +162,7 @@ func (c *command) export(cmd *cobra.Command, _ []string) error {
 }
 
 func (c *command) getChannelDetails(details *accountDetails, flags *flags) error {
-	err := details.getSchemaDetails()
-	if err != nil {
+	if err := details.getSchemaDetails(); err != nil {
 		if err.Error() == protobufErrorMessage {
 			return err
 		}
@@ -174,6 +172,7 @@ func (c *command) getChannelDetails(details *accountDetails, flags *flags) error
 		log.CliLogger.Warnf("Failed to get tags: %v", err)
 	}
 	details.channelDetails.example = nil
+	var err error
 	if flags.consumeExamples {
 		details.channelDetails.example, err = c.getMessageExamples(details.consumer, details.channelDetails.currentTopic.GetTopicName(), details.channelDetails.contentType, details.srClient, flags.valueFormat)
 		if err != nil {
@@ -198,14 +197,13 @@ func (c *command) getChannelDetails(details *accountDetails, flags *flags) error
 
 func (c *command) getAccountDetails(flags *flags) (*accountDetails, error) {
 	details := new(accountDetails)
-	err := c.getClusterDetails(details, flags)
-	if err != nil {
+	if err := c.getClusterDetails(details, flags); err != nil {
 		return nil, err
 	}
-	err = c.getSchemaRegistry(details, flags)
-	if err != nil {
+	if err := c.getSchemaRegistry(details, flags); err != nil {
 		return nil, err
 	}
+	var err error
 	details.subjects, _, err = details.srClient.DefaultApi.List(details.srContext, nil)
 	if err != nil {
 		return nil, err
@@ -242,8 +240,7 @@ func handlePanic() {
 
 func (c command) getMessageExamples(consumer *ckgo.Consumer, topicName, contentType string, srClient *schemaregistry.APIClient, valueFormatFlag string) (any, error) {
 	defer handlePanic()
-	err := consumer.Subscribe(topicName, nil)
-	if err != nil {
+	if err := consumer.Subscribe(topicName, nil); err != nil {
 		return nil, fmt.Errorf(`failed to subscribe to topic "%s": %v`, topicName, err)
 	}
 	message, err := consumer.ReadMessage(10 * time.Second)
@@ -275,8 +272,7 @@ func (c command) getMessageExamples(consumer *ckgo.Consumer, topicName, contentT
 		}
 		// Message body is encoded after 5 bytes of meta information.
 		value = value[messageOffset:]
-		err = deserializationProvider.LoadSchema(schemaPath, referencePathMap)
-		if err != nil {
+		if err := deserializationProvider.LoadSchema(schemaPath, referencePathMap); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/pkg/auth/sso/auth_sso_flow.go
+++ b/internal/pkg/auth/sso/auth_sso_flow.go
@@ -45,26 +45,22 @@ func Login(authURL string, noBrowser bool, connectionName string) (string, strin
 		// we need to start a background HTTP server to support the authorization code flow with PKCE
 		// described at https://auth0.com/docs/flows/guides/auth-code-pkce/call-api-auth-code-pkce
 		server := newServer(state)
-		err = server.startServer()
-		if err != nil {
+		if err := server.startServer(); err != nil {
 			return "", "", errors.Wrap(err, errors.StartHTTPServerErrorMsg)
 		}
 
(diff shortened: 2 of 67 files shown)


internal/cmd/asyncapi/command_export.go
```diff
@@ -174,6 +172,7 @@ func (c *command) getChannelDetails(details *accountDetails, flags *flags) error
 		log.CliLogger.Warnf("Failed to get tags: %v", err)
 	}
 	details.channelDetails.example = nil
+	var err error
 	if flags.consumeExamples {
 		details.channelDetails.example, err = c.getMessageExamples(details.consumer, details.channelDetails.currentTopic.GetTopicName(), details.channelDetails.contentType, details.srClient, flags.valueFormat)
```
brianstrauch, line 177: The more idiomatic approach for cases like this one is to do `example, err :=` which lets us use `:=` to shorten the scope of `err`, followed by an assignment, i.e. `details.channelDetails.example = example`.

internal/pkg/auth/sso/auth_sso_flow.go
```diff
@@ -45,26 +45,22 @@ func Login(authURL string, noBrowser bool, connectionName string) (string, strin
...
-		err = server.startServer()
-		if err != nil {
+		if err := server.startServer(); err != nil {
 			return "", "", errors.Wrap(err, errors.StartHTTPServerErrorMsg)
 		}
 
 		// Get authorization code for making subsequent token request
-		err := browser.OpenURL(state.getAuthorizationCodeUrl(connectionName, isOkta))
-		if err != nil {
+		if err := browser.OpenURL(state.getAuthorizationCodeUrl(connectionName, isOkta)); err != nil {
```
brianstrauch, line 53:
```suggestion
		url := state.getAuthorizationCodeUrl(connectionName, isOkta)
		if err := browser.OpenURL(url); err != nil {
```

internal/pkg/auth/sso/auth_sso_flow.go
```diff
@@ -45,26 +45,22 @@ func Login(authURL string, noBrowser bool, connectionName string) (string, strin
...
 			return "", "", errors.Wrap(err, errors.OpenWebBrowserErrorMsg)
 		}
 
-		err = server.awaitAuthorizationCode(30 * time.Second)
-		if err != nil {
+		if err = server.awaitAuthorizationCode(30 * time.Second); err != nil {
```
brianstrauch, line 57: Unrelated, but I'm thinking we want to remove the timer here, and let the user exit with Ctrl+C. The 30s timeout is arbitrary and sometimes it takes me longer than 30s to get my phone and accept the MFA request.

internal/pkg/config/base_config.go
```diff
@@ -23,11 +23,11 @@ func (v *Version) MarshalJSON() ([]byte, error) {
...
 	var s string
-	err := json.Unmarshal(data, &s)
-	if err != nil {
+	if err := json.Unmarshal(data, &s); err != nil {
 		return err
 	}
 
+	var err error
 	v.Version, err = version.NewVersion(s)
 	return err
```
brianstrauch, line 32:
```suggestion
	version, err := version.NewVersion(s)
	if err != nil {
		return err
	}
	v.Version = version
	
	return nil
```

internal/pkg/secret/encryption_other.go
```diff
@@ -25,11 +25,10 @@ const (
 
 func GenerateRandomBytes(n int) ([]byte, error) {
 	b := make([]byte, n)
-	_, err := rand.Read(b)
-	if err != nil {
+	if _, err := rand.Read(b); err != nil {
 		return []byte{}, err
 	}
-	return b, err
+	return b, nil
```
brianstrauch, line 31: Thanks! Lines like this one could've caused an unintended bug.

internal/pkg/secret/encryption_other.go
```diff
@@ -38,8 +37,7 @@ func DeriveEncryptionKey(salt []byte) ([]byte, error) {
 		return []byte{}, err
 	}
 	userId := strconv.Itoa(os.Getuid())
-	encryptionKey := pbkdf2.Key([]byte(machineId+userId), salt, iterationNumber, keyLength, sha256.New)
-	return encryptionKey, nil
+	return pbkdf2.Key([]byte(machineId+userId), salt, iterationNumber, keyLength, sha256.New), nil
```
brianstrauch, line 40: Can you break the first argument out, and use `fmt.Sprintf()` to create it? We won't have to use `strconv.Itoa()` that way.

internal/pkg/secret/jaas_config_parser.go
```diff
@@ -189,8 +185,7 @@ func (j *JAASParser) parseConfigurationEntry(prefixKey string) (int, int, *prope
 	}
 
 	// Parse Control Flag
-	err = j.parseControlFlag()
-	if err != nil {
+	if err := j.parseControlFlag(); err != nil {
 		return 0, 0, nil, "", err
```
brianstrauch, line 189: Wow, this function returns 5 arguments? 😮

internal/pkg/utils/stripe_utils.go
```diff
@@ -19,9 +19,7 @@ func setStripeKey(isTest bool) {
 	} else {
 		stripe.Key = stripeLiveKey
 	}
-	stripe.DefaultLeveledLogger = &stripe.LeveledLogger{
-		Level: 0,
-	}
+	stripe.DefaultLeveledLogger = &stripe.LeveledLogger{Level: 0}
```
brianstrauch, line 22: This should be equivalent. I would expect this to already be the definition of the default logger though?
```suggestion
	stripe.DefaultLeveledLogger = &stripe.LeveledLogger{}
```

brianstrauch (commented):
Thank you!

brianstrauch:
(I wonder if there's a linter in https://github.com/golangci/golangci-lint that can enforce this?)

internal/pkg/secret/encryption_other.go
sgagniere: Yup. I just happened to notice this while changing other stuff. I wonder if there's some regex string we could write to find these.

internal/pkg/secret/jaas_config_parser.go
sgagniere: Might be something to revisit later for a potential refactor.

sgagniere:
@brianstrauch I didn't see any, but I only skimmed the list. Perhaps we could refine these regex strings and make our own linters?

internal/pkg/utils/stripe_utils.go
sgagniere: The default logger uses `Level: LevelError` which is `1`, we're setting the default to `LevelNull`.

Your suggested change should still be equivalent. `Level` is defined by `type Level int32`, and the default value of `int32` is `0`.

Think we're safe to make this change, or leave it for now?

brianstrauch: Let's leave as-is. It's better to be explicit in this case.

brianstrauch (approved):

sgagniere:
Made a few more changes since the last review, that's why I requested another +1 before merging.

brianstrauch (approved):
Thank you!

PR #1931, opened 2023-05-25
[CLI-2504] Use `UNKNOWN` as the default instead of `DEDICATED` for cluster type

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
An `UNKNOWN` value is already supported, so we should use that instead of assuming that `DEDICATED` is the default case.

References
----------

Test & Review
-------------
Added an integration test.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/test/test-server/utils.go
+++ b/test/test-server/utils.go
@@ -315,9 +315,7 @@ func getCmkUnknownDescribeCluster(id string, name string) *cmkv2.CmkV2Cluster {
 			Availability:           cmkv2.PtrString("SINGLE_ZONE"),
 		},
 		Id: cmkv2.PtrString(id),
-		Status: &cmkv2.CmkV2ClusterStatus{
-			Phase: "PROVISIONED",
-		},
+		Status: &cmkv2.CmkV2ClusterStatus{Phase: "PROVISIONED"},
 	}
 }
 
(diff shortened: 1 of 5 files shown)


test/test-server/utils.go
```diff
@@ -304,6 +304,23 @@ func getCmkDedicatedDescribeCluster(id string, name string, cku int32) *cmkv2.Cm
...
+			Cloud:                  cmkv2.PtrString("aws"),
+			Region:                 cmkv2.PtrString("us-west-2"),
+			KafkaBootstrapEndpoint: cmkv2.PtrString("SASL_SSL://kafka-endpoint"),
+			HttpEndpoint:           cmkv2.PtrString(TestKafkaRestProxyUrl.String()),
+			Availability:           cmkv2.PtrString("SINGLE_ZONE"),
+		},
+		Id: cmkv2.PtrString(id),
+		Status: &cmkv2.CmkV2ClusterStatus{
+			Phase: "PROVISIONED",
+		},
```
brianstrauch, line 320:
```suggestion
		Status: &cmkv2.CmkV2ClusterStatus{Phase: "PROVISIONED"},
```

brianstrauch (approved):

PR #1935, opened 2023-05-30
[CLI-2347, CLI-2349] Add `flink [compute-pool|region|statement]`

brianstrauch:
Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok

What
----
Add a top-level command for Flink, gated by a feature flag

Test & Review
-------------
Integration tests

PR #1937, opened 2023-05-31
Kfs 788/bring sql client history

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Add `confluent flink shell`, merge after #1935.

Test & Review
-------------
Snapshot tests, unit tests


sgagniere (commented):
There are quite a number of changes here that are also in the other PR. Which changes need reviewing? The `shell` additions?

brianstrauch:
> Are the shell additions the only one that needs reviewing here?

Yes, and this'll get merged after #1935. The goal is to merge it by the EA date tomorrow!

MuweiHe (approved):
+1

PR #1939, opened 2023-05-31
Feat/kfs 807 set column width dynamically

brianstrauch:
Checklist
---------

What
----
We now set the column width dynamically based on the window size. This is for both the plain text and the tview table.

The widths are distributed greedily. Each column gets as much space as it wants until there is no more space to distribute. When there's no more space to distribute, the rest of the column widths are set to min(columnWidth, availableSpace / numColumns)

Test & Review
-------------
TODO


YannickPferr (approved):
LGTM!

PR #1946, opened 2023-06-02
chore: remove retry for running/completed statements

YannickPferr:
What
----

This removes the retry logic for completed statements when fetching results. This was originally added as a workaround for a bug in the backend, which is now resolved.


internal/pkg/flink/internal/store/store.go
```diff
@@ -104,34 +104,24 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
...
 	// Process remote statements that are now running or completed
-	pageToken := statement.PageToken
-	runningNoTokenRetries := 5
-	for i := 0; i < runningNoTokenRetries; i++ {
-		// TODO: we need to retry a few times on transient errors
-		statementResultObj, err := s.client.GetStatementResults(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvId(), statement.StatementName, pageToken)
-		if err != nil {
-			return nil, &types.StatementError{Msg: err.Error()}
-		}
+	statementResultObj, err := s.client.GetStatementResults(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvId(), statement.StatementName, statement.PageToken)
```
brianstrauch, line 107: Nit: Remove ambiguity
```suggestion
	statementResultObj, err := s.client.GetStatementResults(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvironmentId(), statement.StatementName, statement.PageToken)
```

brianstrauch:
I assume we should be targeting `main` since the current target branch has already been merged?

internal/pkg/flink/internal/store/store.go
YannickPferr: Done!

internal/pkg/flink/internal/store/store.go
```diff
@@ -104,34 +104,24 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
...
-		statementResults := statementResultObj.GetResults()
-		convertedResults, err := results.ConvertToInternalResults(statementResults.GetData(), statement.ResultSchema)
-		if err != nil {
-			return nil, &types.StatementError{Msg: "Error: " + err.Error()}
-		}
-		statement.StatementResults = convertedResults
+	statementResults := statementResultObj.GetResults()
+	convertedResults, err := results.ConvertToInternalResults(statementResults.GetData(), statement.ResultSchema)
+	if err != nil {
+		return nil, &types.StatementError{Msg: "Error: " + err.Error()}
```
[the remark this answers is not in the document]
YannickPferr: Done!

internal/pkg/flink/internal/store/store.go
```diff
@@ -104,39 +104,29 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
...
-		statementResults := statementResultObj.GetResults()
-		convertedResults, err := results.ConvertToInternalResults(statementResults.GetData(), statement.ResultSchema)
-		if err != nil {
-			return nil, &types.StatementError{Msg: "Error: " + err.Error()}
-		}
-		statement.StatementResults = convertedResults
+	statementResults := statementResultObj.GetResults()
+	convertedResults, err := results.ConvertToInternalResults(statementResults.GetData(), statement.ResultSchema)
+	if err != nil {
+		return nil, &types.StatementError{Msg: fmt.Sprintf("Error: %v", err.Error())}
```
MarTango, line 115:
```suggestion
		return nil, &types.StatementError{Msg: fmt.Sprintf("Error: %v", err)}
```

nit: `.Error()` is unnecessary

MarTango, line 115:
```suggestion
		return nil, &types.StatementError{Msg: fmt.Sprintf("error: %v", err.Error())}
```
nit: error messages should be lowercase (not sure if that's the convention in this repo though)

internal/pkg/flink/internal/store/store.go
```diff
@@ -104,39 +104,29 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
...
-		statement.PageToken = extractedToken
-		if statement.Status == types.COMPLETED || statement.PageToken != "" || len(statementResults.GetData()) > 0 {
-			// We try a few times to get non-empty token for RUNNING statements
-			break
-		}
-		time.Sleep(time.Millisecond * 300)
+	statementMetadata := statementResultObj.GetMetadata()
+	extractedToken, err := extractPageToken(statementMetadata.GetNext())
+	if err != nil {
+		return nil, &types.StatementError{Msg: fmt.Sprintf("Error: %v", err.Error())}
```
MarTango, line 122:
```suggestion
		return nil, &types.StatementError{Msg: fmt.Sprintf("error: %v", err)}
```

same as above

MarTango (approved):

internal/pkg/flink/internal/store/store.go
YannickPferr: It seems to be the convention in this repo (at least from what I have seen so far). Every message that gets shown to the user/printed to the terminal always starts with an uppercase letter

internal/pkg/flink/internal/store/store.go
YannickPferr: It seems to be the convention in this repo (at least from what I have seen so far). Every message that gets shown to the user/printed to the terminal always starts with an uppercase letter

internal/pkg/flink/internal/store/store.go
```diff
@@ -104,39 +104,29 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
...
 	// Process remote statements that are now running or completed
-	pageToken := statement.PageToken
-	runningNoTokenRetries := 5
-	for i := 0; i < runningNoTokenRetries; i++ {
-		// TODO: we need to retry a few times on transient errors
-		statementResultObj, err := s.client.GetStatementResults(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvId(), statement.StatementName, pageToken)
-		if err != nil {
-			return nil, &types.StatementError{Msg: err.Error()}
-		}
+	statementResultObj, err := s.client.GetStatementResults(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvironmentId(), statement.StatementName, statement.PageToken)
```
brianstrauch, line 107: FYI: If you're using the correct flink-gateway client, it should have built-in retry logic if the backend returns 429 or 5xx.

brianstrauch (approved):

internal/pkg/flink/internal/store/store.go
YannickPferr: Oh nice!
In this case though, the retry was just there to get around a backend issue (in some cases it didn't return a page token when there should be one. So when that happened, we retried four times to make sure there really is no token)

PR #1947, opened 2023-06-02
feat: show status detail message to user if available

YannickPferr:
What
----

We want to show the status message to the user when it's available. For successful statements we store in it processedStatement.StatusDetail and for errors we store it in statementError.FailureMessage. 

Additionally, if a statement is in FAILED or FAILING state and it doesn't have the status detail field filled, we fetch the latest exception from the exceptions API and display the exception stack trace instead.


internal/pkg/flink/internal/store/store.go
```diff
@@ -157,6 +157,13 @@ func (s *Store) waitForPendingStatement(ctx context.Context, statementName strin
 			return nil, &types.StatementError{Msg: "Result retrieval aborted. Statement will be deleted.", HttpResponseCode: 499}
 		default:
 			statementObj, err := s.client.GetStatement(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvId(), statementName)
+
+			// print status detail message if available
+			status := statementObj.GetStatus()
```
alexey-lv, line 162: This was added as debug-only block to know the current status. We don't want to show it from the retry loop till we think we're done.

YannickPferr: Got it, will remove!

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -121,6 +121,9 @@ func (c *InputController) RunInteractiveInput() {
 		})
 
 		processedStatement, err = c.store.WaitPendingStatement(ctx, *processedStatement)
+		if processedStatement.StatusDetail != "" {
```
alexey-lv, line 124: processedStatemenent can be nil here, right?

YannickPferr: Good catch!

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -121,6 +121,9 @@ func (c *InputController) RunInteractiveInput() {
 		})
 
 		processedStatement, err = c.store.WaitPendingStatement(ctx, *processedStatement)
+		if processedStatement != nil && processedStatement.StatusDetail != "" {
```
alexey-lv, line 124: But we still want to show the status detail returned from the server ;) there is a bit of refactoring to do. That's why my version added a new return value here.

YannickPferr: Oh okay, now I think I get it. You mean the status detail could be filled when the backend returns an error, in which case we return nil as the processed statement causing it to not be printable in the input controller and that's why you added that extra return value.

Yeah in this case it indeed needs some more refactoring. Sorry, I rushed through this a bit too quickly tbh.. I will look at it next week more thoroughly!

alexey-lv: Sure, some bare minimum is better than nothing. Feel free to release things incrementally. thanks!

internal/pkg/ccloudv2/flink_gateway.go
```diff
@@ -16,6 +16,7 @@ type GatewayClientInterface interface {
 	ListStatements(orgId, environmentId string) ([]flinkgatewayv1alpha1.SqlV1alpha1Statement, error)
 	CreateStatement(orgId, environmentId, computePoolId, identityPoolId, statement string, properties map[string]string) (flinkgatewayv1alpha1.SqlV1alpha1Statement, error)
 	GetStatementResults(orgId, environmentId, statementId, pageToken string) (flinkgatewayv1alpha1.SqlV1alpha1StatementResult, error)
+	GetExceptions(orgId, environmentId, statementId string) (flinkgatewayv1alpha1.SqlV1alpha1StatementExceptionList, error)
```
brianstrauch, line 19: Instead of mocking these functions, your team might be interested in the integration testing framework in the `test/` directory. We mock the backend in `test/test-server`... right now, `flink shell` is the only command not leveraging this!

internal/pkg/ccloudv2/flink_gateway.go
```diff
@@ -79,3 +80,8 @@ func (c *FlinkGatewayClient) GetStatementResults(orgId, environmentId, statement
 	res, r, err := req.Execute()
 	return res, errors.CatchCCloudV2Error(err, r)
 }
+
+func (c *FlinkGatewayClient) GetExceptions(orgId, environmentId, statementId string) (flinkgatewayv1alpha1.SqlV1alpha1StatementExceptionList, error) {
```
brianstrauch, line 84: Nit: We like to pass arguments into the function in the same order as they're used.
```suggestion
func (c *FlinkGatewayClient) GetExceptions(environmentId, statementId, orgId string) (flinkgatewayv1alpha1.SqlV1alpha1StatementExceptionList, error) {
```

internal/pkg/ccloudv2/flink_gateway.go
```diff
@@ -79,3 +80,8 @@ func (c *FlinkGatewayClient) GetStatementResults(orgId, environmentId, statement
...
+	result, r, err := c.StatementExceptionsSqlV1alpha1Api.GetSqlV1alpha1StatementExceptions(c.flinkGatewayApiContext(), environmentId, statementId).OrgId(orgId).Execute()
```
brianstrauch, line 85: I think this is the standard variable name; take a look at other similar files.
```suggestion
	res, r, err := c.StatementExceptionsSqlV1alpha1Api.GetSqlV1alpha1StatementExceptions(c.flinkGatewayApiContext(), environmentId, statementId).OrgId(orgId).Execute()
```

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -128,6 +128,10 @@ func (c *InputController) RunInteractiveInput() {
 			c.isSessionValid(err)
 			continue
 		}
+		// print status detail message if available
+		if processedStatement.StatusDetail != "" {
+			output.Printf("Status detail: %s\n", processedStatement.StatusDetail)
```
brianstrauch, line 133: Do we necessarily need to label this, or should just printing it be sufficient?
```suggestion
			output.Println(processedStatement.StatusDetail)
```

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -198,28 +202,33 @@ func (c *InputController) setInitialBuffer(s string) {
...
+		if processedStatement.Status != "FAILED" {
 			output.Println("Statement successfully submitted. ")
 			output.Println("Fetching results...\n ")
 		} else {
 			output.Println("Error: Statement submission failed. There could a problem with the server right now. Check your statement and try again.")
 		}
+		// print status detail message if available
+		if processedStatement.StatusDetail != "" {
+			output.Printf("Status detail: %s\n", processedStatement.StatusDetail)
+		}
```
brianstrauch, line 229: Duplicate code. Write a function?

internal/pkg/flink/internal/store/store.go
```diff
@@ -68,28 +69,27 @@ func (s *Store) ProcessStatement(statement string) (*types.ProcessedStatement, *
 		s.propsDefault(s.Properties),
 	)
 	if err != nil {
-		return nil, &types.StatementError{Msg: err.Error()}
+		statusDetail := s.getStatusDetail(statementObj)
+		return nil, &types.StatementError{Msg: err.Error(), FailureMessage: statusDetail}
```
brianstrauch, line 73: Split this error across multiple lines since it has more than one field?

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -88,7 +88,7 @@ func (s *Store) processResetStatement(statement string) (*types.ProcessedStateme
 	} else {
 		_, keyExists := s.Properties[configKey]
 		if !keyExists {
-			return nil, &types.StatementError{Msg: fmt.Sprintf("Error: Config key \"%s\" is currently not set.", configKey)}
+			return nil, &types.StatementError{Msg: fmt.Sprintf("Config key \"%s\" is currently not set.", configKey)}
```
brianstrauch, line 91:
```suggestion
			return nil, &types.StatementError{Msg: fmt.Sprintf(`Config key "%s" is not set.`, configKey)}
```

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -136,7 +136,7 @@ func parseSetStatement(statement string) (string, string, error) {
 
 	indexOfSet := strings.Index(strings.ToUpper(statement), config.ConfigOpSet)
 	if indexOfSet == -1 {
-		return "", "", &types.StatementError{Msg: "Error: Invalid syntax for SET. Usage example: SET 'key'='value'."}
+		return "", "", &types.StatementError{Msg: "Invalid syntax for SET. Usage example: SET 'key'='value'."}
```
brianstrauch, line 139: I wonder if the backend could parse this for us, even though it's a frontend-specific command.

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -196,7 +196,7 @@ func parseUseStatement(statement string) (string, string, error) {
 	statement = removeStatementTerminator(statement)
 	words := strings.Fields(statement)
 	if len(words) < 2 {
-		return "", "", &types.StatementError{Msg: "Error: Missing database/catalog name: Usage examples: USE DB1 OR USE CATALOG METADATA."}
+		return "", "", &types.StatementError{Msg: "Missing database/catalog name: Usage examples: USE DB1 OR USE CATALOG METADATA."}
```
brianstrauch, line 199: The "usage example(s)" string is a bit inconsistent. Can we introduce a `Usage []string` field here so we can format usage in a consistent way?

internal/pkg/flink/types/statements.go
```diff
@@ -73,10 +74,20 @@ type MockStatementResult struct {
...
-	return e.Msg
+	if e == nil {
+		return ""
+	}
+	if e.Msg == "" {
+		return "Error"
+	}
+	if e.FailureMessage == "" {
+		return fmt.Sprintf("Error: %s", e.Msg)
+	}
```
brianstrauch, line 89: How are these different?

internal/pkg/flink/types/statements.go
```diff
@@ -73,10 +74,20 @@ type MockStatementResult struct {
...
+	return fmt.Sprintf("Error: %s \nError details: %s", e.Msg, e.FailureMessage)
```
brianstrauch, line 90: Why the space after %s?

internal/pkg/flink/internal/controller/input_controller.go
alexey-lv: I think just printing it could be sufficient. We might want to indicate that this is a response from the server, but is not required.

internal/pkg/flink/internal/store/store.go
```diff
@@ -189,10 +197,36 @@ func (s *Store) waitForPendingStatement(ctx context.Context, statementName strin
...
+
+	if status.GetDetail() != "" {
+		return status.GetDetail()
+	}
+
+	// if the statement is in FAILED or FAILING phase and the status detail field is empty we show the latest exception instead
+	exceptionsResponse, _ := s.client.GetExceptions(s.appOptions.GetOrgResourceId(), s.appOptions.GetEnvId(), statementObj.Spec.GetStatementName())
+	exceptions := exceptionsResponse.GetData()
+	if len(exceptions) < 1 {
+		return status.GetDetail()
```
alexey-lv, line 225: Looks like at this point `status.GetDetail()` would be always empty here, right?

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -128,6 +128,10 @@ func (c *InputController) RunInteractiveInput() {
 			c.isSessionValid(err)
 			continue
```
alexey-lv, line 129: Do we have a test case that gets into this 'if' branch and prints the StatusDetail?

internal/pkg/flink/types/statements.go
YannickPferr: The first would just be "Error" while the second would be "Error: " if e.Msg was empty. 
Maybe "Unknown Error" or something like that would be better for the first? This case should never happen really, but just to be safe..

internal/pkg/flink/internal/store/store.go
YannickPferr: Yes, it would be equivalent to just return empty string here! I can change it if you prefer it that way

internal/pkg/ccloudv2/flink_gateway.go
YannickPferr: Reordered this and all other methods in GW client!

internal/pkg/flink/internal/controller/input_controller.go
YannickPferr: Wrote a helper in ProcessedStatement!

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Done here and in other places I found!

internal/pkg/flink/types/statements.go
YannickPferr: No particular reason, can remove!

internal/pkg/flink/internal/controller/input_controller.go
YannickPferr: Unfortunately no. Our test setup for the input controller is not great at the moment. We haven't found a good way to test the whole STDIN/STDOUT interaction flow, but we have some tickets on our board to address this (current plan is to address it next sprint)

internal/pkg/ccloudv2/flink_gateway.go
YannickPferr: I'll have a look, though we have some tickets on our board regarding test improvements for the client. I'd say we'll try to make use of your IT testing approach when we get to these!

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Done!

internal/pkg/flink/internal/store/store.go
alexey-lv: Let's make it empty string to be more explicit about the intent.

internal/pkg/flink/internal/controller/input_controller.go
alexey-lv: Sanity check: in this case this PR would report the detail, right?

internal/pkg/flink/types/statements.go
brianstrauch: `Error: No message`?

internal/pkg/flink/internal/store/store.go
```diff
@@ -178,10 +191,36 @@ func (s *Store) waitForPendingStatement(ctx context.Context, statementName strin
 
 	var errorsMsg string
 	if len(capturedErrors) > 0 {
-		errorsMsg = fmt.Sprintf(" Captured retriable errors: %s", strings.Join(capturedErrors, "; "))
+		errorsMsg = fmt.Sprintf("Captured retriable errors: %s", strings.Join(capturedErrors, "; "))
+	}
+
+	return nil, &types.StatementError{
+		Msg: fmt.Sprintf("Statement is still pending after %f seconds.\n If you want to increase the timeout for the client, you can run \"SET table.results-timeout=1200;\" to adjust the maximum timeout in seconds.",
```
brianstrauch, line 198: Remove newline? (Or at least the weird space after the newline?)
```suggestion
		Msg: fmt.Sprintf("Statement is still pending after %f seconds. If you want to increase the timeout for the client, you can run \"SET table.results-timeout=1200;\" to adjust the maximum timeout in seconds.",
```

internal/pkg/flink/internal/store/store.go
```diff
@@ -178,10 +191,36 @@ func (s *Store) waitForPendingStatement(ctx context.Context, statementName strin
...
+	if phase != types.FAILED && phase != types.FAILING {
+		return status.GetDetail()
+	}
+
+	if status.GetDetail() != "" {
+		return status.GetDetail()
+	}
+
+	// if the statement is in FAILED or FAILING phase and the status detail field is empty we show the latest exception instead
+	exceptionsResponse, _ := s.client.GetExceptions(s.appOptions.GetEnvironmentId(), statementObj.Spec.GetStatementName(), s.appOptions.GetOrgResourceId())
```
brianstrauch, line 216: Why are we ignoring the error?

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -77,10 +77,12 @@ func TestWaitForPendingStatement3(t *testing.T) {
 		appOptions: &appOptions,
 	}
 
+	statusDetailMessage := "Test status detail message"
 	// Test case 1: Statement is not pending
 	statementObj := flinkgatewayv1alpha1.SqlV1alpha1Statement{
 		Status: &flinkgatewayv1alpha1.SqlV1alpha1StatementStatus{
-			Phase: "COMPLETED",
+			Phase:  "COMPLETED",
+			Detail: &statusDetailMessage,
```
brianstrauch, line 85:
```suggestion
			Detail: flinkgatewayv1alphav1.PtrString("Test status detail message"),
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -105,16 +107,21 @@ func TestWaitForPendingTimesout(t *testing.T) {
...
 	statementObj := flinkgatewayv1alpha1.SqlV1alpha1Statement{
 		Status: &flinkgatewayv1alpha1.SqlV1alpha1StatementStatus{
-			Phase: "PENDING",
+			Phase:  "PENDING",
+			Detail: &statusDetailMessage,
 		},
 	}
+	expectedError := &types.StatementError{
+		Msg:            fmt.Sprintf("Statement is still pending after %f seconds.\n If you want to increase the timeout for the client, you can run \"SET table.results-timeout=1200;\" to adjust the maximum timeout in seconds.", timeout.Seconds()),
+		FailureMessage: fmt.Sprintf("Captured retriable errors: %s", statusDetailMessage),
```
brianstrauch, line 119:
```suggestion
		FailureMessage: fmt.Sprintf("Captured retryable errors: %s", statusDetailMessage),
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -164,16 +174,22 @@ func TestWaitForPendingStatementErrors(t *testing.T) {
...
 		Status: &flinkgatewayv1alpha1.SqlV1alpha1StatementStatus{
-			Phase: "COMPLETED",
+			Phase:  "COMPLETED",
+			Detail: &statusDetailMessage,
 		},
 	}
 
-	expectedErr := errors.New("couldn't get statement!")
-	client.EXPECT().GetStatement("orgId", "envId", statementName).Return(statementObj, expectedErr).Times(1)
+	returnedError := errors.New("couldn't get statement!")
```
brianstrauch, line 185: We try to keep punctuation out of errors in case they are nested!
```suggestion
	returnedError := errors.New("couldn't get statement")
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -542,7 +558,7 @@ func (s *StoreTestSuite) TestProccessHttpErrors() {
 
 	// expect
 	assert.NotNil(s.T(), err)
-	assert.Equal(s.T(), "Invalid syntax: you should provide a table for select", err.Error())
+	assert.Equal(s.T(), "Error: Invalid syntax: you should provide a table for select.", err.Error())
```
brianstrauch, line 561: Errors should not be capitalized and should not contain punctuation
```suggestion
	assert.Equal(s.T(), "Error: invalid syntax: you should provide a table for select", err.Error())
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -799,3 +815,414 @@ func TestTimeout(t *testing.T) {
...
+		},
+		client:     client,
+		appOptions: appOptions,
+	}
+
+	statementName := "Test Statement"
+	statusDetailMessage := "Test status detail message"
+	statementObj := flinkgatewayv1alpha1.SqlV1alpha1Statement{
+		Spec: &flinkgatewayv1alpha1.SqlV1alpha1StatementSpec{
+			StatementName: &statementName,
```
brianstrauch, line 916:
```suggestion
			StatementName: flinkgatewayv1alpha1.PtrString("Test Statement"),
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -799,3 +815,414 @@ func TestTimeout(t *testing.T) {
...
+	require.Nil(s.T(), err)
+	require.Equal(s.T(), &statement, processedStatement)
+}
+
+func (s *StoreTestSuite) TestWaitPendingStatementNoWaitForRunningStatement() {
+	client := mock.NewMockGatewayClientInterface(gomock.NewController(s.T()))
+	store := Store{
+		Properties: map[string]string{
+			"TestProp": "TestVal",
+		},
```
brianstrauch, line 956:
```suggestion
		Properties: map[string]string{"TestProp": "TestVal"},
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -799,3 +815,414 @@ func TestTimeout(t *testing.T) {
...
+		client: client,
+	}
+
+	statement := types.ProcessedStatement{
+		Status: types.PHASE("RUNNING"),
+	}
```
brianstrauch, line 962:
```suggestion
	statement := types.ProcessedStatement{Status: types.PHASE("RUNNING")}
```

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -153,30 +156,45 @@ func parseSetStatement(statement string) (string, string, error) {
...
-		return "", "", &types.StatementError{Msg: "Error: \"=\" should only appear once. Usage example: SET 'key'='value'."}
+		return "", "", &types.StatementError{
+			Msg:   `"=" should only appear once`,
+			Usage: []string{"SET 'key'='value'"},
+		}
 	}
 
 	if keyValuePair[0] != "" && keyValuePair[1] == "" {
-		return "", "", &types.StatementError{Msg: "Error: Value for key not present. If you want to reset a key, use \"RESET 'key'\"."}
+		return "", "", &types.StatementError{Msg: `Value for key not present. If you want to reset a key, use "RESET 'key'"`}
```
brianstrauch, line 175: Looks like the second sentence belongs in `Usage`

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -196,7 +214,10 @@ func parseUseStatement(statement string) (string, string, error) {
 	statement = removeStatementTerminator(statement)
 	words := strings.Fields(statement)
 	if len(words) < 2 {
-		return "", "", &types.StatementError{Msg: "Error: Missing database/catalog name: Usage examples: USE DB1 OR USE CATALOG METADATA."}
+		return "", "", &types.StatementError{
+			Msg:   "Missing database/catalog name",
+			Usage: []string{"USE CATALOG my_catalog or USE my_database"},
```
brianstrauch, line 219: Is "or" a SQL keyword, or should this be two separate usage strings?

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -233,43 +263,52 @@ func parseResetStatement(statement string) (string, error) {
...
 func processHttpErrors(resp *http.Response, err error) error {
 	if err != nil {
-		return &types.StatementError{Msg: "Error: " + err.Error()}
+		return &types.StatementError{Msg: err.Error()}
 	}
 
 	if resp != nil && resp.StatusCode >= 400 {
 		if resp.StatusCode == http.StatusUnauthorized {
-			return &types.StatementError{Msg: "Error: Unauthorized. Please consider running confluent login again.", HttpResponseCode: resp.StatusCode}
+			return &types.StatementError{Msg: "Unauthorized. Please consider running confluent login again", HttpResponseCode: resp.StatusCode}
```
brianstrauch, line 298:
```suggestion
			return &types.StatementError{
				Msg: "Unauthorized. Please run `confluent login`",
				HttpResponseCode: resp.StatusCode,
			}
```

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -233,43 +263,52 @@ func parseResetStatement(statement string) (string, error) {
...
+			Usage: []string{"RESET 'key'"},
+		}
 	}
 
 	isFirstWordReset := strings.ToUpper(words[0]) == config.ConfigOpReset
 	key := strings.ToLower(words[1])
 	if !isFirstWordReset {
-		return "", &types.StatementError{Msg: "Error: Invalid syntax for RESET. Usage example: RESET 'key'."}
+		return "", &types.StatementError{
+			Msg:   "Invalid syntax for RESET",
```
brianstrauch, line 276: Your error messages are inconsistent. Some start with an uppercase letter and some don't. My vote is for starting them with a lowercase letter! https://staticcheck.io/docs/checks#ST1005
```suggestion
			Msg:   "invalid syntax for RESET",
```

internal/pkg/flink/internal/store/store_utils.go
brianstrauch: The second sentence here is more of a suggestion than an error. Should we introduce another field called "Suggestion"? This is what other CLI commands do.

internal/pkg/flink/internal/store/store_utils_test.go
```diff
@@ -141,30 +141,30 @@ func TestProcessResetStatement(t *testing.T) {
...
-		assert.Equal(t, "Configuration has been reset successfully.", result.StatusDetail)
+		assert.Equal(t, "Configuration has been reset successfully", result.StatusDetail)
 		assert.Nil(t, result.StatementResults)
 	})
 
 	t.Run("should return an error message if configKey does not exist", func(t *testing.T) {
 		result, err := s.processResetStatement("reset 'location'")
 		assert.NotNil(t, err)
-		assert.Equal(t, "Error: Config key \"location\" is currently not set.", err.Error())
+		assert.Equal(t, `Error: Config key "location" is not set.`, err.Error())
```
brianstrauch, line 151:
```suggestion
		assert.Equal(t, `Error: configuration key "location" is not set.`, err.Error())
```

internal/pkg/flink/types/statements.go
```diff
@@ -73,10 +75,26 @@ type MockStatementResult struct {
 type StatementError struct {
 	Msg              string
```
brianstrauch, line 76:
```suggestion
	Message          string
```

internal/pkg/flink/types/statements.go
```diff
@@ -73,10 +75,26 @@ type MockStatementResult struct {
...
-	return e.Msg
+	if e == nil {
+		return ""
+	}
+	errStr := "Unknown error."
+	if e.Msg != "" {
+		errStr = fmt.Sprintf("Error: %s.", e.Msg)
+	}
+	if len(e.Usage) > 0 {
+		errStr += fmt.Sprintf("\nUsage: %s.", strings.Join(e.Usage, " or "))
```
brianstrauch, line 91: We have `utils.ArrayToCommaDelimitedString()` for this!

internal/pkg/flink/internal/store/store.go
YannickPferr: Works for me! And I updated it yesterday already 👍

internal/pkg/flink/internal/store/store.go
YannickPferr: Because we don't really need to check it here. When an error happens `exceptionsResponse.GetData()` will still return an empty array and thus trigger the `if` which returns an empty string. 
It works the same as this:
`if err != nil || len(exceptionsResponse.GetData()) < 1`

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Fixed!

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Removed!

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Yeah it should be two separate strings, missed that when refactoring!

internal/pkg/flink/types/statements.go
YannickPferr: Using that now

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Made all of them lowercase and removed the punctuation

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Made all of them lowercase and removed punctuation!

internal/pkg/flink/internal/store/store_utils_test.go
YannickPferr: changed config to configuration in all places!

internal/pkg/flink/types/statements.go
YannickPferr: Using that now!

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Moved it to new Suggestion field

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: Added Suggestion field!

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Fixed!

internal/pkg/flink/internal/store/store_test.go
YannickPferr: In this case I need statementName again in line 923, but I'm using `PtrString()` now wherever I don't need the variable more than once!

internal/pkg/flink/internal/store/store.go
YannickPferr: Okay makes sense, will handle it explicitly!

YannickPferr:
I've added another commit to handle the error explicitly [here](https://github.com/confluentinc/cli/commit/25d619e13a0f9cc9b000470eb91c8bfe44f9e9e4) but it doesn't show up in the PR for some reason (maybe the repo doesn't allow changes after an approval?).

Also, I'm not sure why the CI is failing. I get 404 when trying to check in Semaphore, but locally tests run just fine.

brianstrauch:
> maybe the repo doesn't allow changes after an approval?

That's definitely not the case... we do that all the time- not sure what's going on.

> I'm not sure why the CI is failing

It's failing due to flaky integration tests that @MuweiHe is looking at right now.

> I get 404 when trying to check in Semaphore

Let me get you and your team added to our public Semaphore project.

MarTango (approved):

PR #1948, opened 2023-06-02
[CLI-2519] Post-release Flink fixes

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Fix feature flag, typos in error messages, and data plane auth token.

Test & Review
-------------
Manually tested

### DIFF
--- a/internal/cmd/flink/command_shell.go
+++ b/internal/cmd/flink/command_shell.go
@@ -21,7 +21,8 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
 			return c.startFlinkSqlClient(prerunner, cmd)
 		},
 	}
-	cmd.Flags().String("compute-pool", "", "Flink compute pool ID.")
+
+	c.addComputePoolFlag(cmd)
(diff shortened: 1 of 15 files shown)


internal/cmd/flink/command_shell.go
```diff
@@ -21,7 +21,8 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
 			return c.startFlinkSqlClient(prerunner, cmd)
 		},
 	}
-	cmd.Flags().String("compute-pool", "", "Flink compute pool ID.")
+
+	c.addComputePoolFlag(cmd)
 	cmd.Flags().String("identity-pool", "", "Identity pool ID.")
```
MuweiHe, line 26: are `compute-pool` and `identity-pool` at similar level? Could make the format consistent (use a `addFlag` too)

MuweiHe (approved):

internal/cmd/flink/command_shell.go
brianstrauch: Identity pool has no autocompletion, and is only used in this function, so we don't need a helper function here yet.

PR #1951, opened 2023-06-04
Update `local kafka start` golden file

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
I noticed this while testing something else.

The `local kafka start` integration test can fail if you don't already have the latest image downloaded, because the string `Image is up to date for confluentinc/confluent-local:latest` is replaced with `Downloaded newer image for confluentinc/confluent-local:latest`.

So the test fails the first time you run it if you didn't run the `confluent local start` command before.

Since this golden file already uses regex, this PR just adds a bit more regex to allow for either string.

References
----------

Test & Review
-------------
Tested on Darwin (I commented out `Skip()`) that both patterns give test success now.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/test/fixtures/output/local/start.golden
+++ b/test/fixtures/output/local/start.golden
@@ -2,6 +2,6 @@ The local commands are intended for a single-node development environment only, 
 
 Pulling from confluentinc/confluent-local
 Digest: sha256:[a-zA-Z0-9]*.
-Status: Image is up to date for confluentinc/confluent-local:latest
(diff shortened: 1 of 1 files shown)


MuweiHe (approved):
Thank you!

PR #1954, opened 2023-06-05
[internal] [confluent local] phase 1 patch: code quality and stability of tests

MuweiHe:
Checklist
---------
   * yes: ok

What
----

Address comments from last pr. improve code stability.
Hardcoded ports for tests (assuming ports > 3000 won't work). Shortened wait time. Ran ~10 times and they all passed.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/local/command_kafka_start.go
+++ b/internal/cmd/local/command_kafka_start.go
@@ -82,14 +82,8 @@ func (c *command) kafkaStart(cmd *cobra.Command, args []string) error {
 
 	log.CliLogger.Tracef("Pull confluent-local image success")
 
-	if c.Config.IsTest {
-		if err := c.prepareAndSaveLocalPortsForTest(); err != nil {
-			return err
-		}
-	} else {
-		if err := c.prepareAndSaveLocalPorts(); err != nil {
-			return err
-		}
+	if err := c.prepareAndSaveLocalPorts(c.Config.IsTest); err != nil {
+		return err
 	}
 
 	if c.Config.LocalPorts == nil {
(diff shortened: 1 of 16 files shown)


internal/cmd/local/command_kafka_start.go
```diff
@@ -82,8 +82,14 @@ func (c *command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 	log.CliLogger.Tracef("Pull confluent-local image success")
 
-	if err := c.prepareAndSaveLocalPorts(); err != nil {
-		return err
+	if c.Config.IsTest {
+		if err := c.prepareAndSaveLocalPortsForTest(); err != nil {
+			return err
+		}
+	} else {
+		if err := c.prepareAndSaveLocalPorts(); err != nil {
```
brianstrauch, line 90: Can we pass `isTest` into this function? Looks like the implementation of these two functions is very similar and could be combined.

internal/cmd/local/command_kafka_start.go
```diff
@@ -134,36 +140,40 @@ func (c *command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+		PlaintextPort:  "2997",
+		BrokerPort:     "2998",
+		ControllerPort: "2999",
 	}
 
-	plaintextPort, err := freeport.GetFreePort()
-	if err != nil {
-		return err
+	if err := c.Config.Save(); err != nil {
+		return errors.Wrap(err, "failed to save local ports to config")
```
brianstrauch, line 156:
```suggestion
		return errors.Wrap(err, "failed to save local ports to configuration file")
```

internal/cmd/secret/command_file.go
```diff
@@ -25,6 +25,10 @@ func (c *command) newFileCommand() *cobra.Command {
 	return cmd
 }
 
+func (c *command) addConfigFileFlag(cmd *cobra.Command) {
+	cmd.Flags().String("config-file", "", "Path to the configuration properties file.")
```
brianstrauch, line 29: Does this file have a specific format? YAML, JSON, key=value pairs, etc? We should let the user know.

test/local_test.go
```diff
@@ -20,7 +20,7 @@ func (s *CLITestSuite) TestLocalKafka() {
 		s.runIntegrationTest(tt)
 	}
 
-	time.Sleep(60 * time.Second)
+	time.Sleep(25 * time.Second)
```
brianstrauch, line 23: Would it make more sense to retry the connection every 10s? I've never seen this issue in integration tests before though... is it because we're using Docker?

brianstrauch: I don't think we should be using Docker in integration tests if we are... we should either mock it or move it to system tests. We shouldn't expect anyone to have to run Docker as a prerequisite to running `make test`... `make test` should just work on its own.

MuweiHe: it's running a real container so it needed a little time to set up the connection... 
you have a good point. For stablizing the tests for now i'll merge this pr first, and work on a mock for docker.

internal/cmd/secret/command_file.go
MuweiHe: updated!

internal/cmd/secret/command_file.go
```diff
@@ -25,6 +25,10 @@ func (c *command) newFileCommand() *cobra.Command {
 	return cmd
 }
 
+func (c *command) addConfigFileFlag(cmd *cobra.Command) {
+	cmd.Flags().String("config-file", "", `Path to the configuration properties file, with extension of "json" or "properties".`)
```
brianstrauch, line 29:
```suggestion
	cmd.Flags().String("config-file", "", `Path to the configuration properties file. File extension must be one of ".json" or ".properties" (key=value pairs).`)
```

brianstrauch (approved):

sgagniere (approved):
Thanks!!

PR #1959, opened 2023-06-06
Fix/kfs 814 fix swallowing characters

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This fixes 3 bugs:
- After exiting the client, it was not possible anymore to use CtrlC and some other commands
- Occasionally the first character a user typed (after already having submitted a statement) got swallowed 
- When using "exit" to close the client, the previous output line got deleted 

To fix the first one we need to save the state of STDIN before setting up go-prompt and the ConsoleParser (both of them mess with the STDIN state). Then when we close the application we need to set the STDIN state to the state we saved before. So now now we save the STDIN state once when we init the InputController and then restore it during tear down of the InputController.

To fix the second bug we need to make sure the tear down method of the consoleParser is not called during the execution of a statement. This is to avoid races between `in.TearDown()` and `cancelListenToUserInput()`, which were causing the character swallowing in some cases (when the tear down happened too early). It's unnecessary to set up and tear down the consoleParser in every iteration anyways, so no we just do it once at the start and at the end of the lifetime of the InputController.

**Note: needed to move the interface definitions to the types folder to avoid an import cycle with the mocks. I also fixed the mock generator.**

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/pkg/flink/internal/controller/application_controller.go
```diff
@@ -76,6 +81,10 @@ func (a *ApplicationController) ShowTableView() {
 	a.app.SetRoot(a.tableView, true).EnableMouse(false)
 }
 
+func (a *ApplicationController) AddCleanupFunction(cleanupFunction func()) {
```
swist, line 84: would be nice to make this a real builder pattern so that you get the same thing as the return e.g.

```suggestion
func (a *ApplicationController) AddCleanupFunction(cleanupFunction func()) *ApplicationController {
a.cleanupFunctions = append(a.cleanupFunctions, cleanupFunction)
return a
```

swist (approved):

internal/pkg/flink/internal/controller/application_controller.go
YannickPferr: Done! Needed to move the controller interface definitions to the types folder though because I got an import cycle in the mock app controller

PR #1962, opened 2023-06-07
Fix base64 decoding error

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The new FedRAMP environment returns refresh tokens that have a different format, so the `v1\..*` regex no longer works. This PR prefixes all encrypted strings with "AES/GCM/NoPadding:" (encryption algorithm name) so we know if a refresh token is encrypted or not. The existence of that prefix + the correct platform name (i.e. "confluentgov.com") should be enough to know if we should encrypt a refresh token or not.

Test & Review
-------------
Manual testing with a FedRAMP test account.

### DIFF
--- a/internal/pkg/config/v1/config.go
+++ b/internal/pkg/config/v1/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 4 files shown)


MuweiHe:
i added the regex because sometimes that token is already encrypted, and some methods still get called again. Have you tried e.g. running multiple login / running commands that require config saving or loading after it's been encrypted?

MuweiHe:
encryptContextStateTokens is being called every time you save the config. And it's likely its been encrypted before: there're many places that called `save()`

brianstrauch:
@MuweiHe the new refresh tokens don't have any distinguishing factors (so we can't use a regular expression anymore). How about we introduce a new field called `isEncrypted` to track if we've encrypted the tokens yet during the current execution?

MuweiHe:
> @MuweiHe the new refresh tokens don't have any distinguishing factors (so we can't use a regular expression anymore). How about we introduce a new field called `isEncrypted` to track if we've encrypted the tokens yet during the current execution?

Or add a prefix; what will work

MuweiHe:
i'm working on encrypting context api-key's. Will hold till you finish this pr;

brianstrauch:
> Or add a prefix

Not a bad idea, I'll try both and pick the best

brianstrauch:
I think I'll go with the prefix idea! If we ever need to change the encryption algorithm in the future, this'll make it much easier to maintain backwards compatibility.

brianstrauch:
> Or add a prefix

This won't work because we won't be able to distinguish between the new unencrypted tokens and the old encrypted tokens :( We'll have to keep track of this with a new variable.

DABH (approved):

PR #1967, opened 2023-06-09
[CLI-2528] panic collection implementation

kevin-wu24:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----
Panic collection for the CLI. Now, users will not see a raw stack dump when a `panic` occurs, and will instead receive a normal error message. The relevant stack frame information is sent to `cc-cli-service`.

References
----------
https://confluentinc.atlassian.net/wiki/spaces/DPI/pages/3085239761/1-Pager+Collecting+panics+in+CLI+command+execution

Test & Review
-------------
Unit testing, will add integration tests, need to test end to end


cmd/confluent/main.go
```diff
@@ -38,6 +40,9 @@ func main() {
 	cmd := pcmd.NewConfluentCommand(cfg)
 
 	if err := pcmd.Execute(cmd, os.Args[1:], cfg); err != nil {
+		if _, ok := err.(*ppanic.Panic); ok {
+			output.ErrPrintln(err.Error())
```
MuweiHe, line 44: will traces info be printed out still?

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// ParseStack formats the stack trace resulting from a panic to only include filenames and line numbers up until panic
+func ParseStack(stack string) *[]string {
+	trace := strings.Split(stack, "\n")
+	if trace[len(trace)-1] == "" {
+		trace = trace[:len(trace)-1]
+	}
+	panicIndex := 0
+	for i := range trace {
```
MuweiHe, line 22:
```suggestion
	for item := range trace {
```

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,36 @@
...
+		trace[i] = strings.TrimSpace(trace[i])
+		if strings.Contains(trace[i], "panic.go") {
+			panicIndex = i
+		}
+	}
+	trace = trace[panicIndex:]
+	var result []string
```
MuweiHe, line 29:
```suggestion
	trace = trace[panicIndex:]
	
	var result []string
```

cmd/confluent/main.go
kevin-wu24: No, only a message that is formatted like the other errors in the CLI (i.e. "Error: ..."). Maybe we should make a note as well that a panic occurred and was sent to cli-service?

cmd/confluent/main.go
```diff
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"github.com/confluentinc/cli/internal/pkg/output"
```
brianstrauch, line 4: Run `make lint`

cmd/confluent/main.go
brianstrauch: I think we'll have to add "Error: " here manually, right?

brianstrauch: Might want to call https://pkg.go.dev/github.com/spf13/cobra#CheckErr

internal/cmd/command.go
```diff
@@ -145,7 +146,20 @@ func NewConfluentCommand(cfg *v1.Config) *cobra.Command {
...
 
-func Execute(cmd *cobra.Command, args []string, cfg *v1.Config) error {
+func Execute(cmd *cobra.Command, args []string, cfg *v1.Config) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			//TODO: need to uncomment before releasing
+			// if !cfg.Version.IsReleased() {
+			//	panic(r)
+			// }
+			err = &ppanic.Panic{ErrorMsg: ppanic.FormatPanicMsg(r)}
```
brianstrauch, line 156: This might be the more idiomatic way to do this.
```suggestion
			err = ppanic.NewError(r)
```

internal/cmd/command.go
```diff
@@ -158,19 +172,25 @@ func Execute(cmd *cobra.Command, args []string, cfg *v1.Config) error {
...
 	}
 
-	err := cmd.Execute()
+	err = cmd.Execute()
 	output.ErrPrint(errors.DisplaySuggestionsMessage(err))
 	u.Error = cliv1.PtrBool(err != nil)
 
+	if err := reportUsage(cmd, cfg, u); err != nil {
+		return err
+	}
```
brianstrauch, line 181: These blocks are unrelated, so let's add some whitespace.
```suggestion
	}
	
```

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,88 @@
...
+
+// CollectPanic collects relevant usage data for when panics occur and command execution is not completed.
+func CollectPanic(cmd *cobra.Command, args []string, cfg *v1.Config) *usage.Usage {
+	u := usage.New(cfg.Version.Version)
+	fullCommand, flags, _ := cmd.Find(args)
+	u.Command = cliv1.PtrString(fullCommand.CommandPath())
+	u.Flags = parseFlags(fullCommand, flags)
+	u.Error = cliv1.PtrBool(true)
+	u.StackFrames = parseStack(string(debug.Stack()))
+	return u
```
brianstrauch, line 34: Let's use the following style here?
```suggestion
	return &usage.Usage{
		Command: ...,
	}
```

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,88 @@
...
+	})
+	return &formattedFlags
+}
+
+// parseStack formats the stack trace resulting from a panic to only include line numbers up until panic
+func parseStack(stack string) *[]string {
+	trace := strings.Split(stack, "\n")
+	if trace[len(trace)-1] == "" {
+		trace = trace[:len(trace)-1]
+	}
```
brianstrauch, line 71: Let's call https://pkg.go.dev/strings#TrimRight before `strings.Split()`?

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,88 @@
...
+		trace[idx] = strings.TrimSpace(trace[idx])
+		if strings.Contains(trace[idx], "panic.go") {
+			panicIndex = idx
+		}
+	}
+	trace = trace[panicIndex:]
+
+	var result []string
+	for _, s := range trace {
+		if strings.Contains(s, ".go") {
```
brianstrauch, line 83: Not sure if this is more accurate, but can we be more precise here? Looks like some of our output contains ".goenv" which is a false positive for this check.
```suggestion
		if strings.HasSuffix(s, ".go") {
```

internal/pkg/panic/panic_test.go
```diff
@@ -0,0 +1,17 @@
+package panic
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseStack(t *testing.T) {
+	rawTrace := "runtime/debug.Stack()\n        ~/.goenv/versions/1.20.4/src/runtime/debug/stack.go:24 +0x65\ngithub.com/confluentinc/cli/internal/cmd.Execute.func1()\n        ~/cli/internal/cmd/command.go:157 +0x198\npanic({0x1019356e0, 0xc0004ba990})\n        ~/.goenv/versions/1.20.4/src/runtime/panic.go:884 +0x213\ngithub.com/confluentinc/cli/internal/cmd.Execute(0xc000004600, {0xc000040050?, 0x101d1b4c0?, 0x28?}, 0xc0002c62c0)\n        ~/cli/internal/cmd/command.go:172 +0x17c\nmain.main()\n        ~/cli/cmd/confluent/main.go:40 +0x3f4\n"
```
brianstrauch, line 10: Can you create a raw string with backticks here, so it's easier to read this? The `\n` are distracting...

internal/pkg/panic/panic.go
kevin-wu24: Going to change it to `Contains(s, ".go:")` since the line number strings are formatted like "file path:line number +0xXXX", which will remove the false positive

internal/cmd/command.go
kevin-wu24: https://pkg.go.dev/github.com/spf13/cobra#CheckErr makes it so we don't have to manually add "Error:" and it also automatically exits with code 1.

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,69 @@
...
+	for i := range flags {
+		flags[i] = strings.TrimPrefix(flags[i], "--")
+		flags[i] = strings.TrimPrefix(flags[i], "-")
+	}
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if slices.Contains(flags, flag.Name) {
+			formattedFlags = append(formattedFlags, flag.Name)
+		} else if slices.Contains(flags, flag.Shorthand) {
+			formattedFlags = append(formattedFlags, flag.Name)
+		}
```
sgagniere, line 44:
```suggestion
		if slices.Contains(flags, flag.Name) || slices.Contains(flags, flag.Shorthand) {
			formattedFlags = append(formattedFlags, flag.Name)
		}
```

Since the body of both branches are the same, we can just combine it into a single if condition.

internal/pkg/panic/panic.go
```diff
@@ -0,0 +1,69 @@
...
+		StackFrames: parseStack(string(debug.Stack())),
+	}
+}
+
+// parseFlags collects the flags alongside the panicking command
+func parseFlags(cmd *cobra.Command, flags []string) *[]string {
+	var formattedFlags []string
+	for i := range flags {
+		flags[i] = strings.TrimPrefix(flags[i], "--")
+		flags[i] = strings.TrimPrefix(flags[i], "-")
```
sgagniere, line 37:
```suggestion
		flags[i] = strings.TrimLeft(flags[i], "-")
```
TrimLeft will remove any number of `-`'s so we don't need to code for both cases.

PR #1968, opened 2023-06-12
KFS-906: chore: remove statement cleanup

YannickPferr:
Checklist
---------
   * yes: ok

What
----

We used to delete statements after their execution to save system resources, which should not be an issue anymore. This change introduces a config variable that allows to switch this behavior back on if we need it, but disables it by default.

### DIFF
--- a/internal/pkg/flink/config/results.go
+++ b/internal/pkg/flink/config/results.go
@@ -3,7 +3,8 @@ package config
 import "time"
 
 const (
-	InitialWaitTime        = time.Millisecond * 300
-	WaitTimeIncrease       = 300
-	DefaultTimeoutDuration = 10 * time.Minute
+	InitialWaitTime         = time.Millisecond * 300
+	WaitTimeIncrease        = 300
+	DefaultTimeoutDuration  = 10 * time.Minute
+	ShouldCleanupStatements = false
 )
--- a/internal/pkg/flink/config/results.go
+++ b/internal/pkg/flink/config/results.go
@@ -3,7 +3,7 @@ package config
 import "time"
 
 const (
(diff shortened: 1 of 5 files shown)


internal/pkg/flink/internal/controller/table_controller.go
```diff
@@ -191,6 +197,7 @@ func (t *TableController) refreshResults(ctx context.Context, statement types.Pr
 		for {
 			select {
 			case <-ctx.Done():
+				go t.store.DeleteStatement(statement.StatementName)
```
Guttz, line 200: Was it necessary to move the delete statement here?

Guttz (approved):

internal/pkg/flink/internal/controller/table_controller.go
YannickPferr: I did this to make sure we are still deleting statements that are in RUNNING state. We don't want to delete COMPLETED statements anymore, but I think we should still delete the ones that are running. What do you think?

YannickPferr: Oh yeah absolutely, good catch!

internal/pkg/flink/config/results.go
```diff
@@ -3,7 +3,8 @@ package config
 import "time"
 
 const (
-	InitialWaitTime        = time.Millisecond * 300
-	WaitTimeIncrease       = 300
-	DefaultTimeoutDuration = 10 * time.Minute
+	InitialWaitTime         = time.Millisecond * 300
```
brianstrauch, line 6:
```suggestion
	InitialWaitTime         = 300 * time.Millisecond
```

brianstrauch (approved):

PR #1971, opened 2023-06-12
feat: allow spaces between quotes for SET statements

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This makes it possible to have spaces in keys and values e.g. `SET 'a key' = 'a value'`. 

It also disallows users to use single quotes inside the key or value string. Any other chars are allowed though, which means one can use double quotes to escape some value: `SET '"a key with quotes"' = '"a value with quotes"'`.

### DIFF
--- a/internal/pkg/flink/internal/store/store_test.go
+++ b/internal/pkg/flink/internal/store/store_test.go
@@ -313,8 +313,8 @@ func (s *StoreTestSuite) TestParseSETStatement() {
 	assert.Equal(s.T(), "value", value)
 
 	key, value, _ = parseSetStatement("set 'key    '= '		va  lue'    ")
-	assert.Equal(s.T(), "key", key)
-	assert.Equal(s.T(), "value", value)
+	assert.Equal(s.T(), "key    ", key)
+	assert.Equal(s.T(), "\t\tva  lue", value)
 
 	key, value, _ = parseSetStatement("set 'key' ='value'    ")
 	assert.Equal(s.T(), "key", key)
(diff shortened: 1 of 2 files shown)


internal/pkg/flink/internal/store/store_test.go
```diff
@@ -363,32 +363,66 @@ func (s *StoreTestSuite) TestParseSETStatement() {
 	key, value, _ = parseSetStatement("set 'key= \nvalue'	")
 	assert.Equal(s.T(), "", key)
 	assert.Equal(s.T(), "", value)
+
+	key, value, _ = parseSetStatement(`set ''key''=''value''`)
```
brianstrauch, line 367: This should error, right? Shouldn't we be testing that there's an error, instead of testing that the strings are empty?

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -363,32 +363,66 @@ func (s *StoreTestSuite) TestParseSETStatement() {
...
+	assert.Equal(s.T(), "", key)
+	assert.Equal(s.T(), "", value)
+
+	key, value, _ = parseSetStatement(`set '"key"'='"value"'`)
+	assert.Equal(s.T(), `"key"`, key)
+	assert.Equal(s.T(), `"value"`, value)
 }
 
-func (s *StoreTestSuite) TestParseSETStatementerror() {
+func (s *StoreTestSuite) TestParseSETStatementError() {
```
brianstrauch, line 376:
```suggestion
func (s *StoreTestSuite) TestParseSetStatementError() {
```

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -363,32 +363,66 @@ func (s *StoreTestSuite) TestParseSETStatement() {
 	key, value, _ = parseSetStatement("set 'key= \nvalue'	")
 	assert.Equal(s.T(), "", key)
 	assert.Equal(s.T(), "", value)
+
+	key, value, _ = parseSetStatement(`set ''key''=''value''`)
+	assert.Equal(s.T(), "", key)
+	assert.Equal(s.T(), "", value)
+
+	key, value, _ = parseSetStatement(`set '"key"'='"value"'`)
```
brianstrauch, line 371: Looks like we're only supporting single quotes for the outermost quote, why not support double quotes too?

internal/pkg/flink/internal/store/store_test.go
brianstrauch: What if I want a single quote in my SET statement? Do I have to backslash it?

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -363,32 +363,66 @@ func (s *StoreTestSuite) TestParseSETStatement() {
...
 
 	_, _, err = parseSetStatement("SET =value")
-	assert.NotNil(s.T(), err)
-	assert.Equal(s.T(), "Error: key not present\nUsage: \"SET 'key'='value'\"", err.Error())
+	assert.Equal(s.T(), &types.StatementError{
+		Message: "key not present",
+		Usage:   []string{"SET 'key'='value'"},
+	}, err)
 
 	_, _, err = parseSetStatement("SET ass=value=as")
```
brianstrauch, line 401: :eyes:

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -171,36 +169,59 @@ func parseSetStatement(statement string) (string, string, error) {
...
-			Message: "key and value must be enclosed by single quotes ''",
+			Message: "key and value must be enclosed by single quotes (')",
 			Usage:   []string{"SET 'key'='value'"},
 		}
 	}
 
-	return strings.ReplaceAll(keyValuePair[0], "'", ""), strings.ReplaceAll(keyValuePair[1], "'", ""), nil
+	// remove quotes
+	key := keyWithQuotes[1 : len(keyWithQuotes)-1]
+	value := valueWithQuotes[1 : len(valueWithQuotes)-1]
```
brianstrauch, line 206: I think `gofmt` should catch this, but:
```suggestion
	key := keyWithQuotes[1:len(keyWithQuotes)-1]
	value := valueWithQuotes[1:len(valueWithQuotes)-1]
```

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Done!

internal/pkg/flink/internal/store/store_test.go
YannickPferr: This is how the open source SQL client does it. Single quotes for the outermost quote and then you can use double quotes if you want quotes inside the key/value.

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Haha oops, I guess that's an unfortunate typo. Will rename!

internal/pkg/flink/internal/store/store_utils.go
YannickPferr: My GoLand does this automatically when saving (and uses gofmt in the background). Running gofmt explicitly formats it that way too. Only when I omit the `-1` does it remove the spaces:
`keyWithQuotes[1:len(keyWithQuotes)]`

Guttz: Same behavior on VS Code using gofmt as formatter

internal/pkg/flink/internal/store/store_test.go
brianstrauch: > You would use a double quote inside the key/value string if you want to have a quote in there.

Sounds like there's no way to have a single quote inside the key/value string, though.

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -291,107 +291,161 @@ func (s *StoreTestSuite) TestIsExitStatement() {
...
-	key, value, _ = parseSetStatement("set key= \n'value'	")
+	key, value, err = parseSetStatement("set 'key= \nvalue'	")
 	assert.Equal(s.T(), "", key)
 	assert.Equal(s.T(), "", value)
+	assert.Error(s.T(), err)
 
-	key, value, _ = parseSetStatement("set 'key= \nvalue'	")
+	key, value, err = parseSetStatement(`set ''key''=''value''`)
 	assert.Equal(s.T(), "", key)
 	assert.Equal(s.T(), "", value)
```
brianstrauch, line 387: I think what I was alluding to earlier is, good Go code shouldn't care what the other values are equal to if a function returns an error. It seems weird to be checking these values if we'll never use them in the case of an error.

brianstrauch (approved):

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Yeah, this wouldn't be allowed. I'll double check tomorrow if there's a way to do it with the open source client!

YannickPferr: Okay you can escape a single quote now with another single quote, that's how you can do it in the OS client as well!

internal/pkg/flink/internal/store/store_test.go
YannickPferr: Oh I see, yeah I guess we could just remove those checks, but they probably don't hurt either.

PR #1972, opened 2023-06-12
KFS-859/feat: use fatih/color for output coloring 

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

We now use a package for coloring text instead of hard coded ansi codes for the flink shell.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/flink/components/interactive_input.go
+++ b/internal/pkg/flink/components/interactive_input.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/fatih/color"
 )
 
 func PrintSmartCompletionState(smartCompletion bool, maxCol int) {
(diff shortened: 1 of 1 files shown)


internal/pkg/flink/components/interactive_input.go
```diff
@@ -34,6 +35,6 @@ func PrintWelcomeHeader() {
 	output.Printf("To exit, press Ctrl-Q or type \"exit\". \n\n")
 
 	// Print shortcuts
-	output.Printf("\033[0m%s \033[0;36m%s \033[0m", "[Ctrl-Q]", "Quit")
-	output.Printf("\033[0m%s \033[0;36m%s \033[0m \n", "[Ctrl-S]", "Toggle Smart Completion")
+	output.Printf("[Ctrl-Q] " + color.CyanString("Quit "))
+	output.Printf("[Ctrl-S] " + color.CyanString("Toggle Smart Completion \n"))
```
brianstrauch, line 39: It feels weird to be coloring whitespace, and you've got a weird trailing space. Format strings are probably the way to go here.
```suggestion
	output.Println("[Ctrl-Q] %s [Ctrl-S] %s", color.CyanString("Quit"), color.CyanString("Toggle Smart Completion"))
```

Guttz: Done :) You prob mean printf and not println though. I got an error/warning when using println with the formatting options.

Guttz: image

brianstrauch: Oops, my bad!

brianstrauch: What's up with the trailing space though?

brianstrauch (approved):

swist (approved):

PR #1973, opened 2023-06-12
[FF-10473] Encrypt api key pair for contexts

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Encrypt API key pair in context credentials

Bug Fixes
- 

Checklist
---------
   * yes: ok

What
----

Encrypt the context api-key at saving, and decrypt it in prerunner. Notice it's done for all credentials, as creating a context doesn't set it as current, so we can't just encrypt the current one.

References
----------

https://confluentinc.atlassian.net/browse/FF-10473

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


brianstrauch:
> Bug Fixes
> 
> Encrypt API key pair in context credentials

Not sure I'd classify this as a bug fix, move this to "New Features"?

internal/pkg/config/v1/api_key_pair.go
```diff
@@ -1,7 +1,28 @@
...
+
 // APIKeyPair holds an API Key and Secret.
 type APIKeyPair struct {
 	Key    string `json:"api_key"`
 	Secret string `json:"api_secret"`
+	Salt   []byte `json:"salt,omitempty"`
+	Nonce  []byte `json:"nonce,omitempty"`
+}
+
+func (c *APIKeyPair) DecryptCredentialAPISecret() error {
```
brianstrauch, line 18: The field is named "Secret", so it makes sense to call this function "DecryptSecret"
```suggestion
func (c *APIKeyPair) DecryptSecret() error {
```

internal/pkg/config/v1/config.go
```diff
@@ -154,6 +154,19 @@ func (c *Config) DecryptContextStates() error {
...
 }
 
+func (c *Config) DecryptCredentials() error {
+	if credentials := c.Credentials; c.Credentials != nil {
+		for _, credential := range credentials {
+			if apiKeyPair := credential.APIKeyPair; apiKeyPair != nil {
+				if err := apiKeyPair.DecryptCredentialAPISecret(); err != nil {
+					return err
+				}
+			}
```
brianstrauch, line 164: No reason to create another variable, IMO
```suggestion
			if credential.APIKeyPair != nil {
				if err := credentialAPIKeyPair.DecryptCredentialAPISecret(); err != nil {
					return err
				}
			}
```

internal/pkg/config/v1/config.go
```diff
@@ -217,6 +230,7 @@ func (c *Config) Save() error {
 	tempContext := c.resolveOverwrittenContext()
 	var tempAuthToken string
 	var tempAuthRefreshToken string
+	tempAPISecrets := map[string]string{}
```
brianstrauch, line 233:
```suggestion
	tempApiSecrets := map[string]string{}
```

internal/pkg/config/v1/config.go
```diff
@@ -250,10 +275,40 @@ func (c *Config) Save() error {
 	c.restoreOverwrittenKafka(tempKafka)
 	c.restoreOverwrittenAuthToken(tempAuthToken)
 	c.restoreOverwrittenAuthRefreshToken(tempAuthRefreshToken)
+	c.restoreOverwrittenCredentials(tempAPISecrets)
 
 	return nil
 }
 
+func (c *Config) encryptCredentialsAPISecret(tempAPISecrets map[string]string) error {
```
brianstrauch, line 283: Looks like we're doing more than just encrypting the API secret

internal/pkg/config/v1/config.go
```diff
@@ -326,6 +381,15 @@ func (c *Config) restoreOverwrittenAuthRefreshToken(tempAuthRefreshToken string)
 	}
 }
 
+func (c *Config) restoreOverwrittenCredentials(tempAPISecrets map[string]string) {
+	for name, secret := range tempAPISecrets {
+		if secret != "" {
+			c.Credentials[name].APIKeyPair.Secret = secret
+		}
+	}
+
```
brianstrauch, line 390:
```suggestion
```

internal/pkg/config/v1/config_test.go
```diff
@@ -101,10 +104,7 @@ func SetupTestInputs(isCloud bool) *TestInputs {
 			Name:      "anonymous-cluster",
 			Bootstrap: "http://test",
 			APIKeys: map[string]*APIKeyPair{
```
brianstrauch, line 106: Collapse into one line?

internal/pkg/config/v1/config.go
MuweiHe: it's generating salt and nonce if nil and then encrypt the secret

brianstrauch: Let's give this function a more general name to reflect that?

MuweiHe: I think it's expected to generate nonce and salt during encryption though. What's your suggestion?

brianstrauch: Oh, I didn't fully understand what's going on here. I thought we were in a different function. Can we move the nonce/salt generation to a separate function so we don't have duplicated code?

MuweiHe: refactored into a separate function to generate salt and nonce!

internal/pkg/config/v1/api_key_pair.go
```diff
@@ -1,7 +1,28 @@
...
+	"runtime"
+	"strings"
+
+	"github.com/confluentinc/cli/internal/pkg/secret"
+)
+
 // APIKeyPair holds an API Key and Secret.
 type APIKeyPair struct {
 	Key    string `json:"api_key"`
 	Secret string `json:"api_secret"`
```
brianstrauch, line 13: These should probably be `omitempty` so the configuration file looks a little cleaner.

internal/pkg/config/v1/config.go
```diff
@@ -154,6 +154,19 @@ func (c *Config) DecryptContextStates() error {
 	return c.Validate()
 }
 
+func (c *Config) DecryptCredentials() error {
```
brianstrauch, line 157: This seems a bit wasteful. I don't think we'd ever use all of the stored API keys in a single command, right?

MuweiHe: I think it's ok for apikeys for contexts. the thing is we might need to encrypt/decrypt context apikeys even when they're not the active/current one. But you have a good point. For stored kafka-cluster api-keys, I'll only decrypt them when needed.

internal/pkg/cmd/prerunner.go
```diff
@@ -309,6 +313,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 			return err
 		}
 
+		if err := r.Config.DecryptCredentials(); err != nil {
```
sgagniere, line 316: If this ends up staying in the prerunner (as opposed to being called before use like Brian suggested) then this call (and the one in AuthenticatedWithMDS below) might be redundant since both of these call `r.Anonymous`, which itself calls `r.Config.DecryptCredentials`.

MuweiHe: thanks for catching!

internal/pkg/config/v1/config.go
brianstrauch: > we might need to encrypt/decrypt context apikeys even when they're not the active/current one

Really? What's an example of that?

internal/cmd/asyncapi/command_export.go
```diff
@@ -333,6 +333,12 @@ func (c *command) getClusterDetails(details *accountDetails, flags *flags) error
 	if err != nil {
 		return fmt.Errorf(`failed to find Kafka cluster: %v`, err)
 	}
+
+	err = clusterConfig.DecryptAPIKeys()
+	if err != nil {
```
brianstrauch, line 338:
```suggestion
	if err := clusterConfig.DecryptAPIKeys(); err != nil {
```

internal/pkg/config/v1/api_key_pair.go
```diff
@@ -1,7 +1,47 @@
...
-	Key    string `json:"api_key"`
-	Secret string `json:"api_secret"`
+	Key    string `json:"api_key,omitempty"`
+	Secret string `json:"api_secret,omitempty"`
+	Salt   []byte `json:"salt,omitempty"`
+	Nonce  []byte `json:"nonce,omitempty"`
+}
+
+func (c *APIKeyPair) DecryptSecret() error {
+	if c.Secret != "" && strings.HasPrefix(c.Secret, secret.AesGcm) && (c.Salt != nil || runtime.GOOS == "windows") {
```
brianstrauch, line 19:
```suggestion
	if strings.HasPrefix(c.Secret, secret.AesGcm) && (c.Salt != nil || runtime.GOOS == "windows") {
```

internal/pkg/config/v1/api_key_pair.go
```diff
@@ -1,7 +1,47 @@
...
+	if c.Salt == nil || c.Nonce == nil {
+		salt, nonce, err := secret.GenerateSaltAndNonce()
+		if err != nil {
+			return err
+		}
+		c.Salt = salt
+		c.Nonce = nonce
+	}
+
+	if !strings.HasPrefix(c.Secret, secret.AesGcm) {
```
brianstrauch, line 39: Do we need to check this, or is it possible that secrets are encrypted twice?

internal/pkg/config/v1/config.go
MuweiHe: e.g. creating a new api key for kafka cluster

internal/cmd/asyncapi/command_export.go
```diff
@@ -333,6 +333,11 @@ func (c *command) getClusterDetails(details *accountDetails, flags *flags) error
 	if err != nil {
 		return fmt.Errorf(`failed to find Kafka cluster: %v`, err)
 	}
+
+	if err := clusterConfig.DecryptAPIKeys(); err != nil {
```
brianstrauch, line 337:
```suggestion
	if err := clusterConfig.DecryptApiKeys(); err != nil {
```

brianstrauch: Did we consider only decrypting the API keys we use?

internal/cmd/kafka/command_clientconfig_create.go
```diff
@@ -205,7 +205,10 @@ func (c *createCommand) setKafkaCluster(cmd *cobra.Command, configFile string) (
 			return "", err
 		}
 	}
-
+	err = kafkaCluster.DecryptAPIKeys()
+	if err != nil {
```
brianstrauch, line 209:
```suggestion
	if err := kafkaCluster.DecryptAPIKeys(); err != nil {
```

internal/cmd/kafka/confluent_kafka_configs.go
```diff
@@ -36,11 +40,14 @@ func getCommonConfig(kafka *configv1.KafkaClusterConfig, clientID string) *ckafk
 		"bootstrap.servers":                     kafka.Bootstrap,
 		"sasl.username":                         kafka.APIKey,
 		"sasl.password":                         kafka.APIKeys[kafka.APIKey].Secret,
-	}
+	}, nil
```
brianstrauch, line 43: This formatting looks a bit weird. Create a variable for the ConfigMap?

internal/pkg/auth/auth.go
```diff
@@ -154,15 +149,10 @@ func addOrUpdateContext(config *v1.Config, isCloud bool, credentials *Credential
 		}
 	}
 
-	stateSalt, err := secret.GenerateRandomBytes(secret.SaltLength)
+	stateSalt, stateNonce, err := secret.GenerateSaltAndNonce()
```
brianstrauch, line 152:
```suggestion
	salt, nonce, err := secret.GenerateSaltAndNonce()
```

internal/pkg/config/v1/config_test.go
```diff
@@ -26,17 +26,21 @@ import (
 const (
 	authTokenPlaceholder        = "AUTH_TOKEN_PLACEHOLDER"
 	authRefreshTokenPlaceholder = "AUTH_REFRESH_TOKEN_PLACEHOLDER"
-	saltPlaceholder             = "SALT_PLACEHOLDER"
-	noncePlaceholder            = "NONCE_PLACEHOLDER"
+	saltPlaceholder             = "SALT_TOKEN_PLACEHOLDER"
+	noncePlaceholder            = "NONCE_TOKEN_PLACEHOLDER"
+	apiSecretPlaceholder        = "API_SECRET_PLACEHOLDER"
+	apiSaltPlaceholder          = "c2FsdHBsYWNlaG9sZGVy"
+	apiNoncePlaceholder         = "bm9uY2VwbGFjZWhvbGRlcg=="
```
brianstrauch, line 33: Should we be using actual values here, or can we use UPPERCASE_STRINGS like the other placeholder constants?

brianstrauch (approved):

internal/cmd/asyncapi/command_export.go
MuweiHe: i think here multiple keys might be required depending on the conditions in following lines

internal/pkg/auth/auth.go
MuweiHe: it's been used in the scope

internal/pkg/config/v1/config_test.go
MuweiHe: somehow they're illegal in base64, so i used the encoded version of them

PR #1974, opened 2023-06-12
[cli-2532] fix npd panic when schema registry cluster info is not ready in config

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix bug in getting schema registry cluster information when it's not present in configuration file

Checklist
---------
   * yes: ok 

What
----

Check if sr cluster is nil. If user had deleted sr using cli, it would be nil and cause a panic when trying to get the cluster id.

References
----------

https://confluent.slack.com/archives/C9Y6NAM6X/p1686557957247439

Test & Review
-------------

Delete sr cluster in cli. Create api-key will show `resource not found`. Enable sr in gui, and use the right lsrc id to create api-key, sr cluster info will be updated.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/dynamic-config/dynamic_context.go
+++ b/internal/pkg/dynamic-config/dynamic_context.go
@@ -148,7 +148,7 @@ func (d *DynamicContext) SchemaRegistryCluster(cmd *cobra.Command) (*v1.SchemaRe
 	var clusterChanged bool
 	if resourceType == presource.SchemaRegistryCluster {
 		for _, srCluster := range d.SchemaRegistryClusters {
-			if srCluster.Id == resource {
+			if srCluster != nil && srCluster.Id == resource {
 				cluster = srCluster
 			}
 		}


internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -148,7 +148,7 @@ func (d *DynamicContext) SchemaRegistryCluster(cmd *cobra.Command) (*v1.SchemaRe
 	var clusterChanged bool
 	if resourceType == presource.SchemaRegistryCluster {
 		for _, srCluster := range d.SchemaRegistryClusters {
-			if srCluster.Id == resource {
+			if srCluster != nil && srCluster.Id == resource {
```
brianstrauch, line 151: `srCluster.GetId()`?

brianstrauch: Did you do a global check for other instances of this pattern?

brianstrauch:
> Fix bug in getting schema registry cluster information when it's not present in configuration file

Prevent panic while retrieving Schema Registry cluster information

brianstrauch:
A better long-term fix might be to stop storing Schema Registry information in the configuration file entirely, and fetch it from the relevant API endpoint when we need it.

MuweiHe:
> A better long-term fix might be to stop storing Schema Registry information in the configuration file entirely, and fetch it from the relevant API endpoint when we need it.

User can only delete sr from cli, not in gui, so the approach we have now works. what you suggest will be more safe still. Wdyt?

brianstrauch:
> User can only delete sr from cli, not in gui, so the approach we have now works. what you suggest will be more safe still. Wdyt?

:+1: Can you open a ticket or make a follow-up PR for removing cached Schema Registry information from the configuration file?

internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -148,7 +148,7 @@ func (d *DynamicContext) SchemaRegistryCluster(cmd *cobra.Command) (*v1.SchemaRe
 	var clusterChanged bool
 	if resourceType == presource.SchemaRegistryCluster {
 		for _, srCluster := range d.SchemaRegistryClusters {
-			if srCluster != nil && srCluster.Id == resource {
+			if srCluster != nil && srCluster.GetId() == resource {
```
brianstrauch, line 151:
```suggestion
			if srCluster.GetId() == resource {
```

brianstrauch (approved):

PR #1988, opened 2023-06-13
[internal] debugf information about getting creds from keychain

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

Got one complain where user changed his password and not updated in keychain, and couldn't log in using that saved password. Our login handler can't really tell if the password has been updated but we could tell where we're getting the creds from.

References
----------

https://confluent.slack.com/archives/C010Y0EP5MZ/p1686659584367629

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/errors/strings.go
+++ b/internal/pkg/errors/strings.go
@@ -18,6 +18,8 @@ const (
 		StopNonInteractiveMsg + ".\n"
 	FoundNetrcCredMsg = "Found credentials for user \"%s\" from netrc file \"%s\" " +
 		StopNonInteractiveMsg + ".\n"
+	FoundKeychainCredMsg = "Found credentials for user \"%s\" from keychain " +
+		StopNonInteractiveMsg + ".\n"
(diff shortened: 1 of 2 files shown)


internal/pkg/errors/strings.go
```diff
@@ -18,6 +18,8 @@ const (
 		StopNonInteractiveMsg + ".\n"
 	FoundNetrcCredMsg = "Found credentials for user \"%s\" from netrc file \"%s\" " +
 		StopNonInteractiveMsg + ".\n"
+	FoundKeychainCredMsg = "Found credentials for user \"%s\" from keychain " +
+		StopNonInteractiveMsg + ".\n"
```
brianstrauch, line 22: Does the ".\n" belong in the variable instead of outside of it?

MuweiHe: Following the convention here i think it belongs to the var

brianstrauch (approved):

PR #1990, opened 2023-06-14
Chore/kfs 916 stop results fetching on error

YannickPferr:
Checklist
---------
   * yes: ok

What
----

- removed a workaround that did not stop results fetching when an error occurred (this is fixed in the backend now, so we can stop fetching when an error happens)
- added tests for result fetching loop
- added a wrapper that enables running test cases with a real tview app in headless mode

### DIFF
--- a/internal/pkg/flink/internal/controller/table_controller_test.go
+++ b/internal/pkg/flink/internal/controller/table_controller_test.go
@@ -1,10 +1,12 @@
 package controller
 
 import (
-	"github.com/confluentinc/cli/internal/pkg/flink/components"
-	"github.com/confluentinc/cli/internal/pkg/flink/internal/results"
-	"github.com/confluentinc/cli/internal/pkg/flink/test/mock"
-	"github.com/confluentinc/cli/internal/pkg/flink/types"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
 	"github.com/gdamore/tcell/v2"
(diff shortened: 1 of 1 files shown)


internal/pkg/flink/internal/controller/table_controller_test.go
```diff
@@ -370,3 +386,61 @@ func (s *TableControllerTestSuite) TestDefaultInRowView() {
...
+		// Given
+		table := components.CreateTable()
+		mockStatement := types.ProcessedStatement{PageToken: "NOT_EMPTY"}
+		tableController := TableController{
+			table:         table,
+			appController: s.mockAppController,
+			store:         s.mockStore,
+		}
+		tableController.SetRunInteractiveInputCallback(s.mockInputController.RunInteractiveInput)
+		s.mockAppController.EXPECT().TView().Return(tview).Times(4)
```
Guttz, line 430: where does the 4 comes from? is it a default value? can we import and use it here?

YannickPferr: I counted this myself lol. We expect to have two calls to this `t.appController.TView().QueueUpdateDraw(t.renderTable)`. The first is in the default case of the results fetch loop and then the second one happens when the context is cancelled an we enter the `ctx.Done()` case. Each of those calls counts as 2 calls to `TView()` because t.renderTable also calls `TView()` once.

I could declare it as a variable but you would still need to count this again yourself if the amount of calls changes in the future.

PR #1991, opened 2023-06-14
fix: token refresh not working

YannickPferr:
Checklist
---------
   * yes: ok

What
----

The token refresh in the SQL client was not working correctly. If the client was left open for >= 15mins, the JWT would expire causing errors in subsequent statements. To fix it, you would need to close and reopen the SQL client.

In order for the token refresh to work, we need to first refresh the control plane token and then swap it for a data plane token. This token can then be used in the GatewayClient.

### DIFF
--- a/internal/cmd/flink/command_shell.go
+++ b/internal/cmd/flink/command_shell.go
@@ -41,9 +41,9 @@ func (c *command) authenticated(authenticated func(*cobra.Command, []string) err
 			return err
 		}
 
-		auth := cfg.Context().State.AuthToken
+		authToken := cfg.Context().State.AuthToken
 		authRefreshToken := cfg.Context().State.AuthRefreshToken
-		if err := c.Context.UpdateAuthTokens(auth, authRefreshToken); err != nil {
+		if err := c.Context.UpdateAuthTokens(authToken, authRefreshToken); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


internal/pkg/ccloudv2/flink_gateway.go
```diff
@@ -37,6 +37,20 @@ func NewFlinkGatewayClient(url, userAgent string, unsafeTrace bool, authToken st
 	}
 }
 
+func (c *FlinkGatewayClient) SetAuthToken(authToken string) {
+	if c == nil {
```
brianstrauch, line 41: I don't think the client should ever be `nil`. We always fetch the client or encounter an error first. How about making this variable public and omitting the getter/setter?

internal/cmd/flink/command_shell.go
```diff
@@ -32,18 +34,41 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
...
+
+		if err := authenticated(cmd, nil); err != nil {
+			return err
+		}
+
+		flinkGatewayClient, err := c.GetFlinkGatewayClient()
+		if err != nil {
+			return err
+		}
+		jwtToken := flinkGatewayClient.GetAuthToken()
```
brianstrauch, line 58:
```suggestion
		authToken := flinkGatewayClient.GetAuthToken()
```

internal/pkg/ccloudv2/flink_gateway.go
YannickPferr: Yeah sounds good, made it public now!

internal/cmd/flink/command_shell.go
YannickPferr: Removed this line completely!

internal/cmd/flink/command_shell.go
```diff
@@ -32,18 +34,40 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
...
+
+		flinkGatewayClient, err := c.GetFlinkGatewayClient()
+		if err != nil {
 			return err
 		}
-		return authenticated(cmd, nil)
+
+		jwtCtx := &v1.Context{State: &v1.ContextState{AuthToken: flinkGatewayClient.AuthToken}}
+		if tokenErr := jwtValidator.Validate(jwtCtx); tokenErr != nil {
+			output.Println("Token expired. Refreshing token...")
```
luca-filipponi, line 61: is this something we want to output to customer?

YannickPferr: I thought it could be helpful, but could also delete this. But thanks to your comment I noticed another thing: we would be printing the error twice. Once in line 64 and then again in input_controller line 84!

luca-filipponi (approved):

brianstrauch (approved):
Approving since force push permissions were removed for the cloud surfaces team, investigating why that happened today.

PR #1994, opened 2023-06-15
Feat/kfs 910 add tests for main loop

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Attempt to test the main application loop by mocking go-prompt and piping STDOUT to string and doing assertions on that.

Test coverage of input controller before:
image

Test coverage of input controller now:
image

**Note: added tests commit after commit, so it is probably easiest to review it like that as well**


swist (approved):

internal/pkg/flink/internal/controller/input_controller_test.go
```diff
@@ -1,35 +1,80 @@
...
+
+func TestInputControllerTestSuite(t *testing.T) {
+	suite.Run(t, new(InputControllerTestSuite))
+}
+
+func (s *InputControllerTestSuite) runAndCaptureSTDOUT(test func()) string {
+	// Redirect STDOUT to a buffer
+	old := os.Stdout
+	r, w, _ := os.Pipe()
+	os.Stdout = w
```
swist, line 52: nit for later: does go prompt allow you to inject stdout in any way?

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -195,20 +195,20 @@ func renderMsgAndStatus(processedStatement *types.ProcessedStatement) {
 	}
 
 	if processedStatement.IsLocalStatement {
-		if processedStatement.Status != "FAILED" {
-			output.Println("Statement successfully submitted.\n ")
+		if processedStatement.Status == "FAILED" {
+			output.Println("Error: Couldn't process statement. Please check your statement and try again")
 		} else {
-			output.Println("Error: Couldn't process statement. Please check your statement and try again.")
+			output.Println("Statement successfully submitted.")
```
swist, line 210: oh yess, negative assertions are a smell

internal/pkg/flink/internal/controller/input_controller_test.go
YannickPferr: Not easily at least, I think you would have to implement a custom version of the `ConsoleWriter` interface and then pass that when constructing your Prompt struct: https://github.com/confluentinc/go-prompt/blob/master/option.go#L295

PR #1996, opened 2023-06-16
Remove `pipeline save` feature flag

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The `confluent pipeline save` command has been enabled for a while, so we should be ok to remove the feature flag to give the command some more stability in case of a LaunchDarkly outage.


sgagniere (approved):
The PR description only mentions `save`; the change to `create` and `update` follows the same reasoning, right?

PR #2000, opened 2023-06-20
feat: use different colors for different types of information

Guttz:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

Product wanted us to use different colors in the client for different purposes. Erros should be red, info white and and warnings yellow.

This is now also configurable from a central file.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/flink/config/color.go
+++ b/internal/pkg/flink/config/color.go
@@ -1,6 +1,14 @@
 package config
 
-import "github.com/confluentinc/go-prompt"
+import (
+	"github.com/confluentinc/go-prompt"
+	"github.com/fatih/color"
+)
 
-// We should use this package as a root for coloring variables
-const HighlightColor = prompt.Cyan
+// Package used as a root for color used accross the application
+const PromptAccentColor = prompt.Cyan
+const AccentColor = color.FgCyan
+
+const InfoColor = color.FgWhite
+const ErrorColor = color.FgHiRed
+const WarnColor = color.FgHiYellow
--- a/internal/pkg/flink/config/color.go
+++ b/internal/pkg/flink/config/color.go
@@ -1,14 +1,15 @@
 package config
 
 import (
-	"github.com/confluentinc/go-prompt"
 	"github.com/fatih/color"
+
+	"github.com/confluentinc/go-prompt"
 )
 
-// Package used as a root for color used accross the application
+// Package used as a root for color used across the application
 const PromptAccentColor = prompt.Cyan
 const AccentColor = color.FgCyan
 
 const InfoColor = color.FgWhite
 const ErrorColor = color.FgHiRed
-const WarnColor = color.FgHiYellow
+const WarnColor = color.FgCyan
--- a/internal/pkg/flink/config/color.go
+++ b/internal/pkg/flink/config/color.go
@@ -12,4 +12,4 @@ const AccentColor = color.FgCyan
 
 const InfoColor = color.FgWhite
 const ErrorColor = color.FgHiRed
-const WarnColor = color.FgCyan
+const WarnColor = color.FgHiYellow
--- a/internal/pkg/flink/internal/controller/input_controller_test.go
+++ b/internal/pkg/flink/internal/controller/input_controller_test.go
@@ -204,7 +204,7 @@ func (s *InputControllerTestSuite) TestRenderMsgAndStatusLocalStatements() {
 		{
 			name:      "local failed statement",
 			statement: &types.ProcessedStatement{IsLocalStatement: true, Status: types.FAILED},
-			want:      "Error: Couldn't process statement. Please check your statement and try again.\n",
+			want:      "Error: couldn't process statement, please check your statement and try again\n",
 		},
 		{
 			name:      "local non-failed statement",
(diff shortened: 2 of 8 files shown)


internal/pkg/flink/config/color.go
```diff
@@ -1,6 +1,15 @@
...
+	"github.com/confluentinc/go-prompt"
+)
+
+// Package used as a root for color used across the application
+const PromptAccentColor = prompt.Cyan
+const AccentColor = color.FgCyan
+
+const InfoColor = color.FgWhite
+const ErrorColor = color.FgHiRed
+const WarnColor = color.FgCyan
```
brianstrauch, line 15: Can you move these constants into `pkg/color`? We've been talking about doing this for the whole CLI, and these would be the colors we need.

brianstrauch: > Erros should be red, info white and and warnings yellow.

This doesn't seem consistent by the way- did you mean to use yellow instead of cyan?

internal/pkg/flink/internal/controller/input_controller_test.go
```diff
@@ -204,7 +204,7 @@ func (s *InputControllerTestSuite) TestRenderMsgAndStatusLocalStatements() {
 		{
 			name:      "local failed statement",
 			statement: &types.ProcessedStatement{IsLocalStatement: true, Status: types.FAILED},
-			want:      "Error: Couldn't process statement. Please check your statement and try again\n",
+			want:      "Error: Couldn't process statement. Please check your statement and try again.\n",
```
brianstrauch, line 207: Errors in Go should not be capitalized or end with punctuation
```suggestion
			want:      "Error: couldn't process statement, please check your statement and try again\n",
```

brianstrauch: The second half of this error seems like more of a suggestion than an error. I believe @YannickPferr wrote some code to separate those.

internal/pkg/flink/internal/controller/input_controller_utils.go
```diff
@@ -40,3 +43,23 @@ func restoreStdin(state *term.State) {
...
+
+func outputInfof(s string, args ...any) {
+	c := color.New(config.InfoColor)
+	output.Printf(c.Sprint(s), args...)
+}
+
+func outputWarn(s string) {
+	c := color.New(config.WarnColor)
+	output.Println(c.Sprint(s))
+}
```
brianstrauch, line 65: Can you integrate this with the existing CLI logger? We have support for INFO/DEBUG/WARN/ERROR, which is filtered with the `-v` flags.

brianstrauch: For now, these functions could just call the CLI logger functions

internal/pkg/flink/config/color.go
Guttz: Ops, I typed the wrong color. Thanks

internal/pkg/flink/internal/controller/input_controller_utils.go
Guttz: It makes sense to have a central cli logger but I don't think it makes sense merging these two tho.

The thing here is, even though I named this "info", "warning" and "errors", there's are no use cases where the user don't want to see all infos these. It was mostly about coloring here and not level of importance. All these infos are essential for the SQL Client user. See the image below:

image

The user needs always need all information in white, red and yellow.

Also, even if we somehow do think it makes sense to merge these, I wouldn't want to do this as part of this PR/part of this jira ticket, which is only related to adding different colors to different types of message. I would prefert to do this in interactive steps and create another ticket :)

internal/pkg/flink/internal/controller/input_controller_test.go
Guttz: > Errors in Go should not be capitalized or end with punctuation

Yeah, that's the default behavior. But I think there are multiple reasons for us to format it in the regular human readable way: 

- We just display errors sent from the backend and they format it properly there so we would have inconsistent formatting for local/remote errors:

image


- Also, we're building the UI to end users and I think it makes sense to format it in a more readable way. It's not like a log file that's going to be read sometimes from a dev. If only wrote local errors in lowercase, we would have `error: couldn't process statement, please check your statement and try again` and info and warnings right next to it properly capitalized `Statement successfully submitted`, which wouldn't also really look consistent..

YannickPferr (approved):
LGTM, but I would like to keep the error messages consistent with the rest of the code, which currently means lower case and without punctuation. I think @Guttz has a point though and we could discuss this offline and then possibly create a follow-up ticket to change them back to uppercase etc.

internal/pkg/flink/internal/controller/input_controller_test.go
Guttz: I've reverted, so the errors are now lowercase so we merge this for now. But I still think it's makes sense to format this properly but we can do that later :)

>The second half of this error seems like more of a suggestion than an error. I believe @YannickPferr wrote some code to separate those.

I've used the logic Yannick added and adjusted the messages. Think having` "Suggestion: check your statement and try again"` isn't really helpful so I kept that inside the error.

PR #2002, opened 2023-06-20
Consolidate dataplane token code

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Use the same function for all dataplane token requests. This sets us up for storing and refreshing the dataplane token in a later PR.

### DIFF
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -353,13 +353,13 @@ func (r *PreRun) setCCloudClient(c *AuthenticatedCLICommand) error {
 				return nil, err
 			}
 
-			bearerToken, err := pauth.GetDataplaneToken(state, ctx.Platform.Server, map[string]any{"clusterIds": lkc})
+			dataplaneToken, err := pauth.GetDataplaneToken(state, ctx.Platform.Server, map[string]any{"clusterIds": lkc})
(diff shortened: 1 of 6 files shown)


internal/pkg/cmd/prerunner.go
```diff
@@ -149,7 +149,7 @@ func (c *AuthenticatedCLICommand) GetFlinkGatewayClient() (*ccloudv2.FlinkGatewa
 			return nil, err
 		}
 
-		authToken, err := pauth.GetJwtTokenForV2Client(ctx.GetState(), ctx.GetPlatformServer())
+		authToken, err := pauth.GetDataplaneToken(ctx.GetState(), ctx.GetPlatformServer(), map[string]any{})
```
sgagniere, line 152: Could we replace `map[string[]any{}` with `nil`?

brianstrauch: If it gets encoded to `{}` in JSON, then sure. Let me investigate!

brianstrauch: Just checked and it doesn't work, we need to use an empty map

sgagniere: Well, that's unfortunate.

sgagniere (approved):

PR #2004, opened 2023-06-21
Fix dataplane token flow

brianstrauch:
Checklist
---------
   * yes: ok

What
----
#2002 passed `"clusterIds": "<cluster-id>"` instead of `"clusterIds": {"<cluster-id>"}` for Kafka REST, resulting in a backend error. I'm awaiting confirmation, but it looks like the body is no longer needed, so I deleted it entirely! Also, the header `"content": "application/json"` appears not to be needed as well.

Test & Review
-------------
Verified that system tests are fixed, also verified manually.


sgagniere (approved):
> The body is no longer needed, so I deleted it entirely!

Well that simplifies things!

MuweiHe (approved):

PR #2007, opened 2023-06-22
feat: add demo mode that creates fake data

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This adds a flag `--fake-gateway` if `cfg.isTest` is set to true. With this flag you can start the client with a fake gateway that returns fake data. This is to make it easier to test changes to the client without the need for submitting real requests to the Gateway.

There are currently two queries: `static;` and `dynamic;`. The former brings up a result table that is complete and does not refresh, while the latter simulates a updated counter that refreshes the table.


swist (approved):

internal/cmd/flink/command_shell.go
```diff
@@ -69,6 +73,19 @@ func (c *command) authenticated(authenticated func(*cobra.Command, []string) err
 }
 
 func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Command) error {
+	// if the --fake-gateway flag is set, we start the client with a simulated gateway client that returns fake data
```
brianstrauch, line 76: Our other commands use the integration test framework in the `test/` directory to accomplish the same thing. Have you tried this yet?

YannickPferr: Yeah we've looked at that. I don't think it will help us in this particular case though. We don't want to run a test case here, rather we want to open the client as usual, but instead of making real backend calls we create fake data randomly. Then you can actually check how the table view looks and how the navigation feels etc.

For integration tests though, we might be able to use your framework. I think we would have to create another fake backend that returns non-random data, so we can create the golden file more easily. Only thing I'm unsure about is, how we can issue statements inside the client. I saw that you are using `exec.Command` to run the commands, but I'm not sure if this also works once you are inside the client, and want to write something into the go-prompt buffer.

Guttz: We need this here for faster development, not for testing. Devel/stag is often unstable, either not available or slow and we're often working on features (like coloring), where we want to start the client with a mocked backend so we test stuff.

internal/pkg/flink/test/mock/gateway_client_fake.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	flinkgatewayv1alpha1 "github.com/confluentinc/ccloud-sdk-go-v2/flink-gateway/v1alpha1"
+	"github.com/confluentinc/cli/internal/pkg/ccloudv2"
+	"github.com/confluentinc/cli/internal/pkg/flink/test/generators"
+	"github.com/google/uuid"
+	"pgregory.net/rapid"
+	"time"
+)
+
+const (
```
Guttz, line 13: Can you either add some documentation on all fake-gateway features/modes here or to contributing.md?

Or to a local contributing only inside flink package

YannickPferr: Good idea! I want to avoid having another file in there that needs to be maintained though, so I added some feature descriptions in the comments above the constants 👍.

PR #2011, opened 2023-06-23
Feat/kfs 894 pagination

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This change brings pagination to the client. You can now fetch the next page manually by pressing 'N' and you can go to the latests result/page by pressing 'R'. Toggle auto refresh is now on 'P'. We now also have a fast scroll option that scrolls multiple rows at once (number of rows to scroll is equal to the number of rows your current screen can display). Press 'H' for scrolling up and 'L' for scrolling down.

In addition I've refactored the table controller and split it into two. TableController now does only rendering and handling the table while the new FetchController handles everything regarding data fetching.


Guttz:
Let's adjust the text for quick scroll according to the OS :) You can check it like this: 

```
if runtime.GOOS == "windows"..

if runtime.GOOS == "darwin" ..

if runtime.GOOS == "linux" ..
```

YannickPferr:
> Let's adjust the text for quick scroll according to the OS :) You can check it like this:
> 
> ```
> if runtime.GOOS == "windows"..
> 
> if runtime.GOOS == "darwin" ..
> 
> if runtime.GOOS == "linux" ..
> ```

I actually removed this completely now, because it turns out TView doesn't capture the modifier keys in the mac terminal. I am just using regular keys now. H to scroll up and L to scroll down. Live Results is now on R and toggle refresh is P.

Guttz (approved):

PR #2014, opened 2023-06-23
[CLI-2369] Add `plugin search` and `plugin install` commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add new commands `confluent plugin search` and `confluent plugin install` to list and install available plugins from the official Confluent CLI plugin repository.

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
This initial implementation performs a shallow clone of the `confluentinc/cli-plugins` repository for both commands. The shallow clone (only the latest commit from the main branch) is currently very lightweight. We avoid the Github API's rate limits and the need for Github Actions this way.

If performance ever becomes an issue in the future, we can revisit this and switch to using the Github API.

Draft PR because I have an open PR in the cli-plugins repo that will require changes here.

References
----------

Test & Review
-------------
Manual testing
Unit tests

Open Questions / Follow-ups
---------------------------


brianstrauch:
> List and install plugins from the official Confluent CLI plugin repository with confluent plugin search and confluent plugin install.

We may want to link to the repository when this gets uploaded to the docs page

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/types"
+	"github.com/confluentinc/cli/internal/pkg/utils"
+)
+
+func (c *command) newInstallCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "Install Confluent CLI plugins.",
+		Long:  "Install available Confluent CLI plugins from the confluentinc/cli-plugins repository.",
```
brianstrauch, line 23:
```suggestion
		Short: "Install official Confluent CLI plugins.",
		Long:  "Install official Confluent CLI plugins from the confluentinc/cli-plugins repository.",
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.install,
+	}
+
+	cmd.Flags().String("plugin-directory", "", "The plugin installation directory; this must be a directory in your $PATH. If not specified, a default will be selected based on your OS.")
```
brianstrauch, line 28:
```suggestion
	cmd.Flags().String("plugin-directory", "", "The plugin installation directory. This must be a directory in your $PATH. If not specified, a default will be selected based on your OS.")
```

brianstrauch: Can we tell the user what the default is? I assume /usr/local/bin on macOS?

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+	return cmd
+}
+
+func (c *command) install(cmd *cobra.Command, args []string) error {
+	installDir, err := getPluginInstallDir(cmd)
+	if err != nil {
+		return err
+	}
+
+	dir, err := os.MkdirTemp("", "plugin-install")
```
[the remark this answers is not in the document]
brianstrauch: Might also be worth it to put it in `~/.confluent` instead of a temporary directory?

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
+
+	_, err = clonePluginRepo(dir, cliPluginsUrl)
+	if err != nil {
```
brianstrauch, line 46:
```suggestion
	if _, err = clonePluginRepo(dir, cliPluginsUrl); err != nil {
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+
+func getPluginInstallDir(cmd *cobra.Command) (string, error) {
+	if !cmd.Flags().Changed("plugin-directory") {
+		return getDefaultPluginInstallDir()
+	}
+
+	pluginDir, err := cmd.Flags().GetString("plugin-directory")
+	if err != nil {
+		return "", err
+	}
```
brianstrauch, line 66:
```suggestion
	pluginDir, err := cmd.Flags().GetString("plugin-directory")
	if err != nil {
		return "", err
	}
	if pluginDir == "" {
		return getDefaultPluginInstallDir()
	}
```

brianstrauch: This also handles the `--plugin-dir ""` case, which doesn't make any sense.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+		output.Printf("WARNING: failed to find default directory \"%s\" in your $PATH.\n\n", defaultDir)
+	}
+
+	return defaultDir, nil
+}
+
+func inPath(dir string) bool {
+	pathDirectories := filepath.SplitList(os.Getenv("PATH"))
+	for i := range pathDirectories {
+		absPath, _ := filepath.Abs(pathDirectories[i])
```
brianstrauch, line 116: This function could potentially return `""` when there's an error. I would explicitly check `err == nil`.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		manifest.Name = file.Name()
+
+		return manifest, nil
+	}
+
+	return nil, errors.Errorf("plugin %s not found", pluginName)
+}
+
+func installPlugin(manifest *Manifest, repoDir, installDir string) error {
```
brianstrauch, line 155:
```suggestion
func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
```

internal/cmd/plugin/command_search.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/confluentinc/cli/internal/pkg/utils"
+)
+
+type Manifest struct {
+	Name         string `human:"Name" serialized:"name"`
+	Description  string `yaml:"description" human:"Description" serialized:"description"`
+	Dependencies string `yaml:"dependencies" human:"Dependencies" serialized:"dependencies"`
+}
+
+const cliPluginsUrl = "https://github.com/confluentinc/cli-plugins.git"
```
brianstrauch, line 22: Put this in the top-level command.go file since it's used in command_search.go and command_install.go?

internal/cmd/plugin/command_search.go
```diff
@@ -0,0 +1,98 @@
...
+		list.Add(manifest)
+	}
+
+	return list.Print()
+}
+
+func clonePluginRepo(dir, url string) (*git.Repository, error) {
+	cloneOptions := &git.CloneOptions{
+		URL:          url,
+		SingleBranch: true, // this should be redundant w/ Depth=1, but specify it just in case
```
brianstrauch, line 66: +1 for deleting it if it doesn't do anything

internal/cmd/plugin/install_utils.go
```diff
@@ -0,0 +1,134 @@
+package plugin
```
brianstrauch, line 1: Not a huge fan of files with the name "utils" in it... they can become black holes of random functionality. Can we either put these helper functions in the relevant command_*.go file or the root command.go file?

internal/cmd/plugin/install_utils.go
```diff
@@ -0,0 +1,134 @@
...
+		return "", nil, nil
+	}
+
+	if manifest.Dependencies == "" {
+		return "", nil, nil
+	}
+
+	dependencySlice := strings.Split(strings.ToLower(manifest.Dependencies), " ")
+	if len(dependencySlice) == 1 {
+		return dependencySlice[0], nil, nil
```
brianstrauch, line 35: What is `dependencySlice[0]` and what is `dependencySlice[1]`? Can we put these into variables with better names?

internal/cmd/plugin/install_utils.go
```diff
@@ -0,0 +1,134 @@
...
+func checkGoVersion(ver *version.Version) {
+	versionCmd := exec.NewCommand("go", "version")
+
+	out, err := versionCmd.Output()
+	if err != nil {
+		output.ErrPrintf(programNotFoundMsg, "Go")
+		return
+	}
+
+	re := regexp.MustCompile(`go[1-9]\.[0-9]+[\.0-9]*`)
```
brianstrauch, line 80: Future-proofing :)
```suggestion
	re := regexp.MustCompile(`go[1-9][0-9]*\.[0-9]+(\.[1-9][0-9]*)?`)
```

brianstrauch: (This is untested, btw)

internal/cmd/plugin/command_install.go
sgagniere: Yeah it's `/usr/local/bin` on unix systems and the CLI installation directory on Windows.

That's what I ultimately went with because the parts of the path common to all users is the system root and some windows32 folders, none of which I want to write to. I guess there's no real guarantee that the user will actually have added the CLI install directory to their path, but I issue a warning if the plugin directory isn't in the path.

I'll specify that in the flag description.

internal/cmd/plugin/install_utils.go
sgagniere: The `[0]` entry is the first dependency name, the `[1]` could be either another dependency name or a version number for the first one. Basically, this whole block is me trying to guess what might be in here.

It's why I made that other PR, and I'm going to be rewriting this part anyway to match the new manifest file structure.

internal/cmd/plugin/command_install.go
```diff
@@ -48,7 +48,8 @@ func (c *command) install(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	dir, err := os.MkdirTemp("~/.confluent", "confluent-plugin-install")
+	confluentDir := filepath.Join(os.Getenv("HOME"), ".confluent")
+	dir, err := os.MkdirTemp(confluentDir, "confluent-plugin-install")
```
brianstrauch, line 52: `~/.confluent/plugin-install` is less redundant than `~/.confluent/confluent-plugin-install`.
```suggestion
	dir, err := os.MkdirTemp(confluentDir, "plugin-install")
```

internal/cmd/plugin/command_install.go
```diff
@@ -184,25 +185,22 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
...
 	}
 
-	if manifest.Dependencies == "" {
-		return "", nil
-	}
-
-	dependencySlice := strings.Split(strings.ToLower(manifest.Dependencies), " ")
-	if len(dependencySlice) == 1 {
-		return dependencySlice[0], nil
+	primaryDependency := manifest.Dependencies[0]
```
brianstrauch, line 192: Do we plan to support multiple dependencies in the future?

internal/cmd/plugin/command_search.go
```diff
@@ -13,10 +15,21 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/utils"
 )
 
-type Manifest struct {
+type ManifestOut struct {
 	Name         string `human:"Name" serialized:"name"`
-	Description  string `yaml:"description" human:"Description" serialized:"description"`
-	Dependencies string `yaml:"dependencies" human:"Dependencies" serialized:"dependencies"`
+	Description  string `human:"Description" serialized:"description"`
+	Dependencies string `human:"Dependencies" serialized:"dependencies"`
```
brianstrauch, line 21: `[]string`?

sgagniere: Wait, does table/list handle string slices already?

internal/cmd/plugin/command_install.go
sgagniere: I was thinking that we basically already can, because someone could write a plugin that calls other programs (e.g. a plugin in Go that uses `exec`).

Maybe we should enforce instead that there's only one dependency?

test/fixtures/output/plugin/help-onprem.golden
```diff
@@ -39,7 +39,9 @@ Usage:
   confluent plugin [command]
 
 Available Commands:
+  install     Install official Confluent CLI plugins.
   list        List Confluent CLI plugins in $PATH.
+  search      Search for available Confluent CLI plugins.
```
brianstrauch, line 44:
```suggestion
  search      Search for Confluent CLI plugins.
```

internal/cmd/plugin/install_utils.go
sgagniere: I tested this out in a regex tester; looks good to me. I did add `^` and `$` to match whole strings though, since apparently `MatchString` will return true if it finds a substring match.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,304 @@
...
+func (c *command) newInstallCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "Install official Confluent CLI plugins.",
+		Long:  "Install official Confluent CLI plugins from the confluentinc/cli-plugins repository.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.install,
+	}
+
+	cmd.Flags().String("plugin-directory", "", `The plugin installation directory. This must be a directory in your $PATH. If not specified, the default is the CLI installation directory for Windows, and "usr/local/bin" for MacOS and Linux.`)
```
brianstrauch, line 44: What if we installed all plugins to `~/.confluent/plugins` and make sure to scan that directory even if it isn't part of $PATH?

sgagniere: Oh, I like that a lot.

That will let me remove a chunk of code, too.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,232 @@
...
+	pluginDir := fmt.Sprintf("%s/%s", repositoryDir, name)
+	files, err := os.ReadDir(pluginDir)
+	if err != nil {
+		return err
+	}
+
+	found := false
+	for _, file := range files {
+		if strings.HasPrefix(file.Name(), "confluent-") {
+			found = true
```
brianstrauch, line 204: Hmm, this code should already exist for `confluent plugin list`. Wonder if we can repurpose it in some way.

internal/cmd/plugin/command_install_test.go
```diff
@@ -0,0 +1,55 @@
...
+	dir, _ := filepath.Abs("../../../test/fixtures/input/plugin")
+	manifest, err := getPluginManifest("confluent-test_plugin", dir)
+	assert.NoError(t, err)
+
+	referenceManifest := &Manifest{
+		Name:        "confluent-test_plugin",
+		Description: "Does nothing",
+		Dependencies: []Dependency{
+			{
+				Dependency: "Python",
```
brianstrauch, line 26: Dependency is already the name of the struct, so this may avoid some confusion. `Dependencies[0].Dependency` is less readable than `Dependencies[0].Name`
```suggestion
				Name: "Python",
```

internal/cmd/plugin/command_search.go
```diff
@@ -0,0 +1,127 @@
...
+			manifestFile, err := os.ReadFile(manifestPath)
+			if err != nil {
+				return nil, err
+			}
+
+			manifest := new(Manifest)
+			if err := yaml.Unmarshal(manifestFile, manifest); err != nil {
+				return nil, err
+			}
+			manifestOutList = append(manifestOutList, &ManifestOut{
```
brianstrauch, line 103: Let's initialize this struct outside of the function call?

internal/cmd/plugin/command_search.go
```diff
@@ -0,0 +1,127 @@
...
+				Description:  manifest.Description,
+				Dependencies: dependenciesToString(manifest.Dependencies),
+			})
+		}
+	}
+
+	return manifestOutList, nil
+}
+
+func dependenciesToString(dependencies []Dependency) string {
```
brianstrauch, line 114: This function is really difficult to read. Can we create a slice of dependency strings, and then use `strings.Join()` to comma-delimit them?

brianstrauch: Also, should be super easy to unit test!

internal/pkg/plugin/plugin.go
```diff
@@ -27,10 +27,16 @@ type pluginInfo struct {
 
 // SearchPath goes through the files in the user's $PATH and checks if they are plugins
 func SearchPath(cfg *v1.Config) map[string][]string {
-	log.CliLogger.Debugf("Searching `$PATH` for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
+	log.CliLogger.Debugf("Searching `$PATH` and `~/.confluent/plugins` for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
```
brianstrauch, line 30: Does Windows use `~` for `$HOME`?

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+const (
+	Go     = "go"
+	Python = "python"
+	Shell  = "shell"
+)
+
+const (
+	programNotFoundMsg      = "[WARN] Unable to find %s. Check that it is installed in a directory in your $PATH.\n"
```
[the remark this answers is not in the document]
sgagniere: No, these are warnings if I can't find `go` or `python`, not plugins.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,232 @@
...
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.install,
+	}
+
+	return cmd
+}
+
+func (c *command) install(cmd *cobra.Command, args []string) error {
+	confluentDir := filepath.Join(os.Getenv("HOME"), ".confluent")
+	dir, err := os.MkdirTemp(confluentDir, "plugin-install")
```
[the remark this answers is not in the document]
sgagniere: It does directly clone into here, but I'll rename it to `cli-plugins`.

internal/pkg/plugin/plugin.go
sgagniere: I don't think so, actually. I'll have to do some Windows testing tomorrow.

internal/cmd/plugin/command_install_test.go
sgagniere: Done. I also introduced a PR to make this change in the manifests of the plugin repo.

internal/cmd/plugin/command_install.go
```diff
@@ -107,14 +101,14 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
...
 		return installGoPlugin(manifest.Name)
-	case Python:
+	case "python":
 		checkPythonVersion(ver)
-		return installSimplePlugin(manifest.Name, repositoryDir, installDir, Python)
-	case Shell:
-		return installSimplePlugin(manifest.Name, repositoryDir, installDir, Shell)
+		return installSimplePlugin(manifest.Name, repositoryDir, installDir, "python")
+	case "shell":
+		return installSimplePlugin(manifest.Name, repositoryDir, installDir, "shell")
```
brianstrauch: GitHub won't let me suggest a change here, but we can combine these into one, and pass `language` as the fourth parameter.

internal/cmd/plugin/command_install.go
```diff
@@ -125,26 +119,26 @@ func getLanguage(manifest *Manifest) (string, *version.Version) {
...
+		return language.Dependency, nil
 	}
 
-	return primaryDependency.Dependency, ver
+	return language.Dependency, ver
 }
 
 func checkPythonVersion(ver *version.Version) {
-	versionCmd := exec.NewCommand(Python, "--version")
+	versionCmd := exec.NewCommand("python", "--version")
```
brianstrauch, line 138: This binary is often called `python3` on some machines, so we will want to support that!

internal/cmd/plugin/command_install.go
sgagniere: I set it up this way because the Python case has an additional call to `checkPythonVersion`.

brianstrauch: If we take the interface route, maybe we can call `CheckVersion()` followed by `InstallPlugin()` for each case here, then we won't have to maintain this `switch` statement for future languages.

sgagniere: Good point, I like that. I'll go ahead and make a version of this.

sgagniere: Hmm, well I kept the switch because I still need to pick out the right interface implementations, but at least it's a simpler switch now... what do you think?

internal/pkg/plugin/python_plugin_installer.go
```diff
@@ -0,0 +1,56 @@
...
+	re := regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+\.(0|[1-9][0-9]*)$`)
+	for _, word := range strings.Split(string(out), " ") {
+		if re.MatchString(word) {
+			installedVer, err := version.NewVersion(strings.Trim(word, " \n"))
+			if err != nil {
+				output.ErrPrintf(unableToParseVersionMsg, "python")
+				return
+			}
+			if installedVer.LessThan(ver) {
+				output.ErrPrintf(insufficientVersionMsg, "python", installedVer, ver)
```
brianstrauch, line 47: Let's return an error here instead of printing?

internal/pkg/plugin/go_plugin_installer.go
```diff
@@ -16,13 +16,13 @@ type GoPluginInstaller struct {
...
 
-func (g *GoPluginInstaller) CheckVersion(ver *version.Version) {
+func (g *GoPluginInstaller) CheckVersion(ver *version.Version) error {
 	versionCmd := exec.NewCommand("go", "version")
 
 	out, err := versionCmd.Output()
 	if err != nil {
 		output.ErrPrintf(programNotFoundMsg, "go")
-		return
+		return nil
```
brianstrauch, line 25: We should return an error here too, no?

sgagniere: What I had in mind with this being a warning instead of an error is that they might have just forgotten to put it in their path.

But yeah, come to think of it, if `go` isn't found here then there's no way that the install command will succeed, and at that point they should be errors for all languages to be consistent.

I'll makethe change.

internal/pkg/plugin/plugin.go
sgagniere: Switched it to `%USERPROFILE%` which is Windows' version of `$HOME`, and also fixed the directories in `install` and `search` to use `os.UserHomeProfile` instead of `$HOME`, which didn't work on Windows.

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+
+const (
+	programNotFoundMsg      = "[WARN] Unable to find %s. Check that it is installed in a directory in your $PATH.\n"
+	unableToParseVersionMsg = "[WARN] Unable to parse %s version.\n"
+	insufficientVersionMsg  = "[WARN] Installed %s version %s is less than the required version %s.\n"
+)
+
+func (c *command) newInstallCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "install",
```
brianstrauch, line 27:
```suggestion
		Use:   "install <plugin>",
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func (c *command) install(cmd *cobra.Command, args []string) error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	confluentDir := filepath.Join(home, ".confluent")
+	dir, err := os.MkdirTemp(confluentDir, "cli-plugins")
```
brianstrauch, line 41: Why are we using `MkdirTemp()` instead of `Mkdir()` here? What's the difference?

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+		return err
+	}
+	defer os.RemoveAll(dir)
+
+	installDir := filepath.Join(confluentDir, "plugins")
+	if err := os.MkdirAll(installDir, os.ModePerm); err != nil {
+		return errors.Wrapf(err, "failed to create plugin install directory %s", installDir)
+	}
+
+	if _, err = clonePluginRepo(dir, cliPluginsUrl); err != nil {
```
brianstrauch, line 52:
```suggestion
	if _, err := clonePluginRepo(dir, cliPluginsUrl); err != nil {
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+	manifest, err := getPluginManifest(args[0], dir)
+	if err != nil {
+		return err
+	}
+
+	if err := installPlugin(manifest, dir, installDir); err != nil {
+		return err
+	}
+
+	output.Printf("Installed plugin %s.\n", args[0])
```
brianstrauch, line 65:
```suggestion
	output.Printf("Installed plugin \"%s\".\n", args[0])
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+
+func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
+	language, ver := getLanguage(manifest)
+
+	var pluginInstaller plugin.PluginInstaller
+	switch language {
+	case "go":
+		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Name}
+	case "python":
+		pluginInstaller = &plugin.PythonPluginInstaller{Name: manifest.Name, RepositoryDir: repositoryDir, InstallDir: installDir}
```
brianstrauch, line 111:
```suggestion
		pluginInstaller = &plugin.PythonPluginInstaller{
			Name:          manifest.Name,
			RepositoryDir: repositoryDir,
			InstallDir:    installDir,
		}
```

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if err := pluginInstaller.CheckVersion(ver); err != nil {
+		return err
+	}
+	return pluginInstaller.Install()
+}
+
+func getLanguage(manifest *Manifest) (string, *version.Version) {
+	if manifest == nil || len(manifest.Dependencies) == 0 {
+		return "", nil
```
brianstrauch, line 126: Should we return an error here? Every dependency should have a language, right?

internal/cmd/plugin/command_install.go
```diff
@@ -0,0 +1,141 @@
...
+
+	language := manifest.Dependencies[0]
+	language.Name = strings.ToLower(language.Name)
+	if language.Version == "" {
+		return language.Name, nil
+	}
+
+	ver, err := version.NewVersion(language.Version)
+	if err != nil {
+		return language.Name, nil
```
brianstrauch, line 137: Do we really want to ignore this error? This could cause a plugin to fail without warning.

internal/pkg/plugin/go_plugin_installer.go
```diff
@@ -0,0 +1,51 @@
...
+
+	return nil
+}
+
+func (g *GoPluginInstaller) Install() error {
+	packageName := fmt.Sprintf("github.com/confluentinc/cli-plugins/%s@latest", g.Name)
+	installCmd := exec.NewCommand("go", "install", packageName)
+
+	if _, err := installCmd.Output(); err != nil {
+		return errors.Wrap(err, "failed to run go install command")
```
brianstrauch, line 47:
```suggestion
		return errors.Wrap(err, "failed to run `go install`")
```

internal/pkg/plugin/plugin.go
```diff
@@ -27,18 +27,27 @@ type pluginInfo struct {
 
 // SearchPath goes through the files in the user's $PATH and checks if they are plugins
 func SearchPath(cfg *v1.Config) map[string][]string {
-	log.CliLogger.Debugf("Searching `$PATH` for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
+	if runtime.GOOS == "windows" {
+		log.CliLogger.Debugf("Searching `$PATH` and `%%USERPROFILE%%\\.confluent\\plugins` for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
+	}
+	log.CliLogger.Debugf("Searching `$PATH` and `~/.confluent/plugins` for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
```
brianstrauch, line 33: We're inconsistent on how we format environment variables and file names. My initial reaction is not to format them (and only format commands/flags with backticks, according to https://github.com/confluentinc/cli/blob/main/internal/pkg/errors/README.md#general-message-format)

```suggestion
	log.CliLogger.Debugf("Searching $PATH and ~/.confluent/plugins for plugins. Plugins can be disabled in %s.\n", cfg.GetFilename())
```

internal/pkg/plugin/plugin_installer.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"os"
+
+	"github.com/hashicorp/go-version"
+
+	"github.com/confluentinc/cli/internal/pkg/errors"
+)
+
+const (
+	programNotFoundErrorMsg      = "unable to find %s; check that it is installed in a directory in your $PATH"
```
brianstrauch, line 13: The second half of the error should be part of a suggestion, not the error text.

internal/pkg/plugin/plugin_installer.go
```diff
@@ -0,0 +1,50 @@
...
+func installSimplePlugin(name, repositoryDir, installDir, language string) error {
+	pluginDir := fmt.Sprintf("%s/%s", repositoryDir, name)
+	entries, err := os.ReadDir(pluginDir)
+	if err != nil {
+		return err
+	}
+
+	found := false
+	for _, entry := range entries {
+		if name := PluginFromEntry(entry); name != "" {
```
brianstrauch, line 32: Unused variable
```suggestion
		if PluginFromEntry(entry) != "" {
```

internal/pkg/plugin/python_plugin_installer.go
```diff
@@ -0,0 +1,55 @@
...
+		out, err = versionCmd.Output()
+	}
+	if err != nil {
+		return errors.Errorf(programNotFoundErrorMsg, "python")
+	}
+
+	re := regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+\.(0|[1-9][0-9]*)$`)
+	for _, word := range strings.Split(string(out), " ") {
+		if re.MatchString(word) {
+			installedVer, err := version.NewVersion(strings.Trim(word, " \n"))
```
brianstrauch, line 40: Maybe you want https://pkg.go.dev/strings#TrimSpace?

internal/cmd/plugin/command_install.go
sgagniere: Two reasons: it makes it clearer in the code that it's a temporary directory, and `MkdirTemp` doesn't require folder permissions as an argument.

internal/pkg/plugin/plugin_installer.go
sgagniere: Good point; I forgot to separate this into an error and suggestion when I switched it from a warning to an error.

internal/pkg/plugin/python_plugin_installer.go
sgagniere: Yeah, actually. Thanks.

internal/cmd/plugin/command_install.go
```diff
@@ -115,21 +126,21 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
 	return pluginInstaller.Install()
 }
 
-func getLanguage(manifest *Manifest) (string, *version.Version) {
+func getLanguage(manifest *Manifest) (string, *version.Version, error) {
 	if manifest == nil || len(manifest.Dependencies) == 0 {
-		return "", nil
+		return "", nil, errors.New("invalid plugin manifest: no language found")
```
brianstrauch, line 131: It feels weird to be using the colon in an error message without wrapping another error.
```suggestion
		return "", nil, errors.New("no language found in plugin manifest")
```

internal/pkg/plugin/plugin_installer.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 const (
-	programNotFoundErrorMsg      = "unable to find %s; check that it is installed in a directory in your $PATH"
+	programNotFoundErrorMsg      = "unable to find %s"
+	programNotFoundSuggestions   = "check that it is installed in a directory in your $PATH"
```
brianstrauch, line 14: Suggestions should be full sentences
```suggestion
	programNotFoundSuggestions   = "Check that it is installed in a directory in your $PATH."
```

internal/pkg/plugin/python_plugin_installer.go
```diff
@@ -0,0 +1,55 @@
...
+	v3, _ := version.NewVersion("3.0.0")
+
+	var out []byte
+	var err error
+	if ver.GreaterThanOrEqual(v3) {
+		out, err = version3Cmd.Output()
+	}
+	if err != nil || ver.LessThan(v3) {
+		out, err = versionCmd.Output()
+	}
```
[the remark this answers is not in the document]
sgagniere: Changed it to check if the major version isn't 3 instead of less than 3.

internal/pkg/plugin/shell_plugin_installer.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+type ShellPluginInstaller struct {
+	Name          string
+	RepositoryDir string
+	InstallDir    string
+}
+
+func (s *ShellPluginInstaller) CheckVersion(_ *version.Version) error {
+	return nil
```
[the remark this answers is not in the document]
sgagniere: Previously this was because I didn't want to enforce a specific shell script. But since we're going with just bash scripts now, I'll add a version check for bash.

internal/pkg/plugin/plugin_installer.go
sgagniere: > Suggestions should be full sentences

I know, this was just a copy paste error. Thanks for catching it!

internal/pkg/plugin/python_plugin_installer.go
```diff
@@ -21,14 +21,18 @@ func (p *PythonPluginInstaller) CheckVersion(ver *version.Version) error {
...
 
 	var out []byte
 	var err error
-	if ver.GreaterThanOrEqual(v3) {
+	if majorVer == 3 {
 		out, err = version3Cmd.Output()
 	}
-	if err != nil || ver.LessThan(v3) {
+	if err != nil || majorVer != 3 {
 		out, err = versionCmd.Output()
```
brianstrauch, line 36: The `python` binary could still be Python 3. `python3` is just a common alias for `python`.

sgagniere: Aliases aren't supported with `exec.Command` unless you execute through bash or zsh depending on which `rc` file you put the alias in (also, sometimes `python` is the alias instead of `python3`).

That's why I also check `python` in the case that `python3` failed.

brianstrauch: > Aliases aren't supported with exec.Command

It's not an alias in the sense of the `alias` shell command, it's a real command! You should grab a `python3` binary to test this with since it sounds like this is untested...

sgagniere: I am using a python3 binary.

I'm confused now about what's problematic with this if statement

brianstrauch: Oops, I was thinking `&&` and not `||`, disregard. Tracing the values of `err` here was hurting my head a bit...

sgagniere: Ahh, gotcha.

internal/pkg/plugin/plugin_test.go
```diff
@@ -101,3 +102,29 @@ func TestSearchPath(t *testing.T) {
 	require.True(t, ok)
 	require.Equal(t, fileName, filepath.Base(pluginPaths[0]))
 }
+
+func TestVersionRegex(t *testing.T) {
```
brianstrauch, line 106: FYI, these tests don't add any coverage. It might be better to write a helper function and unit test that!

internal/pkg/plugin/go_plugin_installer.go
```diff
@@ -25,9 +29,8 @@ func (g *GoPluginInstaller) CheckVersion(ver *version.Version) error {
 		return errors.NewErrorWithSuggestions(fmt.Sprintf(programNotFoundErrorMsg, "go"), programNotFoundSuggestions)
 	}
 
-	re := regexp.MustCompile(goVersionPattern)
 	for _, word := range strings.Split(string(out), " ") {
-		if re.MatchString(word) {
+		if g.IsVersionNumber(word) {
 			installedVer, err := version.NewVersion(strings.TrimPrefix(word, "go"))
 			if err != nil {
 				return errors.Errorf(unableToParseVersionErrorMsg, "go")
```
brianstrauch, line 36: I think we may have discussed this earlier, but I wonder if this should actually be a warning. If the user ends up having a old version of a dependency that happens to still work, it would be nice if the plugin still ran.

internal/pkg/plugin/plugin_installer.go
```diff
@@ -17,6 +17,7 @@ const (
 )
 
 type PluginInstaller interface {
+	IsVersionNumber(word string) bool
```
brianstrauch, line 20:
```suggestion
	IsVersion(word string) bool
```

brianstrauch (approved):

PR #2015, opened 2023-06-23
[CLI-2390] Add `confluent kafka link describe`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `confluent kafka link describe`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
The SDKs already support this and we have corresponding `list` commands for cloud and on prem, so it makes sense to add this functionality.

References
----------

Test & Review
-------------
Manual testing
added new integration tests

Open Questions / Follow-ups
---------------------------


test/fixtures/output/kafka/link/describe.golden
```diff
@@ -0,0 +1,8 @@
++---------------------+--------------------+
+| Name                | link-1             |
+| Source Cluster      | lkc-describe-topic |
+| Destination Cluster |                    |
+| State               | AVAILABLE          |
+| Error               |                    |
+| Error Message       |                    |
```
brianstrauch, line 7: `omitempty`?

sgagniere: I ended up making a second struct just for describe that has omitempty on these two fields.

Turns out that while omitempty is ignored in the human list output, it's used in the json and yaml list output.

internal/cmd/kafka/command_link_describe.go
```diff
@@ -3,12 +3,24 @@ package kafka
...
 	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/kafkarest"
 	"github.com/confluentinc/cli/internal/pkg/output"
 )
 
+type describeOut struct {
+	Name                 string `human:"Name" serialized:"link_name"`
+	TopicName            string `human:"Topic Name" serialized:"topic_name"`
+	SourceClusterId      string `human:"Source Cluster" serialized:"source_cluster_id"`
+	DestinationClusterId string `human:"Destination Cluster" serialized:"destination_cluster_id"`
```
brianstrauch, line 18: Why the inconsistency here with the human output not containing "ID"?

brianstrauch: Sounds good, might be worth some investigation to see if other commands do this too

test/fixtures/output/kafka/link/describe.golden
```diff
@@ -0,0 +1,6 @@
++---------------------+--------------------+
```
brianstrauch, line 1: Let's include a test case where "Error" and "Error Detail" are not hidden?

brianstrauch (approved):

PR #2018, opened 2023-06-26
Extending the CLI kill-switch to individual commands/flags

kevin-wu24:
Checklist
---------
   * yes: ok

What
----
Extends the cli.disable feature flag to allow for individual commands/command trees/flags to be disabled in addition to an entire CLI version.

References
----------
Updated 1-pager: https://confluentinc.atlassian.net/wiki/spaces/DPI/pages/2791407966/How+to+Flip+the+Confluent+CLI+Cloud+Kill+Switch

Test & Review
-------------
Tested in LaunchDarkly test environment


internal/pkg/panic-recovery/panic_recovery.go
```diff
@@ -31,9 +31,10 @@ func CollectPanic(cmd *cobra.Command, args []string, cfg *v1.Config) *usage.Usag
 
 // ParseFlags collects the flags of a command after being found with Find()
 func ParseFlags(cmd *cobra.Command, flags []string) *[]string {
-	var flagsTrimmed, formattedFlags []string
+	var formattedFlags []string
+	flagsTrimmed := make([]string, len(flags))
```
brianstrauch, line 35:
```suggestion
	trimmedFlags := make([]string, len(flags))
```

internal/cmd/command.go
```diff
@@ -145,6 +145,7 @@ func NewConfluentCommand(cfg *v1.Config) *cobra.Command {
 
 	changeDefaults(cmd, cfg)
 	deprecateCommandsAndFlags(cmd, cfg)
+	featureflags.Manager.SetCommandAndFlags(cmd, os.Args[1:])
```
brianstrauch, line 148: I think `cmd` contains `os.Args[1:]`?

internal/pkg/featureflags/feature_flags.go
```diff
@@ -42,13 +44,15 @@ const (
...
-var attributes = []string{"user.resource_id", "org.resource_id", "environment.id", "cli.version", "cluster.id", "cluster.physicalClusterId"}
+var attributes = []string{"user.resource_id", "org.resource_id", "environment.id", "cli.version", "cluster.id", "cluster.physicalClusterId", "cli.command", "cli.flags"}
 
 type featureFlagManager interface {
 	BoolVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal bool) bool
 	StringVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal string) string
 	IntVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal int) int
 	JsonVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal any) any
+
+	SetCommandAndFlags(cmd *cobra.Command, args []string)
```
brianstrauch, line 55: No reason to add this here... I don't think we're mocking this function.

internal/pkg/featureflags/feature_flags.go
```diff
@@ -94,6 +100,13 @@ func Init(version *version.Version, isTest, isDisabledConfig bool) {
 	}
 }
 
+func (ld *launchDarklyManager) SetCommandAndFlags(cmd *cobra.Command, args []string) {
+	fullCmd, a, _ := cmd.Find(args)
```
brianstrauch, line 104: What does "a" mean?

internal/pkg/featureflags/feature_flags.go
kevin-wu24: I need to add this here so the function is visible in internal/cmd/command.go, as Manager is of type featureFlagManager.

internal/pkg/featureflags/feature_flags.go
```diff
@@ -57,6 +61,8 @@ type launchDarklyManager struct {
 	isDisabled               bool
 	timeoutSuggestionPrinted bool
 	version                  *version.Version
+	command                  *cobra.Command
+	flags                    []string
```
MuweiHe, line 65: sort the fields? I think things like command and flags should come first than `version`

cmd/lint/main.go
```diff
@@ -48,7 +48,9 @@ var flagRules = []linter.FlagRule{
 	linter.FlagFilter(linter.RequireStringSlicePrefix, linter.ExcludeFlag("property")),
 
 	linter.FlagFilter(linter.RequireFlagUsageMessage, linter.ExcludeFlag("key-deserializer", "value-deserializer")),
-	linter.RequireFlagUsageRealWords(properNouns),
+	linter.FlagFilter(linter.RequireFlagUsageRealWords(properNouns),
```
brianstrauch, line 51: Any idea why we need this all of a sudden?

internal/pkg/panic-recovery/panic_recovery.go
```diff
@@ -23,20 +23,21 @@ func CollectPanic(cmd *cobra.Command, args []string, cfg *v1.Config) *usage.Usag
...
+		Flags:       ParseFlags(fullCommand, flags),
 		Error:       cliv1.PtrBool(true),
 		StackFrames: parseStack(string(debug.Stack())),
 	}
 }
 
-// parseFlags collects the flags alongside the panicking command
-func parseFlags(cmd *cobra.Command, flags []string) *[]string {
+// ParseFlags collects the flags of a command after being found with Find()
+func ParseFlags(cmd *cobra.Command, flags []string) *[]string {
```
brianstrauch, line 33: Any reason we're returning a pointer to a slice? In memory, a slice is already represented by a pointer to the first element so this is doing nothing.

internal/pkg/featureflags/feature_flags.go
```diff
@@ -40,20 +42,15 @@ const (
...
-	IntVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal int) int
-	JsonVariation(key string, ctx *dynamicconfig.DynamicContext, client v1.LaunchDarklyClient, shouldCache bool, defaultVal any) any
-}
+var attributes = []string{"user.resource_id", "org.resource_id", "environment.id", "cli.version", "cluster.id", "cluster.physicalClusterId", "cli.command", "cli.flags"}
 
 type launchDarklyManager struct {
 	cliClient                *sling.Sling
 	ccloudClient             func(v1.LaunchDarklyClient) *sling.Sling
+	command                  *cobra.Command
+	flags                    []string
```
brianstrauch, line 53: Does this disable help text? If we disable a command, but run `confluent help`, this seems like it'll still list that command, since `confluent help` is a separate command.

internal/pkg/panic-recovery/panic_recovery.go
kevin-wu24: In the function above, `CollectPanic`, the usage struct takes in `*[]string` for `Flags` and `StackFrames`. Returning a pointer to a slice here and below in `parseStack` means I don't have to declare a variable in `CollectPanic` for either function call.

kevin-wu24: For whatever reason I can't assign the field in the Usage constructor like `... Flags: &ParseFlags(),` which I think might just be a Go syntax thing

internal/pkg/featureflags/feature_flags.go
kevin-wu24: This does not disable help text. I can run `confluent help` and still see the command, and I can still run the command with the `-h` flag and see help text about it. Muwei said for the output of `confluent help` either we should just prevent the top level command from showing up if it's disabled or append a `[DISABLED]` to its description, and for `-h` it doesn't matter if the help text is shown.

cmd/lint/main.go
brianstrauch: huh, any idea why that line triggered the lint?

internal/pkg/panic-recovery/panic_recovery.go
brianstrauch: > I don't have to declare a variable in CollectPanic for either function call

IMO, doing this definitely sounds better than returning a pointer to a slice

internal/pkg/featureflags/feature_flags.go
brianstrauch: > Muwei said for the output of confluent help either we should just prevent the top level command from showing up if it's disabled

The goal is to prevent any arbitrary command from showing up, not just the top-level command, right?

cmd/lint/main.go
kevin-wu24: no unfortunately. I was just able to isolate that line as causing the lint

internal/pkg/panic-recovery/panic_recovery.go
kevin-wu24: Okay, sounds good!

internal/pkg/featureflags/feature_flags.go
kevin-wu24: Yeah, so if a child command is disabled, running its parent would result in the child now showing up or being marked as disabled.

internal/pkg/featureflags/disable.go
```diff
@@ -0,0 +1,54 @@
...
+func DisableHelpText(command *cobra.Command, flags []string) {
+	if len(flags) == 0 {
+		command.Hidden = true
+	} else {
+		formattedFlags := ppanic.ParseFlags(command, flags)
+		for _, flag := range formattedFlags {
+			err := command.Flags().MarkHidden(flag)
+			if err != nil {
+				return
+			}
```
brianstrauch, line 23: What if one flag doesn't exist but another does?
```suggestion
			_ = command.Flags().MarkHidden(flag)
```

internal/cmd/command.go
```diff
@@ -262,3 +264,16 @@ func deprecateCommandsAndFlags(cmd *cobra.Command, cfg *v1.Config) {
 		}
 	}
 }
+
+func disableCommandAndFlagHelpText(cmd *cobra.Command, cfg *v1.Config) {
+	ctx := dynamicconfig.NewDynamicContext(cfg.Context(), nil)
+	disableResp := featureflags.GetLDDisableMap(ctx)
+	disabledCmdsAndFlags, ok := disableResp["patterns"].([]any)
```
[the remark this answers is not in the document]
kevin-wu24: no, i don't know why but when I try to cast it directly to a `[]string` `ok` is false

cmd/lint/main.go
MuweiHe: i would be curious to know what's the cause here

internal/pkg/featureflags/disable.go
```diff
@@ -0,0 +1,52 @@
...
+	ppanic "github.com/confluentinc/cli/internal/pkg/panic-recovery"
+)
+
+func DisableHelpText(command *cobra.Command, flags []string) {
+	if len(flags) == 0 {
+		command.Hidden = true
+	} else {
+		formattedFlags := ppanic.ParseFlags(command, flags)
+		for _, flag := range formattedFlags {
+			_ = command.Flags().MarkHidden(flag)
```
MuweiHe, line 20: what's being ignored here?

kevin-wu24: Basically, if we accidentally misspell a flag or add a flag that doesn't exist in launchDarkly, ignoring the error here from `MarkHidden` means the other flags will still be marked disabled like normal.

cmd/lint/main.go
kevin-wu24: The short hands for the global persistent flags `unsafe-trace` and `verbose` are not "Real Words" (e.g. `vvvv`, `vvv`, etc.) and contained in the help text of those flags. Maybe help text probably wasn't getting linted before?

MuweiHe (approved):

PR #2021, opened 2023-06-27
[confluent-local] put back and hide old confluent local services commands

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

References
----------

https://confluent.slack.com/archives/CG6BW233L/p1686774180175679

Test & Review
-------------

put back old unit and int tests.

Open Questions / Follow-ups
---------------------------


internal/pkg/cmd/prerunner.go
```diff
@@ -888,6 +888,12 @@ func warnIfConfluentLocal(cmd *cobra.Command) {
 	if strings.HasPrefix(cmd.CommandPath(), "confluent local kafka start") {
 		output.ErrPrintln("The local commands are intended for a single-node development environment only, NOT for production usage. See more: https://docs.confluent.io/current/cli/index.html")
 		output.ErrPrintln()
+		return
+	}
+	if strings.HasPrefix(cmd.CommandPath(), "confluent local") && !strings.HasPrefix(cmd.CommandPath(), "confluent local kafka") {
+		output.ErrPrintln("The local commands are intended for a single-node development environment only, NOT for production usage. See more: https://docs.confluent.io/current/cli/index.html")
+		output.ErrPrintln("As of Confluent Platform 8.0, Java 8 is no longer supported.")
```
sgagniere, line 895: Wait, was this the old message?

CP 8.0 isn't out yet I think so this might be confusing.

internal/cmd/local/command_service.go
```diff
@@ -0,0 +1,765 @@
...
+		c.AddCommand(NewConnectConnectorCommand(prerunner))
+		c.AddCommand(NewConnectPluginCommand(prerunner))
+	case "kafka":
+		c.AddCommand(NewKafkaConsumeCommand(prerunner))
+		c.AddCommand(NewKafkaProduceCommand(prerunner))
+	case "schema-registry":
+		c.AddCommand(NewSchemaRegistryACLCommand(prerunner))
+	}
+
+	return c.Command
```
sgagniere, line 60: Do we need `c.Hidden = true` on this command and/or its subcommands? I see it on `current` `destroy` and `version` but I didn't see it here.

internal/pkg/cmd/prerunner.go
brianstrauch: I think we wanted to warn people in advance?

brianstrauch: https://github.com/confluentinc/cli/pull/1513

sgagniere: That makes sense, but then I would suggest something like `Java 8 will no longer be supported in Confluent Platform 8.0.` to suggest that it's something that will happen, and not something that has already happened.

brianstrauch:
```suggestion
		output.ErrPrintln("As of Confluent Platform 8.0, Java 8 will no longer be supported.")
```

internal/cmd/local/command_service.go
MuweiHe: it's a child cmd of services so it's already hidden

sgagniere: The highlighting is kind of misleading, I was wondering if services is hidden, not schema registry acl.

MuweiHe: hh yes, service is a child of services, so both hidden

sgagniere: Oh, I see, I misunderstood!

sgagniere (approved):

brianstrauch:
This change should probably be added under "Bug Fixes". The current way will cause us to release v4.0.0 next release 😬

PR #2022, opened 2023-06-27
[RCCA-10621] Stop converting resource IDs to numeric IDs for `kafka acl [create|delete|list]`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Reopening #1678. No longer convert resource IDs to numeric IDs when creating ACLs, but continue to convert numeric IDs to resource IDs when listing ACLs.


test/kafka_test.go
```diff
@@ -87,16 +87,13 @@ func (s *CLITestSuite) TestKafka() {
 		{args: "kafka acl list --cluster lkc-acls", fixture: "kafka/acl/list-cloud.golden"},
 		{args: "kafka acl list --cluster lkc-acls -o json", fixture: "kafka/acl/list-json-cloud.golden"},
 		{args: "kafka acl list --cluster lkc-acls -o yaml", fixture: "kafka/acl/list-yaml-cloud.golden"},
-		{args: "kafka acl list --principal User:12345", fixture: "kafka/acl/err-numeric-id.golden", exitCode: 1},
```
MuweiHe, line 90: is `principal` no longer supported?

brianstrauch: It is, but principals containing numeric IDs are not.

MuweiHe: would be good to still have a few `principal` tests?

brianstrauch: We still do!

MuweiHe (approved):

PR #2024, opened 2023-06-28
Autocomplete `api-key create --resource`

brianstrauch:
Release Notes
-------------
New Features
- Autocompletion for `confluent api-key create --resource`

Checklist
---------
   * yes: ok

What
----
Autocomplete the `--resource` flag with "cloud", the list of Kafka cluster IDs, KSQL cluster IDs, and the Schema Registry cluster ID (if it exists). Improve examples to include "cloud" and Schema Registry API keys.

```
% confluent api-key create --resource
lkc-8mm8z0     -- my-kafka-cluster
lksqlc-wzzvnm  -- my-ksql-cluster
lsrc-6oomr2    -- Stream Governance Package
cloud
```

Test & Review
-------------
Integration test

### DIFF
--- a/internal/cmd/api-key/command_use.go
+++ b/internal/cmd/api-key/command_use.go
@@ -21,9 +21,9 @@ func (c *command) newUseCommand() *cobra.Command {
 		RunE:              c.use,
 	}
 
-	cmd.Flags().String(resourceFlagName, "", "The resource ID.")
+	c.addResourceFlag(cmd)
 
-	cobra.CheckErr(cmd.MarkFlagRequired(resourceFlagName))
+	cobra.CheckErr(cmd.MarkFlagRequired("resource"))
 
 	return cmd
 }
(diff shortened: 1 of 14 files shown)


internal/cmd/api-key/command_use.go
```diff
@@ -21,9 +21,9 @@ func (c *command) newUseCommand() *cobra.Command {
 		RunE:              c.use,
 	}
 
-	cmd.Flags().String(resourceFlagName, "", "The resource ID.")
+	c.addResourceFlag(cmd)
```
sgagniere, line 24: It seems like this could end up suggesting invalid values for this flag (like sr clusters) which will result in `functionality not yet available for non-Kafka cluster resources`.

Is this something we want to prevent, or is it even worth preventing?

brianstrauch: I'm planning on dropping this flag altogether soon, since stored API keys are already associated with a Kafka cluster (and hopefully other cluster types in the future) and the user shouldn't need to supply it. For the time-being, I think the autocompletion is more helpful than it is unhelpful, so I'd like to keep it!

sgagniere: Sounds good to me.

sgagniere (approved):
Thanks!

PR #2029, opened 2023-06-29
FLINKCC-100: Create compute pool with wrong-env should fail with descriptive error

pgaref:
Checklist
---------
   * yes: ok

What
----
Make sure all Flink related commands that get --environment as an argument (create, delete, describe and list compute pool) fail with a descriptive message, for example:

```
Tested on STAG: env-mzoyn2  does not exist

mbpro16 :: confluentinc/cli » ./dist/confluent_darwin_amd64_v1/confluent  flink compute-pool list --environment env-mzoyn2                                                    
Error: env-mzoyn2 not found or access forbidden: Forbidden Access: 403 Forbidden

Suggestions:
    List available env-mzoyn2s with `confluent env-mzoyn2 list`.
```

### DIFF
--- a/internal/pkg/ccloudv2/flink.go
+++ b/internal/pkg/ccloudv2/flink.go
@@ -29,12 +29,12 @@ func (c *Client) CreateFlinkComputePool(computePool flinkv2.FcpmV2ComputePool) (
 
 func (c *Client) DeleteFlinkComputePool(id, environment string) error {
 	httpResp, err := c.FlinkClient.ComputePoolsFcpmV2Api.DeleteFcpmV2ComputePool(c.flinkApiContext(), id).Environment(environment).Execute()
-	return errors.CatchCCloudV2Error(err, httpResp)
+	return errors.CatchCCloudV2ResourceNotFoundError(err, environment, httpResp)
 }
 
 func (c *Client) DescribeFlinkComputePool(id, environment string) (flinkv2.FcpmV2ComputePool, error) {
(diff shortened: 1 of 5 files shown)


YannickPferr (approved):
LGTM! Any reason you didn't do it for DELETE and CREATE IAM binding though?

pgaref:
Hey @brianstrauch can I have a +1 please?

internal/pkg/ccloudv2/flink.go
```diff
@@ -88,5 +88,5 @@ func (c *Client) ListFlinkIAMBindings(environmentId, region, cloud, identityPool
 		req = req.IdentityPool(identityPoolId)
 	}
 	res, httpResp, err := req.Execute()
-	return res.GetData(), errors.CatchCCloudV2Error(err, httpResp)
+	return res.GetData(), errors.CatchOrgV2ResourceNotFoundError(err, environmentId, httpResp)
```
brianstrauch, line 91: The function is called CatchOrgV2... but it's being used in a FlinkV2 function. Can we rename it since we're using it for more than just OrgV2 now?

brianstrauch:
```suggestion
	return res.GetData(), errors.CatchCCloudV2ResourceNotFoundError(err, environmentId, httpResp)
```

pgaref: Of course! thanks for the suggestion @brianstrauch ! Updated!

sgagniere (approved):

PR #2030, opened 2023-06-29
Feat/kfs 948 kfs 949 refactoring

YannickPferr:
Checklist
---------
   * yes: ok

What
----

- split up input controller and move main readEvalPrintLoop to application.go
- split out table rendering logic into the table view and make the table controller just a intermediary between the view and the result_fetcher
- improved naming
- we now have a basic and an interactive output controller that implement the same interface


internal/pkg/flink/app/application.go
```diff
@@ -1,62 +1,165 @@
...
+func restoreStdin(state *term.State) {
+	if state != nil {
+		_ = term.Restore(int(os.Stdin.Fd()), state)
+	}
+}
+
+func (a *Application) readEvalPrintLoop() {
+	for a.isAuthenticated() {
+		userInput := a.inputController.GetUserInput()
+		if a.inputController.IsSpecialInput(userInput) {
```
Guttz, line 119: I don't love the name `isSpecialInput` but don't have a better suggestion. But we should add some comments of top of the function since its name doesn't say a lot.

internal/pkg/flink/internal/controller/input_controller.go
```diff
@@ -32,258 +22,85 @@ type InputController struct {
...
+		c.InitialBuffer = ""
 	}
-	return true
+	return c.prompt.Input()
 }
 
-func renderMsgAndStatus(processedStatement *types.ProcessedStatement) {
-	if processedStatement == nil {
-		return
+func (c *InputController) IsSpecialInput(userInput string) bool {
```
Guttz, line 54: Add some comments here because these are kinda of hacks and might be not so intuitive to understand

internal/pkg/flink/app/application.go
```diff
@@ -1,62 +1,165 @@
...
+func (a *Application) isAuthenticated() bool {
+	if authErr := a.tokenRefreshFunc(); authErr != nil {
+		utils.OutputErrf("Error: %v\n", authErr)
+		a.appController.ExitApplication()
+		return false
 	}
+	return true
+}
+
+func (a *Application) fetchInitialResults(executedStatement types.ProcessedStatement) (*types.ProcessedStatement, *types.StatementError) {
```
Guttz, line 147: I like that we differentiate both 👍 

Maybe `fetchFirstPageResults` would be a better name? what do you think

Guttz (approved):
Gj! I think the code has a better structure now :)

Approving since my comments are mostly suggestions 👍

internal/pkg/flink/internal/controller/input_controller.go
YannickPferr: Ah right, especially the empty string return from go-prompt right?

internal/pkg/flink/app/application.go
YannickPferr: yeah I think that sounds good! will rename 👍

internal/pkg/flink/app/application.go
```diff
@@ -1,62 +1,165 @@
...
+		basicOutputController:       basicOutputController,
+		tokenRefreshFunc:            synchronizedTokenRefresh(tokenRefreshFunc),
+	}
+	components.PrintWelcomeHeader()
+	app.readEvalPrintLoop()
+}
 
-	// Event handlers
-	app.SetInputCapture(tableController.AppInputCapture)
+func getStdin() *term.State {
```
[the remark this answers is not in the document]
YannickPferr: yeah sounds good to me! we also have a utils folder now, so I can put it there 👍

internal/pkg/flink/app/application.go
YannickPferr: Hmmm in that case I'll try to think of a better name, because I think a better name would make the comment unnecessary!

internal/pkg/flink/app/application.go
YannickPferr: actually just realized that function might be a bit unnecessary, since it's only used in that one spot.

internal/pkg/flink/app/application.go
YannickPferr: split up the method now, which should make it a lot clearer what's happening 👍

PR #2035, opened 2023-06-29
Fix/kfs 968 tackle zombie queries

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This should fix the issue where some statements would not be deleted correctly. It also fixes future issues, where a request to the Gateway would be made with an invalid token. 

We now ensure that before making a request to the Gateway we at least tried to refresh the token.

### DIFF
--- a/internal/pkg/flink/internal/store/store.go
+++ b/internal/pkg/flink/internal/store/store.go
@@ -32,6 +32,14 @@ type Store struct {
 	exitApplication func()
 	client          ccloudv2.GatewayClientInterface
 	appOptions      *types.ApplicationOptions
+	authenticated   func() error
+}
+
+func (s *Store) authenticatedGatewayClient() ccloudv2.GatewayClientInterface {
+	if authErr := s.authenticated(); authErr != nil {
+		log.CliLogger.Warnf("Failed to refresh token: %v", authErr)
+	}
+	return s.client
 }
 
 func (s *Store) ProcessLocalStatement(statement string) (*types.ProcessedStatement, *types.StatementError) {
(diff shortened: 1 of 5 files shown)


internal/pkg/flink/internal/store/store.go
```diff
@@ -243,12 +252,13 @@ func extractPageToken(nextUrl string) (string, error) {
...
 }
 
-func NewStore(client ccloudv2.GatewayClientInterface, exitApplication func(), appOptions *types.ApplicationOptions) StoreInterface {
+func NewStore(client ccloudv2.GatewayClientInterface, exitApplication func(), appOptions *types.ApplicationOptions, authenticated func() error) StoreInterface {
 	return &Store{
 		Properties:      appOptions.GetDefaultProperties(),
 		client:          client,
 		exitApplication: exitApplication,
 		appOptions:      appOptions,
+		authenticated:   authenticated,
```
swist, line 261: this should be `tokenRefreshFunc`

internal/pkg/flink/internal/store/store.go
```diff
@@ -32,6 +32,14 @@ type Store struct {
 	exitApplication func()
 	client          ccloudv2.GatewayClientInterface
 	appOptions      *types.ApplicationOptions
+	authenticated   func() error
+}
+
+func (s *Store) authenticatedGatewayClient() ccloudv2.GatewayClientInterface {
+	if authErr := s.authenticated(); authErr != nil {
+		log.CliLogger.Warnf("Failed to refresh token: %v", authErr)
```
swist, line 40: I think this should error out. If we can't a good gateway client then we can't do anything?

swist (changes requested):

internal/pkg/flink/internal/store/store.go
YannickPferr: Yes agree, but when this fails to refresh the token, the following request to the GW will error anyways -> and then we return  a proper error  to the user. I think that should be enough

internal/pkg/flink/internal/store/store.go
YannickPferr: Renamed!

internal/pkg/flink/internal/store/store.go
swist: sounds good to me

swist (approved):

PR #2040, opened 2023-06-29
Remove extra API call from `kafka acl list`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
We no longer need to get the list of users, since we can mandate that the "list" endpoint returns resource IDs instead of numeric IDs by passing `UserV2:*` to the backend.

Test & Review
-------------
Manual testing, although still not done

### DIFF
--- a/internal/pkg/acl/acl.go
+++ b/internal/pkg/acl/acl.go
@@ -3,7 +3,6 @@ package acl
 import (
 	"fmt"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/antihax/optional"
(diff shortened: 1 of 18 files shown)


internal/pkg/acl/acl.go
```diff
@@ -317,41 +314,27 @@ func CreateAclRequestDataToAclData(data *AclRequestDataWithError) cpkafkarestv3.
...
+			ResourceName: acl.GetResourceName(),
+			PatternType:  acl.GetPatternType(),
 		})
 	}
 	list.Filter(listFields)
 	return list.Print()
 }
 
-func PrintACLsWithResourceIdMap(cmd *cobra.Command, acls []*ccstructs.ACLBinding, numericIdToResourceId map[int32]string) error {
+func PrintACLsWithResourceIdMap(cmd *cobra.Command, acls []*ccstructs.ACLBinding) error {
```
sgagniere, line 333: We can probably cut `WithResourceIdMap` from this function name.

brianstrauch: Thanks, looks like this function was a duplicate!

sgagniere (approved):
lgtm

PR #2043, opened 2023-06-30
feat: add statements get command and print out statement details and …

swist:
Release Notes
-------------
New Features
- adds a command for viewing statement details and any exceptions that it might have thrown

Checklist
---------
   * yes: ok

What
----
add statements get command and print out statement details and its exceptions if present

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/flink/command_shell.go
+++ b/internal/cmd/flink/command_shell.go
@@ -149,6 +149,8 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
 
 	jwtValidator := pcmd.NewJWTValidator()
 
+	verbose, _ := cmd.Flags().GetCount("verbose")
+
 	client.StartApp(
 		flinkGatewayClient,
 		c.authenticated(prerunner.Authenticated(c.AuthenticatedCLICommand), cmd, jwtValidator),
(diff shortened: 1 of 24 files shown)


internal/cmd/flink/command_statement_describe.go
```diff
@@ -0,0 +1,52 @@
...
+func (c *command) newStatementGetDetailsCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "describe <name>",
+		Short:             "Get details of a Flink SQL statement.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validStatementArgs),
+		RunE:              c.statementDescribe,
+	}
+
+	pcmd.AddForceFlag(cmd)
```
YannickPferr, line 18: Isn't that only used for delete statements?

internal/cmd/flink/command_statement_exceptions_list.go
```diff
@@ -0,0 +1,55 @@
...
+func (c *command) newStatementExceptionsListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "list <statement name>",
+		Short:             "Get exceptions for a Flink SQL statement.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validStatementArgs),
+		RunE:              c.exceptionsList,
+	}
+
+	pcmd.AddForceFlag(cmd)
```
YannickPferr, line 18: Isn't that only used for delete statements?

YannickPferr (approved):
One question about the forceFlag and I think you need to fix the lint errors, otherwise LGTM!

internal/cmd/flink/command_statement_exceptions_list.go
swist: good catch, removed

PR #2044, opened 2023-06-30
fix: sync access to token refresh function

YannickPferr:
Checklist
---------
   * yes: ok

What
----

The token refresh function is not synchronized, but we potentially access it concurrently (e.g. a user clicks fetch next page, while auto refresh is running)

Test & Review
-------------

Added a test case that should trigger the Go race detector (and should also fail in standard test mode)

### DIFF
--- a/internal/pkg/flink/internal/store/store.go
+++ b/internal/pkg/flink/internal/store/store.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	flinkgatewayv1alpha1 "github.com/confluentinc/ccloud-sdk-go-v2/flink-gateway/v1alpha1"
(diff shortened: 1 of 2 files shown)


internal/pkg/flink/internal/store/store.go
```diff
@@ -30,12 +31,16 @@ type StoreInterface interface {
...
 	Properties       map[string]string
 	exitApplication  func()
+	clientLock       sync.Mutex
 	client           ccloudv2.GatewayClientInterface
 	appOptions       *types.ApplicationOptions
 	tokenRefreshFunc func() error
 }
 
 func (s *Store) authenticatedGatewayClient() ccloudv2.GatewayClientInterface {
+	s.clientLock.Lock()
```
luca-filipponi, line 41: what do you need to be thread safe, the tokenRefresh() of the authenticatedGatewayClient()?

swist (approved):

PR #2050, opened 2023-07-04
feat: add statements get command and print out statement details and …

swist:
Release Notes
-------------
New Features
- adds a command for viewing statement details and any exceptions that it might have thrown

Checklist
---------
   * yes: ok

What
----
add statements get command and print out statement details and its exceptions if present

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/flink/command_statement_describe.go
+++ b/internal/cmd/flink/command_statement_describe.go
@@ -1,0 +1,52 @@
+package flink
+
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+func (c *command) newStatementGetDetailsCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "describe <name>",
+		Short:             "Get details of a Flink SQL statement.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validStatementArgs),
+		RunE:              c.statementDescribe,
+	}
+
+	c.addComputePoolFlag(cmd)
+	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
+	pcmd.AddContextFlag(cmd, c.CLICommand)
+
+	return cmd
+}
+
+func (c *command) statementDescribe(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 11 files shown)


internal/cmd/flink/command_statement_describe.go
```diff
@@ -0,0 +1,52 @@
...
+	client, err := c.GetFlinkGatewayClient()
+	if err != nil {
+		return err
+	}
+
+	statement, err := client.GetStatement(environmentId, args[0], c.Context.LastOrgId)
+	if err != nil {
+		return err
+	}
+	list := output.NewList(cmd)
```
YannickPferr, line 41: Are you using the List intentionally, because you could also use the table struct here, since describe should only return a single statement. Here is the example from compute pool describe:
https://github.com/confluentinc/cli/blob/main/internal/cmd/flink/command_compute_pool_describe.go#L46

test/test-server/flink_gateway_router.go
```diff
@@ -64,8 +65,31 @@ func handleSqlEnvironmentsEnvironmentStatements(t *testing.T) http.HandlerFunc {
 	}
 }
 
+func handleSqlEnvironmentsEnvironmentStatementExceptions(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		statementExceptionCreatedAt := time.Date(2022, 1, 1, 0, 0, 0, 0, time.UTC)
+		statementExceptionCreatedAt2 := time.Date(2022, 1, 1, 0, 0, 0, 0, time.UTC)
```
YannickPferr, line 71: You could use a different year for one of them to prove the sort by timestamp works in the test

YannickPferr (approved):
Some nits, but LGTM

internal/cmd/flink/command_statement_describe.go
```diff
@@ -0,0 +1,52 @@
...
+func (c *command) newStatementGetDetailsCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "describe <name>",
+		Short:             "Get details of a Flink SQL statement.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validStatementArgs),
+		RunE:              c.statementDescribe,
+	}
+
+	c.addComputePoolFlag(cmd)
```
YannickPferr, line 19: One more thing: maybe I'm not overseeing it, but I don't think the compute pool flag is used anywhere

PR #2052, opened 2023-07-05
feat: improve shortcuts

YannickPferr:
Checklist
---------
   * yes: ok

What
----

- shortcut text is now dynamic based on the current auto refresh state and the selected mode
- P is now used to pause/play auto refresh
- U/D is now used to jump to top/bottom row of table
- added constants for shortcut keys


internal/pkg/flink/components/table_view_test.go
```diff
@@ -114,3 +115,36 @@ func (s *TableViewTestSuite) TestTableShouldSetTitle() {
...
+
+func (s *TableViewTestSuite) TestTableShortcutsShouldDisplayPause() {
+	materializedStatementResults := getResultsExample()
+
+	actual := s.tableView.getTableShortcuts(materializedStatementResults, true)
+
+	cupaloy.SnapshotT(s.T(), actual)
+}
+
+func (s *TableViewTestSuite) TestTableShortcutsShouldDisplayChangelogMode() {
```
Guttz, line 135: I get that they're "trying to test" two different things, but this test is the same as `TestTableShortcutsShouldDisplayPlay`. We could maybe reduce them to one.

internal/pkg/flink/internal/controller/interactive_output_controller_test.go
```diff
@@ -59,28 +60,28 @@ func (s *InteractiveOutputControllerTestSuite) TestCloseTableViewOnUserInput() {
 func (s *InteractiveOutputControllerTestSuite) TestToggleTableModeOnUserInput() {
 	s.initMockCalls(&types.MaterializedStatementResults{})
 	s.interactiveOutputController.init()
-	input := tcell.NewEventKey(tcell.KeyRune, 'M', tcell.ModNone)
+	input := tcell.NewEventKey(tcell.KeyRune, rune(components.ToggleTableModeShortcut[0]), tcell.ModNone)
```
Guttz, line 63: Nice that these aren't hardcoded anymore

internal/pkg/flink/internal/controller/interactive_output_controller_test.go
```diff
@@ -127,10 +128,7 @@ func getResultsExample() *types.MaterializedStatementResults {
 func (s *InteractiveOutputControllerTestSuite) initMockCalls(materializedStatementResults *types.MaterializedStatementResults) {
 	s.resultFetcher.EXPECT().SetAutoRefreshCallback(gomock.Any())
 	s.resultFetcher.EXPECT().ToggleAutoRefresh()
-	s.resultFetcher.EXPECT().GetMaterializedStatementResults().Return(materializedStatementResults)
-	s.resultFetcher.EXPECT().IsTableMode().Return(materializedStatementResults.IsTableMode())
-	s.resultFetcher.EXPECT().GetFetchState().Return(types.Paused)
-	s.resultFetcher.EXPECT().IsAutoRefreshRunning().Return(false)
+	s.updateTableMockCalls(materializedStatementResults)
```
Guttz, line 131: 👍

Guttz (approved):

internal/pkg/flink/components/table_view_test.go
YannickPferr: yeah makes sense, I merged the 4 tests into 2!

PR #2053, opened 2023-07-05
feat: pause refresh when user triggers scroll event

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Automatically pause the auto refresh when a user triggers a scroll event (arrow up/down, jump keys U/D)

**Note: merge after   #2052 **

### DIFF
--- a/internal/pkg/flink/internal/controller/interactive_output_controller.go
+++ b/internal/pkg/flink/internal/controller/interactive_output_controller.go
@@ -113,14 +113,17 @@ func (t *InteractiveOutputController) inputHandlerTableView(event *tcell.EventKe
 		}
 		return nil
 	case tcell.KeyEscape:
-		t.exitTViewMode()
-		return nil
+		fallthrough
 	case tcell.KeyCtrlQ:
 		t.exitTViewMode()
 		return nil
 	case tcell.KeyEnter:
 		t.renderRowView()
 		return nil
+	case tcell.KeyUp:
+		fallthrough
+	case tcell.KeyDown:
+		return t.handleKeyUpOrDownPress(event)
 	}
 	return event
 }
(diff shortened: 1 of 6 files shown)


internal/pkg/flink/internal/controller/interactive_output_controller_test.go
```diff
@@ -243,3 +241,99 @@ func (s *InteractiveOutputControllerTestSuite) TestTableTitleDisplaysPageSizeAnd
 
 	cupaloy.SnapshotT(s.T(), actual)
 }
+
+func (s *InteractiveOutputControllerTestSuite) TestArrowUpOrDownStopsFetchWhenAutoRefreshIsRunning() {
```
Guttz, line 245: Maybe `TestArrowUpOrDownToggleAutoRefreshWhenAutoRefreshIsRunning` would be a bit more precise, right? We don't/can't really check if it stops.

internal/pkg/flink/internal/controller/interactive_output_controller.go
```diff
@@ -114,30 +113,33 @@ func (t *InteractiveOutputController) inputHandlerTableView(event *tcell.EventKe
...
 		return t.renderAfterAction(t.resultFetcher.ToggleTableMode)
-	case "A":
+	case components.ToggleAutoRefreshShortcut:
 		return t.renderAfterAction(t.resultFetcher.ToggleAutoRefresh)
-	case "H":
-		return t.tableView.FastScrollUp
-	case "L":
-		return t.tableView.FastScrollDown
+	case components.JumpUpShortcut:
+		return t.stopAutoRefreshOrScroll(t.tableView.FastScrollUp)
```
Guttz, line 140: Nice, but hopefully we won't add more layers to this func stacking we have happening here haha

Guttz (approved):

internal/pkg/flink/internal/controller/interactive_output_controller.go
YannickPferr: haha yeah, I just want to avoid the case blocks getting too big by just doing only one thing

internal/pkg/flink/internal/controller/interactive_output_controller_test.go
YannickPferr: Renamed it!

PR #2054, opened 2023-07-05
fixing panic in kafka client-config create

kevin-wu24:
Release Notes
-------------
Bug Fixes

- Prevents panic when trying to run `kafka client-config create` commands

Checklist
---------
   * yes: ok

What
----
Fixes a nil dereference panic that can occur when `kafka client-config create` is run if Schema Registry is not enabled for the current environment

References
----------
https://confluentinc.atlassian.net/browse/CLI-2574

### DIFF
--- a/internal/pkg/dynamic-config/dynamic_context.go
+++ b/internal/pkg/dynamic-config/dynamic_context.go
@@ -183,7 +183,7 @@ func (d *DynamicContext) SchemaRegistryCluster(cmd *cobra.Command) (*v1.SchemaRe
 			if len(srClusters) != 0 {
 				cluster = makeSRCluster(&srClusters[0])
 			} else {
-				cluster = nil
+				return nil, errors.NewSRNotEnabledError()
 			}
 			clusterChanged = true
 		}
--- a/internal/pkg/dynamic-config/dynamic_context.go
+++ b/internal/pkg/dynamic-config/dynamic_context.go
@@ -146,7 +146,7 @@ func (d *DynamicContext) UseAPIKey(apiKey string, clusterId string) error {
 }
 
 // SchemaRegistryCluster returns the SchemaRegistryCluster of the Context,
-// or an empty SchemaRegistryCluster if there is none set,
+// or a SRNotEnabledError if there is none set,
 // or an ErrNotLoggedIn if the user is not logged in.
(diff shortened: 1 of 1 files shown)


internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -183,7 +183,7 @@ func (d *DynamicContext) SchemaRegistryCluster(cmd *cobra.Command) (*v1.SchemaRe
 			if len(srClusters) != 0 {
 				cluster = makeSRCluster(&srClusters[0])
 			} else {
-				cluster = nil
+				return nil, errors.NewSRNotEnabledError()
```
MuweiHe, line 186: if returning right here, it'll pass updating the config starting line 193, when the cluster is missing details but not nil. can you provide more testing / assessment / reasoning that we can just return here?

kevin-wu24: Yeah, looks like we should just store the error instead and allow the updating the config to happen.

sgagniere, line 186: There's a comment at the start of this function saying that it returns an empty SchemaRegistryCluster if none is set. I think we should probably change that to say that this will return an `SRNotEnabledError` if none is set.

I skimmed over the places in the code that call this function and it looks like all of them expect a non-nil SR cluster, and some of them even looked for `SRNotEnabledError`s; can you double check before merge?

kevin-wu24: Sounds good, changed the comment.

Double checked the usages, whenever the function returns a nil SR cluster, it's always coupled with an error so we should be fine there.

sgagniere (approved):
lgtm, but this is a bug fix so please add this back to top of the PR description before you merge and fill out the Bug Fixes section:
```
Release Notes
-------------
<!--
If this PR introduces any user-facing changes, please document them below. Please delete any unused section titles and placeholders.
Please match the style of previous release notes: https://docs.confluent.io/confluent-cli/current/release-notes.html
-->

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER
```

MuweiHe (approved):

PR #2056, opened 2023-07-06
feat: adjust logic when to show basic vs. interactive table

YannickPferr:
Checklist
---------
   * yes: ok

What
----

We now use the interactive table when either of these cases is true for a non-local statement:
- the results include a next page token
- the statement is a SELECT statement 
- the results have more than 3 columns

In all other cases the basic table is used.

**Note: we need the additional isSelectStatement check before deciding based on the number of columns, because of statements like this one:  `SELECT NOW(), COUNT(*) FROM (VALUES (1), (2), (3));`. This is a statement with retractions, but it only has two columns. You definitely want to use the interactive mode here though, so you can check out the results in both the changelog and the table mode. Because you never now what you'll get with a SELECT statement, it is safer to just always use the interactive table.**

### DIFF
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -13,6 +13,8 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/flink/types"
 )
 
+const minNumColumnsToUseInteractiveTable = 4
+
 type Application struct {
 	history                     *history.History
 	store                       types.StoreInterface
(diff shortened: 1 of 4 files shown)


internal/pkg/flink/app/application.go
```diff
@@ -118,10 +120,10 @@ func (a *Application) getOutputController(processedStatementWithResults types.Pr
 	if processedStatementWithResults.IsLocalStatement {
 		return a.basicOutputController
 	}
-	if processedStatementWithResults.PageToken != "" {
+	if processedStatementWithResults.PageToken != "" || processedStatementWithResults.IsSelectStatement {
```
Guttz, line 123: nice

Guttz (approved):

PR #2057, opened 2023-07-06
feat: make results stabler

swist:
Checklist
---------
   * yes: ok

What
----
In best case we expect the retraction for a row to be immediately followed by the update. To limit the number of shuffles we get as a result of new updates coming in, we maintain a queue of elements that are retracted and fill these first with incoming updates.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/flink/internal/controller/.snapshots/TestInteractiveOutputControllerTestSuite-TestTableTitleDisplaysPageSizeAndCacheSizeWithUnsafeTrace
+++ b/internal/pkg/flink/internal/controller/.snapshots/TestInteractiveOutputControllerTestSuite-TestTableTitleDisplaysPageSizeAndCacheSizeWithUnsafeTrace
@@ -1,1 +1,1 @@
- Table mode (auto refresh 1.0s) | last page size: 10 | current cache size: 10/10 | table size: 10 
+ Table mode (auto refresh 1.0s) | last page size: 10 | current cache size: 10/10 | table size: 10 | pending retraction: 0 
--- a/internal/pkg/flink/internal/controller/interactive_output_controller.go
+++ b/internal/pkg/flink/internal/controller/interactive_output_controller.go
@@ -194,13 +194,14 @@ func (t *InteractiveOutputController) getTableTitle() string {
 
 	if t.debug {
 		return fmt.Sprintf(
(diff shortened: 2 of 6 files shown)


internal/pkg/flink/internal/controller/.snapshots/TestInteractiveOutputControllerTestSuite-TestTableTitleDisplaysPageSizeAndCacheSizeWithUnsafeTrace
```diff
@@ -1 +1 @@
- Table mode (auto refresh 1.0s) | last page size: 10 | current cache size: 10/10 | table size: 10 
+ Table mode (auto refresh 1.0s) | last page size: 10 | current cache size: 10/10 | table size: 10 | pending retraction: 0
```
brianstrauch, line 1: Nit: Any reason the first column is capitalized, but the others aren't?

brianstrauch: Also, out of curiosity what does "pending retraction" mean?

swist: Not really. Retraction is a flink specific term telling us “this row is no longer valid, they may be a replacement coming soon”

swist: First column is the only one you see outside of debug mode

sgagniere:
This is for Flink, right? Please remove the Release Notes section in the PR descriptions in these PR's since Flink is still EA, thanks!

brianstrauch:
@swist do you still plan to add this?

brianstrauch:
> @swist do you still plan to merge this?

Bump on the above @swist

swist:
This was already fixed in another way

PR #2058, opened 2023-07-06
feat: persist results of `USE` statements in configs

swist:
Checklist
---------
   * yes: ok

What
----

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/config/v1/context.go
+++ b/internal/pkg/config/v1/context.go
@@ -263,6 +263,40 @@ func (c *Context) SetCurrentFlinkComputePool(id string) error {
 	return nil
 }
 
+func (c *Context) GetCurrentFlinkCatalog() string {
+	if ctx := c.GetCurrentEnvironmentContext(); ctx != nil {
+		return ctx.CurrentFlinkCatalog
+	}
+	return ""
+}
+
+func (c *Context) SetCurrentFlinkCatalog(id string) error {
+	ctx := c.GetCurrentEnvironmentContext()
+	if ctx == nil {
+		return fmt.Errorf("no environment found")
+	}
+
+	ctx.CurrentFlinkCatalog = id
+	return nil
+}
+
+func (c *Context) GetCurrentFlinkDatabase() string {
+	if ctx := c.GetCurrentEnvironmentContext(); ctx != nil {
+		return ctx.CurrentFlinkDatabase
+	}
+	return ""
+}
+
+func (c *Context) SetCurrentFlinkDatabase(id string) error {
+	ctx := c.GetCurrentEnvironmentContext()
+	if ctx == nil {
+		return fmt.Errorf("no environment found")
+	}
+
+	ctx.CurrentFlinkDatabase = id
+	return nil
+}
+
 func (c *Context) GetCurrentIdentityPool() string {
 	if ctx := c.GetCurrentEnvironmentContext(); ctx != nil {
 		return ctx.CurrentIdentityPool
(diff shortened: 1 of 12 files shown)


luca-filipponi (approved):
LGTM

internal/pkg/flink/internal/store/store.go
```diff
@@ -256,25 +256,23 @@ func NewStore(client ccloudv2.GatewayClientInterface, exitApplication func(), ap
 // Set properties default values if not set by the user
 // We probably want to refactor the keys names and where they are stored. Maybe also the default values.
 func (s *Store) propsDefault(propsWithoutDefault map[string]string) map[string]string {
-	properties := make(map[string]string)
+	properties := map[string]string{
+		config.ConfigKeyCatalog:          s.appOptions.GetEnvironmentId(),
+		config.ConfigKeyDatabase:         s.appOptions.GetKafkaClusterId(),
+		config.ConfigKeyOrgResourceId:    s.appOptions.GetOrgResourceId(),
+		config.ConfigKeyExecutionRuntime: "streaming",
```
Guttz, line 263: We needed this some time ago for the demo but I'm pretty sure we can delete this now.

Guttz (changes requested):
Catalog/db doesn't seem to be persisted. Repro steps: 

image

Expected for set to show the db/catalog set in the previous session.

internal/pkg/config/v1/context.go
```diff
@@ -263,6 +263,40 @@ func (c *Context) SetCurrentFlinkComputePool(id string) error {
 	return nil
 }
 
+func (c *Context) GetCurrentFlinkCatalog() string {
+	if ctx := c.GetCurrentEnvironmentContext(); ctx != nil {
+		return ctx.CurrentFlinkCatalog
```
brianstrauch, line 268: Just for my understanding, what is a catalog and where can I read more about it?

internal/pkg/flink/internal/store/store_utils.go
```diff
@@ -110,6 +110,24 @@ func (s *Store) processUseStatement(statement string) (*types.ProcessedStatement
 	}
 
 	s.Properties[configKey] = configVal
+	if s.appOptions.GetContext() != nil {
+		if configKey == config.ConfigKeyCatalog {
+			err := s.appOptions.Context.SetCurrentFlinkCatalog(configVal)
+			if err != nil {
+				return nil, &types.StatementError{Message: err.Error()}
```
brianstrauch, line 117: Every error in this function is wrapped in the same way. Can we wrap it outside of the function, or remove duplication some other way?

internal/pkg/config/v1/context.go
swist: https://nightlies.apache.org/flink/flink-docs-release-1.14/docs/dev/table/catalogs/

internal/pkg/flink/internal/store/store_utils.go
swist: factoring out the error wrap won't dry it up - extra complexity from having a method that literally wraps the repeating return line + it's going to make it harder to read (when you look at statement error now you need to do a jump to find out what the message is) I'm inclined not to touch it for now

internal/pkg/flink/internal/store/.snapshots/TestProcessSetStatement-should_return_all_keys_and_values_from_config_if_configKey_is_empty_after_updates
```diff
@@ -0,0 +1,99 @@
...
+          Value: (string) ""
+        }
+      }
+    },
+    (types.StatementResultRow) {
+      Operation: (types.StatementResultOperation) +I,
+      Fields: ([]types.StatementResultField) (len=2) {
+        (types.AtomicStatementResultField) {
+          Type: (types.StatementResultFieldType) (len=7) "VARCHAR",
+          Value: (string) (len=13) "pipeline.name"
```
Guttz, line 64: Maybe update these old tests to use something like `table.results-timeout` or other key instead of `pipeline.name` just to avoid confusion. The latter doesn't exist anymore, and the user can't pick the pipeline name

Guttz (approved):
Nice 👍

PR #2059, opened 2023-07-06
Refactor/KFS-984 and KFS-982 update out msg and timeout accounts for gateway response time

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

- Adjusted the message sent while waiting for results to be ready
- Also print status
- Print complete/running status after statement is ready
- properly account for gateway response time when calculating timeout

References
----------

Test & Review
-------------

Tests updated

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementPrintsUserInfo
+++ b/internal/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementPrintsUserInfo
@@ -1,5 +1,5 @@
 Statement name: test-statement
 Statement successfully submitted.
-Fetching results...
+Waiting for statement to be ready. Statement phase is .
 status detail message.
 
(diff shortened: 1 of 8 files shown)


YannickPferr (approved):

internal/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementPrintsUserInfo
```diff
@@ -1,5 +1,6 @@
 Statement name: test-statement
 Statement successfully submitted.
-Fetching results...
+Waiting for statement to be ready. Statement phase is COMPLETED.
+Statement phase is COMPLETED.
```
brianstrauch, line 4: Any reason this is printed twice?

internal/pkg/flink/internal/store/store_test.go
```diff
@@ -130,6 +131,39 @@ func TestWaitForPendingTimesout(t *testing.T) {
 	assert.Nil(t, processedStatement)
 }
 
+func TestWaitForPendingHitsErrorRetryLimit(t *testing.T) {
+	statementName := "statementName"
+	timeout := time.Duration(10) * time.Second
```
brianstrauch, line 136:
```suggestion
	timeout := 10 * time.Second
```

internal/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementPrintsUserInfo
Guttz: Oh, true. The test was just non-realistic, returning completed for a select query - selects always return pending. I've updated the test.

PR #2065, opened 2023-07-06
[CLI-2573] Package linux/arm64 binary for CP

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Re-attempting to package linux/arm64 for CP, reverts https://github.com/confluentinc/cli/pull/1533

References
----------
https://confluentinc.atlassian.net/browse/CLI-2573


MuweiHe:
How will we find out if it works?

brianstrauch:
> How will we find out if it works?

A nightly release gets packaged every night, and we'll check to see if this binary was successfully added tomorrow.

DABH (approved):

PR #2068, opened 2023-07-06
Deprecate `confluent api-key use --resource` flag

brianstrauch:
Release Notes
-------------
New Features
- No longer require passing the `--resource` flag to `confluent api-key use`

Checklist
---------
   * yes: ok

What
----
Successfully running `api-key use` requires that the API key has been stored locally (`api-key create --resource` or `api-key store --resource` was used). Therefore, we have enough information to infer which resource owns a given API key when using the `api-key use` command. The new implementation is O(n) instead of O(1), but we don't expect users to have so many Kafka clusters that searching through all of them for an API key takes a meaningful amount of time.

In general, this makes it much easier to provision and use API keys! (and therefore, easier to produce/consume for Confluent Cloud)

Test & Review
-------------
Updated integration tests


internal/cmd/api-key/command_list.go
```diff
@@ -52,13 +52,10 @@ func (c *command) newListCommand() *cobra.Command {
...
 	c.setKeyStoreIfNil()
 
-	resourceType, clusterId, currentKey, err := c.resolveResourceId(cmd, c.V2Client)
+	clusterId, currentKey, err := c.resolveResourceId(cmd, c.V2Client)
 	if err != nil {
 		return err
 	}
-	if resourceType == resource.Cloud {
-		clusterId = resource.Cloud
-	}
```
sgagniere, line 61: It looks like this change will cause clusterId to be empty since `resolveResourceId` still breaks out of the switch in the Cloud case without setting the clusterId; does this affect the `c.V2Client.ListApiKeys` call further down in this file?

sgagniere (approved):

PR #2070, opened 2023-07-07
[KFS-987] feat: add info bar and padding row to table view 

YannickPferr:
Checklist
---------
   * yes: ok

What
----

- adds an empty row add the bottom of the table to indicate when you've reached the end
- adds an info bar at the bottom that displays the current fetch state, the selected row (if refresh is paused) and the last refresh timestamp
- also ensures the jump up/down keeps a focus point (the row you've selected will be at the bottom/top of the table after jumping up/down)

image
image


internal/pkg/flink/components/table_view.go
```diff
@@ -12,15 +13,27 @@ import (
...
 )
 
+type TableViewInterface interface {
+	Init()
+	GetFocusableElement() *tview.Table
+	GetRoot() tview.Primitive
+	GetSelectedRow() *types.StatementResultRow
+	RenderTable(tableTitle string, statementResults *types.MaterializedStatementResults, isAutoRefreshRunning bool, lastRefreshTimestamp *time.Time, fetchState types.FetchState)
+	FastScrollUp()
+	FastScrollDown()
```
Guttz, line 23: Change these names to match what's shown on the screen

internal/pkg/flink/components/table_view.go
```diff
@@ -73,25 +117,65 @@ func (t *TableView) getSelectedRowIdx() int {
...
 
+func (t *TableView) setTableAndColumnWidths(statementResults *types.MaterializedStatementResults) {
 	_, _, tableWidth, _ := t.table.GetInnerRect()
 	t.tableWidth = tableWidth
-	columnWidths := statementResults.GetMaxWidthPerColumn()
-	truncatedColumnWidths := results.GetTruncatedColumnWidths(columnWidths, tableWidth)
+	t.columnWidths = statementResults.GetMaxWidthPerColumn()
+}
+
+func (t *TableView) tableAfterDrawHandler() func(screen tcell.Screen, x int, y int, width int, height int) (int, int, int, int) {
```
Guttz, line 145: Isn't this just a  `TableDrawFunction` or `TableDrawFunctionWrapper`? Handler is also alright, but it wasn't so intuitive to understand what's that for

internal/pkg/flink/components/.snapshots/TestTableViewTestSuite-TestTableInfoBarWhenTableTableHasNoContentAndAutoRefreshIsOff
```diff
@@ -0,0 +1,5 @@
+([]string) (len=3) {
+  (string) (len=19) "Fetch state: paused",
```
Guttz, line 2: Two things about naming:

Fetch state -> Refresh
paused/running -> Paused / Running

Refresh is the same naming they use in the open search version and I think it's just better and shorter than fetch state.

Paused and Running look a bit weird in lower case, because we always have the first letter in upper case. So I would either write the first letter or the whole word in upper case

Guttz (approved):
In general, it looks great! The flink shell is 🔥 :)

internal/pkg/flink/components/table_view.go
YannickPferr: Hmm it's a callback that's invoked after the table has been drawn, so I think the name is fitting, no?

YannickPferr: Will leave it as it is for now

internal/pkg/flink/components/.snapshots/TestTableViewTestSuite-TestTableInfoBarWhenTableTableHasNoContentAndAutoRefreshIsOff
YannickPferr: Changed all states to be uppercase and the prefix from "Fetch state:" to "Refresh:"!

PR #2071, opened 2023-07-07
KGLOBAL-3449: Support bidirectional links

andymg3:
Release Notes
-------------
New Features
- Bidirectional link support for `confluent kafka link`

Checklist
---------
   * yes: ok

What
----
This adds support for bidirectional cluster links. We introduce a few new flags specifically for bidirectional links: `--remote-cluster`, `--remote-bootstrap-server`, `--remote-api-secret`, `--remote-api-key`, `--local-api-key`, `--local-api-secret`. We added these instead of reusing the existing flags ie `--destination/source-bootstrap-server`, `--destination/source-api-key` etc because with bidirectional links no one side is necessarily the destination or source. 

So for bidirectional links we create these new flags to prevent ambiguity. For now we've decided the "remote" flags will just be for bidirectional links. There are a few reasons for that. First, it would require deprecating the other flags which is painful for customers. @lukeknep has said the other flags are also customer friendly. And lastly, in TF we have structures `destination_kafka_cluster` and `source_kafka_cluster` so if we wanted to deprecate the source/destination flags here'd we probably want to do the same in TF but again that requires a painful deprecation process for customers.

References
----------

Test & Review
-------------
This adds some integration tests and I also did some manual testing

Open Questions / Follow-ups
---------------------------
- Should we consolidate on `remote` and `local` in the future?


internal/cmd/kafka/command_link_create.go
```diff
@@ -80,10 +95,16 @@ func (c *linkCommand) newCreateCommand() *cobra.Command {
 	cmd.Flags().String(sourceBootstrapServerFlagName, "", `Bootstrap server address of the source cluster. Can alternatively be set in the configuration file using key "bootstrap.servers".`)
 	cmd.Flags().String(destinationClusterIdFlagName, "", "Destination cluster ID for source initiated cluster links.")
 	cmd.Flags().String(destinationBootstrapServerFlagName, "", `Bootstrap server address of the destination cluster for source initiated cluster links. Can alternatively be set in the configuration file using key "bootstrap.servers".`)
-	cmd.Flags().String(sourceApiKeyFlagName, "", "An API key for the source cluster. For links at destination cluster, this is used for remote cluster authentication. For links at source cluster, this is used for local cluster authentication. "+authHelperMsg)
-	cmd.Flags().String(sourceApiSecretFlagName, "", "An API secret for the source cluster. For links at destination cluster, this is used for remote cluster authentication. For links at source cluster, this is used for local cluster authentication. "+authHelperMsg)
+	cmd.Flags().String(remoteClusterIdFlagName, "", "Remote cluster ID for bidirectional cluster links.")
```
rajinisivaram, line 98: Can't we allow the new `remote` and `local` options to be used with any link mode?

internal/cmd/kafka/command_link_create.go
```diff
@@ -80,10 +95,16 @@ func (c *linkCommand) newCreateCommand() *cobra.Command {
...
+	cmd.Flags().String(sourceApiSecretFlagName, "", "An API secret for the source cluster. For links at destination cluster this is used for remote cluster authentication. For links at source cluster this is used for local cluster authentication. "+authHelperMsg)
 	cmd.Flags().String(destinationApiKeyFlagName, "", "An API key for the destination cluster. This is used for remote cluster authentication links at the source cluster. "+authHelperMsg)
 	cmd.Flags().String(destinationApiSecretFlagName, "", "An API secret for the destination cluster. This is used for remote cluster authentication for links at the source cluster. "+authHelperMsg)
+	cmd.Flags().String(remoteApiKeyFlagName, "", "An API key for the remote cluster for bidirectional links. This is used for remote cluster authentication. "+authHelperMsg)
+	cmd.Flags().String(remoteApiSecretFlagName, "", "An API secret for the remote cluster for bidirectional links. This is used for remote cluster authentication. "+authHelperMsg)
+	cmd.Flags().String(localApiKeyFlagName, "", "An API key for the local cluster for bidirectional links. This is used for local cluster authentication. "+authHelperMsg)
```
rajinisivaram, line 106: used for local cluster authentication if remote link's connection mode is Inbound.

internal/cmd/kafka/command_link_create.go
```diff
@@ -80,10 +95,16 @@ func (c *linkCommand) newCreateCommand() *cobra.Command {
...
+	cmd.Flags().String(localApiSecretFlagName, "", "An API secret for the local cluster for bidirectional links. This is used for local cluster authentication. "+authHelperMsg)
```
rajinisivaram, line 107: as above, used for local cluster authentication if remote link's connection mode is Inbound.

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +193,124 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+	} else if linkMode == Destination {
+		return c.addDestInitiatedLinkSecurityConfigToMap(cmd, configMap)
+	} else if linkMode == Bidirectional {
+		return c.addBidirectionalInitiatedLinkSecurityConfigToMap(cmd, configMap)
+	} else {
+		return unrecognizedLinkModeErr(linkModeMetadata.linkModeStr)
+	}
+}
+
+func (c *linkCommand) addBidirectionalInitiatedLinkSecurityConfigToMap(cmd *cobra.Command, configMap map[string]string) error {
```
rajinisivaram, line 282: remove `Initiated` from method name?

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +193,124 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+	}
+	localApiKey, err := cmd.Flags().GetString(localApiKeyFlagName)
+	if err != nil {
+		return err
+	}
+	localApiSecret, err := cmd.Flags().GetString(localApiSecretFlagName)
+	if err != nil {
+		return err
+	}
+	if localApiKey != "" || localApiSecret != "" {
```
rajinisivaram, line 304: why is this `||` while the equivalent for remote above is `&&`?

rajinisivaram (approved):
@andymg3 Thanks for the PR, looks good. Left a question about whether we should allow all link modes to specify `local`/`remote`. Can you start a thread on the team slack channel and ping Luke to get alignment before merging the PR? Left a few other minor comments.

internal/cmd/kafka/command_link_create.go
andymg3: Hmm I think it should be `&&` here. Will fix.

internal/cmd/kafka/command_link_create.go
```diff
@@ -133,11 +154,16 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
 	configs := toCreateTopicConfigs(configMap)
 	data := kafkarestv3.CreateLinkRequestData{Configs: &configs}
 
+	linkMode := linkModeMetadata.linkMode
 	if remoteClusterId != "" {
 		if linkMode == Destination {
```
brianstrauch, line 159:
```suggestion
		switch linkModeMetadata.linkMode {
		case Destination:
```

internal/cmd/kafka/command_link_create.go
brianstrauch: +1

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +193,124 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+var sensitiveLinkConfigs = map[string]bool{
```
brianstrauch, line 196: How likely is it that we add more in the future? Will we ever need to retroactively mark configs as sensitive?

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +193,124 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+		if _, ok := sensitiveLinkConfigs[key]; ok {
+			filtered[key] = "***"
+		} else {
+			filtered[key] = val
+		}
+	}
+	return createKeyValuePairs(filtered)
+}
+
+func createKeyValuePairs(m map[string]string) string {
```
brianstrauch, line 213: This looks like the inverse of the `toMap()` function in internal/pkg/properties/properties.go. Can we define this function there and unit test it? It seems like we should be able to repurpose this in other areas of the CLI.

internal/cmd/kafka/command_link_create.go
```diff
@@ -206,44 +320,53 @@ func (c *linkCommand) addSecurityConfigToMap(cmd *cobra.Command, linkMode linkMo
...
+	if err != nil {
+		return err
+	}
+	if sourceApiKey != "" && sourceApiSecret != "" {
+		configMap[localListenerPropertyName] = saslSsl
+		configMap[localSecurityProtocolPropertyName] = saslSsl
+		configMap[localSaslMechanismPropertyName] = plain
+		configMap[localSaslJaasConfigPropertyName] = getJaasValue(sourceApiKey, sourceApiSecret)
+	} else if sourceApiKey != "" || sourceApiSecret != "" {
+		return errors.New("--source-api-key and --source-api-secret must be supplied together")
```
brianstrauch, line 347: You can use the built-in function, `cmd.MarkFlagsRequiredTogether("a", "b")`

internal/cmd/kafka/command_link_create.go
```diff
@@ -266,5 +389,22 @@ func (c *linkCommand) getRemoteClusterMetadata(cmd *cobra.Command, linkMode link
 			return "", "", err
 		}
 		return remoteClusterId, bootstrapServer, nil
+	} else if linkMode == Bidirectional {
```
brianstrauch, line 392: `switch` now that there's more than two types?

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +193,124 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
 	}
 }
 
-func (c *linkCommand) addSecurityConfigToMap(cmd *cobra.Command, linkMode linkMode, configMap map[string]string) error {
+func unrecognizedLinkModeErr(linkModeStr string) error {
+	return errors.Errorf(`unrecognized link.mode "%s". Use %s, %s or %s.`, linkModeStr, DESTINATION, SOURCE, BIDIRECTIONAL)
+}
+
+func (c *linkCommand) addSecurityConfigToMap(cmd *cobra.Command, linkModeMetadata *linkModeMetadata, configMap map[string]string) error {
+	linkMode := linkModeMetadata.linkMode
```
brianstrauch, line 270: `switch`?

internal/cmd/kafka/command_link_create.go
andymg3: It's possible. @lukeknep's feedback was to not have multiple ways of doing the same thing. Personally, I'd prefer to consolidate on only having `remote` and `local` but that entails a deprecation process and questions about how consistent the CLI and terraform are wrt to data modeling.

If we're going to keep `destination` and `source` here and in terraform I'm not sure how valuable it is to allow other link modes to also use `remote` and `local`.

internal/cmd/kafka/command_link_create.go
andymg3: It's a good question. It is of course possible. I think it's unlikely though. @rajinisivaram what do you think?

internal/cmd/kafka/command_link_create.go
andymg3: Made this change myself. Thanks.

internal/cmd/kafka/command_link_create.go
andymg3: Agreed. Changed to `switch` here and a few other places.

internal/cmd/kafka/command_link_create.go
brianstrauch: +1 for consolidating on `remote` and `local` and deprecating `destination` and `source`! Can we do this in Terraform too @linouk23?

internal/cmd/kafka/command_link_create.go
andymg3: I copied this function from a Stack Overflow answer - that's not good enough? ;) Aka will add some unit tests.

internal/cmd/kafka/command_link_create.go
andymg3: Per our last discussion we decided not to because it's quite a lot of effort required of our customers. It's also more work on the terraform implementation side. 

A bit of an unfortunate situation :(

internal/cmd/kafka/command_link_create.go
andymg3: Hm are you sure? GoLand doesnt like that method and it also doesnt compile. I get:
```
  ⨯ build failed after 2s                    error=failed to build for darwin_amd64_v1: exit status 1: # github.com/confluentinc/cli/internal/cmd/kafka
internal/cmd/kafka/command_link_create.go:108:14: cmd.Flags().MarkFlagsRequiredTogether undefined (type *pflag.FlagSet has no field or method MarkFlagsRequiredTogether)

make[1]: *** [cli-builder] Error 1
make: *** [build] Error 2
```

internal/cmd/kafka/command_link_create.go
andymg3: Also to be fair this isn't a one-way door decision. We should be able to move to just using `remote` and `local` in the future if we decide on that. I think we'll want to go with this for now to meet our deadlines.

internal/cmd/kafka/command_link_create.go
andymg3: Fixed.

internal/cmd/kafka/command_link_create.go
andymg3: Moved this method to internal/pkg/properties/properties.go and added some tests

internal/cmd/kafka/command_link_create.go
andymg3: Chatted with @rajinisivaram offline. We'll move to an allow list to be on the safe side.

rajinisivaram (approved):
@andymg3 Thanks for the updates, LGTM

internal/cmd/kafka/command_link_create.go
brianstrauch: > We should be able to move to just using remote and local in the future if we decide on that.

Keep me updated! If we want to do this, it would be great to do it before our v4 release at the end of the year.

internal/cmd/kafka/command_link_create.go
brianstrauch: Instead of `cmd.Flags().MarkFlagsRequiredTogether()` try `cmd.MarkFlagsRequiredTogether()`

brianstrauch: (There should be at least one other example of it being used in the CLI, if you want to search for it)

cmd/lint/main.go
```diff
@@ -87,6 +87,12 @@ var flagRules = []linter.FlagRule{
 			"destination-api-secret",
 			"destination-bootstrap-server",
 			"destination-cluster",
+			"remote-api-key",
+			"remote-api-secret",
+			"local-api-key",
```
brianstrauch, line 92: I don't think this needs to be excluded. Looks like < 20 characters?

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +190,129 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+var printableLinkConfigs = map[string]bool{
```
brianstrauch, line 193: We have `types.Set{}` that might be more user friendly to use here (no need to type `true` a bunch of times)

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +190,129 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+			switch linkModeStr {
+			case DESTINATION:
+				linkMode = Destination
+				linkModeStr = DESTINATION
+			case SOURCE:
+				linkMode = Source
+				linkModeStr = SOURCE
+			case BIDIRECTIONAL:
+				linkMode = Bidirectional
+				linkModeStr = BIDIRECTIONAL
```
brianstrauch, line 252: Redundant, since it's already set.
```suggestion
			case DESTINATION:
				linkMode = Destination
			case SOURCE:
				linkMode = Source
			case BIDIRECTIONAL:
				linkMode = Bidirectional
```

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +190,129 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+		configMap := make(map[string]string)
+		configMap["link.mode"] = DESTINATION
 		// Default is destination if no config file is provided.
-		return make(map[string]string), Destination, nil
+		return configMap, &linkModeMetadata{Destination, DESTINATION}, nil
+	}
+}
+
+func unrecognizedLinkModeErr(linkModeStr string) error {
+	return errors.Errorf(`unrecognized link.mode "%s". Use %s, %s or %s.`, linkModeStr, DESTINATION, SOURCE, BIDIRECTIONAL)
```
brianstrauch, line 268:
```suggestion
	return errors.Errorf(`unrecognized link.mode "%s". Use %s, %s, or %s.`, linkModeStr, DESTINATION, SOURCE, BIDIRECTIONAL)
```

internal/cmd/kafka/command_link_create.go
```diff
@@ -266,5 +391,22 @@ func (c *linkCommand) getRemoteClusterMetadata(cmd *cobra.Command, linkMode link
...
+		}
+		return remoteClusterId, bootstrapServer, nil
+	default:
+		return "", "", unrecognizedLinkModeErr(linkModeMetadata.linkModeStr)
 	}
 }
+
+type linkModeMetadata struct {
+	linkMode    linkMode
+	linkModeStr string
```
brianstrauch, line 411: Naming is a bit redundant, i.e. `linkModeMetadata.linkModeString`
```suggestion
	mode linkMode
	name string
```

internal/cmd/kafka/command_link_create.go
andymg3: Oh never mind I found it.

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +195,125 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+	"auto.create.mirror.topics.enable",
+	"auto.create.mirror.topics.filters",
+	"cluster.link.prefix",
+	"topic.config.sync.ms",
+	"topic.config.sync.include",
+	"mirror.start.offset.spec",
+	"security.protocol",
+	"sasl.mechanism",
+	"local.security.protocol",
+	"local.sasl.mechanism")
```
brianstrauch, line 218:
```suggestion
	"local.sasl.mechanism",
)
```

internal/cmd/kafka/command_link_create.go
```diff
@@ -167,36 +195,125 @@ func (c *linkCommand) create(cmd *cobra.Command, args []string) error {
...
+				linkMode = Source
+			case BIDIRECTIONAL:
+				linkMode = Bidirectional
+			default:
+				return nil, &linkModeMetadata{linkMode, linkModeStr}, unrecognizedLinkModeErr(linkModeStr)
+			}
 		}
-		return configMap, linkMode, nil
+		configMap["link.mode"] = linkModeStr
+		return configMap, &linkModeMetadata{linkMode, linkModeStr}, nil
```
brianstrauch, line 259: I feel like we should only have to return `linkMode` from this function, since each linkMode has an associated string that we can look up later.

brianstrauch (approved):
Thank you!

internal/cmd/kafka/command_link_create.go
andymg3: I didn't find an easy way to associate a String with an enum/iota. Do you know of one? I didn't want to maintain a Map and have to deal with error checking if the lookup fails.

brianstrauch: > I didn't want to maintain a Map and have to deal with error checking if the lookup fails.

This is the best way that I know to do it. You can just ignore any errors since the set of inputs are constrained to what's in the enum, and you should never get any other type.

internal/cmd/kafka/command_link_create.go
andymg3: Done.

internal/cmd/kafka/command_link_create.go
andymg3: Yeah that's def a fair option. My personal preference is to avoid having to worry about ignoring errors so will leave as is even if it is a bit clunky.

internal/cmd/kafka/command_link_create.go
linouk23: > Per our last discussion we decided not to because it's quite a lot of effort required of our customers. It's also more work on the terraform implementation side.

+1

PR #2074, opened 2023-07-07
[CLI-2577] adding feedback command and tests

kevin-wu24:
Release Notes
-------------
New Features
- Add `confluent feedback` command for user-submitted feedback

Checklist
---------

What
----
The feedback command allows a user to submit a string containing their feedback regarding the CLI after first acknowledging they will not sensitive data.

References
----------
https://confluentinc.atlassian.net/wiki/spaces/DPI/pages/3090908967/Reviving+confluent+feedback

Test & Review
-------------
Integration tests


internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,80 @@
...
+	prompt := form.NewPrompt(os.Stdin)
+	return NewFeedbackCmdWithPrompt(cfg, prerunner, prompt)
+}
+
+func NewFeedbackCmdWithPrompt(cfg *v1.Config, prerunner pcmd.PreRunner, prompt form.Prompt) *cobra.Command {
+	cmd := pcmd.NewAnonymousCLICommand(
+		&cobra.Command{
+			Use:   "feedback",
+			Short: "Submit feedback about the " + pversion.FullCLIName + ".",
+			RunE: func(cmd *cobra.Command, _ []string) error {
```
brianstrauch, line 29: Please move into a separate function. We should match the style of our other commands.

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			Args: cobra.NoArgs,
+		},
+		prerunner)
+
+	return cmd.Command
+}
+
+func getFeedback(prompt form.Prompt) (string, error) {
+	fmt.Println("Please confirm that your feedback will not contain any " +
```
brianstrauch, line 50:
```suggestion
	output.Println("Please confirm that your feedback will not contain any " +
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,80 @@
...
+		"sensitive information such as API-keys, unencrypted tokens, etc.")
+	fmt.Print("Type \"y\" to confirm or anything else to exit: ")
+	ack, err := prompt.ReadLine()
```
brianstrauch, line 53: Use the `internal/pkg/form` package!

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,80 @@
...
+func sendFeedback(cfg *v1.Config, cmd *cobra.Command, msg string) error {
+	feedback := cliv1.CliV1Feedback{Content: cliv1.PtrString(msg)}
+	unsafeTrace, err := cmd.Flags().GetBool("unsafe-trace")
+	if err != nil {
+		return err
+	}
+	if err := cfg.GetCloudClientV2(unsafeTrace).CreateCliFeedback(feedback); err != nil {
+		log.CliLogger.Warnf("Failed to report CLI feedback: %v", err)
+	}
+	return nil
```
brianstrauch, line 79: IMO, the user should know if their feedback doesn't get submitted.
```suggestion
	return cfg.GetCloudClientV2(unsafeTrace).CreateCliFeedback(feedback)
```

brianstrauch: Also, let's make this an authenticated command, so we can just use `c.ClientV2` here instead of re-creating the client.

test/feedback_test.go
```diff
@@ -0,0 +1,11 @@
+package test
+
+func (s *CLITestSuite) TestFeedback() {
+	tests := []CLITest{
+		{args: "feedback", fixture: "feedback/no-confirm.golden", input: "n\n"},
+		{args: "feedback", fixture: "feedback/received.golden", input: "y\n This CLI is great!\n"},
```
brianstrauch, line 6:
```suggestion
		{args: "feedback", fixture: "feedback/received.golden", input: "y\nThis CLI is great!\n"},
```

test/feedback_test.go
```diff
@@ -0,0 +1,11 @@
...
+	}
+	for _, tt := range tests {
```
brianstrauch, line 8:
```suggestion
	for _, test := range tests {
```

test/fixtures/output/help-onprem.golden
```diff
@@ -10,6 +10,7 @@ Available Commands:
   completion      Print shell completion code.
   connect         Manage Kafka Connect.
   context         Manage CLI configuration contexts.
+  feedback        Submit feedback about the Confluent CLI.
```
brianstrauch, line 13: This command won't work on-prem.

test/feedback_test.go
sgagniere: I think this is for testing that preceding spaces are trimmed away.

brianstrauch: If so, let's unit test that instead!

kevin-wu24: I actually don't know how to get the entered feedback to show up in the golden file. I pass it in as input to the test but it doesn't show up in the actual output of the test. Maybe it would require printing in the feedback command, which we wouldn't want to do.

brianstrauch: > I actually don't know how to get the entered feedback to show up in the golden file.

It currently doesn't show up... there's probably a way to do it but I haven't figured it out yet

brianstrauch: (you can ignore that limitation for now)

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
+	"github.com/confluentinc/cli/internal/pkg/output"
 	pversion "github.com/confluentinc/cli/internal/pkg/version"
 )
 
-func NewFeedbackCmd(cfg *v1.Config, prerunner pcmd.PreRunner) *cobra.Command {
-	prompt := form.NewPrompt(os.Stdin)
-	return NewFeedbackCmdWithPrompt(cfg, prerunner, prompt)
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+	form.Prompt
```
brianstrauch, line 21: Why is this part of the struct?

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
-					}
-					fmt.Println("Thanks for your feedback.")
-				}
-				return nil
-			},
-			Args: cobra.NoArgs,
-		},
-		prerunner)
+const (
+	maxLength = 2000
```
brianstrauch, line 25: Max length of what? Why is this global?

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
+)
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "feedback",
+		Short:       fmt.Sprintf("Submit feedback about the %s.", pversion.FullCLIName),
```
brianstrauch, line 31:
```suggestion
		Short:       fmt.Sprintf("Submit feedback for the %s.", pversion.FullCLIName),
```

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
+	return cmd
 }
 
-func getFeedback(prompt form.Prompt) (string, error) {
-	fmt.Println("Please confirm that your feedback will not contain any " +
-		"sensitive information such as API-keys, unencrypted tokens, etc.")
-	fmt.Print("Type \"y\" to confirm or anything else to exit: ")
-	ack, err := prompt.ReadLine()
+func (c *command) feedback(_ *cobra.Command, _ []string) error {
+	msg, err := getFeedback(c.Prompt)
```
brianstrauch, line 46:
```suggestion
	feedback, err := getFeedback(c.Prompt)
```

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
+		fmt.Println("Thanks for your feedback.")
+	}
+	return nil
+}
+
+func getFeedback(prompt form.Prompt) (string, error) {
+	f := form.New(form.Field{
+		ID: "proceed",
+		Prompt: "Please confirm that your feedback will not contain any " +
+			"sensitive information such as API-keys, unencrypted tokens, etc.",
```
brianstrauch, line 63:
```suggestion
		Prompt: "Please confirm that your feedback does not contain any sensitive information",
```

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
+		IsYesOrNo: true,
+	})
+	if err := f.Prompt(prompt); err != nil || !f.Responses["proceed"].(bool) {
+		output.Println("Not proceeding with feedback")
```
brianstrauch, line 67:
```suggestion
```

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
 		return "", err
 	}
-	fmt.Println("Enter feedback: ")
+	output.Println("Enter feedback (maximum length of 2,000 characters): ")
```
brianstrauch, line 70: 99% of the time, users won't exceed this limit. Let's make sure the backend returns a nice error if this happens though. (Also, if we change this number in the future, we don't want older versions of the CLI to still say 2000)
```suggestion
	output.Print("Enter feedback: ")
```

internal/cmd/feedback/command.go
```diff
@@ -10,71 +10,77 @@ import (
...
 	msg, err := prompt.ReadLine()
 	if err != nil {
 		return "", err
 	}
 	msg = strings.TrimLeft(msg, " ")
 	msg = strings.TrimRight(msg, " \n")
+	if len(msg) > maxLength {
+		return "", errors.New(errors.ExceedsMaxLengthMsg)
+	}
```
brianstrauch, line 79: Always do validation in the backend!

internal/pkg/errors/error_message.go
```diff
@@ -69,6 +69,9 @@ const (
 	NoEnvironmentFoundErrorMsg     = "no environment found"
 	NoEnvironmentFoundSuggestions  = "This issue may occur if this user has no valid role bindings. Contact an Organization Admin to create a role binding for this user."
 
+	// feedback command
+	ExceedsMaxLengthMsg = "Inputted feedback exceeds the maximum character length of 2,000."
+
```
brianstrauch, line 74:
```suggestion
```

test/feedback_test.go
```diff
@@ -1,11 +1,16 @@
...
...t consequatur iure qui sunt deleniti et nisi voluptatum qui consequatur tenetur et praesentium error. Quo quidem quia ut corrupti deserunt non provident odit sit quia placeat aut atque odio in vero galisum eum enim quod. Ut tenetur accusamus et dicta voluptatem At voluptatem quaerat et officiis sapiente ut cupiditate odit non repellendus omnis. Ut minus excepturi vel incidunt veritatis et optio odit a dolorum molestiae sit deserunt quidem. Aut voluptates molestiae eos dolor beatae sit reprehenderit iure. Sed atque molestiae ex voluptatem ullam in error maiores. Non dolor deserunt in obcaecati similique quo omnis repellat id debitis provident! Eos sunt aliquam qui aliquid suscipit ut consequuntur reprehenderit. Ut distinctio commodi eum debitis fugit qui dolor corporis aut voluptas minus ad excepturi iste. Non quia molestias ea velit delectus qui nihil natus et reprehenderit veniam hic harum voluptatum et nesciunt impedit qui natus aliquid. Et reiciendis odit qui obcaecati molestiae non sequi minus et laboriosam placeat ea dignissimos rerum non quisquam accusantium hic quia doloremque. Id illum dolor aut rerum magnam eum consequatur deleniti est placeat ducimus quo dolores velit.\n"
```
brianstrauch, line 6: Let's make the mock backend return an error if the length > 20 (for example), so this test is a bit smaller and more readable.

internal/cmd/feedback/command.go
kevin-wu24: line 20 or line 21?

brianstrauch: Line 21!

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "feedback",
+		Short:       fmt.Sprintf("Submit feedback for the %s.", pversion.FullCLIName),
+		Args:        cobra.NoArgs,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
```
brianstrauch, line 27: These APIs don't support API key auth yet :(
```suggestion
		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonApiKeyCloudLogin},
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	c := &command{
+		AuthenticatedCLICommand: pcmd.NewAuthenticatedCLICommand(cmd, prerunner),
+	}
```
brianstrauch, line 32:
```suggestion
	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+func (c *command) feedback(_ *cobra.Command, _ []string) error {
+	feedback, err := getFeedback(form.NewPrompt(os.Stdin))
+	if err != nil {
+		return err
+	}
+	if len(feedback) > 0 {
+		if err := c.sendFeedback(feedback); err != nil {
+			return err
+		}
+		fmt.Println("Thanks for your feedback.")
```
brianstrauch, line 47:
```suggestion
		output.Println("Thanks for your feedback.")
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return cmd
+}
+
+func (c *command) feedback(_ *cobra.Command, _ []string) error {
+	feedback, err := getFeedback(form.NewPrompt(os.Stdin))
+	if err != nil {
+		return err
+	}
+	if len(feedback) > 0 {
```
brianstrauch, line 43: This is a bit more precise. At first glance I thought `getFeedback()` was returning a slice...
```suggestion
	if feedback != "" {
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+
+func getFeedback(prompt form.Prompt) (string, error) {
+	f := form.New(form.Field{
+		ID:        "proceed",
+		Prompt:    "Please confirm that your feedback does not contain any sensitive information",
+		IsYesOrNo: true,
+	})
+	if err := f.Prompt(prompt); err != nil || !f.Responses["proceed"].(bool) {
+		return "", err
+	}
```
brianstrauch, line 60:
```suggestion
	if err := f.Prompt(prompt); err != nil {
		return "", err
	}
	if !f.Responses["proceed"].(bool) {
		return "", nil
	}
```

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		fmt.Println("Thanks for your feedback.")
+	}
+	return nil
+}
+
+func getFeedback(prompt form.Prompt) (string, error) {
+	f := form.New(form.Field{
+		ID:        "proceed",
+		Prompt:    "Please confirm that your feedback does not contain any sensitive information",
```
brianstrauch, line 55: It seems weird to prompt the user with this before they've actually written the feedback. I think there are two better ways to do this:
1. Print a message reminding them to not include sensitive information without a prompt, then collect their feedback
2. Collect their feedback, then show this prompt (I think this is what the security team wants)

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := f.Prompt(prompt); err != nil || !f.Responses["proceed"].(bool) {
+		return "", err
+	}
+	output.Print("Enter feedback: ")
+	feedback, err := prompt.ReadLine()
+	if err != nil {
+		return "", err
+	}
+	feedback = strings.TrimLeft(feedback, " ")
+	feedback = strings.TrimRight(feedback, " \n")
```
brianstrauch, line 67: https://pkg.go.dev/strings#TrimSpace

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,74 @@
...
+	return feedback, err
+}
+
+func (c *command) sendFeedback(content string) error {
```
brianstrauch, line 71: IMO, this doesn't need to be a separate function since it just calls another function.

test/feedback_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "fmt"
+
+func (s *CLITestSuite) TestFeedback() {
+	feedback := "Lorem ipsum dolor sit amet. Qui amet molestiae eum eaque perferendis\n"
+	tests := []CLITest{
+		{args: "feedback", fixture: "feedback/no-confirm.golden", input: "n\n"},
+		{args: "feedback", fixture: "feedback/received.golden", input: "y\nThis CLI is great!\n"},
+		{args: "feedback", exitCode: 1, fixture: "feedback/too-long.golden", input: fmt.Sprintf("y\n%s", feedback)},
+	}
```
brianstrauch, line 11:
```suggestion
	}
	
```

test/feedback_test.go
```diff
@@ -0,0 +1,16 @@
+package test
+
+import "fmt"
+
+func (s *CLITestSuite) TestFeedback() {
+	feedback := "Lorem ipsum dolor sit amet. Qui amet molestiae eum eaque perferendis\n"
```
brianstrauch, line 6: Move this into line 10? Doesn't look like it's being used anywhere else...

test/test-server/utils.go
```diff
@@ -18,11 +18,12 @@ import (
...
-	roleNameInvalidErrMsg       = `{"status_code":400,"message":"Invalid role name : %s","type":"INVALID REQUEST DATA"}`
-	resourceNotFoundErrMsg      = `{"errors":[{"detail":"resource not found"}], "message":"resource not found"}`
-	badRequestErrMsg            = `{"errors":[{"status":"400","detail":"Bad Request"}]}`
-	userConflictErrMsg          = `{"errors":[{"detail":"This user already exists within the Organization"}]}`
+	serviceAccountInvalidErrMsg    = `{"errors":[{"status":"403","detail":"service account is not valid"}]}`
+	roleNameInvalidErrMsg          = `{"status_code":400,"message":"Invalid role name : %s","type":"INVALID REQUEST DATA"}`
+	resourceNotFoundErrMsg         = `{"errors":[{"detail":"resource not found"}], "message":"resource not found"}`
+	badRequestErrMsg               = `{"errors":[{"status":"400","detail":"Bad Request"}]}`
+	userConflictErrMsg             = `{"errors":[{"detail":"This user already exists within the Organization"}]}`
+	feedbackExceedsMaxLengthErrMsg = `{"errors":[{"status":"403","detail":"feedback exceeds the maximum length"}]}`
```
brianstrauch, line 26: Looks like this only gets used in one place, so let's decrease it's scope (it doesn't need to be a global variable)

test/test-server/utils.go
```diff
@@ -251,6 +252,12 @@ func writeResourceNotFoundError(w http.ResponseWriter) error {
 	return err
 }
 
+func writeFeedbackExceedsMaxLengthError(w http.ResponseWriter) error {
```
brianstrauch, line 255: Looks like this only gets used in one place, so we can move this into the function that uses it.

test/test-server/cli_handler.go
```diff
@@ -0,0 +1,27 @@
...
+// Handler for: /cli/v1/feedbacks
+func handleFeedbacks(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var req cliv1.CliV1Feedback
+		err := json.NewDecoder(r.Body).Decode(&req)
+		require.NoError(t, err)
+		if len(*req.Content) > 20 {
+			err := writeFeedbackExceedsMaxLengthError(w)
+			require.NoError(t, err)
+			return
```
brianstrauch, line 22: This isn't accomplishing anything
```suggestion
```

test/test-server/cli_handler.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/stretchr/testify/require"
+
+	cliv1 "github.com/confluentinc/ccloud-sdk-go-v2/cli/v1"
+)
+
+// Handler for: /cli/v1/feedbacks
+func handleFeedbacks(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var req cliv1.CliV1Feedback
+		err := json.NewDecoder(r.Body).Decode(&req)
```
brianstrauch, line 17: This is a bit more idiomatic. You don't need to use `&` every time you pass it into a function this way.
```suggestion
		req := new(cliv1.CliV1Feedback)
		err := json.NewDecoder(r.Body).Decode(req)
```

internal/cmd/feedback/command.go
kevin-wu24: I think we should do option 2

internal/cmd/feedback/command.go
```diff
@@ -40,35 +38,34 @@ func (c *command) feedback(_ *cobra.Command, _ []string) error {
...
 			return err
 		}
-		fmt.Println("Thanks for your feedback.")
+		output.Println("Thanks for your feedback.")
 	}
 	return nil
 }
 
 func getFeedback(prompt form.Prompt) (string, error) {
+	output.Print("Enter feedback: ")
```
brianstrauch, line 52: Can you add this to the form?

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,71 @@
...
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "feedback",
+		Short:       fmt.Sprintf("Submit feedback for the %s.", pversion.FullCLIName),
+		Args:        cobra.NoArgs,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
sgagniere, line 27: Do we want to limit this to only Cloud logins? Platform users could have feedback as well.

brianstrauch: Sending feedback requires authenticating to api.confluent.cloud, which CP users are unable to do

sgagniere: That's unfortunate.

internal/cmd/feedback/command.go
```diff
@@ -0,0 +1,71 @@
...
+	f := form.New(
+		form.Field{
+			ID:     "feedback",
+			Prompt: "Enter feedback",
+		},
+		form.Field{
+			ID:        "proceed",
+			Prompt:    "Please confirm that your feedback does not contain any sensitive information",
+			IsYesOrNo: true,
+		})
```
brianstrauch, line 61:
```suggestion
		},
	)
```

test/test-server/ccloudv2_router.go
```diff
@@ -15,6 +15,7 @@ var ccloudV2Routes = []route{
 	{"/byok/v1/keys", handleByokKeys},
 	{"/byok/v1/keys/{id}", handleByokKey},
 	{"/billing/v1/costs", handleBillingCosts},
+	{"/cli/v1/feedbacks", handleFeedbacks},
```
brianstrauch, line 18: Please alphabetize!

test/test-server/cli_handler.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Handler for: /cli/v1/feedbacks
+func handleFeedbacks(t *testing.T) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		req := new(cliv1.CliV1Feedback)
+		err := json.NewDecoder(r.Body).Decode(req)
+		require.NoError(t, err)
+		if len(*req.Content) > 20 {
+			w.WriteHeader(http.StatusForbidden)
+			_, err := io.WriteString(w, `{"errors":[{"status":"403","detail":"feedback exceeds the maximum length"}]}`)
```
brianstrauch, line 22: Can write a struct and serialize it to JSON instead of writing a raw JSON string?

test/test-server/cli_handler.go
```diff
@@ -19,7 +19,13 @@ func handleFeedbacks(t *testing.T) http.HandlerFunc {
 		require.NoError(t, err)
 		if len(*req.Content) > 20 {
 			w.WriteHeader(http.StatusForbidden)
-			_, err := io.WriteString(w, `{"errors":[{"status":"403","detail":"feedback exceeds the maximum length"}]}`)
+			errorJson, err := json.Marshal(&struct {
+				Errors []map[string]any `json:"errors"`
+			}{
+				Errors: []map[string]any{{"status": "403", "detail": "feedback exceeds the maximum length"}},
```
brianstrauch, line 25: This should also be a struct :smile:

test/test-server/cli_handler.go
```diff
@@ -20,9 +20,15 @@ func handleFeedbacks(t *testing.T) http.HandlerFunc {
 		if len(*req.Content) > 20 {
 			w.WriteHeader(http.StatusForbidden)
 			errorJson, err := json.Marshal(&struct {
```
brianstrauch, line 22: Now that it works, can you define this struct, so we don't have to use `any`?

kevin-wu24: I basically just copied the json error message string I use here from https://github.com/confluentinc/cli/blob/main/test/test-server/utils.go#L20-L25
In the usages of these strings, it seems more readable than what I'm doing here (also no extra json.Marshall() call). Wdyt?
``` 
w.WriteHeader(http.StatusForbidden)
errorJson, err := json.Marshal(ErrorJson{Errors: []ErrorInfo{{Detail: "feedback exceeds the maximum length", Status: "403"}}})
require.NoError(t, err)
_, err = io.WriteString(w, string(errorJson))
```
VS
```
w.WriteHeader(http.StatusForbidden)
_, err := io.WriteString(w, serviceAccountInvalidErrMsg)
```

brianstrauch: > https://github.com/confluentinc/cli/blob/main/test/test-server/utils.go#L20-L25

This should be avoided, since JSON strings can get out of hand and are really difficult to debug. For a really bad example, look at test/test-server/mds_routes_and_roles.go

test/test-server/utils.go
```diff
@@ -239,28 +236,33 @@ func fillByokStoreV1() map[string]*byokv1.ByokV1Key {
 	return byokStoreV1
 }
 
+func writeErrorJson(w http.ResponseWriter, message string) error {
```
brianstrauch, line 239: Thanks!

test/fixtures/output/feedback/help-onprem.golden
```diff
@@ -0,0 +1,9 @@
+Submit feedback about the Confluent CLI.
```
brianstrauch, line 1: This command shouldn't exist on-prem!

brianstrauch: Did this file get generated incorrectly? (or maybe before you introduced the Cloud-only requirement)

brianstrauch (approved):

test/fixtures/output/feedback/help-onprem.golden
kevin-wu24: introduced before cloud-only, just deleted.

Issue #1963, opened 2023-06-08
General questions about `confluent local kafka start`


Pontius1007:
Hello, 

I have a few general remarks and questions in regard to the docker version for what used to be `confluent local services start`. 

From my understanding of previous use, the bootstrap servers have been default available at localhost:9092. From the logged KafkaRestConfig I can see that: `bootstrap.servers = broker:44535`. Is this documented somewhere or is it expected from the users that we manually check the KafkaRestConfig? 

We have our local development environment running in Docker Containers, and needed to manually get the `docker run` command and modify it with the `--net` flag to facilitate communication. Is there any plan to support additional configuration to avoid this issue? 

This enabled us to get a connection with the bootstrap servers at broker:44535, but we fail to get a response from the Schema Registry which we would expect to be available at `http://localhost:8081`. Any idea why? 

Additionally, we do not get any feedback in the terminal when running `confluent local kafka start`. You are left with a blank terminal until the image is finished downloading, and I miss the printed output previously available as:
`
Starting Zookeeper
Zookeeper is [UP]
Starting Kafka
Kafka is [UP]
Starting Schema Registry
Schema Registry is [UP]
Starting Kafka REST
Kafka REST is [UP]
Starting Connect
Connect is [UP]
Starting KSQL Server
KSQL Server is [UP]
Starting Control Center
Control Center is [UP]
`

And it sometimes fail to create a docker container when running "confluent local kafka start`, and multiple retries are needed without any console output. 

I do also believe that this documentation: https://docs.confluent.io/platform/6.2/quickstart/ce-quickstart.html is outdated after the 3.16 release. 

Thanks in advance! 

BR

Pontius1007:
Hello! 

First of all, thanks for the fast reply and the in-depth answer. 

I do believe that would help! I am not sure how easy it would be, but would it be possible to take in additional configuration via the `--build-args` variable? https://docs.docker.com/engine/reference/commandline/build/#build-arg 

I did some more testing and I guess the "fail to start" was mostly me getting impatiant with the no output and manually disabling the job, which was downloading in the background. Some output on the download progress would be greatly appreciated! 

Hopefully my last message didn't scare you too much!

MuweiHe:
Hi!

It makes sense to me to use default ports if user doesn't specify config! Will try to follow the old convention and will include it in this week's minor release!

PR #2076, opened 2023-07-07
[confluent-local] [CLI-2583] print image downloading progress, print ports and allow overwriting local ports

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Allow changing default port numbers in `confluent local kafka start` with the `--kafka-rest-port` and `--plaintext-port` flags
- Print progress while downloading Docker image in `confluent local kafka start`
- Print the ports that Confluent Local is using when running `confluent local kafka start`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

print the load indicator in image downloading. the same load indicator will be refreshed in the same line, instead of a new line
```
Verifying Checksum
Extracting: [==================================================>]  34.88MB/34.88MB
Extracting: [==================================================>]  1.104kB/1.104kB
Extracting: [==================================================>]     882B/882B
Extracting: [==================================================>]  1.434MB/1.434MB
Extracting: [==================================================>]     131B/131B
Extracting: [==================================================>]     171B/171B
Extracting: [==================================================>]  187.3MB/187.3MB
Extracting: [==================================================>]  4.101kB/4.101kB
Pull complete
Digest: sha256:1aeeebdef8e20548c36ef23df8e10e685b9e4ac2b268b36b045d8d2e3e71c0d9
Started Confluent Local container b7c5dbeba9.nc/confluent-local:latest
To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.
```
allow configuring 2 opening ports (the other 2 are not being used in any commands yet) by flags. default to any valid port if flags are not set.

References
----------

https://github.com/confluentinc/cli/issues/1963

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/cmd/local/command_kafka_start.go
```diff
@@ -25,20 +24,28 @@ import (
 )
 
 var (
-	statusBlacklist = []string{"Pulling fs layer", "Waiting", "Downloading", "Download complete", "Verifying Checksum", "Extracting", "Pull complete"}
+	statusBlacklist = []string{"Pulling fs layer", "Waiting", "Download complete", "Verifying Checksum", "Pull complete"}
```
brianstrauch, line 27: I feel like the user would be interested in seeing all of these? (This is what they're used to seeing when they use Docker)

internal/cmd/local/command_kafka_start.go
```diff
@@ -25,20 +24,28 @@ import (
...
 func (c *Command) newKafkaStartCommand() *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Use:   "start",
 		Short: "Start a single-node instance of Apache Kafka.",
 		Args:  cobra.NoArgs,
 		RunE:  c.kafkaStart,
 	}
+
+	cmd.Flags().String("kafka-rest-port", "", "The port number for kafka REST. If not specified, a random free port will be used.")
```
brianstrauch, line 45:
```suggestion
	cmd.Flags().String("kafka-rest-port", "", "The port number for Kafka REST. If not specified, a random free port will be used.")
```

brianstrauch: Shouldn't we default to 8082, and then a random free port? I think most users are used to Kafka REST being on that port. https://docs.confluent.io/platform/current/kafka-rest/index.html#deployment

internal/cmd/local/command_kafka_start.go
```diff
@@ -58,31 +65,35 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+		err := json.Unmarshal([]byte(text), &response)
+		if err != nil {
+			return err
 		}
+
 		var inBlacklist bool
 		for _, s := range statusBlacklist {
-			if output.Status == s {
+			if response.Status == s {
 				inBlacklist = true
```
sgagniere, line 81: Nitpick:
This block can be condensed with `inBlacklist := types.Contains(statusBlacklist, response.Status)`
You could also roll it into the if below: `if !types.Contains(statusBlacklist, response.Status) {`

internal/cmd/local/command_kafka_start.go
```diff
@@ -58,31 +65,35 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 			}
 		}
+
 		if !inBlacklist {
-			fmt.Printf("%v\n", output.Status)
+			fmt.Printf("%v\n", response)
```
brianstrauch, line 86: I'm pretty sure this is the same thing
```suggestion
			fmt.Println(response)
```

brianstrauch:
Stylistic comments: We should get rid of the leading/trailing `{`/`}` in the output, and try to use ANSI codes to refresh the current line while printing the loading bar. (Our goal should be to imitate the output of Docker)

internal/cmd/local/command_kafka_start.go
sgagniere: Maybe also `output` instead of `fmt`.

brianstrauch:
Release notes: Make sure you use single backticks (we convert these to double backticks only for docs output)! Also, let's capitalize "Confluent Local" instead of using backticks when referring to the product (to differentiate between the product and the command)

internal/cmd/local/command_kafka_start.go
sgagniere: Maybe also `output` instead of `fmt`.

(sorry if you get pinged again; I rewrote this comment)

internal/cmd/local/command_kafka_start.go
MuweiHe: sounds good!

internal/cmd/local/command_kafka_start.go
MuweiHe: could you be more specific? i can't find `Contains` in `go/types` package

sgagniere: Sorry, I meant our own types package `"github.com/confluentinc/cli/internal/pkg/types"`. It's in `pkg/types/slice.go`

MuweiHe:
> Stylistic comments: We should get rid of the leading/trailing `{`/`}` in the output, and try to use ANSI codes to refresh the current line while printing the loading bar. (Our goal should be to imitate the output of Docker)

good idea! it's now printing the loading indicator in the same line for the same target

internal/cmd/local/command_kafka_start.go
MuweiHe: i'm now printing out all outputs so no blacklist is needed now! thanks for the advice though

internal/cmd/local/command_kafka_start.go
```diff
@@ -75,22 +71,22 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 		}
 
-		var inBlacklist bool
-		for _, s := range statusBlacklist {
-			if response.Status == s {
-				inBlacklist = true
-			}
-		}
+		if response.Status == "Downloading" {
+			fmt.Printf("\rDownloading: %s", response.Progress)
```
brianstrauch, line 75:
```suggestion
			output.Printf("\rDownloading: %s", response.Progress)
```

internal/cmd/local/command_kafka_start.go
```diff
@@ -158,20 +154,25 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
 		}
 
 		if kafkaRestPort, err := cmd.Flags().GetString("kafka-rest-port"); err == nil && kafkaRestPort != "" {
 			c.Config.LocalPorts.KafkaRestPort = kafkaRestPort
+		} else {
+			freePort, err := freeport.GetFreePort()
+			if err != nil {
+				return err
+			}
+			c.Config.LocalPorts.KafkaRestPort = strconv.Itoa(freePort)
```
brianstrauch, line 174: Let's print a warn-level message here that says something like: "default kafka rest port 8082 is in use, using XXXX instead"?

internal/cmd/local/command_kafka_start.go
sgagniere: Nice!

internal/cmd/local/command_kafka_start.go
```diff
@@ -158,20 +154,25 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
-			KafkaRestPort:  strconv.Itoa(freePorts[0]),
-			PlaintextPort:  strconv.Itoa(freePorts[1]),
-			BrokerPort:     strconv.Itoa(freePorts[2]),
-			ControllerPort: strconv.Itoa(freePorts[3]),
+			PlaintextPort:  strconv.Itoa(freePorts[0]),
+			BrokerPort:     strconv.Itoa(freePorts[1]),
+			ControllerPort: strconv.Itoa(freePorts[2]),
 		}
 
 		if kafkaRestPort, err := cmd.Flags().GetString("kafka-rest-port"); err == nil && kafkaRestPort != "" {
```
sgagniere, line 167: Since `"8082"` is now the default value for `--kafka-rest-port`, it seems like users need to actively set `--kafka-rest-port ""` to reach the free port branch of this if block.

Do we want to always default to 8082? If so, maybe the else block can just be setting it to 8082 if the user overwrote it with the empty string.

brianstrauch: > Do we want to always default to 8082?

I think this makes the most sense, all of our other tools do this so this is what users expect.

internal/cmd/local/command_kafka_start.go
```diff
@@ -147,16 +153,29 @@ func (c *Command) prepareAndSaveLocalPorts(isTest bool) error {
...
-			KafkaRestPort:  strconv.Itoa(freePorts[0]),
-			PlaintextPort:  strconv.Itoa(freePorts[1]),
-			BrokerPort:     strconv.Itoa(freePorts[2]),
-			ControllerPort: strconv.Itoa(freePorts[3]),
+			PlaintextPort:  strconv.Itoa(freePorts[0]),
+			BrokerPort:     strconv.Itoa(freePorts[1]),
+			ControllerPort: strconv.Itoa(freePorts[2]),
+		}
+
+		if kafkaRestPort, err := cmd.Flags().GetString("kafka-rest-port"); err == nil && kafkaRestPort != "" {
```
brianstrauch, line 167:
```suggestion
		kafkaRestPort, err := cmd.Flags().GetString("kafka-rest-port")
		if err != nil {
			return err
		}
		if kafkaRestPort != "" {
```

internal/cmd/local/command_kafka_start.go
MuweiHe: make sense. or i can just remove the else block, expecting user will not set this flag to "".

MuweiHe: the else block will act as fault tolerance if user unexpectedly input "" for tha flag

internal/cmd/local/command_kafka_start.go
MuweiHe: 👍

internal/cmd/local/command_kafka_start.go
sgagniere: Yup, but it just seems weird to me that it defaults to a free port and not just 8082 if they entered "".

internal/cmd/local/command_kafka_start.go
```diff
@@ -58,31 +61,33 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
-		if err := json.Unmarshal([]byte(ss), &output); err != nil {
-			continue
-		}
-		var inBlacklist bool
-		for _, s := range statusBlacklist {
-			if output.Status == s {
-				inBlacklist = true
-			}
+		err := json.Unmarshal([]byte(text), &response)
+		if err != nil {
```
brianstrauch, line 70:
```suggestion
		if err := json.Unmarshal([]byte(text), &response); err != nil {
```

internal/cmd/local/command_kafka_start.go
```diff
@@ -58,31 +61,33 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 	defer out.Close()
 
-	buf := new(strings.Builder)
-	_, err = io.Copy(buf, out)
-	if err != nil {
-		return err
-	}
+	scanner := bufio.NewScanner(out)
+	for scanner.Scan() {
+		var response ImagePullResponse
```
brianstrauch, line 66: This way, we don't need to use the `&` character whenever we pass this into a function.
```suggestion
		response := new(ImagePullResponse)
```

internal/cmd/local/command_kafka_start.go
brianstrauch: Not sure if you took this suggestion, since it looks like we don't tell the user that port "8082" is taken.

I think the majority of the time, if 8082 is already in use, it probably means that the user is already running an instance of Kafka REST, and they probably want to shut it down to avoid confusion.

internal/cmd/local/command_kafka_start.go
MuweiHe: so do we want it to be still 8082 even if user inputs "", and only a free port when 8082 is not available?

sgagniere: That seems reasonable to me.

internal/cmd/local/command_kafka_start.go
MuweiHe: i added an extra step of validating that port and (potentially) customized plaintext port.

internal/cmd/local/command_kafka_start.go
```diff
@@ -26,21 +27,25 @@ import (
...
 func (c *Command) newKafkaStartCommand() *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Use:   "start",
 		Short: "Start a single-node instance of Apache Kafka.",
 		Args:  cobra.NoArgs,
 		RunE:  c.kafkaStart,
 	}
+
+	cmd.Flags().String("kafka-rest-port", "8082", "The port number for Kafka REST.")
```
brianstrauch, line 45: Let's put this string in a constant and also use it in the localports struct?

internal/cmd/local/command_kafka_start.go
```diff
@@ -180,26 +186,80 @@ func (c *Command) prepareAndSaveLocalPorts(isTest bool) error {
...
+func (c *Command) validateCustomizedPorts() error {
+	kafkaRestLn, err := net.Listen("tcp", ":"+c.Config.LocalPorts.KafkaRestPort)
+	if err != nil {
+		freePort, err := freeport.GetFreePort()
+		if err != nil {
+			return err
+		}
+		invalidKafkaRestPort := c.Config.LocalPorts.KafkaRestPort
+		c.Config.LocalPorts.KafkaRestPort = strconv.Itoa(freePort)
+		log.CliLogger.Warnf("Port %s is not available. Overwrote Kafka REST port to %d", invalidKafkaRestPort, freePort)
```
brianstrauch, line 238:
```suggestion
		log.CliLogger.Warnf("Port %s is not available, using port %d instead.", invalidKafkaRestPort, freePort)
```

internal/pkg/config/v1/local_ports.go
```diff
@@ -1,8 +1,8 @@
 package v1
 
 type LocalPorts struct {
```
brianstrauch, line 3: Do we want to call this struct `out`, and move it to where it's used (command_kafka_start.go) to match the style of our other commands? We may want to add additional fields to this table in the future. As far as I can tell, this has nothing to do with the configuration file, so it shouldn't belong in internal/pkg/config.

internal/cmd/local/command_kafka_start.go
brianstrauch: Looks like this string is used multiple times, let's turn it into a constant

internal/pkg/config/v1/local_ports.go
MuweiHe: this is being saved into config.json, and other commands get it from config as well

internal/cmd/local/command_kafka_start.go
MuweiHe: do you mean const `"kafka-rest-port"`? how do we use it in the struct?

internal/pkg/config/v1/local_ports.go
brianstrauch: > this is being saved into config.json

Why does it need be stored if we have flags? When does it get stored in the first place?

test/fixtures/output/local/kafka/start.golden
```diff
@@ -5,3 +5,9 @@ Digest: sha256:[a-zA-Z0-9]*.
 Status: (?:Image is up to date|Downloaded newer image) for confluentinc/confluent-local:latest
 Started Confluent Local container [a-zA-Z0-9]*.
 To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.
++-----------------+------+
+| Broker Port     | 2998 |
+| Controller Port | 2999 |
```
brianstrauch, line 10: Are these both for Kafka? What about "Kafka Broker Port"?

test/fixtures/output/local/kafka/start.golden
```diff
@@ -5,3 +5,9 @@ Digest: sha256:[a-zA-Z0-9]*.
...
+| Kafka Rest Port | 2996 |
+| Plaintext Port  | 2997 |
```
brianstrauch, line 12: Can we generate ports in the same order as these are listed?

sgagniere (approved):

PR #2084, opened 2023-07-11
check index in range when getting minor version number in update

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

got a panic report that index was out of range (-1) on line 93. that could only happen if `nextMajorVersion` was declared to be `v0` which makes no sense. Return a `failed to obtain correct version info` error if that happens.

References
----------

https://confluent.slack.com/archives/C010Y0EP5MZ/p1689092199750689

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/update/client.go
+++ b/internal/pkg/update/client.go
@@ -68,7 +68,7 @@ func NewClient(params *ClientParams) *client {
 
 // CheckForUpdates checks for new versions in the repo
 func (c *client) CheckForUpdates(cliName, currentVersion string, forceCheck bool) (string, string, error) {
-	if c.DisableCheck {
+	if c.DisableCheck || currentVersion == "v0.0.0" {
 		return "", "", nil
 	}
 
--- a/internal/pkg/update/client.go
+++ b/internal/pkg/update/client.go
@@ -37,6 +37,7 @@ type client struct {
 }
 
 var _ Client = (*client)(nil)
+var defaultVersion = "v0.0.0"
 
 // ClientParams are used to configure the update.Client
 type ClientParams struct {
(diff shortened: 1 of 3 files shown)


brianstrauch:
This is an internal issue that comes from running `go run ...`, so we should remove the release notes. I wonder if we can solve this by not checking for updates when the version is the default (0.0.0)

MuweiHe:
> This is an internal issue that comes from running `go run ...`, so we should remove the release notes. I wonder if we can solve this by not checking for updates when the version is the default (0.0.0)

this could also happen when user run `confluent update`.

brianstrauch:
> this could also happen when user run confluent update

I agree that we should still keep the error (and we should also put a similar statement on line 76 for completeness), but if the version is 0.0.0, the user is a developer and we don't need to be checking for updates in the first place.

internal/pkg/update/client.go
```diff
@@ -68,7 +68,7 @@ func NewClient(params *ClientParams) *client {
 
 // CheckForUpdates checks for new versions in the repo
 func (c *client) CheckForUpdates(cliName, currentVersion string, forceCheck bool) (string, string, error) {
-	if c.DisableCheck {
+	if c.DisableCheck || currentVersion == "v0.0.0" {
```
sgagniere, line 71: Can a user actually end up with `v0.0.0` somehow?

sgagniere (approved):

internal/pkg/update/client.go
brianstrauch: v0.0.0 is the default value if the relevant ldflag isn't passed during the build process. A customer shouldn't ever encounter this, though.

brianstrauch: Maybe we should make a constant for v0.0.0, since it's not obvious what this means

internal/pkg/update/client.go
```diff
@@ -37,6 +37,7 @@ type client struct {
 }
 
 var _ Client = (*client)(nil)
+var defaultVersion = "v0.0.0"
```
brianstrauch, line 40:
```suggestion

const defaultVersion = "v0.0.0"
```

PR #2085, opened 2023-07-11
Autocomplete `kafka topic [produce|consume]`

brianstrauch:
Release Notes
-------------
New Features
- Add autocompletion to `confluent kafka topic produce` and `confluent kafka topic consume`

Checklist
---------
   * yes: ok

What
----
Check in the command function if the current Kafka cluster has an API key, instead of in the prerun function. Allows us to use  `AuthenticatedCLICommand` for produce/consume commands, which allows for autocompletion.

As a follow-up, I plan to switch the `confluent kafka client-config` commands off of the `HasAPIKeyCommand` struct, so we'll have a single struct for all of our authenticated cloud commands.

Test & Review
-------------
Manually tested, and verified that `confluent context create` + `confluent kafka topic [produce|consume]` still works.


internal/pkg/cmd/prerunner.go
```diff
@@ -181,6 +181,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 			return err
 		}
 
+		if r.Config.Context().GetCredentialType() == v1.APIKey {
+			return ErrIfMissingRunRequirement(cmd, r.Config)
```
sgagniere, line 185: Returning here skips `r.setAuthenticatedContext`, which means `command.Context` is nil for api key logins; this causes issues if users supply the `--environment` or `--cluster` flags.

internal/cmd/kafka/command_topic_consume.go
```diff
@@ -34,13 +37,6 @@ func newConsumeCommand(prerunner pcmd.PreRunner, clientId string) *cobra.Command
 		),
 	}
 
-	c := &hasAPIKeyTopicCommand{
-		HasAPIKeyCLICommand: pcmd.NewHasAPIKeyCLICommand(cmd, prerunner),
```
sgagniere, line 38: We lose some warnings about not using `--environment` and `--cluster` flags. We might want to add those warnings elsewhere if we don't want to use this command struct.

internal/cmd/kafka/command_topic_consume.go
```diff
@@ -70,17 +66,33 @@ func newConsumeCommand(prerunner pcmd.PreRunner, clientId string) *cobra.Command
...
+
+		if apiKey != "" && apiSecret != "" {
+			cluster.APIKey = apiKey
+			cluster.APIKeys[cluster.APIKey] = &v1.APIKeyPair{
+				Key:    apiKey,
+				Secret: apiSecret,
+			}
+		} else {
+			return &errors.UnspecifiedAPIKeyError{ClusterID: cluster.ID}
+		}
```
sgagniere, line 95: small suggestion: this block seems to be repeated for produce and consume files; perhaps it can go in a helper function in command_topic.go

sgagniere (commented):
Just one last concern: using those flags with produce | consume now returns an error, while previously it displayed a warning but the commands still worked.

Is this a breaking change, and/or is converting them from warnings to errors something we should put in v4?

brianstrauch:
> using those flags with produce | consume now returns an error, while previously it displayed a warning but the commands still worked.

Oops, I didn't realize we were going from warning -> error here. Good catch, will fix!

sgagniere (approved):
lgtm

PR #2088, opened 2023-07-12
CLI-2589 fix panic when running `api-key store` with resource flag

kevin-wu24:
Release Notes
-------------
Bug Fixes
- Prevents a panic when passing a non-existent Kafka cluster to `confluent api-key store --resource`

Checklist
---------
   * yes: ok

What
----
A panic can occur when the user tries to store an api-key secret for a resource that doesn’t exist anymore for that api-key.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2589

### DIFF
--- a/internal/cmd/api-key/command_store.go
+++ b/internal/cmd/api-key/command_store.go
@@ -112,7 +112,7 @@ func (c *command) store(cmd *cobra.Command, args []string) error {
 		return errors.CatchApiKeyForbiddenAccessError(err, getOperation, httpResp)
 	}
 
-	apiKeyIsValidForTargetCluster := apiKey.Spec.Resource != nil && cluster.ID == apiKey.Spec.Resource.Id
+	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.Spec.Resource.GetId()
 
 	if !apiKeyIsValidForTargetCluster {
 		return errors.NewErrorWithSuggestions(errors.APIKeyNotValidForClusterErrorMsg, errors.APIKeyNotValidForClusterSuggestions)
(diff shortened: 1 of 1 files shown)


internal/cmd/api-key/command_store.go
```diff
@@ -112,7 +112,7 @@ func (c *command) store(cmd *cobra.Command, args []string) error {
 		return errors.CatchApiKeyForbiddenAccessError(err, getOperation, httpResp)
 	}
 
-	apiKeyIsValidForTargetCluster := (cluster.ID == apiKey.Spec.Resource.Id)
+	apiKeyIsValidForTargetCluster := apiKey.Spec.Resource != nil && cluster.ID == apiKey.Spec.Resource.Id
```
brianstrauch, line 115: Can we use panic-safe getter functions here? (or does the SDK not allow it?)
```suggestion
	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.GetSpec().GetResource().GetId()
```

kevin-wu24: The getters are pointer methods, so you could do `apiKey.Spec.Resource.GetID()` which will prevent the same panic as adding the nil check above, so I think I'll do that.

brianstrauch: (We're still not able to chain multiple methods though, right?)

kevin-wu24: Unfortunately no, Go doesn't allow for taking the addresses of pointers in a non-roundabout way (i.e. `apiKey.(&GetSpec()).(&GetResource()).GetID().

brianstrauch (approved):

PR #2089, opened 2023-07-12
Use built-in flag combination primitives from Cobra

brianstrauch:
Checklist
---------
   * yes: ok

What
----
A recent version of Cobra added two functions to mark flags as (1) required together and (2) mutually exclusive. Use the built-in functions instead of an inconsistent hardcoded solution.

Also mark `--output` as deprecated in `confluent kafka topic produce`, since it doesn't get used and can be deleted in the next major version.

### DIFF
--- a/internal/cmd/api-key/command_list.go
+++ b/internal/cmd/api-key/command_list.go
@@ -46,6 +46,8 @@ func (c *command) newListCommand() *cobra.Command {
 	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddOutputFlag(cmd)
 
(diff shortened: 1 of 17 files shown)


sgagniere (commented):
There's also `ExactlyOneSetErrorMsg` used in ACLs. Looks like `principal` and `service-account` can be marked mutually exclusive.

brianstrauch:
> There's also ExactlyOneSetErrorMsg used in ACLs. Looks like principal and service-account can be marked mutually exclusive.

If flags are marked as mutually exclusive, it's ok if neither is passed, which is not the case here.

sgagniere (approved):

PR #2091, opened 2023-07-12
CLI-2083 Accept resource names for all commands and flags

kevin-wu24:
Release Notes
-------------
New Features
- Resource names can now be passed in as arguments for commands and flags

Checklist
---------
   * yes: ok

What
----
Whenever a user passes in a resource name into a command/flag, the CLI gets a list of those resources and makes a mapping from resource name to ID and replaces the name with the correct ID to actually run the command.

Note that if the same name is duplicated among the same kind of resource, the CLI will return an error prompting the user to try again with IDs.

References
----------
https://confluentinc.atlassian.net/wiki/spaces/DPI/pages/3123186482/1-Pager+Accept+Resource+Names+For+Commands+And+Flags+Within+the+CLI
https://confluentinc.atlassian.net/browse/CLI-2083

Test & Review
-------------
Added integration tests that use resource names instead of IDs.


internal/cmd/environment/command_delete.go
```diff
@@ -28,7 +28,10 @@ func (c *command) newDeleteCommand() *cobra.Command {
 }
 
 func (c *command) delete(cmd *cobra.Command, args []string) error {
-	id := args[0]
+	id, err := convertNameToId(args[0], c.AuthenticatedCLICommand.V2Client)
```
brianstrauch, line 31: What if `args[0]` is already an ID? We shouldn't call the list API in that case.

brianstrauch: The `delete` commands will need to be updated after @sgagniere merges #1789

internal/pkg/ccloudv2/org.go
```diff
@@ -29,6 +29,7 @@ func (c *Client) CreateOrgEnvironment(environment orgv2.OrgV2Environment) (orgv2
 }
 
 func (c *Client) GetOrgEnvironment(envId string) (orgv2.OrgV2Environment, error) {
+	//TODO
```
brianstrauch, line 32: Mark this PR as "Draft" if you're still working on it

brianstrauch (changes requested):

brianstrauch:
We probably want to break this PR into a series of PRs, maybe one for each top-level command. If we implement this for *all* commands and flags, this PR is likely going to be thousands of lines.

internal/cmd/environment/command_delete.go
kevin-wu24: We have no idea if the user is passing in an ID or a name, so we would have to call the list API no matter what.

internal/cmd/environment/command.go
```diff
@@ -41,3 +43,16 @@ func (c *command) validArgs(cmd *cobra.Command, args []string) []string {
 
 	return pcmd.AutocompleteEnvironments(c.Client, c.V2Client, c.Context)
 }
+
+func convertNameToId(input string, v2Client *ccloudv2.Client) (string, error) {
+	envs, err := v2Client.ListOrgEnvironments()
```
sgagniere, line 48: This looks like it will call the list endpoint even if the user supplies an `id` instead of a `name`.

kevin-wu24:
> We probably want to break this PR into a series of PRs, maybe one for each top-level command. If we implement this for _all_ commands and flags, this PR is likely going to be thousands of lines.

I'm not sure if that's super necessary. Not all resources have names associated with them, but for the ones that do, they can be grouped under two interfaces from what I've seen, so most of their logic can be handled by a couple of generic functions

internal/cmd/environment/command_delete.go
kevin-wu24: For the `delete` commands I can just not make any changes to them atm.

sgagniere: > We have no idea if the user is passing in an ID or a name, so we would have to call the list API no matter what.

I think all resources either have an ID with a fixed prefix, or else the name itself is the ID (like `topic`). For the former, we can just parse the prefix to see if it's a valid ID for that resource, and for the latter we don't have to do anything.

brianstrauch: Should we call the describe endpoint first? What I'm worried about is the worst case: certain customers can have thousands of a specific resource, (such as 10,000 ACLs) and calling the list endpoint by default might mean 100 paginated calls instead of just 1 call 😬

sgagniere: > We have no idea if the user is passing in an ID or a name, so we would have to call the list API no matter what.

I think all resources either have an ID with a fixed prefix, or else the name itself is the ID (like `topic`). For the former, we can just parse the prefix to see if it's a valid ID for that resource, and for the latter we don't have to do anything.

Edit: Sorry if you get pinged multiple times. Is Github being weird for anyone else? I canceled my pending review because Brian already pointed out the thing I was going to point out, and Github just posted my review anyway 🙄

kevin-wu24: Yeah, from what Steven said, I think if the user passes an ID, since it's prefixed, I can check for that and avoid the list API call. I'll do that then. Thanks for pointing that out!

kevin-wu24: > > We have no idea if the user is passing in an ID or a name, so we would have to call the list API no matter what.
> 
> I think all resources either have an ID with a fixed prefix, or else the name itself is the ID (like `topic`). For the former, we can just parse the prefix to see if it's a valid ID for that resource, and for the latter we don't have to do anything.
> 
> Edit: Sorry if you get pinged multiple times. Is Github being weird for anyone else? I canceled my pending review because Brian already pointed out the thing I was going to point out, and Github just posted my review anyway 🙄

Resources where the name itself is the ID will be unaffected by this PR

sgagniere: We already have some support for this, too:
```
if resource.LookupType(serviceAccountId) != resource.ServiceAccount {
```

brianstrauch: > I think if the user passes an ID, since it's prefixed, I can check for that and avoid the list API call.

What if the user has a Kafka cluster named "lkc-my-kafka-cluster"?

kevin-wu24: > > I think if the user passes an ID, since it's prefixed, I can check for that and avoid the list API call.
> 
> What if the user has a Kafka cluster named "lkc-my-kafka-cluster"?

I'm not sure if we can handle that case unless there's some other rule about IDs after their prefix. If a user were to just name their resource like an ID, we wouldn't have a good way of handling that case (i.e. cluster id: lkc-1234, name: lkc-5678). If there was another cluster with the ID: lkc-5678, and I run `describe lkc-5678`, which one should run? Should we instead place limitations on what users can name their resources (i.e. just don't let them mirror the ID format)?

sgagniere: > Should we instead place limitations on what users can name their resources (i.e. just don't let them mirror the ID format)?
I'm not sure what the best path here is, but this we can't do. There's nothing stopping a user from making these names in the UI, and it's probably not good to introduce new naming limitations anyway.

kevin-wu24: Maybe we should always try to run a command with the assumption of the user using an ID, but if it trying the request with the supposed ID fails, we then can treat it as a name, try to find its corresponding ID, and if that still fails, return the error to the user.

internal/cmd/environment/command_describe.go
```diff
@@ -16,7 +16,7 @@ type out struct {
 
 func (c *command) newDescribeCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "describe [id]",
+		Use:               "describe [id/name]",
```
brianstrauch, line 19:
```suggestion
		Use:               "describe [id|name]",
```

internal/cmd/environment/command_use.go
```diff
@@ -25,13 +25,16 @@ func (c *command) newUseCommand() *cobra.Command {
...
-
+	id := args[0]
 	if _, err := c.V2Client.GetOrgEnvironment(id); err != nil {
-		return errors.NewErrorWithSuggestions(err.Error(), "List available environments with `confluent environment list`.")
+		id, err = convertEnvironmentNameToId(id, c.V2Client)
+		if err != nil {
+			return err
+		}
+		_, err = c.V2Client.GetOrgEnvironment(id)
+		if err != nil {
```
brianstrauch, line 35:
```suggestion
		if _, err = c.V2Client.GetOrgEnvironment(id); err != nil {
```

internal/cmd/kafka/command_cluster.go
```diff
@@ -85,3 +88,16 @@ func (c *clusterCommand) validArgs(cmd *cobra.Command, args []string) []string {
 	}
 	return suggestions
 }
+
+func convertClusterNameToId(input string, environmentId string, v2Client *ccloudv2.Client) (string, error) {
```
brianstrauch, line 92:
```suggestion
func (c *command) convertClusterNameToId(name, environmentId string) (string, error) {
```

kevin-wu24: I think I have to keep it the header this way to make it reusable for the `--cluster` flag, which I'll handle in dynamic_context.go. I think I might move this function to `pkg/utils` since it might result in a circular import

internal/cmd/environment/command_create.go
```diff
@@ -41,7 +41,8 @@ func (c *command) create(cmd *cobra.Command, args []string) error {
 	}
 
 	c.Context.AddEnvironment(environment.GetId())
+	c.Config.ValidEnvNamesToIds[environment.GetDisplayName()] = environment.GetId()
+	c.Config.ValidEnvIdsToNames[environment.GetId()] = environment.GetDisplayName()
```
brianstrauch, line 45: What's the reasoning for storing this in the configuration file? 🤔

kevin-wu24: The environment variable in the dynamic context is just a string, so there's no way to verify that it's actually a valid name corresponding to an ID without calling the list API (which we want to avoid calling if possible) if it is passed into an arbitrary command as a flag. Storing a mapping of environment names -> ids + vice versa in the config makes more sense here since we also want the mapping to persistent across multiple commands after the first time a user tries using the `--environment` flag (doesn't make sense to populate the map before then either because the current environment is assumed in the config).

kevin-wu24: I just realized but I probably want to do the same with `--cluster` too if you think it's worth it to basically "cache" these two mappings in the config to avoid extra list API calls. My concern is that the user's config might become very cluttered by these mappings. But if we choose not to maintain the mappings in some persistent way, every command that is run with a `--cluster/--environment` flag where the value's prefix is not the correct one would result in the list API being called in order to check if the value is a valid resource name. How should we value the time vs space tradeoff here? I think I would prefer to maintain the mapping some way (if we can avoid the config that's probably the best).

brianstrauch: > The environment variable in the dynamic context is just a string

What does this mean?

> Storing a mapping of environment names -> ids + vice versa in the config makes more sense here

We've gotten in trouble for doing this in the past. What if you store an ID -> name mapping in the configuration file, and then change the name of the resource on the website?

kevin-wu24: > > The environment variable in the dynamic context is just a string
> 
> What does this mean?
> 
The Environment field in the context struct is just a string, not the orgV2Environment struct that is returned by the API. Maybe we should change this?

kevin-wu24: > We've gotten in trouble for doing this in the past. What if you store an ID -> name mapping in the configuration file, and then change the name of the resource on the website?

Yeah the config mapping would have no idea about UI changes to resources. I think one way we can avoid this is to retry a command when it fails when using `--cluster` or `--environment` and only then make the API call.

brianstrauch: > The Environment field in the context struct is just a string, not the orgV2Environment struct that is returned by the API.

That's intentional. The idea is that's the only field in the struct that's not subject to change (i.e. by interacting with the UI)

brianstrauch: > This would worst case result in 2 extra list API calls on top of the command being run if both the cluster and environment are passed in as names rather than IDs.

Yeah, I think this is the best we can do. Fortunately, both of these calls will return a few pages of results as the worst case. What we want to watch out for is any resources that a user might potentially have hundreds or thousands of, in which case we might need to make 100s of paginated API calls.

kevin-wu24: For flags it looks like there's only `identity-provider`, `ksql-cluster`, `service-account` besides the two mentioned above.

internal/cmd/kafka/command_cluster.go
```diff
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"fmt"
+	presource "github.com/confluentinc/cli/internal/pkg/name-conversions"
```
brianstrauch, line 5: Which IDE are you using? Most IDEs can sort these imports automatically for you, FYI.

kevin-wu24: I'm using Goland. For import sorting type, do we use `goimports`?

brianstrauch: Yep, `goimports` should get you 95% of the way there. For the other 5% we use `gci` in our linter which unfortunately doesn't have an auto-format tool.

internal/cmd/iam/command_pool.go
```diff
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	pconv "github.com/confluentinc/cli/internal/pkg/name-conversions"
```
brianstrauch, line 7: No reason to rename the package name here, there's no naming conflict as far as I can tell.
```suggestion
	"github.com/confluentinc/cli/internal/pkg/name-conversions"
```

internal/cmd/iam/command_pool.go
```diff
@@ -49,3 +50,15 @@ func (c *identityPoolCommand) validArgs(cmd *cobra.Command, args []string) []str
 	provider, _ := cmd.Flags().GetString("provider")
 	return pcmd.AutocompleteIdentityPools(c.V2Client, provider)
 }
+
+func (c *identityPoolCommand) poolAndProviderNamesToIds(pool string, provider string) (string, string, error) {
```
brianstrauch, line 54: Why are these being grouped in the same function? What if someone passes a pool ID and provider name?

internal/cmd/iam/command_provider_update.go
```diff
@@ -58,7 +59,14 @@ func (c *identityProviderCommand) update(cmd *cobra.Command, args []string) erro
 
 	identityProvider, err := c.V2Client.UpdateIdentityProvider(update)
 	if err != nil {
-		return err
+		provId, err := pconv.ConvertIamProviderNameToId(args[0], c.V2Client)
```
brianstrauch, line 62: I would call this variable `providerId` or just `id`, not a huge fan of shortening names unless it's more common (for example, "directory" -> "dir"). No one looks at "prov" and immediately thinks "provider"

internal/cmd/kafka/command_cluster_describe.go
```diff
@@ -75,11 +74,7 @@ func (c *clusterCommand) describe(cmd *cobra.Command, args []string) error {
 
 	cluster, httpResp, err := c.V2Client.DescribeKafkaCluster(lkc, environmentId)
 	if err != nil {
-		environmentId, err = pconv.ConvertEnvironmentNameToId(environmentId, c.V2Client)
-		if err != nil {
-			return err
-		}
-		lkc, err = pconv.ConvertClusterNameToId(lkc, environmentId, c.V2Client)
+		lkc, environmentId, err = c.clusterAndEnvNamesToIds(lkc, environmentId)
```
brianstrauch, line 77: If we're potentially passing in a name, why is this variable called "lkc"? That's in reference to an ID.

internal/cmd/kafka/command_cluster_update.go
```diff
@@ -56,11 +55,7 @@ func (c *clusterCommand) update(cmd *cobra.Command, args []string) error {
 	clusterID := args[0]
 	currentCluster, _, err := c.V2Client.DescribeKafkaCluster(clusterID, environmentId)
 	if err != nil {
-		environmentId, err = pconv.ConvertEnvironmentNameToId(environmentId, c.V2Client)
-		if err != nil {
-			return err
-		}
-		clusterID, err = pconv.ConvertClusterNameToId(clusterID, environmentId, c.V2Client)
+		clusterID, environmentId, err = c.clusterAndEnvNamesToIds(clusterID, environmentId)
```
brianstrauch, line 58: Please use consistent naming!
```suggestion
		clusterId, environmentId, err = c.clusterAndEnvNamesToIds(clusterId, environmentId)
```

internal/cmd/iam/command_pool.go
kevin-wu24: When we call the Get API for resources like `iam pool` that take in a flag of another resource that can be a name or ID like `iam provider` and the API call fails, we don't know if the error because the `pool` was a name or invalid or if the `provider` flag value was a name or invalid. Therefore we would have to try to convert all of the potential names associated with an execution, starting with the flag(s) since the resource API call depends on the flags. I handle this here: https://github.com/confluentinc/cli/pull/2091/commits/00909f10938acb7cdef362feeb200cbb038aafb7#diff-b793b1ec65c5190a9e06fdb8284e914a0241bb31018ffbc9b81dea7f248cd090R36-R56, and I think first trying a GET request with the resource is the best thing to do here since we don't know if it's valid or not, but if it is, we avoid the list API call.

internal/cmd/environment/command_describe.go
```diff
@@ -1,11 +1,11 @@
 package environment
 
 import (
-	presource "github.com/confluentinc/cli/internal/pkg/name-conversions"
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
 	"github.com/confluentinc/cli/internal/pkg/errors"
+	nameconversions "github.com/confluentinc/cli/internal/pkg/name-conversions"
```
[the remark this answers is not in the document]
kevin-wu24: I might have to keep this in a separate package to prevent a circular import with `dynamic_context` and `resource` later on, but if I can combine it with `resource` I will.

internal/pkg/name-conversions/name-conversions.go
```diff
@@ -0,0 +1,97 @@
...
+	quotas, err := v2Client.ListKafkaQuotas(clusterId, environmentId)
+	if err != nil {
+		return input, err
+	}
+	quotaPtrs := ConvertToPtrSlice(quotas)
+	specPtrs := make([]*v1.KafkaQuotasV1ClientQuotaSpec, len(quotas))
+	for i := range quotas {
+		specPtrs[i] = quotaPtrs[i].Spec
+	}
+	return ConvertSpecNameToId(input, quotaPtrs, specPtrs)
```
brianstrauch, line 96: What's going on here? Are any of these conversions unecessary?

test/fixtures/output/api-key/54.golden
```diff
@@ -28,8 +28,8 @@ Flags:
       --description string       Description of API key.
       --use                      Use the created API key for the provided resource.
       --context string           CLI context name.
-      --environment string       Environment ID.
-      --service-account string   Service account ID.
+      --environment string       Environment ID or name.
```
brianstrauch, line 31: Thoughts on changing this to "Confluent Cloud evironment."?

internal/pkg/dynamic-config/dynamic_context.go
```diff
@@ -31,11 +32,18 @@ func NewDynamicContext(context *v1.Context, v2Client *ccloudv2.Client) *DynamicC
 	}
 }
 
-func (d *DynamicContext) ParseFlagsIntoContext(cmd *cobra.Command) error {
+func (d *DynamicContext) ParseFlagsIntoContext(cmd *cobra.Command, isTest bool) error {
 	if environment, _ := cmd.Flags().GetString("environment"); environment != "" {
 		if d.GetCredentialType() == v1.APIKey {
 			output.ErrPrintln("WARNING: The `--environment` flag is ignored when using API key credentials.")
 		} else {
+			if !isTest {
```
brianstrauch, line 40: Why are we choosing not to test this?

internal/pkg/name-conversions/conversions.go
```diff
@@ -0,0 +1,90 @@
...
+
+type resourcePtr interface {
+	GetId() string
+}
+
+type specPtr interface {
+	GetDisplayName() string
+}
+
+type v2ResourcePtr interface {
```
brianstrauch, line 25: Can you document the difference between `resourcePtr` and `v2ResourcePtr`? Not sure I understand.

internal/pkg/name-conversions/conversions.go
```diff
@@ -0,0 +1,90 @@
...
+func ConvertToPtrSlice[V resource](resources []V) []*V {
+	ptrs := make([]*V, len(resources))
+	for i := range resources {
+		ptrs[i] = &resources[i]
+	}
+	return ptrs
+}
+
+// ConvertSpecNameToId ConvertNamesToID returns a resource spec's name's corresponding ID or returns an error if not found
+func ConvertSpecNameToId[V resourcePtr, T specPtr](input string, resources []V, specs []T) (string, error) {
```
brianstrauch, line 39:
```suggestion
func ConvertSpecNameToId[R resourcePtr, S specPtr](input string, resources []R, specs []S) (string, error) {
```

internal/pkg/name-conversions/name-conversions.go
```diff
@@ -0,0 +1,97 @@
...
+	providers, err := v2client.ListIdentityProviders()
+	if err != nil {
+		return input, err
+	}
+	return ConvertV2NameToId(input, ConvertToPtrSlice(providers))
+}
+
+// IamServiceAccountNameToId attempts to convert from a valid iam service account name to its ID, if it fails it returns an error
+func IamServiceAccountNameToId(input string, v2client *ccloudv2.Client, tryGetFirst bool) (string, error) {
+	if tryGetFirst {
```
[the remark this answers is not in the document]
kevin-wu24: For the actual `service-account` commands (describe, update), if the API call fails, we know that the use did not pass in a valid resource ID, so the `GET` API call here is unnecessary

internal/pkg/name-conversions/name-conversions.go
```diff
@@ -0,0 +1,97 @@
...
+		return input, err
+	}
+	return ConvertV2NameToId(input, ConvertToPtrSlice(serviceAccounts))
+}
+
+// QuotaNameToId attempts to convert from a valid quota name to its ID, if it fails it returns an error
+func QuotaNameToId(input string, clusterId string, environmentId string, v2Client *ccloudv2.Client) (string, error) {
+	quotas, err := v2Client.ListKafkaQuotas(clusterId, environmentId)
+	if err != nil {
+		return input, err
```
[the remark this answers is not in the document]
kevin-wu24: I want to return the `input` string in these conversions when an error occurs so the calling function can print out an informative error (i.e. `No resource with name or ID "input" found....`) with suggestions to the user.

test/fixtures/output/api-key/43.golden
```diff
@@ -8,8 +8,8 @@
           |                    | owned by service account       |          |                       |                 |              |                       
           | MYKEY1             | first-key                      | u11      | bstrauch@confluent.io | kafka           | lkc-bob      | 1999-02-24T00:00:00Z  
           | MYKEY10            |                                | sa-12345 | <service account>     | kafka           | lkc-cool1    | 1999-02-24T00:00:00Z  
-          | MYKEY11            | auditlog-key                   | sa-1337  | <deactivated user>    | kafka           | lkc-cool1    | 1999-02-24T00:00:00Z  
-          | MYKEY12            | auditlog-key                   | sa-1337  | <deactivated user>    | kafka           | lkc-cool1    | 1999-02-24T00:00:00Z  
+          | MYKEY11            | auditlog-key                   | sa-1337  | <service account>     | kafka           | lkc-cool1    | 1999-02-24T00:00:00Z  
+          | MYKEY12            | auditlog-key                   | sa-1337  | <service account>     | kafka           | lkc-cool1    | 1999-02-24T00:00:00Z
```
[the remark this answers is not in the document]
kevin-wu24: I had to add this service-account with ID `sa-1337` to the service account list handler to test passing in resource names for the `--service-account` flag.

internal/pkg/name-conversions/name-conversions.go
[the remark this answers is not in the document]
brianstrauch: > if the API call fails

Which API call are we talking about here?

internal/pkg/dynamic-config/dynamic_context.go
kevin-wu24: This just prevents the `dynamic_context_test.go` unit test from calling `cluster` and `environment` list APIs, which will cause the test to fail because the mocker is not defined in the SDK. However, these blocks of code are still tested in integration tests.

brianstrauch: Should we add those mocks? Better yet, can we delete those tests and replace them with integration tests in a separate PR?

internal/pkg/name-conversions/name-conversions.go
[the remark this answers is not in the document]
kevin-wu24: The Get API call, sorry for not specifying. If it fails once, we don't need to try it again.

internal/pkg/name-conversions/name-conversions.go
kevin-wu24: I have to take the pointers of the resource `specs` because their `GetId()` function takes a pointer receiver.

internal/pkg/name-conversions/conversions.go
```diff
@@ -0,0 +1,90 @@
...
+// GetSpecNamesToIds returns a mapping from spec resource names to their respective IDs
+func GetSpecNamesToIds[V resourcePtr, T specPtr](resources []V, specs []T) (map[string]string, error) {
+	namesToIds := make(map[string]string, len(resources))
+	for i := range resources {
+		name := specs[i].GetDisplayName()
+		if _, ok := namesToIds[name]; !ok {
+			namesToIds[name] = resources[i].GetId()
+		} else {
+			return nil, errors.NewErrorWithSuggestions(fmt.Sprintf(DuplicateResourceNameErrorMsg, name), DuplicateResourceNameErrorSuggestions)
+		}
```
sgagniere, line 60: This will return an error if any duplicated name is found, even if the user is attempting to work with a non-duplicated name.

While not an issue for resources like service accounts or environments where the backend enforces unique names, duplicate names can happen with Kafka clusters.

One way to resolve this is by using a `map[string][]string` where the slice value stores all id's with that name, and then you can throw the duplicate error if the length of the slice that the user is accessing is `> 1`. There's probably more elegant solutions, though.

sgagniere (changes requested):

internal/pkg/name-conversions/name-conversions.go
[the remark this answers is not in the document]
brianstrauch: The counter-argument is that the describe API call is much faster than the list API call in the worst case. (It isn't paginated, while the list API call is)

internal/pkg/name-conversions/name-conversions.go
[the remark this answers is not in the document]
brianstrauch: Can we return that informative error here, directly?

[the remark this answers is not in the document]
kevin-wu24: Yeah I can do that, although for errors with the actual list API call I think we should just return the empty string and the error as suggested. What I meant by the comment above was for later on in the conversion functions.

internal/pkg/dynamic-config/dynamic_context.go
kevin-wu24: I think we should be able to remove this test: https://github.com/confluentinc/cli/blob/e5263f122489f5937216cfc44aba58ecbc658dfb/internal/pkg/dynamic-config/dynamic_context_test.go#L85
Since the functionality of `ParseFlagsIntoContext` is already being relied on in integration tests

internal/pkg/name-conversions/conversions.go
kevin-wu24: Yeah, I didn't think about that case. We definitely don't want to give the user an error unless it is the resource name they are currently using that is shared by two different resource IDs

internal/pkg/dynamic-config/dynamic_context.go
brianstrauch: Let's try to get full coverage of this function with our integration tests... and if it can be written in a way that's easier to unit test, even better!

internal/cmd/environment/command_update.go
```diff
@@ -34,9 +36,14 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 	environment := orgv2.OrgV2Environment{DisplayName: orgv2.PtrString(name)}
-	environment, err = c.V2Client.UpdateOrgEnvironment(args[0], environment)
-	if err != nil {
-		return err
+	if environment, err = c.V2Client.UpdateOrgEnvironment(args[0], environment); err != nil {
+		environmentId, err := nameconversions.EnvironmentNameToId(args[0], c.V2Client, false)
+		if err != nil {
+			return errors.NewErrorWithSuggestions(err.Error(), errors.NotValidEnvironmentIdSuggestions)
+		}
+		if environment, err = c.V2Client.UpdateOrgEnvironment(environmentId, environment); err != nil {
```
sgagniere, line 44: There's a variable name collision here. The `environment` for the request on line 38 gets modified by line 39, and that's what will be sent here.

internal/cmd/environment/command_update.go
```diff
@@ -34,9 +36,14 @@ func (c *command) update(cmd *cobra.Command, args []string) error {
...
 
 	environment := orgv2.OrgV2Environment{DisplayName: orgv2.PtrString(name)}
-	environment, err = c.V2Client.UpdateOrgEnvironment(args[0], environment)
-	if err != nil {
-		return err
+	if environment, err = c.V2Client.UpdateOrgEnvironment(args[0], environment); err != nil {
+		environmentId, err := nameconversions.EnvironmentNameToId(args[0], c.V2Client, false)
+		if err != nil {
+			return errors.NewErrorWithSuggestions(err.Error(), errors.NotValidEnvironmentIdSuggestions)
+		}
```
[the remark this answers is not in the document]
kevin-wu24: I think I might be able to address this with some changes to how the conversion functions will be called. I think it would be better to instead have resource name to ID conversions embedded in our SDK wrapper functions. 

Also, for this particular case, calling `EnvironmentNameToId` function would not change anything if the user supplies a valid ID (i.e. `environmentId == args[0]` and `err == nil`) since the request is not failing because the ID is wrong, but the `UpdateOrgEnvironment` API is called a second time where it would fail again because the `--name "?"` is invalid, but it would return the backend invalid name flag error.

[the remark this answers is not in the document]
sgagniere: Replying to the second paragraph: the issue I see with this pattern is that the code will go into the `NameToId` function even when the `id` is valid but the request is not.

If `tryGetFirst` is false, then these functions will treat the `id` as a `name`, and attempt to map it to its corresponding `id`. Since the `id` itself is not a name, it maps to nothing and we get `ResourceNameNotFoundErrorMsg`.

This causes the cli to error out before it reaches the second API call, which means it won't return the correct error.

Same thing w/ the new version of `UpdateOrgEnvironment`; the second `Execute()` is not reached in this case.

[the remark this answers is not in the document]
kevin-wu24: Oh I see what you mean. When we go into the `NameToId` function for any API call we would need to try the `GET` API call to validate the ID, since the API call could fail for reasons other than an invalid ID (does this apply to `GET` calls from `describe` commands or validating flag resource IDs? or just `update` requests?) 

Do you know if the backend returns a different error code for when the resource ID is invalid vs when other parts of the API request are invalid? If that's the case we can let the second error go through and only catch the first with the `nameToId` conversion.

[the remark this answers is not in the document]
sgagniere: > does this apply to GET calls from describe commands or validating flag resource IDs? or just update requests?
I don't think this problem applies to GET requests, since the only reason those should fail is if the `id` (or other things like the environment `id`) is invalid. Maybe a rate limit error is also possible, though?

For the second question: they should be different codes. For environment, it returns 403 unauthorized when the id is wrong and 400 when it's an otherwise bad request.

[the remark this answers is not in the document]
kevin-wu24: Okay, I think I should be able to do the approach mentioned in my second question, since user flag value input is just validated with a `GET` request, so the only times `GET` requests should be failing is if the resource trying to be retrieved is an invalid ID. And for `update` requests I can check the error code and convert if it's `403`.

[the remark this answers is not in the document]
sgagniere: We can try that, but I hesitate to filter only one one status code because the apis do have other status codes they could return, and some of them might have useful information.

Another idea: what if you build the map so that `id`s are also keys? Like:
```
name-1: [env-1],
env-1:  [env-1],
name-2: [env-2, env-3],
env-2:  [env-2],
env-3:  [env-3],
...
...
```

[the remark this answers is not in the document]
kevin-wu24: I can do some more research into what codes are returned for what situations, but ideally, we would want to only call the list API for a resource if it returns an error code signifying the resource ID in the request was invalid. And hopefully (I really hope lol) that these error codes are standardized across all (or at least a majority) of our backend resources that this feature impacts.

kevin-wu24:
The best way to implement this functionality may not be to convert from resource name to id via calling the `LIST` API when an API request fails as per this thread: https://confluent.slack.com/archives/CG6BW233L/p1689973702098589

PR #2092, opened 2023-07-12
[CLI-2592] Correctly default to Avro when a schema type is not returned

brianstrauch:
Release Notes
-------------
Bug Fixes
- Fix "type" output in `confluent schema-registry schema describe` and properly format Avro schemas

Checklist
---------
   * yes: ok

What
----
Fixing a typo from an earlier PR, this time with updated tests 🤦‍♂️

Test & Review
-------------
Modify tests to match backend response.


sgagniere (approved):
+1

PR #2093, opened 2023-07-13
Fix/kfs 1015 fix set statements

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Refactors how we store user properties in the client. Instead of using a map we now have a custom struct that allows to store default properties which can't be deleted but only reset to the default value. This is useful for the catalog, database and timezone for which the default values will be set at the start of the session. 

It also fixes persisting the catalog name, which did not work correctly before. The persisted catalog was always overwritten in command_shell and could thus never be recovered in the store. We now do the following for catalog and database:
**Catalog**:
- We check if there is a catalog stored in the config, if so we use that
- If not, we use the name of the environment the user provided with the envId flag
- otherwise (if env name can't be retrieved) we return an error and don't start the client

**Database**:
- If the user provided a default name with the --database flag, we use that
- If not, we check if there is a database stored in the config, if so we use that
- if not, we use the current active Kafka cluster name, if set
- otherwise it will be empty


internal/pkg/flink/internal/store/store.go
```diff
@@ -49,6 +50,22 @@ func (s *Store) ProcessLocalStatement(statement string) (*types.ProcessedStateme
 	}
 }
 
+func (s *Store) persistUserProperties() {
+	if s.appOptions.GetContext() != nil {
+		if err := s.appOptions.Context.SetCurrentFlinkCatalog(s.Properties.Get(config.ConfigKeyCatalog)); err != nil {
+			log.CliLogger.Errorf("error setting current flink catalog: %v", err)
```
Guttz, line 56: just a suggestion - if the users sees this 10 times after trying to execute a reset command with another key, he'll ask himself "I'm not trying set the flink catalog. Nitpick tho

```suggestion
			log.CliLogger.Errorf("error persisting current flink catalog: %v", err)
```

internal/pkg/flink/internal/store/store.go
```diff
@@ -49,6 +50,22 @@ func (s *Store) ProcessLocalStatement(statement string) (*types.ProcessedStateme
...
+		}
+
+		if err := s.appOptions.Context.SetCurrentFlinkDatabase(s.Properties.Get(config.ConfigKeyDatabase)); err != nil {
+			log.CliLogger.Errorf("error setting current flink database: %v", err)
```
Guttz, line 60: same as above

Guttz (approved):

internal/pkg/flink/internal/store/user_properties_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (s *UserPropertiesTestSuite) getRandomKey(fromMap map[string]string) string {
+	return rapid.SampledFrom(
+		lo.MapToSlice(fromMap, func(key string, value string) string { return key })).
+		Example()
+}
+
+func (s *UserPropertiesTestSuite) TestMapShouldAddKeys() {
+	standardMap := s.addSomeKeys()
+	standardMap[s.defaultKey] = s.defaultValue
```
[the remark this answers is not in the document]
YannickPferr: yeah I also thought about that, let's do it then!

internal/pkg/flink/internal/store/store.go
YannickPferr: True!

PR #2095, opened 2023-07-13
Deprecate `kafka topic produce --schema-id`

brianstrauch:
Release Notes
-------------
New Features
- Support passing schema IDs to the `--schema` flag in `confluent kafka topic produce`

Checklist
---------
   * yes: ok

What
----
In https://confluentinc.atlassian.net/browse/CLI-2182, we want to support schemas for message keys. Currently, this would require adding `--key-schema` and `--key-schema-id` to mirror `--schema` and `--schema-id`. This PR allows passing IDs to `--schema` so we can deprecate `--schema-id`. (... so in a following PR we only need to add `--key-schema`).

Test & Review
-------------
Updated relevant tests

### DIFF
--- a/cmd/lint/main.go
+++ b/cmd/lint/main.go
@@ -224,6 +224,7 @@ var vocabWords = []string{
 	"env",
 	"eu",
 	"failover",
+	"filepath",
 	"flink",
 	"formatter",
 	"gcp",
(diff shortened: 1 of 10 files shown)


sgagniere (approved):
👍

PR #2096, opened 2023-07-13
DGS-6578 AsyncAPI: Add correct Kafka bindings section

kcorman0:
Release Notes
-------------

Bug Fixes
- Fix `confluent asyncapi import` and `confluent asyncapi export` to follow AsyncAPI documentation for Kafka channel bindings, with explicit fields for `bindingVersion`, `partitions`, and `topicConfiguration`.

Checklist
---------
   * yes: ok

What
----

We are not outputting Kafka bindings in the correct format. Previously, our Kafka bindings struct looked like this (x- in front of a field means that it is custom, not specified in the asyncapi documentation)

```
type confluentBinding struct {
	XPartitions int32             `json:"x-partitions,omitempty"`
	XConfigs    map[string]string `json:"x-configs,omitempty"`
}
```

However, not all of these fields should have been custom. The asyncapi spec does include a Kafka bindings section (under references), its just not exposed by the library we are using to help us convert from Kafka<->Asyncapi.

This PR fixes the following problem by using a new struct that includes all of the fields specified in the latest version of the  asyncapi documentation (keeping x-configs to include fields not listed in asyncapi documentation)

```
type confluentBinding struct {
	BindingVersion string            `json:"bindingVersion,omitempty"`
	Partitions     int32             `json:"partitions,omitempty"`
	TopicConfigs   topicConfig       `json:"topicConfiguration,omitempty"`
	XConfigs       map[string]string `json:"x-configs,omitempty"`
}
```
FF for another example - https://confluentinc.atlassian.net/browse/FF-10843

References
----------

Asyncapi Kafka bindings
https://github.com/asyncapi/bindings/blob/master/kafka/README.md

Test & Review
-------------

Updated integration tests to follow new format

Tested by creating topics/schemas on confluent cloud, exporting them and importing on another cluster (checking that the topic configs are correct).

Open Questions / Follow-ups
---------------------------

Can link https://github.com/asyncapi/bindings/blob/master/kafka/README.md in release notes if seems helpful

Tried but export code maybe could be cleaner

### DIFF
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -35,13 +35,13 @@ type command struct {
 }
 
 type confluentBinding struct {
-	BindingVersion     string             `json:"bindingVersion,omitempty"`
-	Partitions         int32              `json:"partitions,omitempty"`
-	TopicConfiguration topicConfiguration `json:"topicConfiguration,omitempty"`
-	XConfigs           map[string]string  `json:"x-configs,omitempty"`
+	BindingVersion     string                   `json:"bindingVersion,omitempty"`
+	Partitions         int32                    `json:"partitions,omitempty"`
+	TopicConfiguration topicConfigurationExport `json:"topicConfiguration,omitempty"`
+	XConfigs           map[string]string        `json:"x-configs,omitempty"`
 }
 
-type topicConfiguration struct {
+type topicConfigurationExport struct {
 	CleanupPolicy       []string `json:"cleanup.policy,omitempty"`
 	RetentionTime       int64    `json:"retention.ms,omitempty"`
 	RetentionSize       int64    `json:"retention.bytes,omitempty"`
@@ -352,7 +352,7 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
 	}
 
 	// Turn topicConfigMap into correct format
-	topicConfigs := topicConfiguration{}
+	topicConfigs := topicConfigurationExport{}
 	jsonString, err := json.Marshal(topicConfigMap)
 	if err != nil {
 		return nil, err
--- a/internal/cmd/asyncapi/command_export.go
+++ b/internal/cmd/asyncapi/command_export.go
@@ -2,8 +2,10 @@ package asyncapi
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -25,6 +27,7 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/log"
 	"github.com/confluentinc/cli/internal/pkg/output"
 	"github.com/confluentinc/cli/internal/pkg/serdes"
+	"github.com/confluentinc/cli/internal/pkg/types"
 )
 
 type command struct {
@@ -32,8 +35,18 @@ type command struct {
 }
 
 type confluentBinding struct {
-	XPartitions int32             `json:"x-partitions,omitempty"`
-	XConfigs    map[string]string `json:"x-configs,omitempty"`
+	BindingVersion string            `json:"bindingVersion,omitempty"`
+	Partitions     int32             `json:"partitions,omitempty"`
+	TopicConfigs   topicConfig       `json:"topicConfiguration,omitempty"`
+	XConfigs       map[string]string `json:"x-configs,omitempty"`
+}
+
+type topicConfig struct {
+	CleanupPolicy       []string `json:"cleanup.policy,omitempty"`
+	RetentionTime       int64    `json:"retention.ms,omitempty"`
+	RetentionSize       int64    `json:"retention.bytes,omitempty"`
+	DeleteRetentionTime int64    `json:"delete.retention.ms,omitempty"`
+	MaxMessageSize      int32    `json:"max.message.bytes,omitempty"`
 }
 
 type bindings struct {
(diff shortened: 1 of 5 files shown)


internal/cmd/asyncapi/command_export.go
```diff
@@ -25,15 +27,26 @@ import (
...
 type command struct {
 	*pcmd.AuthenticatedCLICommand
 }
 
 type confluentBinding struct {
-	XPartitions int32             `json:"x-partitions,omitempty"`
-	XConfigs    map[string]string `json:"x-configs,omitempty"`
+	BindingVersion string            `json:"bindingVersion,omitempty"`
+	Partitions     int32             `json:"partitions,omitempty"`
+	TopicConfigs   topicConfig       `json:"topicConfiguration,omitempty"`
```
brianstrauch, line 40: Let's match with the JSON name for clarity
```suggestion
	TopicConfiguration   topicConfiguration       `json:"topicConfiguration,omitempty"`
```

internal/cmd/asyncapi/command_export.go
```diff
@@ -306,13 +319,53 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
 	if partitionsResp.Data != nil {
 		numPartitions = int32(len(partitionsResp.Data))
 	}
-	configsMap := make(map[string]string)
+	customConfigMap := make(map[string]string)
+	topicConfigMap := make(map[string]interface{})
```
brianstrauch, line 323:
```suggestion
	topicConfigMap := make(map[string]any)
```

internal/cmd/asyncapi/command_export.go
```diff
@@ -306,13 +319,53 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
...
+		"cleanup.policy",
+		"delete.retention.ms",
+		"max.message.bytes",
+		"retention.bytes",
+		"retention.ms",
+	)
 	for _, config := range configs.Data {
-		configsMap[config.GetName()] = config.GetValue()
+		if topicConfigOptions.Contains(config.GetName()) {
+			if config.GetName() == "cleanup.policy" {
```
brianstrauch, line 335: `switch`?

internal/cmd/asyncapi/command_export.go
```diff
@@ -306,13 +319,53 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
...
+	}
+
+	// Turn topicConfigMap into correct format
+	topicConfigs := topicConfig{}
+	jsonString, err := json.Marshal(topicConfigMap)
+	if err != nil {
+		return nil, err
 	}
+	err = json.Unmarshal(jsonString, &topicConfigs)
+	if err != nil {
```
brianstrauch, line 360:
```suggestion
	if err := json.Unmarshal(jsonString, &topicConfigs); err != nil {
```

internal/cmd/asyncapi/command_import.go
```diff
@@ -41,8 +42,18 @@ type flagsImport struct {
 }
 
 type kafkaBinding struct {
-	XPartitions int32             `yaml:"x-partitions"`
-	XConfigs    map[string]string `yaml:"x-configs"`
+	BindingVersion string            `yaml:"bindingVersion"`
+	Partitions     int32             `yaml:"partitions"`
+	TopicConfigs   topicConfigPtr    `yaml:"topicConfiguration"`
```
brianstrauch, line 47:
```suggestion
	TopicConfiguration   topicConfiguration    `yaml:"topicConfiguration"`
```

internal/cmd/asyncapi/command_import.go
```diff
@@ -330,6 +341,41 @@ func (c *command) updateTopic(details *accountDetails, topicName string, kafkaBi
...
 
+// TopicConfigs and XConfigs are both specifying Kafka configs, XConfigs is for those not specified in async api.
+// This function combines TopicConfigs and XConfigs
+func combineTopicConfigs(kafkaBinding kafkaBinding) map[string]string {
+	var configs map[string]string
+	if kafkaBinding.XConfigs != nil {
+		configs = kafkaBinding.XConfigs
+	} else {
+		configs = make(map[string]string)
+	}
```
brianstrauch, line 352:
```suggestion
	configs := make(map[string]string)
	if kafkaBinding.XConfigs != nil {
		configs = kafkaBinding.XConfigs
	}
```

internal/cmd/asyncapi/command_import.go
```diff
@@ -330,6 +341,41 @@ func (c *command) updateTopic(details *accountDetails, topicName string, kafkaBi
...
+		if !fieldValue.IsNil() {
+			elem := fieldValue.Elem()
+			if elem.Kind() == reflect.Slice {
+				ss := make([]string, elem.Len())
+				for i := 0; i < elem.Len(); i++ {
+					ss[i] = elem.Index(i).String()
+				}
+				value = strings.Join(ss, ",")
+			} else {
+				value = strconv.FormatInt(elem.Int(), 10)
```
brianstrauch, line 371: So it looks like we're converting slices to a comma-delimited string, and we're converting integers to strings? Do we really need to use reflection here?

internal/cmd/asyncapi/command_import.go
```diff
@@ -42,13 +41,13 @@ type flagsImport struct {
...
-	TopicConfigs   topicConfigPtr    `yaml:"topicConfiguration"`
-	XConfigs       map[string]string `yaml:"x-configs"`
+	BindingVersion     string                `yaml:"bindingVersion"`
+	Partitions         int32                 `yaml:"partitions"`
+	TopicConfiguration topicConfigurationPtr `yaml:"topicConfiguration"`
+	XConfigs           map[string]string     `yaml:"x-configs"`
 }
 
-type topicConfigPtr struct {
+type topicConfigurationPtr struct {
```
brianstrauch, line 50: Why does this end with "ptr"? The `type` itself is not a pointer, so is it named this way because it contains pointers?

internal/cmd/asyncapi/command_import.go
```diff
@@ -341,38 +340,31 @@ func (c *command) updateTopic(details *accountDetails, topicName string, kafkaBi
...
-			configs[fieldType.Tag.Get("yaml")] = value
-		}
 	}
+
+	topicConfig := kafkaBinding.TopicConfiguration
+	if topicConfig.CleanupPolicy != nil {
+		configs["cleanup.policy"] = strings.Join(*topicConfig.CleanupPolicy, ",")
+	}
+	if topicConfig.RetentionTime != nil {
+		configs["retention.ms"] = strconv.FormatInt(*topicConfig.RetentionTime, 10)
```
brianstrauch, line 356: Let's pick a different function that doesn't involve changing the base of the integer (I assume we always want base 10 here 😆). I think these should be equivalent:

```suggestion
		configs["retention.ms"] = fmt.Sprint(*topicConfig.RetentionTime)
```

```suggestion
		configs["retention.ms"] = fmt.Sprintf("%d", *topicConfig.RetentionTime)
```

internal/cmd/asyncapi/command_import.go
kcorman0: @brianstrauch topicConfiguration is used in command_export, but cant reuse because the fields have to be pointers here (nil when empty instead of default 0, they can define it to 0 cant tell difference)

not sure on a good name, topicConfigurationImport? haha

brianstrauch: > not sure on a good name, topicConfigurationImport?

Sure! `topicConfigurationImport` and `topicConfigurationExport` could be good for helping to differentiate between the two.

internal/cmd/asyncapi/command_export.go
```diff
@@ -306,13 +319,53 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
 	if partitionsResp.Data != nil {
 		numPartitions = int32(len(partitionsResp.Data))
 	}
-	configsMap := make(map[string]string)
+	customConfigMap := make(map[string]string)
+	topicConfigMap := make(map[string]any)
+
+	// Determine whether the given config value can be put into the AsyncAPI Kakfa bindings or put into our custom struct for extra configs
```
brianstrauch, line 325:
```suggestion
	// Determine whether the given config value can be put into the AsyncAPI Kafka bindings or put into our custom struct for extra configs
```

internal/cmd/asyncapi/command_export.go
```diff
@@ -306,13 +319,53 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
...
+		if topicConfigOptions.Contains(config.GetName()) {
+			switch config.GetName() {
+			case "cleanup.policy":
+				topicConfigMap[config.GetName()] = strings.Split(config.GetValue(), ",")
+			case "max.message.bytes":
+				topicConfigMap[config.GetName()], err = strconv.ParseInt(config.GetValue(), 10, 32)
+				if err != nil {
+					return nil, err
+				}
+			default:
```
brianstrauch, line 343: Does this encompass "delete.retention.ms", "retention.bytes", and "retention.ms"? Can we write all of those to be more explicit?

That'll let us use the `default` block to add to the custom config map, and then we won't need to use a `set` anymore!

internal/cmd/asyncapi/command_export.go
```diff
@@ -322,31 +321,22 @@ func (c *command) getBindings(clusterId, topicName string) (*bindings, error) {
...
-				}
+		switch config.GetName() {
+		case "cleanup.policy":
+			topicConfigMap[config.GetName()] = strings.Split(config.GetValue(), ",")
+		case "max.message.bytes":
+			topicConfigMap[config.GetName()], err = strconv.ParseInt(config.GetValue(), 10, 32)
+			if err != nil {
+				return nil, err
+			}
+		case "delete.retention.ms", "retention.bytes", "retention.ms":
```
brianstrauch, line 334: Thanks for making this change! I think `switch` uses a HashSet under the hood, so the runtime should be the same :smile:

brianstrauch (approved):

PR #2097, opened 2023-07-13
notify context is cloud when run on-prem cmds after cloud login

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Clear error message when user runs on premise commands when logged into Confluent Cloud

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

if logged in to cloud but runs a cp cmd, will notify:
```
mhe@C02DV0KVMD6T cli % confluent cluster list

Suggestions:
    Log in with `confluent login --url <mds-url>`.
    Run command with "--help" to see available commands.
```

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/config/v1/config.go
+++ b/internal/pkg/config/v1/config.go
@@ -64,7 +64,7 @@ var (
 		"you must log in to Confluent Platform to use this command",
 		"Log in to Confluent Platform with `confluent login --url <mds-url>`.",
 	)
-	CloudLoginRequireOnPremLoginErr = errors.NewErrorWithSuggestions(
+	RunningOnPremCommandInCloudErr = errors.NewErrorWithSuggestions(
 		"this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
 		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Run command with "--help" to see available commands.`,
 	)
@@ -655,7 +655,7 @@ func (c *Config) CheckIsOnPremLogin() error {
 		if !c.isCloud() {
 			return nil
 		} else {
-			return CloudLoginRequireOnPremLoginErr
+			return RunningOnPremCommandInCloudErr
 		}
 	}
 	return RequireOnPremLoginErr
--- a/internal/pkg/config/v1/config.go
+++ b/internal/pkg/config/v1/config.go
@@ -66,7 +66,7 @@ var (
 	)
 	RunningOnPremCommandInCloudErr = errors.NewErrorWithSuggestions(
 		"this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
-		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Run command with "--help" to see available commands.`,
+		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Use the "--help" flag to see available commands.`,
 	)
 )
 
--- a/internal/pkg/config/v1/config.go
+++ b/internal/pkg/config/v1/config.go
@@ -62,7 +62,11 @@ var (
 	)
 	RequireOnPremLoginErr = errors.NewErrorWithSuggestions(
 		"you must log in to Confluent Platform to use this command",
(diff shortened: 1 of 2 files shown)


internal/pkg/config/v1/config.go
```diff
@@ -62,7 +62,11 @@ var (
 	)
 	RequireOnPremLoginErr = errors.NewErrorWithSuggestions(
 		"you must log in to Confluent Platform to use this command",
-		"Log in with `confluent login --url <mds-url>`.",
+		"Log in to Confluent Platform with `confluent login --url <mds-url>`.",
+	)
+	CloudLoginRequireOnPremLoginErr = errors.NewErrorWithSuggestions(
+		"this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
```
sgagniere, line 68: The second half of this error is the same as `RequireOnPremLoginErr`; is the first half here in case the user doesn't realize that they're logged into cloud?

MuweiHe: yes. they're logged into cloud but tried running a cp cmd. i'm unhappy with my naming. do you have suggestions?

sgagniere: How about `RunningOnPremCommandInCloudErr`?

sgagniere: Alternatively, you could just get rid of this one and just reuse `RequireOnPremLoginErr`. Hopefully even without the note about Cloud, that error should cause people to double check their login.

internal/pkg/config/v1/config.go
```diff
@@ -647,8 +651,12 @@ func (c *Config) GetFilename() string {
 
 func (c *Config) CheckIsOnPremLogin() error {
 	ctx := c.Context()
-	if ctx != nil && ctx.PlatformName != "" && !c.isCloud() {
-		return nil
+	if ctx != nil && ctx.PlatformName != "" {
+		if !c.isCloud() {
+			return nil
+		} else {
+			return CloudLoginRequireOnPremLoginErr
```
brianstrauch, line 658: We should still print this if the user isn't logged in to **either** Cloud or CP.

brianstrauch (changes requested):

internal/pkg/config/v1/config.go
MuweiHe: it'll be handled by `CheckIsNonAPIKeyCloudLoginOrOnPremLogin()`. This way i'm keeping the original logic and only adding a smaller branch

internal/pkg/config/v1/config.go
MuweiHe: it's getting used in a few places so easier to name this a separate variable

internal/pkg/config/v1/config.go
```diff
@@ -62,7 +62,11 @@ var (
 	)
 	RequireOnPremLoginErr = errors.NewErrorWithSuggestions(
 		"you must log in to Confluent Platform to use this command",
-		"Log in with `confluent login --url <mds-url>`.",
+		"Log in to Confluent Platform with `confluent login --url <mds-url>`.",
+	)
+	RunningOnPremCommandInCloudErr = errors.NewErrorWithSuggestions(
+		"this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
+		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Run command with "--help" to see available commands.`,
```
sgagniere, line 69:
```suggestion
		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Run this command with "--help" to see available commands.`,
```

Or maybe `Use the "--help" flag to...`.

sgagniere (approved):

internal/pkg/config/v1/config.go
MuweiHe: i like `use the help flag` better, cz if you're running `confluent cluster --help` it's no point anyways

sgagniere (approved):

PR #2099, opened 2023-07-14
[CLI-2625] Add `flink statement create`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Add `confluent flink statement create`

Test & Review
-------------
Integration tests


Guttz:
Nice that you're picking that up!

Feedback for flags, you need
 
- compute-pool (can be set via use, but nice to have as flag)
- identity-pool (can be set via use, but nice to have as flag)
- environment (can be set via use, but nice to have as flag)
- database
- name - there is a discussion in the thread

 You also need environment (ID) to use the GatewayClient. 

What we did for flink shell is that we have the `environment` flag and we fetch the name of the environment and use is as the `catalog` value instead of also having the catalog flag.

brianstrauch:
> Feedback for flags, you need

Already added the first three! Will add `--database` and `--name` today, and it sounds like I should omit the `--catalog` flag and use the current environment name?

pkg/dynamic-config/dynamic_context.go
```diff
@@ -60,6 +60,14 @@ func (d *DynamicContext) ParseFlagsIntoContext(cmd *cobra.Command) error {
 		}
 	}
 
+	if identityPool, _ := cmd.Flags().GetString("identity-pool"); identityPool != "" {
+		ctx := d.Config.Context()
+		d.Config.SetOverwrittenIdentityPool(ctx.GetCurrentFlinkComputePool())
```
kevin-wu24, line 65:
```suggestion
		d.Config.SetOverwrittenIdentityPool(ctx.GetCurrentIdentityPool())
```

internal/flink/command.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/featureflags"
 )
 
+const defaultDatabasePlaceholder = "<current-kafka-cluster>"
```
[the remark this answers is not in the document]
brianstrauch: This allows for better documentation. When you type `confluent flink statement create -h` you see that "<current-kafka-cluster>" is the default value instead of "".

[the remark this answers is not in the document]
kevin-wu24: Oh ok gotcha.

internal/flink/command_statement_create.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return err
+	}
+
+	database, err := cmd.Flags().GetString("database")
+	if err != nil {
+		return err
+	}
+	if database == defaultDatabasePlaceholder {
+		database = c.Context.KafkaClusterContext.GetActiveKafkaClusterId()
```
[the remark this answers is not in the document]
brianstrauch: > What happens when I have 2 clusters with the same display name?

I asked the same question to the Flink team, and it looks like they may be changing this in the backend. Stay tuned...

internal/flink/command_statement_list.go
```diff
@@ -35,12 +36,12 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	computePoolId, err := cmd.Flags().GetString("compute-pool")
-	if err != nil {
-		return err
+	computePool := c.Context.GetCurrentFlinkComputePool()
+	if computePool == "" {
+		return errors.NewErrorWithSuggestions("no compute pool selected", "Select a compute pool with `confluent flink compute-pool use` or `--compute-pool`.")
```
sgagniere, line 41: These strings seem to be used multiple times, maybe we can put them in errors package.

sgagniere (approved):

internal/flink/command_statement_list.go
brianstrauch: No longer being used!

test/fixtures/output/local/services/help-onprem.golden
```diff
@@ -5,6 +5,7 @@ Usage:
 
 Available Commands:
   connect         Manage Connect.
+  control-center  Manage Control Center.
```
brianstrauch, line 8:
```suggestion
```

test/fixtures/output/local/services/help.golden
```diff
@@ -5,6 +5,7 @@ Usage:
 
 Available Commands:
   connect         Manage Connect.
+  control-center  Manage Control Center.
```
brianstrauch, line 8:
```suggestion
```

PR #2101, opened 2023-07-14
suggest `confluent kafka cluster` when run `confluent cluster` while logged in to cloud

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add a suggestion to use `confluent kafka cluster` when users run the on-premises `confluent cluster` command while logged in to Confluent Cloud

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

```
mhe@C02DV0KVMD6T cli % confluent cluster list --context login-mhe@confluent.io-https://confluent.cloud
Error: `confluent cluster list` is not a Confluent Cloud command. Did you mean `confluent kafka cluster list`?

Suggestions:
    If you are a Confluent Cloud user, run `confluent kafka cluster list` instead.
    If you are attempting to connect to Confluent Platform, login with `confluent login --url <mds-url>` to use `confluent cluster list`.
```
```
mhe@C02DV0KVMD6T cli % confluent audit-log config

Suggestions:
    Log in to Confluent Platform with `confluent login --url <mds-url>`.
    Use the "--help" flag to see available commands.
```

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -448,6 +448,9 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 
 		// Even if there was an error while setting the context, notify the user about any unmet run requirements first.
 		if err := ErrIfMissingRunRequirement(cmd, r.Config); err != nil {
+			if err == v1.RunningOnPremCommandInCloudErr && strings.Contains(cmd.CommandPath(), "confluent cluster") {
+				return v1.RunningClusterCommandInCloudErr
+			}
 			return err
 		}
 
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -34,9 +34,12 @@ type wrongLoginCommandError struct {
 	suggestionString string
 }
 
-var wrongLoginCommandErrors = map[string]wrongLoginCommandError{
-	"confluent cluster": {"`%s` is not a Confluent Cloud command. Did you mean `confluent kafka cluster %s`?",
-		"Log in to Confluent Platform with `confluent login --url <mds-url>` to use `%s`."},
+var wrongLoginCommandErrorWithSuggestion = wrongLoginCommandError{
+	"`%s` is not a Confluent Cloud command. Did you mean `%s`?",
+	"Log in to Confluent Platform with `confluent login --url <mds-url>` to use `%s`."}
+
+var wrongLoginCommandsMap = map[string]string{
+	"confluent cluster": "confluent kafka cluster",
 }
 
 // PreRun is a helper class for automatically setting up Cobra PersistentPreRun commands
@@ -459,11 +462,11 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 		// Even if there was an error while setting the context, notify the user about any unmet run requirements first.
 		if err := ErrIfMissingRunRequirement(cmd, r.Config); err != nil {
 			if err == v1.RunningOnPremCommandInCloudErr {
-				for topLevelCmd, errorStruct := range wrongLoginCommandErrors {
+				for topLevelCmd, suggestCmd := range wrongLoginCommandsMap {
 					if strings.HasPrefix(cmd.CommandPath(), topLevelCmd) {
-						trimmed := strings.TrimPrefix(cmd.CommandPath(), topLevelCmd+" ")
-						return errors.NewErrorWithSuggestions(fmt.Sprintf(errorStruct.errorString, cmd.CommandPath(), trimmed),
-							fmt.Sprintf(errorStruct.suggestionString, cmd.CommandPath()))
+						suggestCmdPath := strings.Replace(cmd.CommandPath(), topLevelCmd, suggestCmd, 1)
+						return errors.NewErrorWithSuggestions(fmt.Sprintf(wrongLoginCommandErrorWithSuggestion.errorString, cmd.CommandPath(), suggestCmdPath),
+							fmt.Sprintf(wrongLoginCommandErrorWithSuggestion.suggestionString, cmd.CommandPath()))
 					}
 				}
 			}
--- a/internal/pkg/cmd/prerunner.go
+++ b/internal/pkg/cmd/prerunner.go
@@ -36,7 +36,8 @@ type wrongLoginCommandError struct {
 
 var wrongLoginCommandErrorWithSuggestion = wrongLoginCommandError{
 	"`%s` is not a Confluent Cloud command. Did you mean `%s`?",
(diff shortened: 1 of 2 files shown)


sgagniere (approved):

internal/pkg/cmd/prerunner.go
```diff
@@ -448,6 +448,9 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 
 		// Even if there was an error while setting the context, notify the user about any unmet run requirements first.
 		if err := ErrIfMissingRunRequirement(cmd, r.Config); err != nil {
+			if err == v1.RunningOnPremCommandInCloudErr && strings.Contains(cmd.CommandPath(), "confluent cluster") {
```
brianstrauch, line 451: Let's make a `map` here for future "did you mean..." suggestions?

internal/pkg/cmd/prerunner.go
```diff
@@ -448,6 +448,9 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
...
+				return v1.RunningClusterCommandInCloudErr
```
brianstrauch, line 452: Why is this variable global/public? It can probably just be returned here without being defined.

internal/pkg/cmd/prerunner.go
brianstrauch, line 451:
```suggestion
			if err == v1.RunningOnPremCommandInCloudErr && strings.HasPrefix(cmd.CommandPath(), "confluent cluster") {
```

DABH (changes requested):
Can we suggest the equivalent cloud command?

internal/pkg/config/v1/config.go
```diff
@@ -68,6 +68,10 @@ var (
 		"this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
 		"Log in to Confluent Platform with `confluent login --url <mds-url>`.\n"+`Use the "--help" flag to see available commands.`,
 	)
+	RunningClusterCommandInCloudErr = errors.NewErrorWithSuggestions(
+		"`confluent cluster` is not a Confluent Cloud command. Did you mean `confluent kafka cluster`?",
+		"Log in to Confluent Platform with `confluent login --url <mds-url>` to use `confluent cluster`.",
```
DABH, line 73:
```suggestion
		"Log in to Confluent Platform with `confluent login --url <mds-url>` to use `confluent cluster`.",
		"If you are a Confluent Cloud user, try `confluent kafka cluster` instead.",
```

internal/pkg/cmd/prerunner.go
MuweiHe: similar errors are all defined in config.go so i want to put them together

brianstrauch: The errors in that file are related to being logged in to the wrong context, but this specific suggestion is about a potential typo that the user may have made while typing. Those should be separate, IMO.

Is there a way we can take the existing error and append a suggestion to it so we can avoid duplication? I noticed you copied and pasted this message "Log in to Confluent Platform with `confluent login --url <mds-url>` to use `confluent cluster`." which makes it more work to maintain in the future.

internal/pkg/config/v1/config.go
DABH: If we could append the command they tried, that would be even cooler, i.e. if we could dynamically suggest e.g. confluent kafka cluster list so users can literally just copy paste

sgagniere: "Did you mean `confluent kafka cluster`?" in the first part of this suggestion already mentions `confluent kafka cluster`.

brianstrauch: > so users can literally just copy paste

... and to make it even better, we should include the subcommands, arguments, and flags they passed:

```
$ confluent cluster list --all
Did you mean "confluent kafka cluster list --all"?
```

DABH: Given the feedback we got, I think the move here is to hit users over the head -- repeating a suggestion in both the error message and the suggestion is not a bad thing, if that potentially makes it clearer for some users

sgagniere: Makes sense. In this case, only the suggestion would have the dynamic command, right?

I had a small concern that reproducing the whole set of arguments and flags would make it harder to spot the difference (as in, users might gloss right over `kafka` if the suggestion string is long). But having just `confluent kafka cluster` in the error text itself would draw attention to it.

MuweiHe: i updated it to print something as:
```
mhe@C02DV0KVMD6T cli % confluent cluster list --context login-mhe@confluent.io-https://confluent.cloud
Error: `confluent cluster list` is not a Confluent Cloud command. Did you mean `confluent kafka cluster list`?

Suggestions:
    Log in to Confluent Platform with `confluent login --url <mds-url>` to use `confluent cluster list`.
```
i can repeat it in the suggestion if you prefer.
it'll be more complex to include arguments and flags cause `cmd.CommandPath()` don't include them and i'll need to construct a new struct to include all them

internal/pkg/cmd/prerunner.go
```diff
@@ -29,6 +29,16 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/version"
 )
 
+type wrongLoginCommandError struct {
+	errorString      string
+	suggestionString string
+}
+
+var wrongLoginCommandErrors = map[string]wrongLoginCommandError{
```
brianstrauch, line 37: I was thinking this mapping could just be from `confluent cluster` to `confluent kafka cluster`, not the whole suggestion. Odds are we'll want a similar suggestion for each mapping, and we don't want to have to write out the whole suggestion every time.

internal/pkg/cmd/prerunner.go
```diff
@@ -448,6 +458,15 @@ func (r *PreRun) AuthenticatedWithMDS(command *AuthenticatedCLICommand) func(*co
 
 		// Even if there was an error while setting the context, notify the user about any unmet run requirements first.
 		if err := ErrIfMissingRunRequirement(cmd, r.Config); err != nil {
+			if err == v1.RunningOnPremCommandInCloudErr {
+				for topLevelCmd, errorStruct := range wrongLoginCommandErrors {
+					if strings.HasPrefix(cmd.CommandPath(), topLevelCmd) {
+						trimmed := strings.TrimPrefix(cmd.CommandPath(), topLevelCmd+" ")
```
brianstrauch, line 464: Let's just use `strings.Replace()` here.

internal/pkg/cmd/prerunner.go
MuweiHe: that sounds better

DABH (approved):

PR #2103, opened 2023-07-14
Fix autocompletion for `kafka cluster create`, formatting improvements

brianstrauch:
Release Notes
-------------
Bug Fixes
- Fix autocompletion of `--availability` and `--type` flags in `confluent kafka cluster create`

Checklist
---------
   * yes: ok

What
----
Fix autocompletion, consolidate Kafka REST helper function, improve overall formatting


sgagniere (approved):
+1

We should probably increase autocomplete integration test coverage at some point to catch weird things like this.

brianstrauch:
> We should probably increase autocomplete integration test coverage at some point

Forgot about this, added!

PR #2111, opened 2023-07-17
[DO NOT MERGE UNTIL GA] [CIAM-2710] Remove identity admin LD flag

lucy-fan:
Checklist
---------
   * yes: ok

What
----
In preparation for GA, we need to take out the identity admin LD flag so that roles with the `identity` namespace will show up in the CLI when listing or describing roles. 

*Please do not merge this until we GA the new roles!

Test & Review
-------------
Tested locally to confirm it works.


internal/cmd/iam/command_rbac_role_describe.go
```diff
@@ -51,14 +51,10 @@ func (c *roleCommand) ccloudDescribe(cmd *cobra.Command, role string) error {
 		ksqlNamespace.Value(),
 		publicNamespace.Value(),
 		streamCatalogNamespace.Value(),
+		identityNamespace.Value(),
```
brianstrauch, line 54: Please alphabetize

brianstrauch (approved):

PR #2112, opened 2023-07-18
fix: make basic output the default regardless of number of columns

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This makes it so that we only use the interactive output when we have a next page token or the statement is a select statement. We do not factor in the number of columns in this decision anymore. 

The main reason for this is the DESCRIBE statement. With the results of the DESCRIBE statement being displayed in the interactive table, the user needs to run the DESCRIBE, then memorize the field names and schema, and then write their query in one go. It is much easier for the user, if we just print the fields and the schema, so that the subsequent queries can be written without going back and forth between the terminal and the interactive table.

### DIFF
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -13,8 +13,6 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/flink/types"
 )
 
-const minNumColumnsToUseInteractiveTable = 4
-
 type Application struct {
 	history                     *history.History
 	store                       types.StoreInterface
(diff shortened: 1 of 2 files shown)


YannickPferr:
> I think we both agree this is not the last/final solution, approving with two things I think we should think about:
> 
> 1 - What's going to happen when statements like show jobs are paginated (too many results)? 2 - We should have an adr where we document these changes and link the tickets or else in 2 months or a year when we have the same discussion, it's going to be hard to refer back to why the decisions were made.

Yeah that makes sense, I'm going to set that up tomorrow to document our decisions!

PR #2113, opened 2023-07-19
change CatchCCloudV2Error to always return an errorWithSuggestion including the status code

luca-filipponi:
…luding the status code

Checklist
---------
   * yes: ok

What
----
Extended the errorWithSuggestion type to include the status code. Updated CatchCCloudV2Error to always return an errortWithSuggestion so there will always be a status code available from SDK errors. 

This is needed to implement custom logic for different status codes (e.g. retry)

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/errors/catcher.go
+++ b/internal/pkg/errors/catcher.go
@@ -139,9 +139,8 @@ func catchCCloudBackendUnmarshallingError(err error) error {
 }
 
 /*
-	CCLOUD-SDK-GO CLIENT ERROR CATCHING
+CCLOUD-SDK-GO CLIENT ERROR CATCHING
 */
-
 func CatchCCloudV2Error(err error, r *http.Response) error {
 	if err == nil {
 		return nil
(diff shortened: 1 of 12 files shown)


internal/pkg/errors/catcher.go
```diff
@@ -154,31 +151,33 @@ func CatchCCloudV2Error(err error, r *http.Response) error {
 	body, _ := io.ReadAll(r.Body)
 	var resBody errorResponseBody
 	_ = json.Unmarshal(body, &resBody)
+	statusCode := r.StatusCode
+
 	if len(resBody.Errors) > 0 {
 		detail := resBody.Errors[0].Detail
 		if ok, _ := regexp.MatchString(quotaExceededRegex, detail); ok {
-			return NewWrapErrorWithSuggestions(err, detail, QuotaExceededSuggestions)
+			return NewWrapErrorWithSuggestionsAndCode(err, detail, QuotaExceededSuggestions, statusCode)
```
brianstrauch, line 159: All of our API calls go through this function, and while it may be desired to print the status code for Flink, we generally consider it to be something we should abstract away from the user. If a user needs to see the status code, they can run `--unsafe-trace` to see the full HTTP response.

Futhermore, there are a few Confluent Cloud APIs that currently return improperly formatted errors... for example, certain Connect APIs will actually return the error code as the error message body, and this PR would cause the error code to be printed twice in those cases, which looks like a bug on our end.

brianstrauch:
> This is needed to implement custom logic for different status codes (e.g. retry)

Can you expand on this? We already have retry logic set up for all of our Confluent Cloud API requests.

internal/pkg/errors/catcher.go
luca-filipponi: We don't need user to see status code, the reason why I've done this is that we need the status code back from the gateway client: https://github.com/confluentinc/cli/blob/ad396fc0ddd0cb67decdfb4cd2b18ae5213952ef/internal/pkg/ccloudv2/flink_gateway.go#L13

Knowing the status code is needed for implementing custom logic in certain scenarios (@YannickPferr can expand on this)

Will this change bubble up status code to customers? I didn't see this happening in tests? I can see that only the message is used: https://github.com/confluentinc/cli/blob/ffd25f1ad80c664fc2289affaf9154e073754631/internal/pkg/errors/error_with_suggestions.go#L67

In which way would you like this to be done? Happy to add a third param on the gateway client or to have a new error type for the gateway and a different error catcher for our use case.

luca-filipponi:
> > This is needed to implement custom logic for different status codes (e.g. retry)
> 
> Can you expand on this? We already have retry logic set up for all of our Confluent Cloud API requests.

Is not just about retry, but really to have the status code back from the gateway client, like this: https://github.com/confluentinc/cli/blob/fa3e088b89e2feabb2dda2e574a29eae991a4906/internal/pkg/flink/internal/store/store.go#L95. We might want to retry on status codes < 500 (like 401), because there could be situations where you check the validity of the token and it's still valid, but the request takes a couple seconds so it might expire during that time

internal/pkg/errors/catcher.go
brianstrauch: Apologies for not reading the PR more thoroughly. The `errorWithSuggestion` type is user-facing, so I assumed the status code would also be shown to the user. I still think we should make this Flink-specific to avoid cluttering the error package with structs that aren't used globally. Can we make a `FlinkErrorWithSuggestion` struct and keep it inside the Flink package?

luca-filipponi: Yep, that sounds good to me! Tomorrow will update the pr

luca-filipponi:
Closed in favour of: https://github.com/confluentinc/cli/pull/2128

PR #2115, opened 2023-07-19
Autocomplete `kafka client-config`, remove `HasAPIKeyCLICommand` struct

brianstrauch:
Release Notes
-------------
New Features
- Autocomplete flags in `confluent kafka client-config create` commands

Checklist
---------
   * yes: ok

What
----
The `HasAPIKeyCLICommand` struct wasn't be used as it was originally intended, and is not actually needed. Replacing it removes a bit of tech debt and allows us to add autocompletion to all of the flags in `kafka client-config create`!


internal/pkg/config/v1/kafka_cluster_config.go
```diff
@@ -41,3 +41,12 @@ func (k *KafkaClusterConfig) EncryptAPIKeys() error {
 	}
 	return nil
 }
+
+func (k *KafkaClusterConfig) GetApiSecret() string {
+	if k.APIKeys != nil {
```
sgagniere, line 46: Just to be extra safe, maybe we can also check `k != nil` first.

internal/cmd/kafka/command_clientconfig_create.go
```diff
@@ -186,38 +170,44 @@ func (c *createCommand) create(configId string, srApiAvailable bool) func(cmd *c
...
+
 	// Only validate that the key pair matches with the cluster if it's passed via the flag.
 	// This is because currently "api-key store" does not check if the secret is valid. Therefore, if users
 	// choose to use the key pair stored in the context, we should use it without doing a validation.
 	flagKey, flagSecret, err := c.Config.Context().KeyAndSecretFlags(cmd)
 	if err != nil {
 		return "", err
 	}
-	if flagKey != "" && flagSecret != "" {
+	if flagKey != "" || flagSecret != "" {
```
sgagniere, line 191: `KeyAndSecretFlags` returns an error if the key is empty and the secret isn't, so I think we can just replace this with `if flagKey != "" {` since `false, true` seems unreachable.

sgagniere (approved):
+1 w/ small nitpicks/comments.

internal/cmd/kafka/command_clientconfig_create.go
brianstrauch:
```suggestion
	if flagKey != "" {
```

internal/pkg/config/v1/kafka_cluster_config.go
brianstrauch:
```suggestion
	if k != nil && k.APIKeys != nil {
```

PR #2116, opened 2023-07-19
Schema Registry refactors

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Setting up the Schema Registry code to be easier to work with.

### DIFF
--- a/internal/cmd/schema-registry/command_schema_create.go
+++ b/internal/cmd/schema-registry/command_schema_create.go
@@ -126,7 +126,7 @@ func (c *command) schemaCreate(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
-	srClient, ctx, err := getApiClient(cmd, c.srClient, c.Config, c.Version)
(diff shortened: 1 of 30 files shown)


internal/cmd/schema-registry/command_schema_create.go
```diff
@@ -107,19 +91,26 @@ func (c *command) schemaCreate(cmd *cobra.Command, _ []string) error {
...
-	request := srsdk.RegisterSchemaRequest{
-		Schema:     string(schema),
+	normalize, err := cmd.Flags().GetBool("normalize")
+	if err != nil {
+		return err
+	}
+
+	cfg := &RegisterSchemaConfigs{
+		Subject:    subject,
+		SchemaPath: &schema,
```
sgagniere, line 101: Can this be changed to just a string, instead of a pointer to string? As far as I can tell, `SchemaPath` only looks like it's used in one place as an argument to `os.ReadFile`, which takes a string anyway.

sgagniere (approved):

PR #2122, opened 2023-07-21
feat/KFS-1058-exit attached mode upon pressing ControlC, ControlD, ControlQ..

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

When submitting persistent queries like INSERT INTO that run in the background we attach to the job and listen for errors and the user can exit this mode now pressing also ctrlC, ctrlD and so on, which are the regular cancelEvents for the other statements. 
 
References
----------

Test & Review
-------------

unit tests
Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/flink/internal/controller/statement_controller.go
+++ b/internal/pkg/flink/internal/controller/statement_controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	fColor "github.com/fatih/color"
+	"github.com/samber/lo"
 
 	"github.com/confluentinc/go-prompt"
 
@@ -129,17 +130,9 @@ func (c *StatementController) userInputIsOneOf(keyEvents ...func(key prompt.Key)
 }
 
 func isCancelEvent(key prompt.Key) bool {
-	switch key {
-	case prompt.ControlC, prompt.ControlD, prompt.ControlQ, prompt.Escape:
-		return true
-	}
-	return false
+	return lo.Contains([]prompt.Key{prompt.ControlC, prompt.ControlD, prompt.ControlQ, prompt.Escape}, key)
 }
 
 func isDetachEvent(key prompt.Key) bool {
-	switch key {
-	case prompt.ControlM, prompt.Enter:
-		return true
-	}
-	return false
+	return lo.Contains([]prompt.Key{prompt.ControlM, prompt.Enter}, key)
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/flink/internal/controller/statement_controller.go
```diff
@@ -125,14 +113,33 @@ func (c *StatementController) waitForStatementToBeInTerminalStateOrError(process
...
+		return false
+	}
+}
+
+func isCancelEvent(key prompt.Key) bool {
+	switch key {
+	case prompt.ControlC, prompt.ControlD, prompt.ControlQ, prompt.Escape:
+		return true
+	}
+	return false
```
brianstrauch, line 136:
```suggestion
	return types.Contains([]string{prompt.ControlC, prompt.ControlD, prompt.ControlQ, prompt.Escape}, key)
```

internal/pkg/flink/internal/controller/statement_controller_test.go
```diff
@@ -315,3 +315,75 @@ func (s *StatementControllerTestSuite) TestRenderMsgAndStatusNonLocalNonFailedSt
...
+			want: true,
+		},
+		{
+			name: "Other",
+			key:  prompt.ShiftDown, // Just an example of a key that is not in the switch cases
+			want: false,
+		},
+	}
+
+	for _, tt := range tests {
```
brianstrauch, line 352:
```suggestion
	for _, test := range tests {
```

internal/pkg/flink/internal/controller/statement_controller.go
YannickPferr: Neat one-liner! You could also do it with lo like this:
`lo.Contains([]prompt.Key{prompt.ControlC, prompt.ControlD, prompt.ControlQ, prompt.Escape}, pressedKey)`

internal/pkg/flink/internal/controller/statement_controller.go
```diff
@@ -125,14 +113,33 @@ func (c *StatementController) waitForStatementToBeInTerminalStateOrError(process
...
+}
+
+func isDetachEvent(key prompt.Key) bool {
+	switch key {
```
YannickPferr, line 140: and then you could also reduce this to one line:
`lo.Contains([]prompt.Key{prompt.ControlM, prompt.Enter}, key)`

YannickPferr (approved):
Thanks!

PR #2124, opened 2023-07-21
Use dataplane token instead of API key for Schema Registry auth

brianstrauch:
Release Notes
-------------
New Features
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
All Schema Registry commands require that a user creates an API key scoped to their Schema Registry cluster. We don't actually need to do this- we can generate a dataplane token instead! This is the style of Kafka REST / Flink REST and others, and results in a huge UX improvement.

Will apply this auth style to all other Schema Registry commands in a follow-up PR!

Test & Review
-------------
Updated integration tests and hardcoded a port for the mock Schema Registry backend.


sgagniere (approved):
Should the api key and secret flags for `cluster describe` be deprecated?

brianstrauch:
> Should the api key and secret flags for cluster describe be deprecated?

Yep, just realized this while testing!

sgagniere (approved):

PR #2126, opened 2023-07-22
[CLI-2564] Remove misleading feature flag warning when running test/lint/docs

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
`"WARNING: Failed to fetch feature flags."` shows up multiple times when running `lint` or `docs`, and occasionally causes flakiness in the integration tests.

This PR just skips reporting the warning and suggestion if `cfg.IsTest` is true.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/pkg/featureflags/feature_flags.go
+++ b/internal/pkg/featureflags/feature_flags.go
@@ -47,13 +47,14 @@ var Manager launchDarklyManager
 var attributes = []string{"user.resource_id", "org.resource_id", "environment.id", "cli.version", "cluster.id", "cluster.physicalClusterId", "cli.command", "cli.flags"}
 
 type launchDarklyManager struct {
-	cliClient                *sling.Sling
-	ccloudClient             func(v1.LaunchDarklyClient) *sling.Sling
-	Command                  *cobra.Command
-	flags                    []string
-	isDisabled               bool
-	timeoutSuggestionPrinted bool
-	version                  *version.Version
+	cliClient             *sling.Sling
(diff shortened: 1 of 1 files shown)


internal/pkg/featureflags/feature_flags.go
```diff
@@ -183,10 +185,10 @@ func (ld *launchDarklyManager) fetchFlags(user lduser.User, client v1.LaunchDark
 	}
 	if err != nil {
 		log.CliLogger.Debug(resp)
-		if !ld.timeoutSuggestionPrinted {
+		if !ld.hideTimeoutWarning && !ld.timeoutWarningPrinted {
```
brianstrauch, line 188: It makes sense why this warning was showing up during `make lint` and `make docs`, but do you know why it was showing up during `make test`? We should always mock the LaunchDarkly backend while testing...

sgagniere: We use two LD clients, `cliClient` (which is mocked by the backend) and `ccloudClient` which isn't. `confluent iam rbac [describe | list]` attempt to pull the real flags using `ccloudClient`, which is why they can fail and generate this warning (and screw up the comparison to the golden files).

brianstrauch: Let's merge these client in a separate PR, and make sure both are mocked?

brianstrauch (approved):

PR #2127, opened 2023-07-24
feat: cleanup statement on ungraceful table view shutdown

YannickPferr:
Checklist
---------
   * yes: ok

What
----

This change ensures we cleanup statements in case of panics inside the table view, or when the table view is stopped with CtrlC. When a panic occurs the user will see the message "Error: internal error occurred while in table view" and can proceed to submit statement instead of being kicked out of the client. We also still send the usage report to `cc-cli-usage-service`.

### DIFF
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -36,7 +36,7 @@ func synchronizedTokenRefresh(tokenRefreshFunc func() error) func() error {
 	}
 }
 
-func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() error, appOptions types.ApplicationOptions) {
+func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() error, appOptions types.ApplicationOptions, reportUsageFunc func()) {
 	// Load history of previous commands from cache file
 	historyStore := history.LoadHistory()
 
@@ -58,7 +58,7 @@ func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() er
 	// Instantiate Component Controllers
 	inputController := controller.NewInputController(historyStore)
 	statementController := controller.NewStatementController(appController, dataStore, consoleParser)
-	interactiveOutputController := controller.NewInteractiveOutputController(components.NewTableView(), resultFetcher, appOptions.GetVerbose())
+	interactiveOutputController := controller.NewInteractiveOutputController(components.NewTableView(), resultFetcher, reportUsageFunc, appOptions.GetVerbose())
 	basicOutputController := controller.NewBasicOutputController(resultFetcher, inputController.GetWindowWidth)
 
 	app := Application{
--- a/internal/pkg/flink/internal/controller/interactive_output_controller.go
+++ b/internal/pkg/flink/internal/controller/interactive_output_controller.go
@@ -15,34 +15,54 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/output"
 )
 
+const errorDuringTableView = "Error: internal error occurred while in table view"
+
 type InteractiveOutputController struct {
 	app           *tview.Application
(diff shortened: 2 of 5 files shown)


internal/pkg/flink/internal/controller/interactive_output_controller.go
```diff
@@ -15,36 +15,56 @@ import (
...
 func (t *InteractiveOutputController) VisualizeResults() {
+	defer t.panicRecovery()
 	t.init()
-	t.start()
+	t.startTView()
+	t.close()
+}
+
+func (t *InteractiveOutputController) panicRecovery() {
+	if r := recover(); r != nil {
```
brianstrauch, line 47: Surprised that this isn't caught by the main `recover()` function

YannickPferr: Oh no that does work! It will get caught by the man recover(), it's just that we cannot delete the statement we created there and this can leave dangling resources (because currently there is no cancel API for statements, so we have to delete statements that would otherwise run forever).

So it just gives us a bit more control over how we recover in case of a panic.

internal/pkg/flink/app/application.go
```diff
@@ -58,7 +58,7 @@ func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() er
 	// Instantiate Component Controllers
 	inputController := controller.NewInputController(historyStore)
 	statementController := controller.NewStatementController(appController, dataStore, consoleParser)
-	interactiveOutputController := controller.NewInteractiveOutputController(components.NewTableView(), resultFetcher, appOptions.GetVerbose())
+	interactiveOutputController := controller.NewInteractiveOutputController(components.NewTableView(), resultFetcher, reportUsageFunc, appOptions.GetVerbose())
 	basicOutputController := controller.NewBasicOutputController(resultFetcher, inputController.GetWindowWidth)
 
 	app := Application{
```
Guttz, line 64: Panic recovery for interactive output works great 👍 

a.statementController.ExecuteStatement(userInput), in the input mode, can however still panic leading to statements not being deleted before the results were fetched. I think we should have some panic recovery for that case, so we delete the statement if the shell panics while the statement is being created/waiting to be ready/we're listening for errors.

Guttz (changes requested):

internal/pkg/flink/app/application.go
YannickPferr: Yeah that makes sense! I've refactored it now, so that the panic recovery happens in the main app loop. That way, regardless of where it panics, if we created a statement it will get deleted.

Guttz: nice!

Guttz:
Thanks for refactoring to make sure we can clean up in all stages! Unfortunately, I pulled the code again but couldn't make it work in input mode

image

YannickPferr:
> Thanks for refactoring to make sure we can clean up in all stages! Unfortunately, I pulled the code again but couldn't make it work in input mode (I'm triggering the panic with ctrlc, so maybe because it listens for userinput in a different go routine?, bit weird tho)
> 
> image

Yeah the recover only works in the goroutine which calls the panic and in this case we just cover the main goroutine. 
I think this is fine, because I don't think we should just spam the recover everywhere 😄

Guttz:
@swist suggested we could have a wrapper to spawn go routines so we make sure we cleanup if those fail. I think that's a good idea.

I think you can either o this here or create a new ticket for it

YannickPferr:
> @swist suggested we could have a wrapper to spawn go routines so we make sure we cleanup if those fail. I think that's a good idea.
> 
> I think you can either o this here or create a new ticket for it

This will probably be a bit of work to make it right, so I created a ticket: https://confluentinc.atlassian.net/jira/software/projects/KFS/boards/691?selectedIssue=KFS-1071

Guttz (approved):

PR #2125, opened 2023-07-21
Schema Registry client migration (part 1)

brianstrauch:
Release Notes
-------------
New Features
- PLACEHOLDER

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok

What
----
Continuation of #2124, for:
* `confluent schema-registry cluster update`
* `confluent schema-registry compatibility validate`
* `confluent schema-registry config describe`

Also set up SR client for CP, which allowed for a few implementations to be combined.

Test & Review
-------------
Updated all integration tests!

PR #2134, opened 2023-07-25
Schema Registry client migration (part 2)

brianstrauch:
Release Notes
-------------
New Features
- When authenticated, no longer require an API key for Schema Registry operations

Checklist
---------
   * yes: ok

What
----
A continuation of #2125, for the the rest of the Schema Registry commands, as well as produce/consume and AsyncAPI

Test & Review
-------------
Updated relevant integration tests


internal/cmd/asyncapi/command_export.go
```diff
@@ -218,22 +214,28 @@ func (c *command) getAccountDetails(flags *flags) (*accountDetails, error) {
 	if err := c.getClusterDetails(details, flags); err != nil {
 		return nil, err
 	}
-	if err := c.getSchemaRegistry(details, flags); err != nil {
+
+	client, err := c.GetSchemaRegistryClient()
```
kevin-wu24:
```suggestion
	details.srClient, err := c.GetSchemaRegistryClient()
```

internal/cmd/asyncapi/command_export.go
```diff
@@ -218,22 +214,28 @@ func (c *command) getAccountDetails(flags *flags) (*accountDetails, error) {
...
 		return nil, err
 	}
-	subjects, _, err := details.srClient.DefaultApi.List(details.srContext, nil)
+	details.srClient = client
+
+	subjects, err := details.srClient.List(nil)
 	if err != nil {
 		return nil, err
 	}
 	details.subjects = subjects
```
kevin-wu24, line 228:
```suggestion
	details.subjects, err := details.srClient.List(nil)
	if err != nil {
		return nil, err
	}
```

internal/cmd/asyncapi/command_import.go
```diff
@@ -455,7 +454,7 @@ func addSchemaTags(details *accountDetails, components Components, topicName str
 			tagConfigs = append(tagConfigs, srsdk.Tag{
 				TypeName:   tag.Name,
 				EntityType: "sr_schema",
-				EntityName: fmt.Sprintf("%s:.:%s", details.srCluster.GetId(), strconv.Itoa(int(schemaId))),
+				EntityName: fmt.Sprintf("%s:.:%s", "TODO", strconv.Itoa(int(schemaId))),
```
kevin-wu24: what does `TODO` mean here?

internal/cmd/asyncapi/command_export.go
brianstrauch: This won't work, since you can only use `:=` with variables that don't exist yet

kevin-wu24: can we replace `:=` with `=`, since `err` is already defined above and we know `err == nil` up until this point?

kevin-wu24 (approved):

sgagniere (approved):

internal/cmd/asyncapi/command.go
```diff
@@ -6,15 +6,21 @@ import (
...
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
 func New(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:         "asyncapi",
 		Short:       "Manage AsyncAPI document tooling.",
-		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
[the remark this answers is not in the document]
brianstrauch: This is a bug fix! The command depends on the user's Schema Registry ID, which can only be fetched via ccloud-sdk-go-v2, hence the "NonApiKey". The command was actually relying on the ID being present in the configuration file... if it wasn't, the command would malfunction.

PR #2140, opened 2023-07-27
[CLI-829] Add `confluent kafka consumer-group [describe | list]` and `confluent kafka consumer list` commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `confluent kafka consumer group describe`
- Add `confluent kafka consumer group list`
- Add `confluent kafka consumer list`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
This follows the design of the 1-pager except that I split one command into two.

- Update the `consumer-group list` command
- Split the `consumer-group describe` command into `consumer-group describe` and `consumer list`. The previous iteration of `consumer-group describe` displayed the `consumer list` output in the human readable output of `consumer-group describe` but not in the serialized output.
- The on-prem endpoints are also GA, so I went ahead and added the on-prem versions of these three commands.

References
----------

Test & Review
-------------
Added integration tests for cloud & on-prem. Also added two autocomplete integration tests.

Open Questions / Follow-ups
---------------------------
`confluent kafka consumer-group lag` commands are next.


internal/cmd/kafka/command_consumer_list.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (c *consumerCommand) list(cmd *cobra.Command, _ []string) error {
+	kafkaREST, err := c.GetKafkaREST()
+	if kafkaREST == nil {
+		if err != nil {
+			return err
+		}
+		return errors.New(errors.RestProxyNotAvailableMsg)
+	}
```
brianstrauch, line 45: I've been using this style instead for other commands. Is it really possible that both return values are `nil`, and if so, can we make `GetKafkaREST()` return the not available error in that case?
```suggestion
	if err != nil {
		return err
	}
```

internal/cmd/kafka/command_consumer_list.go
```diff
@@ -0,0 +1,72 @@
...
+
+	consumerGroup, err := cmd.Flags().GetString("consumer-group")
+	if err != nil {
+		return err
+	}
+
+	consumerDataList, httpResp, err := kafkaREST.CloudClient.ListKafkaConsumers(cluster.ID, consumerGroup)
+	if err != nil {
+		return kafkarest.NewError(kafkaREST.CloudClient.GetUrl(), err, httpResp)
```
brianstrauch, line 59: Can we move this line into the `ListKafkaConsumers()` function? (I did this for the `kafka cluster configuration` commands, and seems like a good thing to do to abstract away this logic)

internal/cmd/kafka/command_consumer_list_onprem.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/examples"
+	"github.com/confluentinc/cli/internal/pkg/kafkarest"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+func (c *consumerCommand) newListCommandOnPrem() *cobra.Command {
```
brianstrauch, line 12: Any way we can nicely combine the Cloud + On-Prem implementations in the same file?

internal/cmd/kafka/command_consumer_list_onprem.go
```diff
@@ -0,0 +1,72 @@
...
+	consumerDataList, resp, err := restClient.ConsumerGroupV3Api.ListKafkaConsumers(restContext, clusterId, consumerGroup)
+	if err != nil {
+		return kafkarest.NewError(restClient.GetConfig().BasePath, err, resp)
+	}
+
+	list := output.NewList(cmd)
+	for _, consumer := range consumerDataList.Data {
+		instanceId := ""
+		if consumer.InstanceId != nil {
+			instanceId = *consumer.InstanceId
```
brianstrauch, line 61: No `GetInstanceId()`? :(

test/fixtures/output/kafka/consumer-group/describe.golden
```diff
@@ -0,0 +1,8 @@
++--------------------+------------------------------------------------------+
+| Cluster            | lkc-1234                                             |
+| Consumer Group     | consumer-group-1                                     |
+| Coordinator        | broker-1                                             |
+| Simple             | true                                                 |
+| Partition Assignor | org.apache.kafka.clients.consumer.RoundRobinAssignor |
```
brianstrauch, line 6: Assigner?

test/kafka_test.go
```diff
@@ -483,8 +483,53 @@ func (s *CLITestSuite) TestKafkaQuota() {
...
+		login:   "cloud",
+		args:    "kafka consumer list --consumer-group consumer-group-1 --cluster lkc-1234",
+		fixture: "kafka/consumer/list.golden",
+	}
+
+	s.runIntegrationTest(test)
+
+	test = CLITest{
+		login:   "onprem",
+		args:    fmt.Sprintf("kafka consumer list --consumer-group consumer-group-1 --url %s", s.TestBackend.GetKafkaRestUrl()),
```
brianstrauch, line 497: Do we need to pass `--url` if the login is "onprem"?

internal/cmd/kafka/command_consumer_list_onprem.go
brianstrauch: I wonder how easy it would be to add that, considering that it's also based on an OpenAPI spec, just like ccloud-sdk-go-v2 is.

test/fixtures/output/kafka/consumer-group/describe.golden
sgagniere: They're both words (with similar but different meanings), and assignor is the one that's used in the API.

internal/cmd/kafka/command_consumer_list_onprem.go
sgagniere: With the way that we currently initialize kafka clients for cloud and on-prem, I don't think so. Plus, `kafka-rest-sdk-go` doesn't have `Get` methods. So the kafka client initialization and the output code will both be in `if else` blocks.

internal/cmd/kafka/command_consumer_list.go
sgagniere: Looks like it can happen if a cluster's rest endpoint is missing. Is that possible?

Either way, I followed your suggestion and moved the error into the function. Link, mirror, and a few other Kafka commands still use this style and I'll update them in a followup PR after this one.

internal/cmd/kafka/command_consumer_list_onprem.go
sgagniere: I asked Kafka eng and tagged you in the post.

internal/cmd/kafka/command_consumer_list.go
brianstrauch: Awesome, thank you! (Shouldn't be possible, but you never know...)

test/kafka_test.go
sgagniere: Yes and no. This url is the rest proxy endpoint and not the mds url, so we still need to provide it after logging in unless we set the `CONFLUENT_REST_URL` environment variable.

A did a quick survey though and it seems that most of our Kafka onprem tests set the env variable instead of using the `--url` flag, so I've switched it over just for consistency.

brianstrauch: I think there's a ticket for it, but I wonder if in the future, we can infer the Kafka REST URL from `confluent kafka cluster use` or `--cluster` on the on-prem side.

internal/cmd/kafka/command_consumer.go
```diff
@@ -0,0 +1,54 @@
...
+
+type consumerCommand struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+type consumerOut struct {
+	ConsumerGroupId string `human:"Consumer Group" serialized:"consumer_group"`
+	ConsumerId      string `human:"Consumer" serialized:"consumer"`
+	InstanceId      string `human:"Instance" serialized:"instance"`
+	ClientId        string `human:"Client" serialized:"client"`
```
brianstrauch, line 18: I think across the board, we suffix the serialized field name with "_id" when the struct field name ends with "Id"

internal/cmd/kafka/command_consumergroup_describe.go
```diff
@@ -22,11 +21,10 @@ func (c *consumerGroupCommand) newDescribeCommand() *cobra.Command {
 		RunE:              c.describe,
 		Example: examples.BuildExampleString(
 			examples.Example{
-				Text: `Describe the "my-consumer-group" consumer group:`,
+				Text: `Describe the "my-consumer-group" consumer group.`,
 				Code: "confluent kafka consumer-group describe my-consumer-group",
 			},
```
brianstrauch, line 26: My vote is not to include an example, since the usage is straightforward.

internal/cmd/kafka/command_consumergroup_list.go
```diff
@@ -33,18 +31,23 @@ func (c *consumerGroupCommand) newListCommand() *cobra.Command {
...
 	}
 
-	groupCmdResp, httpResp, err := kafkaREST.CloudClient.ListKafkaConsumerGroups(lkc)
+	cluster, err := c.Context.GetKafkaClusterForCommand()
 	if err != nil {
-		return kafkarest.NewError(kafkaREST.CloudClient.GetUrl(), err, httpResp)
+		return err
+	}
+
+	consumerGroupDataList, err := kafkaREST.CloudClient.ListKafkaConsumerGroups(cluster.ID)
```
brianstrauch, line 44:
```suggestion
	consumerGroups, err := kafkaREST.CloudClient.ListKafkaConsumerGroups(cluster.ID)
```

brianstrauch:
Waiting on the decision between `confluent kafka consumer group` and `confluent kafka consumer-group`... The first looks a bit cleaner, but the second looks more discoverable, so I could go either way.

internal/cmd/kafka/command_consumergroup_describe_onprem.go
```diff
@@ -0,0 +1,67 @@
...
+	table := output.NewTable(cmd)
+	table.Add(&consumerGroupOut{
+		ClusterId:         consumerGroup.ClusterId,
+		ConsumerGroupId:   consumerGroup.ConsumerGroupId,
+		Coordinator:       getStringBrokerOnPrem(consumerGroup.Coordinator),
+		IsSimple:          consumerGroup.IsSimple,
+		PartitionAssignor: consumerGroup.PartitionAssignor,
+		State:             consumerGroup.State,
+	})
+	return table.Print()
```
kevin-wu24, line 55: Since this code is the same as the cloud version, can we put this logic in a shared function?

internal/cmd/kafka/command_consumergroup_describe_onprem.go
```diff
@@ -0,0 +1,67 @@
...
+
+func getStringBrokerOnPrem(relationship kafkarestv3.Relationship) string {
+	// relationship.Related will look like ".../v3/clusters/{cluster_id}/brokers/{broker_id}
+	splitString := strings.SplitAfter(relationship.Related, "brokers/")
+	// if relationship was an empty string or did not contain "brokers/"
+	if len(splitString) < 2 {
+		return ""
+	}
+	// returning brokerId
+	return splitString[1]
```
kevin-wu24, line 66: Similar to above. Since this is the same as `getStringBroker`, let's delete this and move `getStringBroker` to `internal/cmd/kafka/command_consumergroup.go`?

internal/cmd/kafka/command_consumergroup_list_onprem.go
```diff
@@ -0,0 +1,54 @@
...
+	list := output.NewList(cmd)
+	for _, group := range consumerGroup.Data {
+		list.Add(&consumerGroupOut{
+			ClusterId:         group.ClusterId,
+			ConsumerGroupId:   group.ConsumerGroupId,
+			Coordinator:       getStringBrokerOnPrem(group.Coordinator),
+			IsSimple:          group.IsSimple,
+			PartitionAssignor: group.PartitionAssignor,
+			State:             group.State,
+		})
```
kevin-wu24, line 51: Not sure if this is considered idiomatic Go or not, but since a lot of CLI code in general has often has these large `[resource]Out` struct literals in the implementation of `describe, update, list` commands, wouldn't it be more readable to have like "constructors" for these `[resource]Out` structs we have?

I guess we do an extra pointer copy from that approach? What do u think?

internal/cmd/kafka/command_consumergroup_describe_onprem.go
sgagniere: We'll be able to do that after we update the on-premises `kafka-rest-sdk-go` to have `Get()` functions. But currently, it doesn't support them :(

sgagniere:
@ahuang98 We'd like your opinion on some changes I made from the design of your 1-pager.

We generally don't want multiple tables/lists being output for one command, so I split the consumer list data out of `confluent kafka consumer-group describe` and into its own `confluent kafka consumer list` command.

@brianstrauch then raised the point that perhaps we should have a `consumer` command with `group` being a subcommand of that, instead of `consumer-group`, which is cleaner.

In other words, we'd have:
`confluent kafka consumer list`
`confluent kafka consumer group describe`
`confluent kafka consumer group list`
and (in a followup PR)
`confluent kafka consumer group lag` commands.

What do you think?

ahuang98:
Thanks for picking this up! 
> We generally don't want multiple tables/lists being output for one command, so I split the consumer list data out of `confluent kafka consumer-group describe` and into its own `confluent kafka consumer list` command.

I originally designed it this way to more closely match the output of kafka-consumer-groups --describe (which is what customers would be familiar with). I do agree it's cleaner to just have one schema returned per command and I'm not strongly opposed towards the suggestions made. 

Would `confluent kafka consumer list` also have a `--groupId` argument to specify which group to list consumers from and perhaps `--all-groups` to specify listing all consumers? I'm a bit worried that having `confluent kafka consumer list` default to return all consumers would be spammy

internal/cmd/kafka/command_consumer.go
```diff
@@ -0,0 +1,54 @@
...
+		c.PersistentPreRunE = prerunner.InitializeOnPremKafkaRest(c.AuthenticatedCLICommand)
+
+		cmd.AddCommand(c.newListCommandOnPrem())
+	}
+
+	return cmd
+}
+
+func (c *consumerCommand) addConsumerGroupFlag(cmd *cobra.Command) {
+	cmd.Flags().String("consumer-group", "", "Consumer group ID.")
```
brianstrauch, line 45:
```suggestion
	cmd.Flags().String("group", "", "Consumer group ID.")
```

internal/kafka/command_consumer_group.go
```diff
@@ -0,0 +1,47 @@
...
+	IsSimple          bool   `human:"Simple" serialized:"is_simple"`
+	PartitionAssignor string `human:"Partition Assignor" serialized:"partition_assignor"`
+	State             string `human:"State" serialized:"state"`
+}
+
+func (c *consumerCommand) newGroupCommand(cfg *config.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "group",
+		Short:       "Manage Kafka consumer groups.",
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLoginOrOnPremLogin},
```
brianstrauch, line 23: Same as parent, so not needed.
```suggestion
```

internal/kafka/command_consumer_group_list.go
```diff
@@ -7,12 +7,12 @@ import (
...
 )
 
-func (c *consumerGroupCommand) newListCommand() *cobra.Command {
+func (c *consumerCommand) newGroupListCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "list",
 		Short: "List Kafka consumer groups.",
 		Args:  cobra.NoArgs,
-		RunE:  c.list,
+		RunE:  c.listGroups,
```
brianstrauch, line 15: Follow the format of our other commands!
```suggestion
		RunE:  c.groupList,
```

internal/kafka/command_consumer_list.go
```diff
@@ -39,7 +39,7 @@ func (c *consumerCommand) list(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	consumerGroup, err := cmd.Flags().GetString("consumer-group")
+	consumerGroup, err := cmd.Flags().GetString("group")
```
brianstrauch, line 42:
```suggestion
	group, err := cmd.Flags().GetString("group")
```

internal/kafka/command_consumer.go
```diff
@@ -0,0 +1,55 @@
...
+
+type consumerCommand struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+type consumerOut struct {
+	ConsumerGroupId string `human:"Consumer Group" serialized:"consumer_group_id"`
+	ConsumerId      string `human:"Consumer" serialized:"consumer"`
+	InstanceId      string `human:"Instance" serialized:"instance"`
+	ClientId        string `human:"Client" serialized:"client"`
```
brianstrauch, line 18: Should these also include "_id" in the serialized name, since the first one does?

internal/kafka/command_consumer_group_describe.go
```diff
@@ -0,0 +1,61 @@
...
+	group, err := kafkaREST.CloudClient.GetKafkaConsumerGroup(args[0])
+	if err != nil {
+		return err
+	}
+
+	table := output.NewTable(cmd)
+	table.Add(&consumerGroupOut{
+		ClusterId:         group.GetClusterId(),
+		ConsumerGroupId:   group.GetConsumerGroupId(),
+		Coordinator:       getStringBroker(group.GetCoordinator().Related),
```
brianstrauch, line 44:
```suggestion
		Coordinator:       getStringBroker(group.Coordinator.GetRelated()),
```

internal/kafka/command_consumer_group_describe.go
```diff
@@ -0,0 +1,61 @@
...
+		IsSimple:          group.GetIsSimple(),
+		PartitionAssignor: group.GetPartitionAssignor(),
+		State:             group.GetState(),
+	})
+	return table.Print()
+}
+
+func getStringBroker(relationship string) string {
+	// relationship will look like ".../v3/clusters/{cluster_id}/brokers/{broker_id}
+	splitString := strings.SplitAfter(relationship, "brokers/")
```
brianstrauch, line 54: Split by "/" and take the last element?

brianstrauch: Should be very easy to write unit tests for this function.

internal/kafka/command_consumer_list.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (c *consumerCommand) newListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List Kafka consumers.",
+		Args:  cobra.NoArgs,
+		RunE:  c.list,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `List all consumers for consumer group "my-consumer-group".`,
```
brianstrauch, line 19:
```suggestion
				Text: `List all consumers in consumer group "my-consumer-group".`,
```

internal/kafka/command_consumer_list.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+
+	consumers, err := kafkaREST.CloudClient.ListKafkaConsumers(group)
+	if err != nil {
+		return err
+	}
+
+	list := output.NewList(cmd)
+	for _, consumer := range consumers.Data {
```
brianstrauch, line 53:
```suggestion
	for _, consumer := range consumers.GetData() {
```

internal/kafka/command_consumer_group_describe.go
sgagniere: I think this is fine, because it also serves as a check that we're actually grabbing the right value (since it must be after `brokers`).

I did add the unit test, though.

brianstrauch (approved):

test/fixtures/output/local/services/help-onprem.golden
```diff
@@ -5,6 +5,7 @@ Usage:
 
 Available Commands:
   connect         Manage Connect.
+  control-center  Manage Control Center.
```
brianstrauch, line 8:
```suggestion
```

test/fixtures/output/local/services/help.golden
```diff
@@ -5,6 +5,7 @@ Usage:
 
 Available Commands:
   connect         Manage Connect.
+  control-center  Manage Control Center.
```
brianstrauch, line 8:
```suggestion
```

PR #2142, opened 2023-07-28
Adding SSO group mapping CRUDL commands

kevin-wu24:
Release Notes
-------------
New Features
  - `confluent iam group-mapping` CRUDL commands

Checklist
---------

What
----
Implementing these commands according to api spec. 
Currently, these resources are prefixed with `pool-` but this will change to `group-` before GA. However, since the CLI is needed for LA, we can go back and fix the CLI examples retroactively.

These commands will be placed under the cli.iam.group-mapping feature flag.

This PR also cleans up some code for `iam provider, pool, service-account`. 

References
----------
https://confluentinc.atlassian.net/wiki/spaces/AE/pages/3058335992/New+Trust-Service+Endpoints+for+AUPM+Flink#AUPM

Test & Review
-------------
Integration tests for AUPM group-mapping CRUDL commands.


internal/cmd/iam/command.go
```diff
@@ -29,6 +29,9 @@ func New(cfg *v1.Config, prerunner pcmd.PreRunner) *cobra.Command {
 		cmd.AddCommand(newPoolCommand(prerunner))
 		cmd.AddCommand(newProviderCommand(prerunner))
 	}
+	if cfg.IsTest || featureflags.Manager.BoolVariation("cli.group-mapping", dc.Context(), v1.CliLaunchDarklyClient, true, false) {
```
brianstrauch, line 32: Have you made this feature flag yet, or do you need me to make it?
```suggestion
	if cfg.IsTest || featureflags.Manager.BoolVariation("cli.iam.group-mapping", dc.Context(), v1.CliLaunchDarklyClient, true, false) {
```

internal/cmd/iam/command.go
```diff
@@ -29,6 +29,9 @@ func New(cfg *v1.Config, prerunner pcmd.PreRunner) *cobra.Command {
 		cmd.AddCommand(newPoolCommand(prerunner))
```
brianstrauch, line 29: I wonder if we can remove the feature flag for this command now?

go.mod
```diff
@@ -189,3 +190,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/confluentinc/ccloud-sdk-go-v2/sso => ../ccloud-sdk-go-v2-internal/sso
```
brianstrauch, line 194: Mark as "Draft" until it's ready?

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,56 @@
...
+func (c *groupMappingCommand) newCreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create <name>",
+		Short: "Create a group mapping.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.create,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Create a group mapping named "DemoGroupMapping".`,
+				Code: `confluent iam group-mapping create DemoGroupMapping --description "description of group mapping" --filter "\"demo\" in claims.group"`,
```
brianstrauch, line 21: What does the filter mean?

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,56 @@
...
+			},
+		),
+	}
+
+	cmd.Flags().String("description", "", "Description of the group mapping.")
+	cmd.Flags().String("filter", "", "CEL-compliant filter for this group mapping")
```
brianstrauch, line 27:
```suggestion
	cmd.Flags().String("filter", "", "CEL-compliant filter for this group mapping.")
```

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,56 @@
...
+	pcmd.AddOutputFlag(cmd)
+
+	return cmd
+}
+
+func (c *groupMappingCommand) create(cmd *cobra.Command, args []string) error {
+	name := args[0]
```
brianstrauch, line 34: IMO, no need to make a variable here, especially since this only gets used in one spot.

internal/pkg/resource/resource.go
```diff
@@ -19,6 +19,7 @@ const (
 	FlinkRegion           = "Flink region"
 	FlinkIamBinding       = "Flink IAM binding"
 	FlinkStatement        = "Flink SQL statement"
+	GroupMapping          = "group mapping"
```
brianstrauch, line 22: Should the output include "SSO"?
```suggestion
	SsoGroupMapping          = "SSO group mapping"
```

internal/cmd/iam/command.go
kevin-wu24: I need you to make it haha

internal/cmd/iam/command_groupmapping_create.go
kevin-wu24: From the `openapi.html` file: 

A filter expression in [Supported Common Expression Language (CEL)](https://docs.confluent.io/cloud/current/access-management/authenticate/oauth/identity-pools.html#supported-common-expression-language-cel-filters) that specifies which identities can authenticate using your group mapping (see [Set identity pool filters](https://docs.confluent.io/cloud/current/access-management/authenticate/oauth/identity-pools.html#set-identity-pool-filters) for more details).

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,54 @@
...
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Create a group mapping named "DemoGroupMapping".`,
+				Code: `confluent iam group-mapping create DemoGroupMapping --description "description of group mapping" --filter "\"demo\" in claims.group"`,
+			},
+		),
+	}
+
+	cmd.Flags().String("description", "", "Description of the group mapping.")
+	cmd.Flags().String("filter", "true", "A supported Common Expression Language filter expression for group mappings.")
```
brianstrauch, line 27:
```suggestion
	cmd.Flags().String("filter", "true", "A supported Common Expression Language (CEL) filter expression for group mappings.")
```

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return err
+	}
+
+	filter, err := cmd.Flags().GetString("filter")
+	if err != nil {
+		return err
+	}
+
+	newGroupMapping := ssov2.IamV2SsoGroupMapping{
```
brianstrauch, line 44: It's obviously "new" if you're using `:=`!
```suggestion
	groupMapping := ssov2.IamV2SsoGroupMapping{
```

internal/cmd/iam/command_groupmapping_create.go
```diff
@@ -0,0 +1,54 @@
...
+		DisplayName: ssov2.PtrString(args[0]),
+		Description: ssov2.PtrString(description),
+		Filter:      ssov2.PtrString(filter),
+	}
+	groupMapping, err := c.V2Client.CreateGroupMapping(newGroupMapping)
```
brianstrauch, line 49: (Based on the suggested change above)
```suggestion
	groupMapping, err = c.V2Client.CreateGroupMapping(newGroupMapping)
```

internal/cmd/iam/command_groupmapping_delete.go
```diff
@@ -0,0 +1,53 @@
...
+	cmd := &cobra.Command{
+		Use:               "delete <id>",
+		Short:             "Delete a group mapping.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE:              c.delete,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Delete group mapping "pool-12345":`,
+				Code: "confluent iam group-mapping delete pool-12345",
```
brianstrauch, line 26: I thought the "pool" prefix was for identity pools?

internal/cmd/iam/command_groupmapping_list.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (c *groupMappingCommand) newListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List group mappings.",
+		Args:  cobra.NoArgs,
+		RunE:  c.list,
+	}
+
+	pcmd.AddOutputFlag(cmd)
```
brianstrauch, line 18: Please add `--environment` and `--context`. All Cloud commands have these flags.

internal/cmd/iam/command_groupmapping_update.go
```diff
@@ -0,0 +1,72 @@
...
+	cmd := &cobra.Command{
+		Use:               "update <id>",
+		Short:             "Update a group mapping.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE:              c.update,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Update the description of group mapping "pool-123456".`,
+				Code: `confluent iam group-mapping update pool-123456 --description "Update demo group mapping information."`,
```
brianstrauch, line 23: Let's keep our examples concise so they don't take up multiple lines
```suggestion
				Code: `confluent iam group-mapping update pool-123456 --description "updated description"`,
```

internal/cmd/iam/command_groupmapping_update.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+		),
+	}
+
+	cmd.Flags().String("name", "", "Name of the group mapping.")
+	cmd.Flags().String("description", "", "Description of the group mapping.")
+	cmd.Flags().String("filter", "", "A supported Common Expression Language filter expression for group mappings.")
```
brianstrauch, line 30:
```suggestion
	cmd.Flags().String("filter", "", "A supported Common Expression Language (CEL) filter expression for group mappings.")
```

internal/pkg/errors/error_message.go
```diff
@@ -76,6 +76,9 @@ const (
 	InvalidOperationValueErrorMsg = "invalid operation value: %s"
 	ExactlyOneSetErrorMsg         = "exactly one of %v must be set"
 
+	// iam group-mapping commands
+	GroupMappingNoOpUpdateErrorMsg = "one of `--description`, `--name`, or `--filter` must be set"
```
brianstrauch, line 80: If this only gets used in one package, let's define it in that package, not globally.

test/test-server/ccloud_handlers.go
```diff
@@ -51,6 +51,7 @@ var (
 const (
 	serviceAccountID           = int32(12345)
 	serviceAccountResourceID   = "sa-12345"
+	groupMappingResourceId     = "pool-abc"
```
brianstrauch, line 54: Either change all the other variables here to use "Id", or call this variable "ID". Consistency is always the better approach.

internal/cmd/iam/command_groupmapping_delete.go
kevin-wu24: It is, and it's planned to be changed to `group` in Q4. I think for the LA release in September it will still be `pool`. Would it be better to change it to `group` now or retroactively update it when the API is updated?

brianstrauch: Let's keep it as `pool`, I guess. Can you open a ticket for this so we don't forget?

internal/cmd/iam/command_groupmapping_list.go
kevin-wu24: I took a look at other `iam` resources (non-RBAC and ACL) like `service-account`, `provider`, and `pool`. And none of them have an `--environment` or `--context` flag for their list command. 

I guess the `--context` flag makes sense here for `iam` resources because I could list out these resources for a different platform without having to explicitly change contexts first, but I don't think these issues are env-level, so why would I need an `--environment` flag?

brianstrauch: Oh, are these resources all at the organization level?

kevin-wu24: Yeah the ones mentioned above and `group-mapping` are org level according to their api-specs: https://github.com/confluentinc/api/blob/master/sso/minispec.yaml#L42
https://github.com/confluentinc/api/blob/master/identity-provider/minispec.yaml#L42
https://github.com/confluentinc/api/blob/master/iam/minispec.yaml#L86

internal/cmd/iam/command_groupmapping_update.go
kevin-wu24: sounds good! looks like `iam pool` uses this too

internal/cmd/iam/command_groupmapping_delete.go
```diff
@@ -0,0 +1,53 @@
...
+			},
+		),
+	}
+
+	pcmd.AddForceFlag(cmd)
```
brianstrauch, line 31: `--context`?

internal/cmd/iam/command_groupmapping_update.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/spf13/cobra"
+
+	ssov2 "github.com/confluentinc/ccloud-sdk-go-v2/sso/v2"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/examples"
+)
+
+const groupMappingNoOpUpdateErrorMsg = "one of `--description`, `--name`, or `--filter` must be set"
```
brianstrauch, line 13: Do we have a global format string that we can use for all of our update commands?

internal/cmd/iam/command_groupmapping_update.go
```diff
@@ -0,0 +1,74 @@
...
+	cmd := &cobra.Command{
+		Use:               "update <id>",
+		Short:             "Update a group mapping.",
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE:              c.update,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Update the description of group mapping "pool-123456".`,
+				Code: `confluent iam group-mapping update pool-123456 --description "Updated description"`,
```
brianstrauch, line 25: Nit: Keep everything lowercase, easier to type!
```suggestion
				Code: `confluent iam group-mapping update pool-123456 --description "updated description"`,
```

internal/cmd/iam/command_pool_create.go
```diff
@@ -18,19 +17,18 @@ func (c *identityPoolCommand) newCreateCommand() *cobra.Command {
 		RunE:  c.create,
 		Example: examples.BuildExampleString(
 			examples.Example{
-				Text: `Create an identity pool named "DemoIdentityPool" with provider "op-12345":`,
-				Code: `confluent iam pool create DemoIdentityPool --provider op-12345 --description new-description --identity-claim claims.sub --filter 'claims.iss=="https://my.issuer.com"'`,
+				Text: `Create an identity pool named "demo-identity-pool" with provider "op-12345":`,
```
brianstrauch, line 20: What kind of a provider is this? Wondering if there's more specific terminology we can use
```suggestion
				Text: `Create an identity pool named "demo-identity-pool" with identity provider "op-12345":`,
```

test/fixtures/output/iam/group-mapping/create.golden
```diff
@@ -0,0 +1,9 @@
++-------------+--------------------------------+
+| ID          | pool-12345                     |
+| Name        | group_mapping                  |
+| Description | new-group-description          |
```
brianstrauch, line 4: We have a ticket to include this output in our public-facing documentation, so let's make sure the formatting matches the style of our examples.
```suggestion
| Name        | my-group-mapping               |
| Description | new-description                |
```

internal/cmd/iam/command_groupmapping_update.go
kevin-wu24: I don't think we have one? A no-op error only makes sense when all the fields in a resource's `update` API call are optional, so we can just avoid an unnecessary API call, which isn't the case for a lot of resources' `update` commands.

From a initial look-around, it looks like the `group-mapping, provider, pool` resources involved in this PR could use a global format string for an `update` no-op error msg, and also `pipeline`, `kafka cluster`, `kafka quota`, and `context`. Since there are quite a few resources that could use this and this PR is already getting quite large, I think I'll just make a ticket and do it in another PR.

internal/cmd/iam/command_pool_create.go
kevin-wu24: According to: https://docs.confluent.io/cloud/current/api.html#tag/Identity-Providers-(iamv2) they are "OAuth/OpenID Connect" identity providers.

test/fixtures/output/iam/group-mapping/create.golden
kevin-wu24: Can you link the ticket? Also IIUC, you mean our CRUDL integration test output should match what is in our examples?

brianstrauch: I mean the examples and integration tests should all be formatted with lower-kebab-case and have similar fake resource names / IDs.

kevin-wu24: oh ok gotcha

internal/cmd/iam/command_provider_update.go
```diff
@@ -8,9 +8,10 @@ import (
 	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
 	"github.com/confluentinc/cli/internal/pkg/errors"
 	"github.com/confluentinc/cli/internal/pkg/examples"
-	"github.com/confluentinc/cli/internal/pkg/output"
 )
 
+const identityProviderNoOpUpdateErrorMsg = "one of `--description` or `--name` must be set"
```
brianstrauch, line 13: Does this get used in multiple places in this package? If not, no need to create a constant (it just makes you have to jump around more when you're reading the code)

internal/pkg/resource/resource.go
```diff
@@ -19,6 +19,7 @@ const (
 	FlinkRegion           = "Flink region"
 	FlinkIamBinding       = "Flink IAM binding"
 	FlinkStatement        = "Flink SQL statement"
+	SsoGroupMapping       = "SSO group mapping"
```
brianstrauch, line 22: Alphabetize!

test/test-server/utils.go
```diff
@@ -349,6 +351,37 @@ func buildIamInvitation(id, email, userId, status string) iamv2.IamV2Invitation
 	}
 }
 
+func buildIamGroupMapping(id, name, description, filter string) ssov2.IamV2SsoGroupMapping {
```
brianstrauch, line 354: Does this get used in more than one file in this package? If not, there's no need to put it in a different file than where it get used. (Results in needing to jump around more while reading code, and files with generic names like utils.go often have the tendency to get really big... like hundreds of lines)

brianstrauch (approved):

internal/cmd/iam/command_groupmapping_update.go
```diff
@@ -22,7 +20,7 @@ func (c *groupMappingCommand) newUpdateCommand() *cobra.Command {
 		Example: examples.BuildExampleString(
 			examples.Example{
 				Text: `Update the description of group mapping "pool-123456".`,
-				Code: `confluent iam group-mapping update pool-123456 --description "updated description"`,
+				Code: "confluent iam group-mapping update pool-123456 --description updated-description",
```
brianstrauch, line 23: Sorry for being ambiguous earlier. Our goal is to replicate what users would most likely want to input here. I think the majority of the time, users want to have spaces in their description. Any flag value or argument (resource) should ideally be lower-kebab-case since that's what's easiest to type.
```suggestion
				Code: `confluent iam group-mapping update pool-123456 --description "updated description"`,
```

kevin-wu24: All good, just gonna revert the last commit then.

PR #2144, opened 2023-07-28
Cloud Kafka REST cleanup

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
- Move (most) calls to `kafkarest.NewError` into the ccloudv2 sdk file.
- Move `RestProxyNotAvailableMsg` into `GetKafkaREST` and remove checks for `kafkaREST == nil` as a result.
- Get rid of some pointless code.
- Switch all `confluent kafka mirror` to using the cloud kafkarest sdk instead of the on-prem sdk.
- Add a few more integration tests for `kafka mirror [failover | pause | resume]`

References
----------

Test & Review
-------------
all tests pass
manual testing of mirror

Open Questions / Follow-ups
---------------------------


internal/cmd/asyncapi/command_import.go
```diff
@@ -288,16 +286,8 @@ func (c *command) createTopic(details *accountDetails, topicName string, kafkaBi
...
 	}
-	if _, httpResp, err := kafkaRest.CloudClient.CreateKafkaTopic(details.kafkaClusterId,
-		createTopicRequestData); err != nil {
-		restErr, parseErr := kafkarest.ParseOpenAPIErrorCloud(err)
-		if parseErr == nil && restErr.Code == ccloudv2.BadRequestErrorCode {
-			// Print partition limit error w/ suggestion
-			if strings.Contains(restErr.Message, "partitions will exceed") {
-				return false, errors.NewErrorWithSuggestions(restErr.Message, errors.ExceedPartitionLimitSuggestions)
-			}
-		}
```
brianstrauch, line 299: Are we sure we want to move all of this logic into the SDK wrappers file? IMO, the SDK wrappers should be straightforward and potentially able to be auto-generated in the future.

sgagniere: I can move all suggestion wrapping back into the command files. That probably shouldn't be the job of the sdk wrappers...

But I think we can't make it completely straightforward in the ccloudv2 file, because error message replacement logic should be baked in like with `DeleteKafkaTopic`. I'm not sure how much that affects future attempts to auto generate it.

internal/cmd/kafka/command_mirror_list.go
```diff
@@ -55,53 +56,52 @@ func (c *mirrorCommand) list(cmd *cobra.Command, _ []string) error {
...
 	cluster, err := c.Context.GetKafkaClusterForCommand()
 	if err != nil {
 		return err
 	}
 
-	mirrorStatusOpt := optional.EmptyInterface()
+	var mirrorTopicStatus *kafkarestv3.MirrorTopicStatus
 	if mirrorStatus != "" {
-		mirrorStatusOpt = optional.NewInterface(kafkarestv3.MirrorTopicStatus(mirrorStatus))
+		mirrorTopicStatus, err = kafkarestv3.NewMirrorTopicStatusFromValue(strings.ToUpper(mirrorStatus))
```
sgagniere, line 70: The `kafka-rest-sdk-go` version accepted lowercase strings here, while the `ccloud-sdk-go-v2` version does not. I added `strings.ToUpper` here so that this doesn't cause a breaking change, but we could remove this in V4 and enforce uppercase.

internal/pkg/kafkarest/kafkarest.go
```diff
@@ -64,6 +64,10 @@ func ParseOpenAPIErrorCloud(err error) (*V3Error, error) {
 		if err := json.Unmarshal(openAPIError.Body(), &decodedError); err != nil {
 			return nil, err
 		}
+		if decodedError.Message == "" && decodedError.Code == 0 {
+			// Sometimes the SDK puts the error message in `error` instead of `body`
+			decodedError.Message = openAPIError.Error()
+		}
```
sgagniere, line 70: Specifically, if kafkarest SDK function returns a struct, then it puts whatever the current state of the return struct is in the `body` field.

It's a weird design, but unless/until we modify the SDK we'll have to work around that to catch these errors.

internal/cmd/kafka/command_consumergroup.go
```diff
@@ -83,14 +82,19 @@ func (c *consumerGroupCommand) autocompleteConsumerGroups() []string {
...
 	if err != nil {
 		return nil, err
 	}
 
-	groupCmdResp, httpResp, err := kafkaREST.CloudClient.ListKafkaConsumerGroups(lkc)
+	cluster, err := flagCmd.Context.GetKafkaClusterForCommand()
 	if err != nil {
-		return nil, kafkarest.NewError(kafkaREST.CloudClient.GetUrl(), err, httpResp)
+		return nil, err
+	}
```
brianstrauch, line 93: Thanks for splitting this!

internal/cmd/kafka/command_consumergroup_lag_summarize.go
```diff
@@ -48,14 +47,19 @@ func (c *lagCommand) newSummarizeCommand() *cobra.Command {
...
 	}
 
-	summary, httpResp, err := kafkaREST.CloudClient.GetKafkaConsumerGroupLagSummary(lkc, consumerGroupId)
+	cluster, err := c.Context.GetKafkaClusterForCommand()
 	if err != nil {
-		return kafkarest.NewError(kafkaREST.CloudClient.GetUrl(), err, httpResp)
+		return err
+	}
+
+	summary, err := kafkaREST.CloudClient.GetKafkaConsumerGroupLagSummary(cluster.ID, consumerGroupId)
```
brianstrauch, line 60: Does **every** Kafka REST function require a cluster ID? If so, should we add it to the Kafka REST client struct so we don't need to pass it here?

internal/cmd/kafka/command_link_configuration_list.go
```diff
@@ -39,43 +37,40 @@ func (c *linkCommand) configurationList(cmd *cobra.Command, args []string) error
...
 	list := output.NewList(cmd)
-	if len(listLinkConfigsRespData.Data) == 0 {
+	if len(configs.GetData()) == 0 {
 		return list.Print()
 	}
 
 	list.Add(&linkConfigurationOut{
 		ConfigName:  "dest.cluster.id",
-		ConfigValue: listLinkConfigsRespData.Data[0].ClusterId,
+		ConfigValue: configs.GetData()[0].GetClusterId(),
```
brianstrauch, line 61: This seems a bit weird, and forces us to add the special case on line 55. Can we just use `cluster.ID` here?

brianstrauch:
```suggestion
		ConfigValue: cluster.ID,
```

internal/cmd/kafka/command_link_list_onprem.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/confluentinc/cli/internal/pkg/output"
 )
 
-func newLinkOnPrem(data kafkarestv3.ListLinksResponseData, topic string) *link {
-	l := &link{
+func newLinkOnPrem(data kafkarestv3.ListLinksResponseData, topic string) *listOut {
+	l := &listOut{
```
brianstrauch, line 13: Can we come up with a better name than `l`?

internal/cmd/kafka/command_mirror_list.go
```diff
@@ -28,13 +29,13 @@ func (c *mirrorCommand) newListCommand() *cobra.Command {
 			},
 			examples.Example{
 				Text: `List all active mirror topics under "my-link":`,
-				Code: "confluent kafka mirror list --link my-link --mirror-status active",
+				Code: "confluent kafka mirror list --link my-link --mirror-status ACTIVE",
```
brianstrauch, line 32: Can we just upper-case whatever the user passes in? I'd like to keep all of our examples lower-kebab-case.

internal/cmd/kafka/command_mirror_list.go
```diff
@@ -55,53 +56,52 @@ func (c *mirrorCommand) list(cmd *cobra.Command, _ []string) error {
...
 	}
 
 	list := output.NewList(cmd)
-	for _, mirror := range listMirrorTopicsResponseDataList.Data {
+	for _, mirror := range mirrors.GetData() {
 		var maxLag int64 = 0
-		for _, mirrorLag := range mirror.MirrorLags {
-			if mirrorLag.Lag > maxLag {
-				maxLag = mirrorLag.Lag
+		for _, mirrorLag := range mirror.GetMirrorLags().Items {
```
brianstrauch, line 92:
```suggestion
		for _, mirrorLag := range mirror.MirrorLags.GetItems() {
```

internal/pkg/ccloudv2/kafkarest.go
```diff
@@ -58,15 +58,17 @@ func (c *KafkaRestClient) UpdateKafkaClusterConfigs(clusterId string, req kafkar
...
+func (c *KafkaRestClient) BatchCreateKafkaAcls(clusterId string, list kafkarestv3.CreateAclRequestDataList) error {
+	httpResp, err := c.ACLV3Api.BatchCreateKafkaAcls(c.context(), clusterId).CreateAclRequestDataList(list).Execute()
+	return kafkarest.NewError(c.GetUrl(), err, httpResp)
 }
 
-func (c *KafkaRestClient) CreateKafkaAcls(clusterId string, data kafkarestv3.CreateAclRequestData) (*http.Response, error) {
-	return c.ACLV3Api.CreateKafkaAcls(c.context(), clusterId).CreateAclRequestData(data).Execute()
+func (c *KafkaRestClient) CreateKafkaAcls(clusterId string, data kafkarestv3.CreateAclRequestData) error {
+	httpResp, err := c.ACLV3Api.CreateKafkaAcls(c.context(), clusterId).CreateAclRequestData(data).Execute()
+	return kafkarest.NewError(c.GetUrl(), err, httpResp)
```
brianstrauch, line 68: Thoughts on not including this in the error?
```suggestion
	return kafkarest.NewError(err, httpResp)
```

internal/pkg/cmd/prerunner.go
```diff
@@ -368,7 +368,7 @@ func (r *PreRun) setCCloudClient(c *AuthenticatedCLICommand) error {
 
 			return kafkaRest, nil
 		}
-		return nil, nil
+		return nil, errors.New("Kafka REST is not enabled: the operation is only supported with Kafka REST proxy.")
```
brianstrauch, line 371: It's hard to tell when this gets returned... can we invert the above `if`?

test/fixtures/output/kafka/mirror/list-help.golden
```diff
@@ -10,11 +10,11 @@ List all mirror topics in the cluster:
 
 List all active mirror topics under "my-link":
 
-  $ confluent kafka mirror list --link my-link --mirror-status active
+  $ confluent kafka mirror list --link my-link --mirror-status ACTIVE
 
 Flags:
       --link string            Name of cluster link.
-      --mirror-status string   Mirror topic status. Can be one of [active, failed, paused, stopped, pending_stopped]. If not specified, list all mirror topics.
+      --mirror-status string   Mirror topic status. Can be one of "ACTIVE", "FAILED", "PAUSED", "STOPPED", or "PENDING_STOPPED". If not specified, list all mirror topics.
```
brianstrauch, line 17:
```suggestion
      --mirror-status string   Mirror topic status. Can be one of "active", "failed", "paused", "stopped", or "pending-stopped". If not specified, list all mirror topics.
```

test/fixtures/output/kafka/mirror/resume-mirror.golden
```diff
@@ -0,0 +1,5 @@
+  Mirror Topic Name | Partition | Partition Mirror Lag | Last Source Fetch Offset | Error Message  | Error Code
```
brianstrauch, line 1: How about "Topic Name" instead of "Mirror Topic Name"? The command is called `kafka mirror resume` so the user knows this is a mirror topic.

test/test-server/kafka_rest_router.go
```diff
@@ -935,59 +938,219 @@ func handleKafkaRPLagSummary(t *testing.T) http.HandlerFunc {
...
+					},
+				},
+			}})
+			require.NoError(t, err)
+		}
+	}
+}
+
+// Handler for: "/kafka/v3/clusters/{cluster_id}/links/{link_name}/mirrors:resume"
+func handleKafkaRPMirrorsResume(t *testing.T) http.HandlerFunc {
```
brianstrauch, line 1111: A bit easier to understand than wondering what "RP" stands for.
```suggestion
func handleKafkaRestMirrorsResume(t *testing.T) http.HandlerFunc {
```

internal/cmd/kafka/command_mirror_list.go
sgagniere: There is no `GetItems()` method.

test/fixtures/output/kafka/mirror/resume-mirror.golden
sgagniere: Let's save that change for v4, so that the human and serialized labels stay consistent in the meantime.

internal/cmd/kafka/command_consumergroup_lag_summarize.go
sgagniere: Good call. It looks like that's the case.

Also, `GetKafkaREST` already calls `GetKafkaClusterForCommand` so we'll be able to remove a redundant call from many commands and make them a bit more efficient.

internal/pkg/ccloudv2/kafkarest.go
sgagniere: I think it could provide useful information, so I'm not sure about removing it.

brianstrauch (approved):
Thanks!!

PR #2152, opened 2023-08-01
[CLI-2576] Detect Homebrew installation on `confluent update`

brianstrauch:
Release Notes
-------------
New Features
- Suggest updating with `brew upgrade` if installed with Homebrew

Checklist
---------
   * yes: ok

What
----
If updates are disabled (CP or Homebrew), mark `confluent update` as hidden and return a helpful suggestion on execution.
* For Homebrew, suggest `brew upgrade`
* For CP, suggest independently installing a binary according to https://docs.confluent.io/platform/current/installation/versions-interoperability.html#confluent-cli

Test & Review
-------------
Manually tested all of the negative paths, added tracing to debug the positive path, if needed.

### DIFF
--- a/internal/cmd/update/command.go
+++ b/internal/cmd/update/command.go
@@ -91,7 +91,7 @@ func (c *command) update(cmd *cobra.Command, _ []string) error {
 
 		return errors.NewErrorWithSuggestions(
 			message,
-			"If installed with Confluent Platform, you may install an independent, updated version of the CLI, as long as it maintains compatibility: https://docs.confluent.io/platform/current/installation/versions-interoperability.html#confluent-cli",
+			"Use a package manager to update the binary, if applicable. Otherwise, consider deleting this binary and re-installing a newer version.",
 		)
 	}
 
--- a/internal/cmd/update/command.go
+++ b/internal/cmd/update/command.go
@@ -3,6 +3,7 @@ package update
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/cmd/update/command.go
```diff
@@ -74,6 +80,21 @@ func NewClient(cliName string, disableUpdateCheck bool) update.Client {
...
+	if c.Config.DisableUpdates {
+		message := "updates are disabled for this binary"
+		if isHomebrew() {
+			return errors.NewErrorWithSuggestions(
+				message,
+				fmt.Sprintf("If installed with Homebrew, run `brew upgrade %s`.", homebrewFormula),
+			)
+		}
+
+		return errors.NewErrorWithSuggestions(
```
sgagniere, line 92: The only thing I'm wondering here is if we want to always default to the assumption that they're using the CP installed CLI, or if we want to try and detect this too. Mainly just in case other checks before this report false negatives (e.g. `EvalSymLinks` can fail in some cases, though I'm not sure if they're applicable to us).

There's already logic in `connect plugin install` that examines the CLI's path from `os.Executable()` to determine if it's part of a package manager or archive CP installation.

sgagniere (approved):

internal/cmd/update/command.go
brianstrauch: > EvalSymLinks can fail in some cases

Good point, we should show a generic error in this case instead of a CP-specific error. I'll let CP detection be a separate PR if we want to go down that route.

internal/cmd/update/command.go
```diff
@@ -74,6 +80,21 @@ func NewClient(cliName string, disableUpdateCheck bool) update.Client {
...
+			message,
+			"If installed with Confluent Platform, you may install an independent, updated version of the CLI, as long as it maintains compatibility: https://docs.confluent.io/platform/current/installation/versions-interoperability.html#confluent-cli",
```
brianstrauch, line 94:
```suggestion
			"Use a package manager to update the binary, if applicable. Otherwise, consider deleting this binary and re-installing a newer version.",
```

sgagniere (approved):
Lgtm.

PR #2154, opened 2023-08-01
CLI-2667 commit offset for consumer group and start consuming from it the next time the same group is specified

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- `confluent kafka topic consume --group` will start consuming from the offset that it exited last time

Checklist
---------
   * yes: ok 

What
----

consuming with an existing consumer group will commit the offset it consumed to to the broker.
when specifying an existing consumer group in consumer:
when offset is specified, start consuming from that offset;
when --from-beginning is specified, start consuming from the beginning;
when neither flags are specified, start consuming from where it left off when it exited.

```
mhe@C02DV0KVMD6T ~ % confluent kafka topic consume b --group confluent_cli_consumer_eafe1ee7-7ab0-4557-b7b6-5a9597e485ae
Starting Kafka Consumer. Use Ctrl-C to exit.
15
17
18
14
16 // these are new produced msgs after the last time this group was consuming

mhe@C02DV0KVMD6T ~ % confluent kafka topic consume b --group confluent_cli_consumer_eafe1ee7-7ab0-4557-b7b6-5a9597e485ae --offset 1
Starting Kafka Consumer. Use Ctrl-C to exit.
13
8
9
6
11
15
17
18
2
12
14
16
```

References
----------

https://confluent.slack.com/archives/C9Y6NAM6X/p1690906607470689

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/cmd/kafka/command_topic_consume.go
```diff
@@ -80,8 +80,11 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	var commitGroupOffset bool
 	if !cmd.Flags().Changed("group") {
 		group = fmt.Sprintf("confluent_cli_consumer_%s", uuid.New())
+	} else {
+		commitGroupOffset = true
```
brianstrauch, line 87: Potential edge case: What if someone uses a default group ID the first time they call `confluent kafka topic consume`, and then they run `confluent kafka topic consume --group <default-group-id-from-the-first-run>`?

internal/cmd/kafka/command_topic_consume.go
```diff
@@ -144,6 +147,9 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
 	}
 
 	rebalanceCallback := GetRebalanceCallback(offset, partitionFilter)
+	if commitGroupOffset && !cmd.Flags().Changed("from-beginning") && !cmd.Flags().Changed("offset") {
+		rebalanceCallback = nil
```
brianstrauch, line 151: Do you mind explaining how the rebalance callback works? I thought this was related to when a cluster adds or subtracts partitions?

internal/cmd/kafka/confluent_kafka_configs.go
```diff
@@ -71,6 +71,14 @@ func getConsumerConfigMap(group string, kafka *configv1.KafkaClusterConfig, clie
 		return nil, err
 	}
 	log.CliLogger.Debugf("Created consumer group: %s", group)
+	if commitGroupOffset {
+		if err := configMap.SetKey("enable.auto.commit", true); err != nil {
+			return nil, err
+		}
```
brianstrauch, line 77: The documentation says this is the default, pretty sure we can remove it. https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md
```suggestion
```

internal/cmd/kafka/confluent_kafka_configs.go
```diff
@@ -71,6 +71,14 @@ func getConsumerConfigMap(group string, kafka *configv1.KafkaClusterConfig, clie
...
+		if err := configMap.SetKey("auto.commit.interval.ms", 100); err != nil {
```
brianstrauch, line 78: Ah, I think this is what I was doing wrong earlier. The default is 5s...

brianstrauch (commented):
Thank you!!

internal/cmd/kafka/command_topic_consume.go
brianstrauch: In other words, maybe we should always commit the group offset?

internal/cmd/kafka/confluent_kafka_configs.go
brianstrauch: What are your thoughts on manually committing? What if the client closes before the 100ms timer goes off?

brianstrauch: (I think that's what `kcat` does)

internal/cmd/kafka/command_topic_consume.go
MuweiHe: The rebalanceCallback is a function that gets called when the Kafka client needs to rebalance its consumer group, for example when a new consumer joins the group, or an existing consumer leaves the group. it allows me to define a specific offset to consume from when user passes in `--offset`

internal/cmd/kafka/command_topic_consume.go
MuweiHe: if it's set to true by default then i think yes, we don't need to change that. i should rename this var to be `consume from committed offset instead of rebalanceCallback`

internal/cmd/kafka/confluent_kafka_configs.go
MuweiHe: how to manually commit? hitting a key or running a command?

brianstrauch: We can capture the interrupt from the ctrl-c key, or maybe even use a `defer` in case the program errors?

MuweiHe: `defer` sounds safer. changing that!

MuweiHe: or we can call `commit` every time a message is being consumed.

MuweiHe: added a defer in `runConsumer`

brianstrauch: :+1: Make sure it also runs when Ctrl-C is pressed!

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -238,6 +239,12 @@ func consumeMessage(e *ckafka.Message, h *GroupHandler) error {
 }
 
 func RunConsumer(consumer *ckafka.Consumer, groupHandler *GroupHandler) error {
+	defer func() {
+		_, err := consumer.Commit()
+		if err != nil {
```
brianstrauch, line 244:
```suggestion
		if _, err := consumer.Commit(); err != nil {
```

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -238,6 +239,12 @@ func consumeMessage(e *ckafka.Message, h *GroupHandler) error {
...
+			log.CliLogger.Warn("Failed to commit current offset of the consumer.")
```
brianstrauch, line 245:
```suggestion
			log.CliLogger.Warnf("Failed to commit current offset of the consumer: %v", err)
```

internal/cmd/kafka/confluent_kafka_configs.go
brianstrauch: (Also, do we still need to turn on auto-commit if we're manually committing?)

MuweiHe: let's keep auto-commit on but use a longer interval so it doesn't impact the performance.

brianstrauch: What's the reason for keeping auto-commit on?

MuweiHe: on the second thought. auto-commit is for convenience. the way we're doing now is more precise control.
turning it off.

internal/cmd/kafka/command_topic_consume.go
```diff
@@ -80,8 +80,11 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	var consumeFromGroupOffset bool
 	if !cmd.Flags().Changed("group") {
 		group = fmt.Sprintf("confluent_cli_consumer_%s", uuid.New())
+	} else {
+		consumeFromGroupOffset = true
 	}
```
brianstrauch, line 88:
```suggestion
	if cmd.Flags().Changed("group") {
		consumeFromGroupOffset = true
	} else {
		group = fmt.Sprintf("confluent_cli_consumer_%s", uuid.New())
	}
```

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -238,13 +239,21 @@ func consumeMessage(e *ckafka.Message, h *GroupHandler) error {
 }
 
 func RunConsumer(consumer *ckafka.Consumer, groupHandler *GroupHandler) error {
+	defer func() {
+		if _, err := consumer.Commit(); err != nil {
+			log.CliLogger.Warnf("Failed to commit current offset of the consumer: %v", err)
```
brianstrauch, line 244:
```suggestion
			log.CliLogger.Warnf("Failed to commit current consumer offset: %v", err)
```

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -238,13 +239,21 @@ func consumeMessage(e *ckafka.Message, h *GroupHandler) error {
...
+	}()
 	run := true
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt)
 	for run {
 		select {
 		case <-signals: // Trap SIGINT to trigger a shutdown.
 			output.ErrPrintln(errors.StoppingConsumerMsg)
+			if _, err := consumer.Commit(); err != nil {
+				log.CliLogger.Warnf("Failed to commit current offset of the consumer: %v", err)
```
brianstrauch, line 255:
```suggestion
				log.CliLogger.Warnf("Failed to commit current consumer offset: %v", err)
```

internal/cmd/kafka/confluent_kafka_configs.go
```diff
@@ -40,6 +40,7 @@ func getCommonConfig(kafka *configv1.KafkaClusterConfig, clientId string) (*ckaf
 		"bootstrap.servers":                     kafka.Bootstrap,
 		"sasl.username":                         kafka.APIKey,
 		"sasl.password":                         kafka.GetApiSecret(),
+		"auto.commit.interval.ms":               1000,
```
brianstrauch, line 43:
```suggestion
```

internal/cmd/kafka/confluent_kafka_configs.go
```diff
@@ -72,6 +73,10 @@ func getConsumerConfigMap(group string, kafka *configv1.KafkaClusterConfig, clie
 	}
 	log.CliLogger.Debugf("Created consumer group: %s", group)
 
+	if err := configMap.SetKey("enable.auto.commit", false): err!=nil {
```
brianstrauch, line 76: Doesn't look like this code compiles. Has it been tested?
```suggestion
	if err := configMap.SetKey("enable.auto.commit", false); err != nil {
```

MuweiHe: thanks! fixed build and tested once again

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -245,6 +246,9 @@ func RunConsumer(consumer *ckafka.Consumer, groupHandler *GroupHandler) error {
 		select {
 		case <-signals: // Trap SIGINT to trigger a shutdown.
 			output.ErrPrintln(errors.StoppingConsumerMsg)
+			if _, err := consumer.Commit(); err != nil {
+				log.CliLogger.Warnf("Failed to commit current consumer offset: %v", err)
+			}
 			consumer.Close()
```
MuweiHe, line 252: @brianstrauch defer wouldn't work as consumer has been closed and connection is gone. so i'm adding a `commit` with timeout on line 262, as `commit` would hang if there's no connection.

internal/cmd/kafka/confluent_kafka.go
```diff
@@ -255,6 +259,20 @@ func RunConsumer(consumer *ckafka.Consumer, groupHandler *GroupHandler) error {
...
+					go func() {
+						_, err := consumer.Commit()
+						commitErrCh <- err
+					}()
+					select {
+					case commitErr := <-commitErrCh:
+						if commitErr != nil {
+							log.CliLogger.Warnf("Failed to commit current consumer offset: %v", commitErr)
+						}
+					case <-time.After(5 * time.Second): // Timeout in case consumer has lost connection to kafka and commit would hang
```
brianstrauch, line 272: Line is way too long
```suggestion
					// Time out in case consumer has lost connection to Kafka and commit would hang
					case <-time.After(5 * time.Second):
```

brianstrauch (approved):
Looks like we either commit on Ctrl+C, or on an error. As long as that covers 100% of the cases, +1

PR #2155, opened 2023-08-02
Deprecate "version" field in configuration file

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The configuration file version is no longer being used (all usages were along dead code paths), so we can mark it as "omitempty" just so it can still be deserialized. I changed the type to "string" to avoid any potential panics that could occur, out of an abundance of caution.

In a follow-up PR, I'm planning to rename the `config/v1` package to `config`

Test & Review
-------------
Tests still pass


sgagniere (approved):
👍

PR #2156, opened 2023-08-02
CLI automatically uses LD test env when not logged into prod

kevin-wu24:
Checklist
---------
   * yes: ok

What
----
Now the CLI automatically uses the LaunchDarkly Test environment to fetch feature flags when the user is not logged into the prod environment. Setting the `XX_LAUNCH_DARKLY_TEST_ENV` variable is now no longer needed.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2624

Test & Review
-------------

### DIFF
--- a/internal/pkg/cmd/prerunner_test.go
+++ b/internal/pkg/cmd/prerunner_test.go
@@ -123,7 +123,8 @@ func getPreRunBase() *pcmd.PreRun {
 }
 
 func TestPreRun_Anonymous_SetLoggingLevel(t *testing.T) {
-	featureflags.Init(nil, true, false)
+	cfg := &v1.Config{Version: nil, IsTest: true, DisableFeatureFlags: false}
+	featureflags.Init(cfg)
 
 	tests := map[string]log.Level{
 		"":      log.ERROR,
--- a/internal/pkg/cmd/prerunner_test.go
+++ b/internal/pkg/cmd/prerunner_test.go
@@ -123,7 +123,7 @@ func getPreRunBase() *pcmd.PreRun {
 }
 
 func TestPreRun_Anonymous_SetLoggingLevel(t *testing.T) {
-	cfg := &v1.Config{Version: nil, IsTest: true, DisableFeatureFlags: false}
+	cfg := &v1.Config{IsTest: true, Contexts: map[string]*v1.Context{}}
 	featureflags.Init(cfg)
 
 	tests := map[string]log.Level{
--- a/internal/pkg/featureflags/feature_flags.go
+++ b/internal/pkg/featureflags/feature_flags.go
@@ -31,9 +31,10 @@ const (
 	userPath = "users/%s"
 )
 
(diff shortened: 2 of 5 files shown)


internal/pkg/cmd/prerunner_test.go
```diff
@@ -123,7 +123,8 @@ func getPreRunBase() *pcmd.PreRun {
 }
 
 func TestPreRun_Anonymous_SetLoggingLevel(t *testing.T) {
-	featureflags.Init(nil, true, false)
+	cfg := &v1.Config{Version: nil, IsTest: true, DisableFeatureFlags: false}
```
brianstrauch, line 126:
```suggestion
	cfg := &v1.Config{IsTest: true}
```

internal/pkg/featureflags/feature_flags.go
```diff
@@ -34,6 +33,7 @@ const (
 
 const (
 	cliProdEnvClientId     = "61af57740127630ce47de5be"
+	cliProdEnv             = "confluent.cloud"
```
brianstrauch, line 36: Let's move this to a different `const` block than the client IDs?

internal/pkg/featureflags/feature_flags.go
```diff
@@ -57,16 +57,25 @@ type launchDarklyManager struct {
 	version               *version.Version
 }
 
-func Init(version *version.Version, isTest, isDisabledConfig bool) {
+func Init(cfg *v1.Config) {
+	var platformName string
+	if !cfg.IsTest {
```
brianstrauch, line 62: I don't think we need to add this `if`. If it's a test, we can retrieve the platform name, but we just won't use it.

internal/pkg/featureflags/feature_flags.go
```diff
@@ -57,16 +57,25 @@ type launchDarklyManager struct {
...
+		currentContext, err := cfg.FindContext(cfg.CurrentContext)
+		if err != nil {
+			log.CliLogger.Warnf("Current context %s not found", cfg.CurrentContext)
+		} else {
+			platformName = currentContext.PlatformName
+		}
```
brianstrauch, line 68: can't we just do `cfg.Context.GetPlatformName()` here?

internal/pkg/featureflags/feature_flags.go
kevin-wu24: Without this `if` this function will panic when called in `test_helpers.go` and `prerunner_test.go` since we make the config that is passed into `Init`. I'm not sure if it's worth it to populate the test configs of these usages with a fake context map since it's not being used in a test anyways.

internal/pkg/featureflags/feature_flags.go
kevin-wu24: `cfg.Context()` and then calling another method isn't panic-safe if the user edits their config.

brianstrauch (approved):

PR #2157, opened 2023-08-03
[CLI-2673] Fix cross-compilation for linux

brianstrauch:
Checklist
---------
   * yes: ok

What
----
`go install` should always use the GOOS and GOARCH of the current system, so default to `""` for both.


sgagniere (approved):
Thanks; this was a very annoying error

PR #2158, opened 2023-08-03
[CLI-2627] Clean up feature flag code and mock ccloud flags in test

kevin-wu24:
Checklist
---------
   * yes: ok

What
----
Cleaning up feature flag code. Added Ccloud feature flags to LaunchDarkly mock

I'm not sure if we should merge the two clients, as they do go to separate endpoints outside of testing. I actually think passing in an `enum` value corresponding to the client is more clean than explicitly building the `Sling` object every time flags are fetched.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2627

Test & Review
-------------
feature flag unit tests and integration tests


test/test-server/ccloud_handlers.go
```diff
@@ -377,6 +377,9 @@ func handleLaunchDarkly(t *testing.T) http.HandlerFunc {
 			"cli.deprecation_notices":                []map[string]any{},
 			"cli.client_quotas.enable":               true,
 			"cli.stream_designer.source_code.enable": true,
+			"flink.rbac.namespace.cli.enable":        true,
+			"auth.rbac.identity_admin.enable":        false,
```
brianstrauch, line 381: Should we make this true so we can test it?

internal/pkg/featureflags/feature_flags.go
```diff
@@ -31,15 +30,18 @@ const (
 	userPath = "users/%s"
 )
 
-const cliProdEnv = "confluent.cloud"
+const (
+	ccloudUrl = "https://confluent.cloud"
+	devel     = "devel.cpdev.cloud"
+	stag      = "stag.cpdev.cloud"
```
brianstrauch, line 36: How come these don't start with https:// but `ccloudUrl` does?

internal/pkg/featureflags/feature_flags.go
```diff
@@ -59,42 +61,44 @@ type launchDarklyManager struct {
...
-			ccloudBasePath = fmt.Sprintf(baseURL, auth.CCloudURL, ccloudProdEnvClientId)
+		ccloudBasePath = cliBasePath
+	} else {
+		switch cfg.Context().GetPlatform().GetName() {
+		case devel:
+			ccloudBasePath = fmt.Sprintf(baseURL, ccloudUrl, ccloudDevelEnvClientId)
+			cliBasePath = fmt.Sprintf(baseURL, ccloudUrl, cliTestEnvClientId)
+		case stag:
+			ccloudBasePath = fmt.Sprintf(baseURL, ccloudUrl, ccloudStagEnvClientId)
+			cliBasePath = fmt.Sprintf(baseURL, ccloudUrl, cliTestEnvClientId)
```
brianstrauch, line 76: There's a better way to write this that doesn't involve duplicate code

test/test-server/ccloud_handlers.go
kevin-wu24: I just copied the value from `cpd`, since that is where we were originally fetching the flag from.

internal/pkg/featureflags/feature_flags.go
kevin-wu24: I'm going to change this. I was trying to get all of our launchdarkly code into the featureflags pkg, but that caused circular imports.

test/test-server/ccloud_handlers.go
brianstrauch: got it... we should probably still make it true so we can have the proper test coverage. if it breaks tests, let's make a separate pr where we make this true and fix the tests

internal/pkg/featureflags/feature_flags.go
kevin-wu24: Wouldn't we always want to default to `prod`

test/fixtures/output/iam/rbac/role/list-json-cloud.golden
```diff
@@ -756,6 +756,327 @@
       }
     ]
   },
+  {
```
brianstrauch, line 759: This is testing the same code as the YAML version of this file (it's adding no new coverage). Let's delete this test?

brianstrauch (approved):
Thank you!!

PR #2160, opened 2023-08-04
[CLI-2674] Panic: confluent api-key store v3.25.1 

kevin-wu24:
Release Notes
-------------
Bug Fixes
- fixing panic with `api-key store --resource`

Checklist
---------
   * yes: ok

What
----
Previous change to this line was not using panic-safe `GetSpec()` method.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2674

### DIFF
--- a/internal/cmd/api-key/command_store.go
+++ b/internal/cmd/api-key/command_store.go
@@ -111,7 +111,7 @@ func (c *command) store(cmd *cobra.Command, args []string) error {
 		return errors.CatchApiKeyForbiddenAccessError(err, getOperation, httpResp)
 	}
 
-	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.Spec.Resource.GetId()
+	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.GetSpec().Resource.GetId()
 
 	if !apiKeyIsValidForTargetCluster {
 		return errors.NewErrorWithSuggestions(errors.APIKeyNotValidForClusterErrorMsg, errors.APIKeyNotValidForClusterSuggestions)


brianstrauch (approved):

internal/cmd/api-key/command_store.go
```diff
@@ -111,7 +111,7 @@ func (c *command) store(cmd *cobra.Command, args []string) error {
 		return errors.CatchApiKeyForbiddenAccessError(err, getOperation, httpResp)
 	}
 
-	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.Spec.Resource.GetId()
+	apiKeyIsValidForTargetCluster := cluster.ID == apiKey.GetSpec().Resource.GetId()
```
brianstrauch, line 114: Wait... if `apiKey` is `nil`, then `apiKey.GetSpec()` will also be `nil`, and we'll still panic on `apiKey.GetSpec().Resource`.

brianstrauch (changes requested):

internal/cmd/api-key/command_store.go
brianstrauch: Our backend services have been known to hit an internal error and return `{}` as the response, so `apiKey == nil` is definitely possible.

kevin-wu24: GetSpec()` would prevent a panic when dereferencing `.Resource` in the case `apiKey == nil` occurs but the request succeeds.

kevin-wu24: It looks like `cluster` can be `nil` from this line without causing the function to return an error
https://github.com/confluentinc/cli/blob/a4f2f2cb52c363c4fbfff1238df3ce7edc6d88d6/internal/pkg/dynamic-config/dynamic_context.go#L110

MuweiHe (approved):

PR #2161, opened 2023-08-04
Tests should not write to actual `config` file

kevin-wu24:
Checklist
---------
   * yes: ok

What
----
Updating unit tests so that they do not create a persistent context in the user's config file.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2672

Test & Review
-------------
manually tested


internal/pkg/featureflags/feature_flags.go
```diff
@@ -322,13 +322,11 @@ func writeFlagsToConfig(ctx *dynamicconfig.DynamicContext, key string, vals map[
...
-func TryInvalidateCache(cfg *v1.Config, url string) error {
-	if cfg.CurrentContext == "" || strings.Contains(cfg.CurrentContext, url) {
-		return nil
-	}
-	if oldContext, _ := cfg.FindContext(cfg.CurrentContext); oldContext != nil {
-		oldContext.FeatureFlags = nil
-		if err := oldContext.Save(); err != nil {
+func ClearTestContextCache(cfg *v1.Config) error {
+	platformName := cfg.Context().GetPlatform().GetName()
+	if platformName == testserver.TestCloudUrl.String() {
```
brianstrauch, line 326: Can we just write our tests in such a way that they don't interfere with the user's configuration file?

brianstrauch: Most of our tests create a temporary HOME directory where the configuration file can live. Can we track down the specific test that's adding this platform name and make sure it's not messing with your configuration file.

brianstrauch (changes requested):

internal/pkg/featureflags/feature_flags.go
kevin-wu24: Yeah that sounds like a better approach. I thought this bug was affecting non-test contexts but it seems like it's just the mocked contexts.

kevin-wu24: It looks like this is only the case for integration tests: https://github.com/confluentinc/cli/blob/f3d163b0ae2d6839c968af628eac20587af11cc6/test/cli_test.go#L96 
I suspect that the `login` unit tests are what are writing to the actual `config`

kevin-wu24: The temporary HOME directory is only set for integration tests. I've added it to the relevant unit tests that I could find, but I've ran into some issues with this suggested approach.

It seems there are multiple unit tests that mess with the config file. It's pretty difficult to pinpoint all of them (I can't find a shared entry point like with integration tests), so a context with name `login-prompt-user@confluent.io-http://localhost:8090` still persists after running `make test`. 

What do you think about just clearing out the contexts after the unit tests run? What happens now is that only a single `test` context persists in the `config` anyways after running tests, so this would have the same affect as not writing the `test` context.

brianstrauch: > It's pretty difficult to pinpoint all of them

Try running tests individually and see if they change the configuration file?

kevin-wu24: Yeah I guess there's no way around it

internal/pkg/config/v1/mock.go
```diff
@@ -220,3 +222,11 @@ func setUpConfig(conf *Config, ctx *Context, platform *Platform, credential *Cre
 		panic(err)
 	}
 }
+
+func SetTempHomeDir() {
+	// Temporarily change $HOME, so the current config file isn't altered.
```
brianstrauch, line 227:
```suggestion
	// Temporarily change $HOME, so the current configuration file isn't altered.
```

internal/pkg/config/v1/mock.go
```diff
@@ -220,3 +222,11 @@ func setUpConfig(conf *Config, ctx *Context, platform *Platform, credential *Cre
 		panic(err)
 	}
 }
+
+func SetTempHomeDir() {
```
brianstrauch, line 226: Are there any other areas of the codebase that could use this function? Run a global search for `os.Setenv("HOME"`?

MuweiHe:
do you need to revert `$HOME` after running tests?

kevin-wu24:
> do you need to revert `$HOME` after running tests?

No, since `os.Setenv()` only changes the environment variable in the current process and not the parent process, when `make test` exits, `$HOME` will remain what it was before tests were run.
https://stackoverflow.com/questions/24938877/how-to-set-environment-variables-that-last

brianstrauch (approved):
Thanks!!

PR #2162, opened 2023-08-04
Add plaintext support for `kafka topic consume` with unauthenticated Schema Registry

brianstrauch:
Release Notes
-------------
New Features
- Support `--protocol PLAINTEXT` in `confluent kafka topic consume`
- Enable Schema Registry to be used without authentication on-premises, for the purpose of demos 

Checklist
---------
   * yes: ok

What
----
Both of these tasks are needed before we can use `confluent` in https://developer.confluent.io/tutorials/create-stateful-aggregation-count/flinksql.html, instead of `kafka-avro-console-consumer` and `kafka-configs`.

Test & Review
-------------
Manually tested in demo.

### DIFF
--- a/internal/pkg/cmd/authenticated_cli_command.go
+++ b/internal/pkg/cmd/authenticated_cli_command.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"net/http"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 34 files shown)


internal/pkg/cmd/authenticated_cli_command.go
```diff
@@ -198,15 +201,15 @@ func (c *AuthenticatedCLICommand) GetSchemaRegistryClient() (*schemaregistry.Cli
...
 				if err != nil {
 					return nil, err
 				}
 
-				c.schemaRegistryClient = schemaregistry.NewClient(configuration, dataplaneToken)
+				c.schemaRegistryClient = schemaregistry.NewClientWithToken(configuration, dataplaneToken)
 			} else {
 				// Used by `asyncapi export`, `asyncapi import`, `kafka client-config create`, `kafka topic consume`, and `kafka topic produce`
-				schemaRegistryEndpoint, err := c.Flags().GetString("schema-registry-endpoint")
+				schemaRegistryEndpoint, err := cmd.Flags().GetString("schema-registry-endpoint")
```
sgagniere, line 212: Why do we need to pass in `cmd` instead of using `c`?

sgagniere (approved):
+1, with small question

internal/pkg/cmd/authenticated_cli_command.go
brianstrauch: `c` is created in `kafka topic` which doesn't have `--schema-registry-endpoint`, but `cmd` refers to `kafka topic consume`, which does.

PR #2164, opened 2023-08-04
Fix api secret decryption issue on Windows

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix authentication issue in `confluent kafka topic [produce | consume]` on Windows

Checklist
---------
   * yes: ok

What
----
Encrypted api secrets don't have an `AesGcm` prefix, so we shouldn't check for that before decrypting secrets on Windows.

References
----------

Test & Review
-------------
Manual testing on Windows, Linux, and MacOS.

Open Questions / Follow-ups
---------------------------
Next: fix an issue on Windows with `confluent kafka topic produce` where `Ctrl-C` causes a panic and `Ctrl-D` does nothing. In other words, the only way to exit out of `consume` on Windows is to panic.

This doesn't appear to be harmful, but we definitely want to exit gracefully.

### DIFF
--- a/internal/pkg/config/v1/api_key_pair.go
+++ b/internal/pkg/config/v1/api_key_pair.go
@@ -16,7 +16,7 @@ type APIKeyPair struct {
 }
 
 func (c *APIKeyPair) DecryptSecret() error {
-	if strings.HasPrefix(c.Secret, secret.AesGcm) && (c.Salt != nil || runtime.GOOS == "windows") {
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):

MuweiHe (approved):
thank you!

PR #2173, opened 2023-08-08
[CLI-788] Update CLI version in muckrake during release

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Discussed with the quality eng team, and agreed we should transition to testing the CLI against CP in nightly muckrake runs. They no longer have the capacity to keep .x branches stable, so our tests will be run nightly on the branch builder tool. This matches the testing strategy of our Confluent Cloud tests, although those run more frequently (every hour).

Example PR: https://github.com/confluentinc/muckrake/pull/2081

Test & Review
-------------
Ran with `DRY_RUN=true`.


mk-files/release.mk
```diff
@@ -148,3 +150,32 @@ download-licenses:
...
+.PHONY: update-muckrake
+update-muckrake:
+	$(eval DIR=$(shell mktemp -d))
+	$(eval CLI_RELEASE=$(DIR)/cli-release)
+	$(eval MUCKRAKE=$(DIR)/muckrake)
+
+	git clone git@github.com:confluentinc/cli-release.git $(CLI_RELEASE) && \
+	cd $(CLI_RELEASE) && \
+	version=$$(ls release-notes | $(SED) "s/.json$$//" | sort --version-sort | tail -1) && \
+	git clone git@github.com:confluentinc/muckrake.git $(MUCKRAKE) && \
```
MuweiHe, line 163: muckrake is being cloned under cli-release?

brianstrauch: Yup, we need to clone it before we can open a PR.

MuweiHe: i'm confused it looks like muckrake is being clones inside of cli-release, but from `$(eval CLI_RELEASE=$(DIR)/cli-release) $(eval MUCKRAKE=$(DIR)/muckrake)` they're at the same level of dir

brianstrauch: The second argument to `git clone` is the directory that we're cloning into!

MuweiHe: oh haha

MuweiHe (approved):
thanks!

PR #2174, opened 2023-08-09
Remove extraneous GET requests in `confluent api-key list`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
When listing api keys with `confluent api-key list`, we call `c.getEmail` for every single key.

This function first checks if the user ID belongs to a service account. If it doesn't, then this function will next check if it belongs to an auditlog service account, and it does this by making a request to `/api/me`. It then finally checks if the ID belongs to a user account.

The problem here is that this will make a redundant http request for every key attached to a user, which can be quite a lot. A simple fix is to just move the user account check ahead of the auditlog check, but that still leaves open the possibility of calling this endpoint multiple times.

So this PR moves the request into its own `c.getAuditLog` method, calls it before the loop, and passes the audit log into `c.getEmail`.

References
----------

Test & Review
-------------
Verified manually with `--unsafe-trace`that the loop no longer produces a bunch of GET requests to `/api/me`.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/api-key/command_list.go
+++ b/internal/cmd/api-key/command_list.go
@@ -105,6 +105,8 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
 	serviceAccountsMap := getServiceAccountsMap(serviceAccounts)
 	usersMap := getUsersMap(allUsers)
 
+	auditLog := c.getAuditLog()
+
 	list := output.NewList(cmd)
 	for _, apiKey := range apiKeys {
 		// ignore keys owned by Confluent-internal user (healthcheck, etc)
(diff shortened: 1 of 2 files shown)


internal/cmd/api-key/command_list.go
```diff
@@ -164,17 +165,15 @@ func mapResourceIdToUserId(users []*ccloudv1.User) map[string]int32 {
...
 		return "<service account>"
 	}
 
 	userId := resourceIdToUserIdMap[resourceId]
 
-	if user, err := c.Client.Auth.User(); err == nil {
-		if auditLog := user.GetOrganization().GetAuditLog(); auditLog != nil && auditLog.GetServiceAccountId() == userId {
-			return "<auditlog service account>"
-		}
+	if auditLog != nil && auditLog.GetServiceAccountId() == userId {
```
brianstrauch, line 175: Why pass the whole audit log struct if we just need the audit log service account ID?

brianstrauch (approved):

internal/cmd/api-key/command_list.go
sgagniere: Right, good point. I'll make this change tomorrow morning before I merge.

MuweiHe (approved):

PR #2177, opened 2023-08-09
[CLI-2642] Resolve panic when exiting `confluent kafka topic produce`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Resolve a panic that can happen when exiting `confluent kafka topic produce` with Ctrl-C

Checklist
---------
   * yes: ok

What
----
When the user enters `Ctrl-C`, this input is occasionally handled by the scanner goroutine before the signals goroutine. The consequence is that the former will close the `input` channel. Afterwards, the `signals` channel is unblocked and the second goroutine will then attempt to close `input` again, resulting in a double close panic.

This PR uses a `select` block in the signals goroutine to check if `input` is closed, and only closes it if it is not already closed.

Additionally, since `Ctrl-D` is not a valid way to exit the command prompt in Windows, I've removed it from the exiting instructions on Windows.

References
----------

Test & Review
-------------
Manual testing in Windows (where I was able to reproduce the problem consistently) and MacOS.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/kafka/command_topic_produce.go
+++ b/internal/cmd/kafka/command_topic_produce.go
@@ -117,7 +117,11 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	output.ErrPrintln(errors.StartingProducerMsg)
+	if runtime.GOOS == "windows" {
+		output.ErrPrintln("Starting Kafka Producer. Use Ctrl-C to exit.")
+	} else {
+		output.ErrPrintln(errors.StartingProducerMsg)
+	}
 
 	var scanErr error
 	input, scan := PrepareInputChannel(&scanErr)
@@ -127,7 +131,11 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 	signal.Notify(signals, os.Interrupt)
 	go func() {
 		<-signals
-		close(input)
+		select {
+		case <-input:
+		default:
+			close(input)
+		}
 	}()
 	// Prime reader
 	go scan()
--- a/internal/cmd/kafka/command_topic_produce.go
+++ b/internal/cmd/kafka/command_topic_produce.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"runtime"
 	"strconv"
 	"strings"
 


internal/cmd/kafka/command_topic_produce.go
```diff
@@ -117,7 +118,11 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	output.ErrPrintln(errors.StartingProducerMsg)
+	if runtime.GOOS == "windows" {
+		output.ErrPrintln("Starting Kafka Producer. Use Ctrl-C to exit.")
```
brianstrauch, line 122: Can we write a format string here and replace "Ctrl-C" with "%s"?

internal/cmd/kafka/command_topic_produce.go
```diff
@@ -127,7 +132,11 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 	signal.Notify(signals, os.Interrupt)
 	go func() {
 		<-signals
-		close(input)
+		select {
+		case <-input:
+		default:
+			close(input)
```
brianstrauch, line 138: Is it still possible to have a double close? I thought the solution would involve removing one of the `close()` functions.

sgagniere: It might be possible with very precise timing.

And both of these `close()` calls appear to be necessary without a larger refactor: without closing the input in the signal goroutine, the scanner goroutine treats `Ctrl-C` like a regular empty line on Unix, which means that it doesn't exit (on Windows, the scanner apparently treats `Ctrl-C` like EOF so it would still end there).

Without closing the input in the scanner goroutine, `Ctrl-D` no longer exits on Unix since it isn't an interrupt signal.

The easiest solution for now is to just repeat this check for all `close` calls.

sgagniere: Nevermind; that idea simply introduces its own possibilities of double closure panic.

At this point I'm just going to resell this as fixing a panic that occurs in `produce` for Windows (since it happens nearly 100% of the time as far as I can tell) and leave the CLI-2642 ticket open.

brianstrauch (approved):

PR #2178, opened 2023-08-09
Support configuration files and configuration key=value pairs with `--config`

brianstrauch:
Release Notes
-------------
Breaking Changes
- PLACEHOLDER

New Features
- Add support for both strings and files for the `--config` flag, in all relevant `confluent kafka` and `confluent schema-registry` commands

Checklist
---------
   * yes: ok

What
----
Certain commands only supported comma-delimited strings (`--config`) and others only supported files (`--config-file`). Combine both into `--config` and use everywhere.

Test & Review
-------------
Add missing integration tests where needed

### DIFF
--- a/internal/cmd/kafka/command_cluster_configuration_update.go
+++ b/internal/cmd/kafka/command_cluster_configuration_update.go
@@ -50,6 +50,7 @@ func (c *clusterCommand) configurationUpdate(cmd *cobra.Command, _ []string) err
 		return err
 	}
 
+	// Deprecated
 	configFile, err := cmd.Flags().GetString(configFileFlagName)
 	if err != nil {
 		return err
--- a/internal/cmd/kafka/command_cluster_configuration_update.go
+++ b/internal/cmd/kafka/command_cluster_configuration_update.go
@@ -29,7 +29,7 @@ func (c *clusterCommand) newConfigurationUpdateCommand() *cobra.Command {
 		),
 	}
 
-	cmd.Flags().StringSlice("config", nil, `A comma-separated list of configuration overrides with form "key=value".`)
+	pcmd.AddConfigFlag(cmd)
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddClusterFlag(cmd, c.AuthenticatedCLICommand)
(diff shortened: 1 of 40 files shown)


internal/cmd/kafka/command_cluster_configuration_update.go
```diff
@@ -50,7 +50,16 @@ func (c *clusterCommand) configurationUpdate(cmd *cobra.Command, _ []string) err
 		return err
 	}
 
-	configMap, err := properties.ConfigFlagToMap(config)
+	// Deprecated
+	configFile, err := cmd.Flags().GetString(configFileFlagName)
```
sgagniere: `config-file` doesn't look like a flag for this command.

internal/pkg/properties/properties.go
```diff
@@ -6,10 +6,21 @@ import (
...
 
-// FileToMap reads key=value pairs from a properties file, ignoring comments and empty lines.
-func FileToMap(filename string) (map[string]string, error) {
+// GetMap reads newline-separated configuration files or comma-separated lists of key=value pairs, and supports configuration values containing commas.
+func GetMap(config []string) (map[string]string, error) {
+	if len(config) == 1 && utils.FileExists(config[0]) {
+		return fileToMap(config[0])
+	}
+
+	return ConfigFlagToMap(config)
```
kevin-wu24, line 19: What happens if a user passes in a config file name that doesn't exist, but has a `=` in it (i.e. `config[0] = "temp=config.json"`)? Seems like we'll actually convert this into a `map[string]string` without erroring, but we should throw a file doesn't exist error.

sgagniere (approved):

internal/pkg/properties/properties.go
sgagniere: Hmm, I would expect a backend error in this case.

kevin-wu24: Yeah, looks like for the current usages of `GetMap` this should be fine (I still think we should error if the file isn't found found so we don't make unnecessary API requests).

If we decide to bring back `config` commands and have the `set` command accept `key=value` pairs as arguments to mirror this syntax, this bug will probably resurface anyways.

internal/cmd/schema-registry/command_exporter_create.go
```diff
@@ -48,6 +48,11 @@ func (c *command) newExporterCreateCommand(cfg *v1.Config) *cobra.Command {
 	}
 	pcmd.AddOutputFlag(cmd)
 
+	// Deprecated
+	cmd.Flags().String("config-file", "", "Exporter configuration file.")
```
MuweiHe, line 52: may use the description `"Name of a file with additional topic configuration. Each property should be on its own line with the format: key=value."` here as well to be more informative

brianstrauch: This flag is deprecated (hidden), so users won't see this description. I just copied the original declaration from above down here.

PR #2182, opened 2023-08-10
Use go1.21

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Use new built-in functions and `slices` package. Plus, a speed boost in build times!

Test & Review
-------------
Tests still pass


brianstrauch:
(Looks like go1.21 isn't supported by Semaphore yet)

sgagniere (approved):
Very nice, it looks much cleaner now!

Is the lack of 1.21 support in Semaphore a blocker for merging this PR?

brianstrauch:
> Is the lack of 1.21 support in Semaphore a blocker for merging this PR?

Yeah, usually it takes a few weeks for them to add it, IIRC

pkg/types/slice.go
```diff
@@ -17,8 +17,3 @@ func Remove(haystack []string, needle string) []string {
 	}
```
kevin-wu24, line 17: `slices` has a `Contains` and `DeleteFunc` functions, so we can remove the `Contains` and `Remove` functions above.

pkg/utils/utils_test.go
```diff
@@ -19,12 +18,12 @@ const (
...
 func TestContains(t *testing.T) {
 	req := require.New(t)
-	req.True(types.Contains([]string{"a"}, "a"))
+	req.True(slices.Contains([]string{"a"}, "a"))
 }
 
 func TestDoesNotContain(t *testing.T) {
 	req := require.New(t)
-	req.False(types.Contains([]string{}, "a"))
+	req.False(slices.Contains([]string{}, "a"))
```
kevin-wu24, line 26: I don't think we need these tests anymore.

pkg/panic-recovery/panic_recovery.go
```diff
@@ -3,11 +3,11 @@ package panic_recovery
 import (
 	"runtime"
 	"runtime/debug"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"golang.org/x/exp/slices"
```
kevin-wu24, line 10: The `configuration/command.go` that I merged today also import this, so if you could replace that too that'd be nice!

pkg/utils/utils.go
```diff
@@ -13,13 +13,6 @@ import (
...
-func Max(x, y int64) int64 {
-	if x > y {
-		return x
-	}
-	return y
-}
-
 func TestEq(a, b []string) bool {
 	// If one is nil, the other must also be nil.
 	if (a == nil) != (b == nil) {
```
kevin-wu24, line 18: `slices` has a `Equal` function now, so we can remove this.

pkg/types/slice.go
brianstrauch: Thanks, I deleted these but they got re-added in a bad merge 🤦‍♂️

pkg/utils/utils.go
brianstrauch: Nice find!

kevin-wu24 (approved):

PR #2185, opened 2023-08-11
Recover from panic when opening dev/ttv and format error

kevin-wu24:
Release Notes
-------------
Bug Fixes
- Resolve `confluent flink shell` panic when `/dev/tty` is inaccessible

Checklist
---------
   * yes: ok

What
----
When starting a flink application via the `confluent flink shell` command, [this function](https://github.com/confluentinc/go-prompt/blob/master/input_posix.go#L69) can panic if the file `/dev/tty` is inaccessible (e.g. running cli in containerized environment, or cli doesn’t have permission). Since the panicking function is used outside of CLI source, we should `recover()` and format an error for the user.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2683

### DIFF
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -124,7 +124,7 @@ func (a *Application) panicRecovery() {
 
 func (a *Application) isAuthenticated() bool {
 	if authErr := a.refreshToken(); authErr != nil {
-		utils.OutputErrf("Error: %v", authErr)
+		utils.OutputErrf("Error: %v\n", authErr)
 		a.appController.ExitApplication()
 		return false
 	}
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -53,7 +53,7 @@ func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() er
 	stdinBefore := utils.GetStdin()
 	consoleParser := utils.GetConsoleParser()
 	if consoleParser == nil {
-		utils.OutputErr("Error: failed to set initialize console parser")
+		utils.OutputErr("Error: failed to initialize console parser")
 		return
 	}
 	appController.AddCleanupFunction(func() {
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -52,6 +52,10 @@ func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() er
 
 	stdinBefore := utils.GetStdin()
 	consoleParser := utils.GetConsoleParser()
+	if consoleParser == nil {
+		utils.OutputErr("Error: failed to set initialize console parser")
+		return
+	}
 	appController.AddCleanupFunction(func() {
 		utils.TearDownConsoleParser(consoleParser)
 		utils.RestoreStdin(stdinBefore)
--- a/internal/pkg/flink/app/application.go
+++ b/internal/pkg/flink/app/application.go
@@ -124,7 +124,7 @@ func (a *Application) panicRecovery() {
 
 func (a *Application) isAuthenticated() bool {
 	if authErr := a.refreshToken(); authErr != nil {
-		utils.OutputErrf("Error: %v\n", authErr)
+		utils.OutputErrf("Error: %v", authErr)
 		a.appController.ExitApplication()
 		return false
 	}
--- a/internal/pkg/flink/internal/history/history.go
+++ b/internal/pkg/flink/internal/history/history.go
@@ -34,11 +34,11 @@ func loadFromPath(history *History) *History {
 	}
 
 	if err != nil {
-		log.CliLogger.Warnf("Couldn't load past statements history: unable to read file Error: " + err.Error())
+		log.CliLogger.Warnf("Couldn't load past statements history: unable to read file Error: %v", err)
 	}
 
 	if err := json.Unmarshal(jsonFile, history); err != nil {
-		log.CliLogger.Warnf("Couldn't load past statements history. Error: " + err.Error())
+		log.CliLogger.Warnf("Couldn't load past statements history. Error: %v", err)
 	}
 	return history
 }
@@ -46,7 +46,7 @@ func loadFromPath(history *History) *History {
 func initPath() *History {
 	home, osHomedirErr := os.UserHomeDir()
 	if osHomedirErr != nil {
-		log.CliLogger.Warnf("Couldn't get homedir with os.UserHomeDir(). Error: " + osHomedirErr.Error())
+		log.CliLogger.Warnf("Couldn't get homedir with os.UserHomeDir(). Error: %v", osHomedirErr)
 		return nil
 	}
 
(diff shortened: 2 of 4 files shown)


internal/pkg/flink/internal/utils/input.go
```diff
@@ -20,6 +20,11 @@ func GetStdin() *term.State {
 }
 
 func GetConsoleParser() prompt.ConsoleParser {
+	defer func() {
+		if r := recover(); r != nil {
+			log.CliLogger.Warnf("Couldn't open \"/dev/ttv\" file. Error: %v\n", r)
```
brianstrauch, line 25: Any way we can catch this **before** the panic occurs?

kevin-wu24: Oh yeah I guess we can check that this file 1. exists and 2. we have read permission before calling `Open`.

internal/pkg/flink/internal/utils/input.go
```diff
@@ -20,6 +20,10 @@ func GetStdin() *term.State {
 }
 
 func GetConsoleParser() prompt.ConsoleParser {
+	if fileInfo, err := os.Stat("/dev/tty"); err != nil || fileInfo.Mode().Perm()&0444 == 0 {
```
kevin-wu24, line 23: Should I split this up and specify the error for each situation?

internal/pkg/flink/internal/utils/input.go
```diff
@@ -20,11 +20,12 @@ func GetStdin() *term.State {
 }
 
 func GetConsoleParser() prompt.ConsoleParser {
-	defer func() {
-		if r := recover(); r != nil {
-			log.CliLogger.Warnf("Couldn't open \"/dev/ttv\" file. Error: %v\n", r)
-		}
-	}()
+	if fileInfo, err := os.Stat("/dev/tty"); err != nil {
+		log.CliLogger.Warnf("Couldn't open \"/dev/tty\" file because it doesn't exist. Error: %v\n", err)
```
brianstrauch:
```suggestion
		log.CliLogger.Warnf(`Couldn't open "/dev/tty" file because it doesn't exist. Error: %v`, err)
```

internal/pkg/flink/internal/utils/input.go
```diff
@@ -20,11 +20,12 @@ func GetStdin() *term.State {
...
+	} else if fileInfo.Mode().Perm()&0444 != 0 {
```
brianstrauch: Shouldn't this equal 0 if no read permissions are set? Also, why are we checking all three (user, group, other)?

kevin-wu24: It is equal 0 if no read permissions are set. I fixed that in the most recent commit.

My thinking in checking all 3 (user, group, other) is that the CLI could be containerized or on a shared machine where the user of the CLI may not be the owner of the `\dev\tty` file.

kevin-wu24: On docker:
```
f8ae59e12f7d:/$ ls dev/tty -l
crw-rw-rw-    1 root     root        5,   0 Aug 11 22:09 dev/tty
f8ae59e12f7d:/$ whoami
confluent
```

internal/pkg/flink/internal/utils/input.go
MuweiHe: splitting them up sounds like a better idea

internal/pkg/flink/internal/utils/input.go
```diff
@@ -13,25 +13,32 @@ import (
 func GetStdin() *term.State {
 	state, err := term.GetState(int(os.Stdin.Fd()))
 	if err != nil {
-		log.CliLogger.Warnf("Couldn't get stdin state with term.GetState. Error: %v\n", err)
+		log.CliLogger.Warn("Couldn't get stdin state with term.GetState. Error: " + err.Error())
```
brianstrauch, line 16: Why aren't we using `Warnf()` anymore?

internal/pkg/flink/app/application.go
```diff
@@ -52,6 +52,10 @@ func StartApp(client ccloudv2.GatewayClientInterface, tokenRefreshFunc func() er
 
 	stdinBefore := utils.GetStdin()
 	consoleParser := utils.GetConsoleParser()
+	if consoleParser == nil {
+		utils.OutputErr("Error: failed to initialize console parser")
```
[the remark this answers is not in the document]
kevin-wu24: I wanted to follow the style of how other `pkg/flink` code handles errors, since they just log them to `WARN`. Maybe we should clean that up in another PR?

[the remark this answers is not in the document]
kevin-wu24: We already log the error resulting in `consoleParser` being `nil` to WARN, but if nothing is printed out afterwards, the command will just silently exit.

Until the CLI prints out `WARN` messages by default, I think it would be nice for the user to know their `flink shell` command failed.

internal/pkg/flink/internal/utils/input.go
kevin-wu24: Apparently [`fmt.Sprintf` is slower than direct concatenation of strings](https://forum.golangbridge.org/t/fmt-sprintf-vs-string-concatenation/23006). 
Since concatenating here doesn't take away from the readability of the string, we should just use `Warn`.

internal/pkg/flink/app/application.go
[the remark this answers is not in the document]
brianstrauch: > Until the CLI prints out WARN messages by default

Got it. Can you open a ticket for this in that case?

internal/pkg/flink/internal/utils/input.go
brianstrauch: I don't think performance matters here, especially since the difference between the two is on the order of nanoseconds. Let's just try to standardize the format throughout the code as much as possible.

internal/pkg/flink/internal/history/history.go
```diff
@@ -73,24 +73,24 @@ func (history *History) Save() {
 	// Convert struct to JSON
 	b, err := json.Marshal(history)
 	if err != nil {
-		log.CliLogger.Warnf("Couldn't save past statements history: couldn't marhsal history. Error: " + err.Error())
+		log.CliLogger.Warnf("Couldn't save past statements history: couldn't marhsal history. Error: %v", err)
```
brianstrauch, line 76:
```suggestion
		log.CliLogger.Warnf("Couldn't save past statements history: couldn't marshal history. Error: %v", err)
```

internal/pkg/flink/internal/utils/input.go
```diff
@@ -13,25 +13,32 @@ import (
...
+		log.CliLogger.Warnf("Couldn't setup console parser. Error: %v", err)
 	}
 	return consoleParser
 }
 
 func TearDownConsoleParser(consoleParser prompt.ConsoleParser) {
 	err := consoleParser.TearDown()
 	if err != nil {
-		log.CliLogger.Warnf("Couldn't tear down console parser. Error: %v\n", err)
+		log.CliLogger.Warnf("Couldn't tear down console parser. Error: %v", err)
```
brianstrauch, line 41: It seems wrong to be writing "error" in a warning.
```suggestion
		log.CliLogger.Warnf("Couldn't tear down console parser: %v", err)
```

brianstrauch (approved):

PR #2186, opened 2023-08-11
[FF-11969] Add `--all` flag to `confluent kafka acl list` to include ACLs with only integer IDs

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `--all` to `confluent kafka acl list` to include deleted ACLs whose principals are integer IDs

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Some deleted principals do not have resource IDs, so the backend returns integer IDs even if we query specifically for resource IDs.

This PR adds a `--all` flag to show these, and otherwise hides them by default. Currently, the CLI is showing these by default which is not intended.

References
----------

Test & Review
-------------
Added a new integration test

Open Questions / Follow-ups
---------------------------


internal/pkg/acl/acl.go
```diff
@@ -317,86 +315,52 @@ func CreateAclRequestDataToAclData(data *AclRequestDataWithError) cpkafkarestv3.
...
-	prefix := x[0]
 	suffix := x[1]
 
-	// The principal has a resource ID
-	resources := []string{
-		resource.IdentityPool,
-		resource.ServiceAccount,
-		resource.User,
-	}
-	if types.Contains(resources, resource.LookupType(suffix)) {
```
MuweiHe, line 384: for those normal acl's we don't validate the resource type anymore?

MuweiHe: we might still wanna keep that. (it might just be github not showing the code change properly?

brianstrauch:
> Add --all to confluent kafka acl list to include ACLs which only have integer IDs

"... whose principals are integer IDs"

internal/cmd/kafka/command_acl_list.go
```diff
@@ -25,6 +25,7 @@ func (c *aclCommand) newListCommand() *cobra.Command {
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
 	cmd.Flags().String("principal", "", `Principal for this operation, prefixed with "User:".`)
+	cmd.Flags().Bool("all", false, "Include ACLs for principals with integer IDs. These ACLs are associated with deleted principals.")
```
brianstrauch, line 28:
```suggestion
	cmd.Flags().Bool("all", false, "Include ACLs for deleted principals with integer IDs.")
```

internal/pkg/acl/acl.go
```diff
@@ -317,86 +315,52 @@ func CreateAclRequestDataToAclData(data *AclRequestDataWithError) cpkafkarestv3.
...
-		list.Add(&out{
-			Principal:    prefix + ":" + resourceId,
-			Permission:   acl.Permission,
-			Operation:    acl.Operation,
-			ResourceType: string(acl.ResourceType),
-			ResourceName: acl.ResourceName,
-			PatternType:  acl.PatternType,
-		})
+func PrintACLsFromKafkaRestResponse(cmd *cobra.Command, acls []cckafkarestv3.AclData) error {
+	showAll, err := cmd.Flags().GetBool("all")
```
brianstrauch, line 319:
```suggestion
	all, err := cmd.Flags().GetBool("all")
```

internal/pkg/acl/acl.go
```diff
@@ -317,86 +315,52 @@ func CreateAclRequestDataToAclData(data *AclRequestDataWithError) cpkafkarestv3.
...
+			ResourceName: acl.GetResourceName(),
+			PatternType:  acl.GetPatternType(),
 		})
 	}
 	list.Filter(listFields)
 	return list.Print()
 }
 
-func getPrefixAndResourceIdFromPrincipal(principal string, numericIdToResourceId map[int32]string) (string, string, error) {
+func principalHasIntegerID(principal string) (bool, error) {
```
brianstrauch, line 347:
```suggestion
func principalHasIntegerId(principal string) (bool, error) {
```

internal/pkg/acl/acl.go
sgagniere: This is just a diff from reverting.

The main goal of this part was to determine if we needed to translate it or not (whether or not it had a resource id already). We don't want to do that anymore even for the leftover integer IDs being returned, so we don't need this check.

internal/cmd/kafka/command_acl_list.go
sgagniere: I like this.

It explains what the integer principals mean but stops short of implying that all of the deleted principals will have integer ids. Thanks.

internal/pkg/acl/acl.go
sgagniere: Also this whole diff disappeared after I fixed a merge conflict...

pkg/acl/acl.go
```diff
@@ -330,6 +344,25 @@ func PrintACLsFromKafkaRestResponse(cmd *cobra.Command, acls []cckafkarestv3.Acl
...
+	}
+
+	x := strings.Split(principal, ":")
+	if len(x) < 2 {
+		return false, errors.Errorf("unrecognized principal format %s", principal)
+	}
+	suffix := x[1]
+
+	// The principal has a numeric ID
+	if _, err := strconv.ParseInt(suffix, 10, 32); err == nil {
```
MuweiHe, line 359: if this is the only condition for returning true, this function actually only needs two branches: this one, and the unrecognized error one. all other circumstances will return false.

MuweiHe (approved):

pkg/acl/acl.go
```diff
@@ -315,10 +316,23 @@ func CreateAclRequestDataToAclData(data *AclRequestDataWithError) cpkafkarestv3.
...
+	all, err := cmd.Flags().GetBool("all")
+	if err != nil {
+		return err
+	}
+
 	list := output.NewList(cmd)
 	for _, acl := range acls {
+		principal := acl.GetPrincipal()
+		if !all {
+			if integerId, err := principalHasIntegerId(principal); err != nil {
```
brianstrauch, line 328: Let's make this clear it's a boolean
```suggestion
			if hasIntegerId, err := principalHasIntegerId(principal); err != nil {
```

pkg/acl/acl.go
```diff
@@ -330,6 +344,25 @@ func PrintACLsFromKafkaRestResponse(cmd *cobra.Command, acls []cckafkarestv3.Acl
...
+		return false, errors.Errorf("unrecognized principal format %s", principal)
+	}
+	suffix := x[1]
+
+	// The principal has a numeric ID
+	if _, err := strconv.ParseInt(suffix, 10, 32); err == nil {
+		return true, nil
+	}
+
+	return false, nil
```
brianstrauch, line 363:
```suggestion
	_, err := strconv.ParseInt(suffix, 10, 32)
	return err == nil, nil
```

test/test-server/kafka_rest_router.go
```diff
@@ -109,6 +109,9 @@ func handleKafkaRestClusters(t *testing.T) http.HandlerFunc {
 // Handler for: "/kafka/v3/clusters/{cluster}/acls"
 func handleKafkaRestACLs(t *testing.T) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		clusterName := vars["cluster"]
```
brianstrauch, line 113: Looks like this is a cluster ID if it's sometimes equal to "lkc-acls"
```suggestion
		clusterId := vars["cluster"]
```

brianstrauch (approved):

PR #2191, opened 2023-08-14
Use Schema Registry client in `kafka client-config`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Use the Schema Registry client for better error handling. This will help debug a flaky system test.


sgagniere (approved):

brianstrauch:
It looks like the Schema Registry API key may sometimes not be available for over a minute... waiting to merge until we get an answer about how much lag to expect from the Schema Registry team.

PR #2192, opened 2023-08-14
[CLI-2681] Adding `config` commands

kevin-wu24:
Release Notes
-------------
New Features
- Add `confluent configuration` commands for configuration file management

Checklist
---------
   * yes: ok

What
----
This PR adds the `describe`, `update`, and `list` subcommands for `confluent configuration`. 

References
----------
[1-pager](https://confluentinc.atlassian.net/wiki/spaces/DPI/pages/3141799632/1-Pager+Bring+back+confluent+config+commands)

Test & Review
-------------
Added integration and unit tests.


internal/cmd/configuration/command.go
```diff
@@ -0,0 +1,74 @@
...
+	Name     string `human:"Name" serialized:"name"`
+	Value    string `human:"Value" serialized:"value"`
+	ReadOnly bool   `human:"Read-Only" serialized:"read_only"`
+}
+
+func New(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "configuration",
+		Aliases: []string{"config"},
+		Short:   "Manage CLI configuration fields.",
```
brianstrauch, line 29:
```suggestion
		Short:   "Configure the Confluent CLI.",
```

internal/cmd/configuration/command.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	c := &command{
+		CLICommand: pcmd.NewAnonymousCLICommand(cmd, prerunner),
+		config:     cfg,
```
brianstrauch, line 34:
```suggestion
		cfg:     cfg,
```

internal/cmd/configuration/command.go
```diff
@@ -0,0 +1,74 @@
...
+
+func getJsonFieldName(field reflect.StructField) string {
+	jsonTag := field.Tag.Get("json")
+	if jsonTag == "-" {
+		return ""
+	}
+	if strings.Contains(jsonTag, ",") {
+		jsonTag, _, _ = strings.Cut(jsonTag, ",")
+	}
+	if runtime.GOOS != "windows" && jsonTag == "disable_plugins_once" {
```
brianstrauch, line 70: Any way we can get around hardcoding this specific case?

internal/cmd/configuration/command_set.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/examples"
+	"github.com/confluentinc/cli/internal/pkg/form"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+func (c *command) newSetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set <config-field-1=value-1> ... [config-field-n=value-n]",
```
brianstrauch, line 19: To match our other commands?
```suggestion
		Use:   "update <config-field-1=value-1> ... [config-field-n=value-n]",
```

test/configuration_test.go
```diff
@@ -0,0 +1,42 @@
...
+	if err := cfg.CreateContext("context2", "http://test", "costa", "rica"); err != nil {
+		panic(err)
+	}
+	if err := cfg.Save(); err != nil {
+		panic(err)
+	}
+
+	tests := []CLITest{
+		{args: "configuration set disable_update_check=true", fixture: "configuration/set-one.golden"},
+		{args: "configuration set disable_update_check=false disable_plugins=true current_context=context2", fixture: "configuration/set-multiple.golden"},
```
brianstrauch, line 33: Do you think editing multiple configurations is an actual use case?

test/fixtures/output/configuration/list-windows.golden
```diff
@@ -0,0 +1,20 @@
+          Name          |                       Value                        | Read-Only
+------------------------+----------------------------------------------------+------------
+  context_states        | map[string]*config.ContextState                    | true
+                        | with keys: [context1]                              |
+  contexts              | map[string]*config.Context                         | true
+                        | with keys: [context1]                              |
+  credentials           | map[string]*config.Credential with keys:           | true
+                        | [username-cli-mock-email@confluent.io-http://test] |
```
brianstrauch, line 8: I don't think users need to see this information.

test/fixtures/output/configuration/list-windows.golden
```diff
@@ -0,0 +1,20 @@
...
+  current_context       | context1                                           | false
+  disable_feature_flags | false                                              | false
+  disable_plugins       | false                                              | false
+  disable_plugins_once  | false                                              | false
+  disable_update_check  | false                                              | false
+  disable_updates       | false                                              | false
+  local_ports           |                                                    | true
```
brianstrauch, line 15: Looks like this is empty?

internal/cmd/configuration/command.go
kevin-wu24: I don't think so unfortunately. Hardcoding this case makes it so that `disable_plugins_once` doesn't show up in the output of `config list` and is not settable unless the user is running Windows. 
If a Mac/Unix user saw this in `config list`, they could reasonably infer that setting this flag would disable plugins one time, but in actuality setting it would do nothing.
Maybe we should instead append a `_windows` to this config field to avoid confusion?

internal/cmd/configuration/command_set.go
```diff
@@ -0,0 +1,119 @@
...
+					return err
+				} else if !ok {
+					continue
+				}
+			}
+			oldValue.SetBool(v)
+		case string:
+			oldValue.SetString(v)
+		}
+		updates = append(updates, fmt.Sprintf(errors.UpdateSuccessMsg, "value", "config field", keys[i], values[i]))
```
sgagniere, line 58: This seems like it has the potential to be misleading. If the user tries to update the `current_context` to a bad value, this line would report success.

But then we would error out on the `Validate()` all below since that function checks that the `current_context` is in the list of contexts.

Also, I think perhaps we should not let users update the context here since we already have `confluent context use`. Perhaps we could include a message directing them to use that instead?

test/fixtures/output/configuration/list-windows.golden
kevin-wu24: The mock `config` I'm using for tests doesn't have `local` setup. I have to look at `confluent local` to see if there's a mock or just make one myself.

internal/cmd/configuration/command_set.go
kevin-wu24: `updates` only print out if the `Validate()` and `Save()` functions return normally.

Yeah, I think if they attempt to set `current_context`, I'll just direct them to `context use`.

sgagniere: Ah, I misread. Nice!

test/configuration_test.go
kevin-wu24: I mainly decided this since it follows from the style of the `cluster config update` and `topic update`. I do think since `update` commands across the CLI allow for the user to update multiple things at a time, this command should too for consistency (especially if it's called `update`).

If we only want to set 1 field at time, then we probably should call the command `set` and instead take in 2 args: `key` and `value`.

brianstrauch: > it follows from the style of the cluster config update and topic update

Those commands use a `--config` flag though, which you can pass a comma-delimited list to

brianstrauch: > If we only want to set 1 field at time, then we probably should call the command set and instead take in 2 args: key and value.

This sounds a bit better and matches https://jamesdefabia.github.io/docs/user-guide/kubectl/kubectl_config_set/

test/fixtures/output/configuration/list-windows.golden
brianstrauch: (It should say `[]` if it doesn't exist)

internal/cmd/configuration/command.go
brianstrauch: Ah, makes sense.

> Maybe we should instead append a _windows to this config field to avoid confusion?

That's a good idea, but doing so would be a breaking change. Feel free to open a ticket for this under "v4" though!

internal/cmd/configuration/command.go
```diff
@@ -0,0 +1,74 @@
...
+		Short:   "Manage CLI configuration fields.",
+	}
+
+	c := &command{
+		CLICommand: pcmd.NewAnonymousCLICommand(cmd, prerunner),
+		config:     cfg,
+	}
+
+	cmd.AddCommand(c.newSetCommand())
+	cmd.AddCommand(c.newListCommand())
```
MuweiHe, line 38: will you consider adding a getter?

internal/cmd/configuration/command_list.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/internal/pkg/cmd"
+	"github.com/confluentinc/cli/internal/pkg/output"
+)
+
+func (c *command) newListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List configuration fields in ~/.confluent/config.json.",
```
MuweiHe, line 16: list `configurable` config fields? make it clear it not all configs that'll be listed

kevin-wu24: Yeah, I think it would be better to list out only the config fields that we're fine with users modifying, since we want to abstract away most of the file's contents from users.

internal/cmd/configuration/command.go
kevin-wu24: I'm still split on whether or not we should. `list` already shows all the configuration fields in a format where it's clear to the user what the value of each field is. 

Especially if `list` is refactored to only display the `configurable` fields (i.e. the ones modifiable with set), I think `get` doesn't really provide any new functionality.

internal/cmd/configuration/command_list.go
brianstrauch: > I think it would be better to list out only the config fields that we're fine with users modifying via set

:+1:

kevin-wu24: Yeah, they will still be able to see the value of `disable_updates`.

test/fixtures/output/configuration/help-onprem.golden
```diff
@@ -7,8 +7,8 @@ Aliases:
   configuration, config
 
 Available Commands:
-  list        List configuration fields in ~/.confluent/config.json.
-  set         Set a field's value in ~/.confluent/config.json.
+  set         Set a user-configurable field's value.
+  view        View user-configurable fields in ~/.confluent/config.json.
```
brianstrauch, line 11: `update` and `list`?

internal/cmd/configuration/command_list.go
kevin-wu24: Why would CP users not be able to update their binary? Also is there anywhere in CLI code that checks if a user is a CP user?

test/fixtures/output/configuration/help-onprem.golden
kevin-wu24: I talked about this in the 1-pager, but I think `set` and `view` are more fitting command names that also align with `kubectl`. `list` in particular implies a user can have multiple of a resource, but they only have 1 config.

internal/cmd/configuration/command_list.go
brianstrauch: CP is packaged for DEB/RPM, which does not allow self-updating. Here's where it's enforced (currently, not very well): https://github.com/confluentinc/cli/blob/10cee7c4ed6e72ce62ff1d44b77049796920f28b/packaging/confluent.sh#L60

kevin-wu24: I'm not sure if we can actually know if a CLI binary is used for CP or not, since it looks like[ it's not stored alongside other CP tools](https://docs.confluent.io/platform/current/installation/cli-reference.html#confluent-cli-tool).

test/fixtures/output/configuration/help-onprem.golden
brianstrauch: Alignment with our other commands is more important, IMO!

> list in particular implies a user can have multiple of a resource, but they only have 1 config.

I think the configuration key=value pair is the resource in question here

internal/cmd/configuration/command_list.go
brianstrauch: True... we will in the future though, once we package the non-updating CLI with CP (instead of using the hacky workaround I described above)

kevin-wu24: > True... we will in the future though, once we package the non-updating CLI with CP (instead of using the hacky workaround I described above)

Yeah sounds good, I'll probably just make a ticket for that.

internal/configuration/command.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/confluentinc/cli/v3/internal/update"
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/config"
+	"github.com/confluentinc/cli/v3/pkg/types"
+)
+
+const fieldNotConfigurableError = `config field "%s" either doesn't exist or is not configurable`
```
brianstrauch, line 17:
```suggestion
const fieldNotConfigurableError = `configuration field "%s" does not exist or is not configurable`
```

internal/configuration/command.go
```diff
@@ -0,0 +1,105 @@
...
+			if jsonFieldName := getJsonFieldName(field, cfg.IsTest); jsonFieldName != "" {
+				jsonFieldToType[jsonFieldName] = kind
+				jsonFieldToName[jsonFieldName] = field.Name
+			}
+		}
+	}
+	return jsonFieldToType, jsonFieldToName
+}
+
+func getJsonFieldName(field reflect.StructField, isTest bool) string {
```
brianstrauch, line 69: There's a lot going on here... let's write some unit tests?

internal/configuration/command.go
```diff
@@ -0,0 +1,105 @@
...
+	return jsonTag
+}
+
+func (c *command) newConfigurationOut(field string) *configurationOut {
+	value := reflect.ValueOf(c.cfg).Elem().FieldByName(c.jsonFieldToName[field])
+	configOut := &configurationOut{
+		Name:  field,
+		Value: fmt.Sprintf("%v", value),
+	}
+	return configOut
```
brianstrauch, line 92:
```suggestion
	return &configurationOut{
		Name:  field,
		Value: fmt.Sprintf("%v", value),
	}
```

internal/configuration/command_describe.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+func (c *command) newDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "describe [config-field]",
```
brianstrauch, line 15: Should we refer to these as "keys", as in key-value pairs?
```suggestion
		Use:               "describe [key]",
```

internal/configuration/command_describe.go
```diff
@@ -0,0 +1,39 @@
...
+			},
+		),
+	}
+
+	return cmd
+}
+
+func (c *command) describe(cmd *cobra.Command, args []string) error {
+	field := args[0]
+	if _, ok := c.jsonFieldToType[field]; !ok {
```
brianstrauch, line 33: Should we create a whitelist? What if we add a new field but decide we don't want to display it here?

internal/configuration/command_update.go
```diff
@@ -0,0 +1,97 @@
...
+	cmd := &cobra.Command{
+		Use:               "update [config-field] [value]",
+		Short:             "Update a user-configurable field's value.",
+		Args:              cobra.ExactArgs(2),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE:              c.update,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Disable plugins by setting "disable_plugins" to true.`,
+				Code: `confluent configuration update disable_plugins true`,
```
brianstrauch, line 27:
```suggestion
				Code: "confluent configuration update disable_plugins true",
```

internal/configuration/command_update.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return cmd
+}
+
+func (c *command) update(_ *cobra.Command, args []string) error {
+	field := args[0]
+	value, err := convertValue(field, args[1], c.jsonFieldToType)
+	if err != nil {
+		if field == "current_context" {
+			err = errors.NewErrorWithSuggestions(err.Error(), "Please use `confluent context use` to set the current context.")
```
brianstrauch, line 40:
```suggestion
			return errors.NewErrorWithSuggestions(err.Error(), "Please use `confluent context use` to set the current context.")
```

pkg/errors/error_message.go
```diff
@@ -267,7 +267,7 @@ const (
 
 	// config package
 	CorruptedConfigErrorPrefix = "corrupted CLI config"
-	CorruptedConfigSuggestions = `Your CLI config file "%s" is corrupted.\n` +
+	CorruptedConfigSuggestions = "Your CLI config file \"%s\" is corrupted.\n" +
```
brianstrauch, line 270:
```suggestion
	CorruptedConfigSuggestions = "Your configuration file \"%s\" is corrupted.\n" +
```

internal/configuration/command_describe.go
kevin-wu24: The maps `jsonFieldToType` and `jsonFieldToName` technically are whitelists. If a new field is added to the `config` it also needs to be added to these maps by editing the `getSettableConfigFields` function to be visible to these commands.

The alternative would be having these maps as literals that we maintain alongside the definition of the `config` struct. I honestly think this might be more maintainable and make the code a lot easier to understand than having to modify a function every time you want a field to be visible to these commands. Wdyt?

internal/configuration/command.go
```diff
@@ -60,39 +62,46 @@ func getConfigWhiteList(cfg *config.Config) map[string]*configInfo {
...
-		return ""
-	}
 	return jsonTag
 }
 
+func isReadOnly(jsonField string, isTest bool) bool {
+	if jsonField == "disable_updates" && !isTest && update.IsHomebrew() {
+		return true
+	}
+	return false
```
brianstrauch, line 96: Why do we include `!isTest` in here?
```suggestion
	return jsonField == "disable_updates" && !isTest && update.IsHomebrew()
```

kevin-wu24: The `IsHomebrew()` function is quite slow, and causes the integration tests to time out because the whitelist is generated as part of the `command` struct, which is called in every test. However, I shouldn't have `isTest` here anymore.

The whitelist should instead be generated lazily (only when using the `config` subcommands), rather than whenever `New()` is called to create these commands (which is always).

internal/configuration/configuration_test.go
```diff
@@ -16,16 +16,16 @@ func TestGetConfigWhitelist(t *testing.T) {
 		panic(err)
 	}
 
-	expected := map[string]*configInfo{
+	expected := map[string]*configFieldInfo{
 		"disable_feature_flags": {name: "DisableFeatureFlags", kind: reflect.Bool},
```
brianstrauch, line 20: What's the UpperCamelCase formatting for?

kevin-wu24: Oh if you mean the `name:` field values, they have to match the `config` struct name exactly, so since the `config` fields are exported, these have to UpperCamelCase.

kevin-wu24: I think it is better to include the entire configuration field's information (field name in Go + type), as if any of these fields are renamed or their type is changed, this test would fail (which is good in this case!). 

If a modification happens to any of these fields, and we just check that the JSON tags are in the whitelist, this test would still pass, but using the `describe/update` subcommand could panic when trying to retrieve the `Value` of the field by name.

MuweiHe (approved):

internal/configuration/command.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/confluentinc/cli/v3/internal/update"
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/config"
+	"github.com/confluentinc/cli/v3/pkg/types"
+)
+
+const fieldNotConfigurableError = `configuration field "%s" either does not exist or is not configurable`
```
brianstrauch, line 17:
```suggestion
const fieldNotConfigurableError = `configuration field "%s" does not exist or is not configurable`
```

internal/configuration/command_describe.go
```diff
@@ -0,0 +1,38 @@
...
+			examples.Example{
+				Text: `See if update checks are enabled by describing "disable_update_check".`,
+				Code: "confluent configuration describe disable_update_check",
+			},
+		),
+	}
+}
+
+func (c *command) describe(cmd *cobra.Command, args []string) error {
+	configWhitelist := getConfigWhitelist(c.cfg)
```
brianstrauch, line 30:
```suggestion
	whitelist := getWhitelist(c.cfg)
```

pkg/config/config.go
```diff
@@ -81,10 +81,6 @@ type Config struct {
 	SavedCredentials    map[string]*LoginCredential `json:"saved_credentials,omitempty"`
 	LocalPorts          *LocalPorts                 `json:"local_ports,omitempty"`
 
-	// Deprecated
-	AnonymousId string `json:"anonymous_id,omitempty"`
-	Ver         string `json:"version,omitempty"`
```
brianstrauch, line 86: These fields can't be deleted. Users who still have them in their configuration file will get a serialization error upon running any command.

brianstrauch (changes requested):

pkg/config/config.go
kevin-wu24: Oh okay gotcha. Do we want these fields to be in the whitelist? (I'm assuming no since they're deprecated).

internal/configuration/command_describe.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (c *command) describe(cmd *cobra.Command, args []string) error {
+	whitelist := getWhitelist(c.cfg)
+	field := args[0]
+	if _, ok := whitelist[field]; !ok {
+		return fmt.Errorf(fieldNotConfigurableError, field)
+	}
+	table := output.NewTable(cmd)
+	table.Add(c.newFieldOut(field, whitelist))
+	return table.Print()
```
brianstrauch, line 37: Adding newlines around related code blocks will help readability!
```suggestion
	whitelist := getWhitelist(c.cfg)
	field := args[0]
	
	if _, ok := whitelist[field]; !ok {
		return fmt.Errorf(fieldNotConfigurableError, field)
	}
	
	table := output.NewTable(cmd)
	table.Add(c.newFieldOut(field, whitelist))
	return table.Print()
```

internal/configuration/command_list.go
```diff
@@ -0,0 +1,30 @@
...
+		RunE:  c.list,
+	}
+
+	pcmd.AddOutputFlag(cmd)
+
+	return cmd
+}
+
+func (c *command) list(cmd *cobra.Command, _ []string) error {
+	whitelist := getWhitelist(c.cfg)
```
brianstrauch, line 24:
```suggestion
	whitelist := getWhitelist(c.cfg)
	
```

internal/configuration/command_update.go
```diff
@@ -0,0 +1,96 @@
...
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/form"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+func (c *command) newUpdateCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:               "update [key] [value]",
```
brianstrauch, line 19: These are required, so we use `<>`. We should write a linter rule to catch this...
```suggestion
		Use:               "update <key> <value>",
```

internal/configuration/command_update.go
```diff
@@ -0,0 +1,96 @@
...
+		Short:             "Update a user-configurable field's value.",
+		Args:              cobra.ExactArgs(2),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE:              c.update,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Disable plugins by setting "disable_plugins" to true.`,
```
brianstrauch, line 26:
```suggestion
				Text: `Set the "disable_plugins" configuration to "true".`,
```

pkg/config/local_ports.go
```diff
@@ -1,8 +1,17 @@
...
+import "fmt"
+
 type LocalPorts struct {
 	BrokerPort     string `human:"Broker Port" json:"broker_port"`
 	ControllerPort string `human:"Controller Port" json:"controller_port"`
 	KafkaRestPort  string `human:"Kafka Rest Port" json:"kafka_rest_port"`
 	PlaintextPort  string `human:"Plaintext Port" json:"plaintext_port"`
 }
+
+func (p *LocalPorts) String() string {
```
brianstrauch, line 12: Are we using this still?

internal/configuration/command.go
```diff
@@ -0,0 +1,112 @@
...
+	return cmd
+}
+
+func getWhitelist(cfg *config.Config) map[string]*fieldInfo {
+	whitelist := make(map[string]*fieldInfo)
+	t := reflect.TypeOf(*cfg)
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		// currently only boolean fields are part of this whitelist, but this may change in the future
+		if kind := field.Type.Kind(); kind == reflect.Bool {
```
brianstrauch, line 61: What if we add a boolean field that we don't want to show up here? Can we make an actual whitelist slice of fields that we want to display?

kevin-wu24: If we add a boolean field to the `config` struct, it will fail the `getWhitelist` unit test, so if we don't want it to be added we would have to modify this function. 

Yeah, I guess we could make an actual whitelist along side the `config` struct, but we can't avoid having to add `disable_plugins_once` to it at runtime.

brianstrauch: > we can't avoid having to add disable_plugins_once to it at runtime

Why not?

kevin-wu24: We only want to add this field if the CLI is on Windows. I think I mentioned this before, but if a Unix/mac user sees this field, they may think setting it will actually do something, but it won't. 

One alternative would be to include it in the whitelist, but make it read-only for non-Windows users.

brianstrauch: What if we just define a slice of whitelisted configurations, and only add that field if `runtime.GOOS == "windows"`?

kevin-wu24: My wording was probably confusing earlier. I meant that `disable_plugins_once` cannot be included in the map literal's declaration, and instead must be added at runtime if `runtime.GOOS == "windows"`.

What u said is exactly what I'm doing haha. Sorry for the misunderstanding.

pkg/config/config.go
```diff
@@ -65,6 +65,15 @@ var (
 	)
 )
 
+// Whitelist is the configuration fields that are visible by the `config` subcommands.
+var Whitelist = []string{
```
brianstrauch, line 69: Seems like we should define this in the `getWhitelist()` function, and not globally?

pkg/config/config.go
brianstrauch: Yep, let's leave them out!

pkg/config/config.go
kevin-wu24: I defined this globally so it could be next to the `Config` struct. That way, when modifications are made to the struct, a developer can easily add/remove a field from the whitelist. Otherwise, they would have to find the `getWhitelist` function in another directory and modify that.

internal/configuration/command_update.go
```diff
@@ -0,0 +1,97 @@
...
+		return err
+	}
+	output.Print(fmt.Sprintf(errors.UpdateSuccessMsg, "value", "config field", field, value))
+	return nil
+}
+
+func convertValue(field, value string, whitelist map[string]*fieldInfo) (any, error) {
+	info, ok := whitelist[field]
+	if !ok || info.readOnly {
+		return nil, fmt.Errorf(fieldNotConfigurableError, field)
```
brianstrauch, line 67: Can we split this into two cases? One error for "configuration key does not exist" and another for "configuration is read-only"?

internal/configuration/command.go
```diff
@@ -0,0 +1,101 @@
...
+	if runtime.GOOS == "windows" {
+		config.Whitelist = append(config.Whitelist, "disable_plugins_once")
+	}
+
+	whitelist := make(map[string]*fieldInfo, len(config.Whitelist))
+	t := reflect.TypeOf(*cfg)
+
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		jsonTag, _, _ := strings.Cut(field.Tag.Get("json"), ",")
```
brianstrauch, line 66: splitting by "," and taking the first value might be more readable. I don't see `Cut()` being used very often...

brianstrauch (approved):

PR #2194, opened 2023-08-15
Show single-resource API keys for multicluster enabled orgs

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Some system tests recently failed because API keys with a single resource don't show up in `confluent api-key list`. The cause is that the associated resource is listed in the `Resource` field of the http response and not in the `Resources` field, which is given preference for multicluster enabled orgs.

This PR only uses the list in `Resources` if it is non-empty.

I also added a check on the test handler for `/srcm/v2/clusters/{id}` to verify that the `{id}` matches the test backend SR cluster id. There was previously a confusing golden file where the SR cluster given to the `--resource` flag did not match the actual cluster printed in the output.

References
----------

Test & Review
-------------
Tested using a stag org and a prod org.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/cmd/api-key/command_list.go
+++ b/internal/cmd/api-key/command_list.go
@@ -119,7 +119,7 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
 		resources := []apikeysv2.ObjectReference{apiKey.Spec.GetResource()}
 
 		// Check if multicluster keys are enabled, and if so check the resources field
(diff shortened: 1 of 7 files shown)


brianstrauch (approved):

brianstrauch:
Release notes?

sgagniere:
@brianstrauch This is only a bug fix for multicluster enabled orgs and that feature doesn't appear to be GA, so I don't think we should write a release note for this.

PR #2195, opened 2023-08-15
[confluent-local] Phase 2: add multi-broker option and broker management commands

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `--brokers` flag in `confluent local kafka start` to allow starting a multi-broker Confluent Local instance
- Add `confluent local kafka broker` command
- Add `confluent local kafka cluster configuration` command

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

When running `confluent local kafka start`, can add a `multi-broker` flag to start the cluster as 2 brokers. Each broker runs on its own container. 
Added `broker list/describe/update` commands.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

In discussion: do we add a prompt for switching from single-node to 2-node when user runs `local start` and there's already an instance running.
Descriptions of `confluent local` should all change. It's not `single-node` necessarily anymore.
How to we help users to distinguish the 2 sets of old/new confluent local commands? how do we let them know that `confluent local broker` doesn't work with the old `services`?


internal/cmd/local/command_kafka_start.go
```diff
@@ -127,46 +127,109 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	)
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return err
+	}
+
+	config1 := &container.Config{
 		Image:    dockerImageName,
-		Hostname: "broker",
-		Cmd:      strslice.StrSlice{"bash", "-c", "'/etc/confluent/docker/run'"},
+		Hostname: "broker1",
```
davetroiano, line 149: suggest hyphenating `broker-<ID>`.  reads a little more clearly. might be more consistent with other examples too -- I doubt we're consistent everywhere, but a couple of public multibroker examples that I checked do hyphenate (MRC [example](https://github.com/confluentinc/examples/blob/7.4.0-post/multiregion/docker-compose.yml#L67), kafka-images KRaft [example](https://github.com/confluentinc/kafka-images/blob/master/examples/confluent-server-kraft/docker-compose.yml#L20))

internal/cmd/local/command_kafka_start.go
```diff
@@ -227,59 +290,67 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
 		"KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS=0",
 		"KAFKA_TRANSACTION_STATE_LOG_MIN_ISR=1",
 		"KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR=1",
 		"KAFKA_PROCESS_ROLES=broker,controller",
-		"KAFKA_NODE_ID=1",
-		fmt.Sprintf("KAFKA_CONTROLLER_QUORUM_VOTERS=1@broker:%s", ports.ControllerPort),
-		fmt.Sprintf("KAFKA_LISTENERS=PLAINTEXT://broker:%s,CONTROLLER://broker:%s,PLAINTEXT_HOST://0.0.0.0:%s", ports.BrokerPort, ports.ControllerPort, ports.PlaintextPort),
+		fmt.Sprintf("KAFKA_NODE_ID=%d", brokerId),
+		fmt.Sprintf("KAFKA_CONTROLLER_QUORUM_VOTERS=1@broker1:%s,2@broker2:%s", ports.ControllerPorts[0], ports.ControllerPorts[1]),
+		fmt.Sprintf("KAFKA_LISTENERS=PLAINTEXT://broker%d:%s,CONTROLLER://broker%d:%s,PLAINTEXT_HOST://0.0.0.0:%s", brokerId, ports.BrokerPorts[idx], brokerId, ports.ControllerPorts[idx], ports.PlaintextPorts[idx]),
```
davetroiano, line 342: update these if broker name gets hyphenated

internal/cmd/local/command_kafka_start.go
```diff
@@ -127,46 +127,109 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	}
+	log.CliLogger.Tracef("Create confluent-local container for broker 1 success")
+	err = dockerClient.ContainerStart(context.Background(), createResp1.ID, types.ContainerStartOptions{})
 	if err != nil {
 		return err
 	}
-	log.CliLogger.Tracef("Create confluent-local container success")
 
-	err = dockerClient.ContainerStart(context.Background(), createResp.ID, types.ContainerStartOptions{})
+	multiBroker, err := cmd.Flags().GetBool("multi-broker")
```
davetroiano, line 213: I think that we should go for configurable N brokers (1 <= N <= 5 or 10 or something) rather than a flag that gives you 2. I suspect users will want to play with other numbers, with 2 and 3 (common replication factor 3), maybe 4 being the common cases

internal/cmd/local/command_kafka_start.go
```diff
@@ -127,46 +127,109 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+		log.CliLogger.Tracef("Create confluent-local container for broker 2 success")
+		err = dockerClient.ContainerStart(context.Background(), createResp2.ID, types.ContainerStartOptions{})
+		if err != nil {
+			return err
+		}
 
-	output.Printf("Started Confluent Local container %v.\nTo continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.\n", getShortenedContainerId(createResp.ID))
+		output.Printf("Started Confluent Local containers %v, %v.\nTo continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.\n", getShortenedContainerId(createResp1.ID), getShortenedContainerId(createResp2.ID))
+	} else {
+		output.Printf("Started Confluent Local containers %v.\nTo continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.\n", getShortenedContainerId(createResp1.ID))
```
davetroiano, line 230: This PR didnt introduce this, but in testing this I think that  the second sentence should be the last thing output, and maybe add a couple of newlines, e.g.:
```
Started Confluent Local containers da0e273845, 934727acfa.

+------------------+---------------+
| Broker Ports     | [58583 58584] |
| Controller Ports | [58585 58586] |
| Kafka Rest Ports | [8082 9092]   |
| Plaintext Ports  | [58581 58582] |
+------------------+---------------+

To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.
```

internal/pkg/config/local_ports.go
```diff
@@ -1,8 +1,8 @@
 package config
 
 type LocalPorts struct {
-	BrokerPort     string `human:"Broker Port" json:"broker_port"`
-	ControllerPort string `human:"Controller Port" json:"controller_port"`
-	KafkaRestPort  string `human:"Kafka Rest Port" json:"kafka_rest_port"`
-	PlaintextPort  string `human:"Plaintext Port" json:"plaintext_port"`
+	BrokerPorts     []string `human:"Broker Ports" json:"broker_ports"`
+	ControllerPorts []string `human:"Controller Ports" json:"controller_ports"`
+	KafkaRestPorts  []string `human:"Kafka Rest Ports" json:"kafka_rest_ports"`
```
davetroiano, line 6: `Kafka REST Ports`?

davetroiano (commented):
suggest updating the stop command as part of this PR. currently it's this for multi-broker:
```
% ./confluent local kafka stop
Confluent Local has been stopped: removed container "934727acfa".
Confluent Local has been stopped: removed container "da0e273845".
```
suggest collecting the container IDs and printing one line:
```
% ./confluent local kafka stop
Confluent Local has been stopped: removed containers 934727acfa, da0e273845.
```

internal/pkg/broker/broker_utils.go
```diff
@@ -0,0 +1,130 @@
+package broker
```
brianstrauch, line 1: Just calling this file `broker.go` is probably fine. Suffixing file names with `utils` makes it easy just to throw any function into it, and can quickly grow out of control (100+ lines of code)

internal/cmd/kafka/command_broker_describe.go
```diff
@@ -53,7 +38,7 @@ func (c *brokerCommand) newDescribeCommand() *cobra.Command {
 }
 
 func (c *brokerCommand) describe(cmd *cobra.Command, args []string) error {
-	brokerId, all, err := checkAllOrBrokerIdSpecified(cmd, args)
+	brokerId, all, err := broker.CheckAllOrBrokerIdSpecified(cmd, args)
```
brianstrauch, line 41: Since this is in the `broker` package, probably don't need to write "broker" in the function name
```suggestion
	brokerId, all, err := broker.CheckAllOrIdSpecified(cmd, args)
```

brianstrauch: In the future, I want to get rid of the `--all` flag. For example,

```confluent kafka broker configuration list --all```

should probably be written as:

```confluent kafka cluster configuration list```

Comment (brianstrauch) [reply internal/cmd/kafka/command_broker_describe.go:41]: (Not sure how you're using `--all` in the Confluent Local commands, but maybe we should consider removing it pre-emptively)

File: internal/cmd/local/command_broker.go [hunk 8]
```diff
@@ -0,0 +1,18 @@
+package local
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (c *Command) newBrokerCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "broker",
+		Short: "Manage brokers in Confluent Local cluster",
```
brianstrauch, line 10:
```suggestion
		Short: "Manage Kafka brokers in Confluent Local.",
```

internal/cmd/local/command_broker.go
```diff
@@ -0,0 +1,18 @@
...
+	}
+
+	cmd.AddCommand(c.newBrokerListCommand())
+	cmd.AddCommand(c.newBrokerDescribeCommand())
```
brianstrauch, line 14:
```suggestion
	cmd.AddCommand(c.newBrokerDescribeCommand())
	cmd.AddCommand(c.newBrokerListCommand())
```

internal/cmd/local/command_broker_describe.go
```diff
@@ -0,0 +1,83 @@
...
+				Code: "confluent local broker describe 1 --config-name min.insync.replicas",
+			},
+			examples.Example{
+				Text: "Describe the non-default cluster-wide broker configuration values.",
+				Code: "confluent local broker describe --all",
+			},
+		),
+	}
+
+	cmd.Flags().Bool("all", false, "Get cluster-wide broker configurations (non-default values only).")
```
brianstrauch, line 35: As I mentioned above, I would like to delete this and replace this with `confluent local kafka cluster configuration list` to match the existing CP command.

internal/cmd/local/command_broker_update.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/confluentinc/cli/internal/pkg/kafkarest"
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/properties"
+	"github.com/confluentinc/kafka-rest-sdk-go/kafkarestv3"
+)
+
+func (c *Command) newBrokerUpdateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update [id]",
+		Short: "Update a Confluent local broker configurations.",
```
brianstrauch, line 22: I believe the product "Confluent Local" is a proper noun that should be capitalized in all cases?
```suggestion
		Short: "Update Confluent Local broker configurations.",
```

internal/cmd/local/command_kafka.go
```diff
@@ -20,7 +21,7 @@ const (
 func (c *Command) newKafkaCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "kafka",
-		Short: "Manage a single-node instance of Apache Kafka.",
+		Short: "Manage a single-node or two-node instance of Apache Kafka.",
```
brianstrauch, line 24: The distinction between a single- and multi-node cluster should probably just be part of the description of `--multi-broker`
```suggestion
		Short: "Manage a local instance of Apache Kafka.",
```

internal/cmd/local/command_kafka_start.go
```diff
@@ -37,14 +37,14 @@ type ImagePullResponse struct {
...
 		Args:  cobra.NoArgs,
 		RunE:  c.kafkaStart,
 	}
 
-	cmd.Flags().String("kafka-rest-port", "8082", "The port number for Kafka REST.")
-	cmd.Flags().String("plaintext-port", "", "The port number for plaintext producer and consumer clients. If not specified, a random free port will be used.")
-
+	cmd.Flags().StringSlice("kafka-rest-ports", []string{"8082", "9092"}, "The port number for Kafka REST for brokers.")
+	cmd.Flags().StringSlice("plaintext-ports", nil, "The port number for plaintext producer and consumer clients for brokers. If not specified, a random free port will be used.")
+	cmd.Flags().Bool("multi-broker", false, "Start Confluent Local cluster with two brokers.")
```
brianstrauch, line 47:
```suggestion
	cmd.Flags().Bool("multi-broker", false, "Use two Kafka brokers instead of one.")
```

internal/cmd/local/command_kafka_start.go
```diff
@@ -127,46 +127,109 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 	}
 
-	createResp, err := dockerClient.ContainerCreate(context.Background(), config, hostConfig, nil, platform, confluentLocalContainerName)
+	createResp1, err := dockerClient.ContainerCreate(context.Background(), config1, hostConfig1, nil, platform, "broker1")
+	if err != nil {
+		return err
+	}
+	log.CliLogger.Tracef("Create confluent-local container for broker 1 success")
+	err = dockerClient.ContainerStart(context.Background(), createResp1.ID, types.ContainerStartOptions{})
 	if err != nil {
```
brianstrauch, line 209:
```suggestion
	if err := dockerClient.ContainerStart(context.Background(), createResp1.ID, types.ContainerStartOptions{}); err != nil {
```

internal/cmd/local/command_kafka_start.go
```diff
@@ -227,59 +290,67 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
 }
 
 func (c *Command) validateCustomizedPorts() error {
-	kafkaRestLn, err := net.Listen("tcp", ":"+c.Config.LocalPorts.KafkaRestPort)
-	if err != nil {
-		freePort, err := freeport.GetFreePort()
+	for idx := 0; idx < len(c.Config.LocalPorts.KafkaRestPorts); idx++ {
```
brianstrauch, line 293:
```suggestion
	for i, port := range c.Config.LocalPorts.KafkaRestPorts {
```

internal/pkg/broker/broker_utils.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/confluentinc/cli/internal/pkg/errors"
+	"github.com/confluentinc/cli/internal/pkg/kafkarest"
+	"github.com/confluentinc/kafka-rest-sdk-go/kafkarestv3"
+	"github.com/spf13/cobra"
+)
+
+const AbbreviationLength = 25
```
brianstrauch, line 14: Abbreviation length of what? The broker ID? The broker name?

brianstrauch (changes requested):

internal/cmd/local/command_broker_list.go
```diff
@@ -0,0 +1,55 @@
...
+	// Get Brokers
+	brokersGetResp, resp, err := restClient.BrokerV3Api.ClustersClusterIdBrokersGet(context.Background(), clusterId)
+	if err != nil {
+		return kafkarest.NewError(restClient.GetConfig().BasePath, err, resp)
+	}
+
+	list := output.NewList(cmd)
+	for _, data := range brokersGetResp.Data {
+		broker := &broker.BrokerOut{
+			ClusterId: data.ClusterId,
```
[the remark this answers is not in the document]
MuweiHe: for example cluster id would be just `4L6g3nShT-eMCtK--X86sw`. I'm ok with removing it since we only have one cluster.

internal/cmd/local/command_kafka_start.go
MuweiHe: agree. This is also being used as the name of the containers too, so I'm thinking something more explicit like `confluent-local-broker-1`

internal/cmd/local/command_broker_describe.go
MuweiHe: we have this flag in CP `confluent kafka broker describe --all` too. Are cluster configs and a cluster-wide broker config the same?

internal/pkg/broker/broker_utils.go
MuweiHe: when describing broker configs, the abbreved version of config name (e.g. ssl.trustmanager.algorith...

MuweiHe:
@davetroiano @brianstrauch update the pr to allow configuring 1~4 brokers.

internal/local/command_kafka_start.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/output"
 )
 
+var defaultKafkaRestPorts = []string{"8082", "9092", "10012", "20012"}
```
brianstrauch, line 30: 9092 is typically used for the broker, not Kafka REST: https://docs.confluent.io/platform/current/installation/system-requirements.html#ports

internal/local/command_kafka_start.go
```diff
@@ -144,106 +149,61 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
-				},
-			},
-			natPlaintextPorts[0]: []nat.PortBinding{
-				{
-					HostIP:   localhost,
-					HostPort: ports.PlaintextPorts[0],
-				},
+	var containerIds []string
+	for idx := int32(0); idx < numOfBrokers; idx++ {
+		brokerId := idx + 1
```
brianstrauch, line 157:
```suggestion
	for idx := 0; idx < numOfBrokers; idx++ {
		brokerId := int32(idx + 1)
```

internal/local/command_kafka_start.go
```diff
@@ -144,106 +149,61 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	output.Printf("Started Confluent Local containers %s.\nTo continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.\n", strings.Join(containerIds, ","))
+
 	return nil
 }
 
-func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) error {
+func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, numOfBrokers int32, isTest bool) error {
```
brianstrauch, line 209:
```suggestion
func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, numBrokers int32, isTest bool) error {
```

internal/local/command_kafka_start.go
```diff
@@ -256,16 +216,16 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
 			PlaintextPorts:  []string{"3002", "3003"},
 		}
 	} else {
-		freePorts, err := freeport.GetFreePorts(6)
+		freePorts, err := freeport.GetFreePorts(int(3 * numOfBrokers))
```
brianstrauch, line 222: Why `* 3`? Does every broker have its own Kafka REST instance? That shouldn't be the case... right?

MuweiHe: you're right. only one krest port is needed.

MuweiHe: will update

internal/local/command_kafka_start.go
```diff
@@ -115,10 +117,13 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	output.Println("\r")
-
 	log.CliLogger.Tracef("Successfully pulled Confluent Local image")
 
-	if err := c.prepareAndSaveLocalPorts(cmd, c.Config.IsTest); err != nil {
+	numOfBrokers, err := cmd.Flags().GetInt32("brokers")
```
sgagniere, line 122: Based on the `// range: [1, 4]` comment, should that limit be enforced here? At a quick glance I didn't see anything that seems to stop users from choosing higher numbers.

MuweiHe: that's right, no enforcement added yet!

internal/local/command_kafka_start.go
MuweiHe: krest is always 8082, but controller / broker / plaintext are 3 different ports

internal/local/command_kafka_start.go
MuweiHe: `numOfBrokers` is an int32

cmd/lint/main.go
```diff
@@ -24,7 +24,7 @@ var commandRules = []linter.CommandRule{
 
 	linter.Filter(linter.RequireCapitalizeProperNouns("Short", properNouns), linter.ExcludeCommand("local current")),
 	linter.RequireEndWithPunctuation("Short", false),
-	linter.Filter(linter.RequireNotTitleCase("Short", properNouns)),
+	linter.Filter(linter.RequireNotTitleCase("Short", properNouns), linter.ExcludeCommandContains("local kafka broker")),
```
brianstrauch, line 27: You shouldn't have to add this rule. Should we have added "Confluent Local" to the list of proper nouns?

internal/kafka/command_broker_list.go
```diff
@@ -48,7 +42,7 @@ func (c *brokerCommand) list(cmd *cobra.Command, _ []string) error {
 
 	list := output.NewList(cmd)
 	for _, data := range brokersGetResp.Data {
-		broker := &brokerOut{
+		broker := &broker.BrokerOut{
```
brianstrauch, line 45:
```suggestion
		broker := &broker.Out{
```

internal/local/command_kafka_broker_describe.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/utils"
+)
+
+func (c *Command) newKafkaBrokerDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe [id]",
+		Short: "Describe a local Kafka broker.",
+		Long:  "Describe cluster-wide or per-broker configuration values.",
```
brianstrauch, line 20: I think we should align with the future state of the `broker` commands, that is:
1. This command should strictly be for printing `broker` configurations, not cluster-wide configurations (remove the `--all` flag)
2. Cluster-wide configurations should be printed with `confluent local kafka cluster configuration list`.

I was waiting for v4 to make this change, but seems like we have an opportunity to make it early here.

internal/local/command_kafka_broker_describe.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Get Broker Configs
+	var data []*broker.ConfigOut
+	if all { // fetch cluster-wide configs
+		clusterConfig, err := broker.GetClusterWideConfigs(restClient, context.Background(), clusterId, configName)
+		if err != nil {
+			return err
+		}
+		data = broker.ParseClusterConfigData(clusterConfig)
+	} else { // fetch individual broker configs
```
brianstrauch, line 66:
```suggestion
	} else {
		// fetch individual broker configs
```

internal/local/command_kafka_broker_update.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/confluentinc/cli/v3/pkg/broker"
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/kafkarest"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/properties"
+)
+
+func (c *Command) newKafkaBrokerUpdateCommand() *cobra.Command {
```
brianstrauch, line 20: These broker commands seem completely identical to the existing broker commands. Do you think we could call the main function of those commands directly?

internal/local/command_kafka_start.go
```diff
@@ -37,14 +37,14 @@ type ImagePullResponse struct {
...
 		Args:  cobra.NoArgs,
 		RunE:  c.kafkaStart,
 	}
 
-	cmd.Flags().String("kafka-rest-port", "8082", "The port number for Kafka REST.")
-	cmd.Flags().String("plaintext-port", "", "The port number for plaintext producer and consumer clients. If not specified, a random free port will be used.")
-
+	cmd.Flags().String("kafka-rest-port", "8082", "Kafka REST port number.")
+	cmd.Flags().StringSlice("plaintext-ports", nil, "A comma-separated list of port numbers for plaintext producer and consumer clients for brokers. If not specified, random free ports will be used.")
+	cmd.Flags().Int32("brokers", 1, "Number of brokers in the Confluent Local cluster.") // range: [1, 4]? do we enforce a range?
```
brianstrauch, line 47: :eyes:

brianstrauch (changes requested):

internal/kafka/command_broker_list.go
MuweiHe: there're 2 types or `out` structs, config or broker

internal/local/command_kafka_broker_describe.go
MuweiHe: sounds good. Adding this change.

internal/local/command_kafka_broker_update.go
MuweiHe: how do i do this? kafka broker commands are of type `*pcmd.AuthenticatedCLICommand` 🤔

brianstrauch: Sounds like we'll have to create the client outside of the function, and pass it in to a the shared function?

internal/kafka/command_broker_list.go
brianstrauch: The package is called broker, so we can have: `broker.Out` and `broker.ConfigurationOut`

brianstrauch: (`broker.BrokerOut` is a bit redundant, IMO)

internal/kafka/command_broker_describe.go
```diff
@@ -45,15 +30,15 @@ func (c *brokerCommand) newDescribeCommand() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("all", false, "Get cluster-wide broker configurations (non-default values only).")
-	cmd.Flags().String("config-name", "", "Get a specific configuration value (pair with --all to see a cluster-wide configuration.")
+	cmd.Flags().String("config-name", "", "Get a specific configuration value (pair with --all to see a cluster-wide configuration).")
```
brianstrauch, line 33: Good catch here. We can't use backticks here due to a limitation in Cobra, but regular quotes are better than nothing.
```suggestion
	cmd.Flags().String("config-name", "", `Get a specific configuration value (pair with "--all" to see a cluster-wide configuration).`)
```

internal/local/command_kafka_broker_describe.go
```diff
@@ -0,0 +1,70 @@
...
+		RunE:  c.brokerDescribe,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Describe the "min.insync.replicas" configuration for broker 1.`,
+				Code: "confluent local broker describe 1 --config-name min.insync.replicas",
+			},
+		),
+	}
+
+	cmd.Flags().String("config-name", "", "Get a specific configuration value")
```
brianstrauch, line 32: The linter should check for this, right?
```suggestion
	cmd.Flags().String("config-name", "", "Get a specific configuration value.")
```

internal/local/command_kafka_broker_update.go
```diff
@@ -0,0 +1,49 @@
...
+			},
+			examples.Example{
+				Text: "Update configuration values for all brokers in the cluster.",
+				Code: "confluent kafka broker update --all --config min.insync.replicas=2,num.partitions=2",
+			},
+		),
+	}
+
+	pcmd.AddConfigFlag(cmd)
+	cmd.Flags().Bool("all", false, "Apply configuration update to all brokers in the cluster.")
```
brianstrauch, line 34: I think we agreed to delete this in another comment, and use the `confluent kafka cluster configuration update` commands instead?

internal/local/command_kafka_cluster.go
```diff
@@ -0,0 +1,16 @@
+package local
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (c *Command) newKafkaClusterCommand() *cobra.Command {
```
brianstrauch, line 7: Any reason why `Command` is public?
```suggestion
func (c *command) newKafkaClusterCommand() *cobra.Command {
```

internal/local/command_kafka_start.go
```diff
@@ -127,79 +130,104 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	natPlaintextPorts := getNatPlaintextPorts(ports, numOfBrokers)
+	containerStartCmd := strslice.StrSlice{"bash", "-c", "'/etc/confluent/docker/run'"}
+
+	_, err = dockerClient.NetworkCreate(
+		context.Background(),
+		"confluent-local-network",
+		types.NetworkCreate{
+			CheckDuplicate: true,
+			Driver:         "bridge",
 		},
```
brianstrauch, line 144: Can you define this outside of the function call?

brianstrauch (commented):
Looks like there are still comments from my previous review that haven't been addressed.

internal/local/command_kafka_start.go
MuweiHe: Haven't decided on the value we should set here. @davetroiano what's a proper range of number of brokers that users should be able to set?

internal/local/command_kafka_broker_update.go
MuweiHe: oops

internal/local/command_kafka_cluster.go
MuweiHe: Good catch! It was not necessary

internal/local/command_kafka_start.go
MuweiHe: got it

internal/cmd/local/command_kafka_start.go
MuweiHe: changed to only display krest port and plaintext ports as others are internal.
used comma-delimited format

internal/local/command_kafka_start.go
```diff
@@ -127,79 +130,104 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 	table.Add(c.Config.LocalPorts)
-	if err := table.Print(); err != nil {
+	err = table.Print()
+	if err != nil {
 		return err
 	}
-
-	output.ErrPrintln("To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.")
-
+	output.Printf("Started Confluent Local containers %s.\n", strings.Join(containerIds, ","))
```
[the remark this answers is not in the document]
MuweiHe: Started Confluent Local containers "152f97f38c", "564f151276", "5f5ddca7d1",  "8c704f07df".

internal/kafka/command_topic.go
```diff
@@ -202,7 +202,7 @@ func ProduceToTopic(cmd *cobra.Command, keyMetaInfo []byte, valueMetaInfo []byte
 		} else if data == EOF {
 			break
 		}
-
+		fmt.Println("meta", keyMetaInfo, valueMetaInfo)
```
brianstrauch, line 205: :eyes:

internal/local/command_kafka_broker_describe.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+)
+
+func (c *command) newKafkaBrokerDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe [id]",
+		Short: "Describe a local Kafka broker.",
+		Long:  "Describe per-broker configuration values.",
+		Args:  cobra.MaximumNArgs(1),
```
brianstrauch, line 19: Should be required now that `--all` is removed.
```suggestion
		Use:   "describe <id>",
		Short: "Describe a local Kafka broker.",
		Long:  "Describe per-broker configuration values.",
		Args:  cobra.ExactArgs(1),
```

internal/local/command_kafka_start.go
```diff
@@ -127,79 +141,106 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
-	createResp, err := dockerClient.ContainerCreate(context.Background(), config, hostConfig, nil, platform, confluentLocalContainerName)
-	if err != nil {
+	natPlaintextPorts := getNatPlaintextPorts(ports, numBrokers)
+	containerStartCmd := strslice.StrSlice{"bash", "-c", "'/etc/confluent/docker/run'"}
+
+	_, err = dockerClient.NetworkCreate(
+		context.Background(),
+		"confluent-local-network",
+		networkCreate,
+	)
```
brianstrauch, line 153:
```suggestion
	_, err = dockerClient.NetworkCreate(context.Background(), "confluent-local-network", networkCreate)
```

internal/local/command_kafka_start.go
```diff
@@ -127,79 +141,106 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	var containerIds []string
+	for idx := int32(0); idx < numBrokers; idx++ {
+		brokerId := idx + 1
+		config := &container.Config{
+			Image:    dockerImageName,
+			Hostname: fmt.Sprintf("confluent-local-broker-%d", brokerId),
+			Cmd:      containerStartCmd,
+			ExposedPorts: nat.PortSet{
+				natPlaintextPorts[idx]: struct{}{},
+			},
```
brianstrauch, line 167:
```suggestion
			ExposedPorts: nat.PortSet{natPlaintextPorts[idx]: struct{}{}},
```

internal/local/command_kafka_start.go
```diff
@@ -210,12 +251,12 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
 		}
 
-		plaintextPort, err := cmd.Flags().GetString("plaintext-port")
+		plaintextPorts, err := cmd.Flags().GetStringSlice("plaintext-ports")
 		if err != nil {
 			return err
 		}
-		if plaintextPort != "" {
-			c.Config.LocalPorts.PlaintextPort = plaintextPort
+		if len(plaintextPorts) != 0 {
```
brianstrauch, line 258:
```suggestion
		if len(plaintextPorts) > 0 {
```

internal/local/command_kafka_start.go
```diff
@@ -230,60 +271,90 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
+	for i := 0; i < int(numBrokers); i++ {
+		res = append(res, nat.Port(ports.PlaintextPorts[i]+"/tcp"))
+	}
+	return res
+}
+
+func getKafkaControllerQuorumVoters(ports *config.LocalPorts, numBrokers int32) string {
+	voters := fmt.Sprintf("KAFKA_CONTROLLER_QUORUM_VOTERS=1@confluent-local-broker-1:%s", ports.ControllerPorts[0])
+	for i := int32(1); i < numBrokers; i++ {
+		voters += fmt.Sprintf(",%v@confluent-local-broker-%v:%s", i+1, i+1, ports.ControllerPorts[i])
```
brianstrauch, line 347:
```suggestion
		voters += fmt.Sprintf(",%d@confluent-local-broker-%d:%s", i+1, i+1, ports.ControllerPorts[i])
```

internal/local/command_kafka_topic.go
```diff
@@ -61,3 +62,9 @@ func initKafkaRest(c *pcmd.CLICommand, cmd *cobra.Command) (*kafkarestv3.APIClie
 
 	return kafkaRestClient, clusterListData.Data[0].ClusterId, nil
 }
+
+func (c *command) getPlaintextBootstrapServers() string {
+	servers := strings.Join(c.Config.LocalPorts.PlaintextPorts, ",:")
+	fmt.Println("bs:", ":"+servers)
```
brianstrauch, line 68: :eyes:

internal/local/command_kafka_start.go
```diff
@@ -230,60 +271,90 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
 			return err
 		}
 	}
 
-	plaintextLn, err := net.Listen("tcp", ":"+c.Config.LocalPorts.PlaintextPort)
-	if err != nil {
-		freePort, err := freeport.GetFreePort()
+	for idx, port := range c.Config.LocalPorts.PlaintextPorts {
+		brokerId := idx + 1
+		plaintextLn, err := net.Listen("tcp", ":"+port)
```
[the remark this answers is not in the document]
MuweiHe: yes

MuweiHe:
@brianstrauch thanks again for the review! fixed tests and lint, refactored broker functions, added `cluster configuration` commands, and removed println's :)

DABH (approved):
Hitting approve so I don’t get alerts about this but please work with the others for approval, don’t interpret my review as an actual approval

internal/local/command_kafka_start.go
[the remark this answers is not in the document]
brianstrauch:
```suggestion
		plaintextListener, err := net.Listen("tcp", ":"+port)
```

internal/local/command_kafka_broker_update.go
```diff
@@ -0,0 +1,44 @@
...
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+)
+
+func (c *command) newKafkaBrokerUpdateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update <id>",
+		Short: "Update local Kafka broker configurations.",
+		Long:  "Update per-broker configurations.",
```
brianstrauch, line 18: `Long` is shorter than `Short` :thinking:

internal/local/command_kafka_cluster_configuration_update.go
```diff
@@ -0,0 +1,77 @@
...
+	resp, err := restClient.ConfigsV3Api.UpdateKafkaClusterConfigs(context.Background(), clusterId,
+		&kafkarestv3.UpdateKafkaClusterConfigsOpts{
+			AlterConfigBatchRequestData: optional.NewInterface(data),
+		})
+	if err != nil {
+		return kafkarest.NewError(restClient.GetConfig().BasePath, err, resp)
+	}
+
+	if output.GetFormat(cmd) == output.Human {
+		output.Printf("Updated the following broker configurations for cluster \"%s\":\n", clusterId)
```
brianstrauch, line 65: Do we necessarily need this?

internal/local/command_kafka_start.go
```diff
@@ -25,30 +25,41 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/form"
 	"github.com/confluentinc/cli/v3/pkg/log"
 	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/utils"
 )
 
+var networkCreate = types.NetworkCreate{
```
brianstrauch, line 31: This only gets used in one function. Let's define it there so it's not global in this package?

internal/local/command_kafka_start.go
```diff
@@ -127,79 +141,103 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
+	table.Add(portsData)
+	err = table.Print()
+	if err != nil {
 		return err
 	}
 
-	output.ErrPrintln("To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.")
-
+	output.Printf("Started Confluent Local containers %s.\n", utils.ArrayToCommaDelimitedString(containerIds, ""))
+	output.Println("To continue your Confluent Local experience, run `confluent local kafka topic create test` and `confluent local kafka topic produce test`.")
```
brianstrauch, line 214:
```suggestion
	output.Println("To continue your Confluent Local experience, run `confluent local kafka topic create <topic>` and `confluent local kafka topic produce <topic>`.")
```

internal/local/command_kafka_start.go
```diff
@@ -230,60 +268,90 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
+		res = append(res, nat.Port(ports.PlaintextPorts[i]+"/tcp"))
+	}
+	return res
+}
+
+func getKafkaControllerQuorumVoters(ports *config.LocalPorts, numBrokers int32) string {
+	voters := fmt.Sprintf("KAFKA_CONTROLLER_QUORUM_VOTERS=1@confluent-local-broker-1:%s", ports.ControllerPorts[0])
+	for i := int32(1); i < numBrokers; i++ {
+		voters += fmt.Sprintf(",%d@confluent-local-broker-%d:%s", i+1, i+1, ports.ControllerPorts[i])
+	}
```
brianstrauch, line 345: Make a list of strings, and then use `strings.Join()`?

internal/local/command_kafka_start.go
```diff
@@ -230,60 +268,90 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
+	return voters
+}
+
+func getKafkaRestBootstrapServers(ports *config.LocalPorts, numBrokers int32) string {
```
brianstrauch, line 349: Seems like we could generalize this function and the one above?

internal/local/command_kafka_topic.go
```diff
@@ -61,3 +62,8 @@ func initKafkaRest(c *pcmd.CLICommand, cmd *cobra.Command) (*kafkarestv3.APIClie
 
 	return kafkaRestClient, clusterListData.Data[0].ClusterId, nil
 }
+
+func (c *command) getPlaintextBootstrapServers() string {
+	servers := strings.Join(c.Config.LocalPorts.PlaintextPorts, ",:")
+	return ":" + servers
```
brianstrauch, line 68: Can we loop through the array and add ":" to each of the elements, and then use `Join()`? This solution works, but seems really hacky.

internal/local/command_kafka_start.go
MuweiHe: i took it out as per last comment said `define it outside the function`. I'm ok either way though

internal/local/command_kafka_start.go
MuweiHe: they're using different ports and different ways to construct the string. it's clearer to keep them separate.

internal/local/command_kafka_cluster_configuration_update.go
MuweiHe: removed cluster id!

internal/local/command_kafka_broker_update.go
MuweiHe: `"Update per-broker configurations for local Kafka brokers."`

internal/local/command_kafka_start.go
```diff
@@ -115,91 +133,121 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
...
 	log.CliLogger.Tracef("Successfully pulled Confluent Local image")
 
-	if err := c.prepareAndSaveLocalPorts(cmd, c.Config.IsTest); err != nil {
+	numBrokers, err := cmd.Flags().GetInt32("brokers")
+	if err != nil {
 		return err
 	}
+	if numBrokers < 1 || numBrokers > 4 {
+		return errors.NewErrorWithSuggestions("Input number of brokers out of range.", "Choose the number of brokers between 1 and 4.") // wording?
+	}
```
MuweiHe, line 144: @brianstrauch  I added a range check for between 1 and 4, but not sure about the error wording here. also is there other examples for enforcing the range of a int flag in our code?

brianstrauch: How about the following?

> `--num-brokers` must be an integer between 1 and 4, inclusive.

internal/local/command_kafka_start.go
brianstrauch: Hopefully that's not something I said 😅 Let's keep its scope as small as possible!

internal/local/command_kafka_start.go
sgagniere: +1 to the flag wording.

As for the error, how about just reflecting the flag description, like: `number of brokers must be between 1 and 4, inclusive` and leave off the suggestion.

MuweiHe: 👍

internal/local/command_kafka_start.go
```diff
@@ -25,30 +25,48 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/form"
 	"github.com/confluentinc/cli/v3/pkg/log"
 	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/utils"
 )
 
+var (
```
brianstrauch, line 31:
```suggestion
const (
```

internal/local/command_kafka_start.go
brianstrauch: :eyes:

internal/local/command_kafka_start.go
```diff
@@ -25,30 +25,48 @@ import (
...
+	confluentLocalNetworkName = "confluent-local-network"
+)
+
+var networkCreate = types.NetworkCreate{
+	CheckDuplicate: true,
+	Driver:         "bridge",
+}
+
+type portsOut struct {
+	KafkaRESTPort string `human:"Kafka Rest Port" json:"kafka_rest_port"`
```
brianstrauch, line 44:
```suggestion
	KafkaRestPort string `human:"Kafka REST Port" json:"kafka_rest_port"`
```

internal/local/command_kafka_start.go
```diff
@@ -25,30 +25,48 @@ import (
...
+	PlaintextPort string `human:"Plaintext Ports" json:"plaintext_ports"`
```
brianstrauch, line 45:
```suggestion
	PlaintextPorts string `human:"Plaintext Ports" json:"plaintext_ports"`
```

internal/local/command_kafka_start.go
```diff
@@ -25,30 +25,48 @@ import (
...
 		Args:  cobra.NoArgs,
 		RunE:  c.kafkaStart,
 	}
 
-	cmd.Flags().String("kafka-rest-port", "8082", "The port number for Kafka REST.")
-	cmd.Flags().String("plaintext-port", "", "The port number for plaintext producer and consumer clients. If not specified, a random free port will be used.")
-
+	cmd.Flags().String("kafka-rest-port", "8082", "Kafka REST port number.")
+	cmd.Flags().StringSlice("plaintext-ports", nil, "A comma-separated list of port numbers for plaintext producer and consumer clients for brokers. If not specified, random free ports will be used.")
+	cmd.Flags().Int32("brokers", 1, "Number of brokers (between 1 and 4) in the Confluent Local cluster.")
```
brianstrauch, line 65:
```suggestion
	cmd.Flags().Int32("brokers", 1, "Number of brokers (between 1 and 4) in the Confluent Local Kafka cluster.")
```

internal/local/command_kafka_start.go
```diff
@@ -115,91 +133,121 @@ func (c *Command) kafkaStart(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	output.Println("\r")
-
 	log.CliLogger.Tracef("Successfully pulled Confluent Local image")
 
-	if err := c.prepareAndSaveLocalPorts(cmd, c.Config.IsTest); err != nil {
+	numBrokers, err := cmd.Flags().GetInt32("brokers")
```
brianstrauch, line 138:
```suggestion
	brokers, err := cmd.Flags().GetInt32("brokers")
```

internal/local/command_kafka_start.go
```diff
@@ -230,60 +278,90 @@ func (c *Command) prepareAndSaveLocalPorts(cmd *cobra.Command, isTest bool) erro
...
+		envs = append(envs, getKafkaRestBootstrapServers(ports, numBrokers))
+	}
+	return envs
+}
+
+func getNatPlaintextPorts(ports *config.LocalPorts, numBrokers int32) []nat.Port {
+	res := []nat.Port{}
+	for i := 0; i < int(numBrokers); i++ {
+		res = append(res, nat.Port(ports.PlaintextPorts[i]+"/tcp"))
+	}
```
brianstrauch, line 347:
```suggestion
	for _, port := range ports.PlaintextPorts {
		res = append(res, port+"/tcp")
	}
```

internal/local/command_kafka_start.go
MuweiHe: want to rename this flag as "num-brokers"

brianstrauch: Up to you! Maybe check our other `int` and `int32` flags to see if there's a precendent?

test/fixtures/output/local/kafka/start-help.golden
```diff
@@ -1,11 +1,12 @@
...
 
 Usage:
   confluent local kafka start [flags]
 
 Flags:
-      --kafka-rest-port string   The port number for Kafka REST. (default "8082")
-      --plaintext-port string    The port number for plaintext producer and consumer clients. If not specified, a random free port will be used.
+      --kafka-rest-port string    Kafka REST port number. (default "8082")
+      --plaintext-ports strings   A comma-separated list of port numbers for plaintext producer and consumer clients for brokers. If not specified, random free ports will be used.
+      --brokers int32             Number of brokers (between 1 and 4, inclusive) in the Confluent Local Kafka cluster. (default 1)
```
brianstrauch, line 9: `--num-brokers`?

test/fixtures/output/local/kafka/help.golden
```diff
@@ -1,10 +1,12 @@
-Manage a single-node instance of Apache Kafka.
+Manage a local instance of Apache Kafka.
 
 Usage:
   confluent local kafka [command]
 
 Available Commands:
-  start       Start a single-node instance of Apache Kafka.
+  broker      Manage Kafka brokers in Confluent Local.
+  cluster     Manage Kafka cluster in Confluent Local.
```
brianstrauch, line 8: Let's match the style of the other descriptions?
```suggestion
  broker      Manage local Kafka brokers.
  cluster     Manage local Kafka clusters.
```

pkg/broker/broker_test.go
```diff
@@ -1,43 +1,54 @@
...
 
 	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
-func (suite *KafkaClusterTestSuite) TestBroker_checkAllOrBrokerIdSpecified() {
+type KafkaClusterTestSuite struct {
+	suite.Suite
+}
+
+func (suite *KafkaClusterTestSuite) TestBroker_CheckAllOrIdSpecified() {
```
brianstrauch, line 17: This is a style of testing that we're trying to move away from. Can we use integration tests instead here?

test/fixtures/output/local/kafka/start-help.golden
MuweiHe: other flags like `partitions` `cku` don't have that `--num` prefix so i decided to do it this way

brianstrauch: Awesome, shorter is better! (I agree)

pkg/broker/broker_test.go
MuweiHe: we have broker tests for --all flag. I'll just remove this one.!

internal/local/command_kafka_cluster.go
```diff
@@ -7,7 +7,7 @@ import (
 func (c *command) newKafkaClusterCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "cluster",
-		Short: "Manage Kafka cluster in Confluent Local.",
+		Short: "Manage local Kafka cluster.",
```
brianstrauch, line 10: ... or if it makes more sense: "Manage the local Kafka cluster."
```suggestion
		Short: "Manage local Kafka clusters.",
```

MuweiHe: "Manage the local Kafka cluster." sounds good

brianstrauch (approved):
thanks!!

MuweiHe:
> thanks!!

Ty!

PR #2199, opened 2023-08-16
[Draft] `topic produce` panics from ctrl-c

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Here's one way of fixing these Ctrl-C panics without a larger refactor. Mostly opening this PR to highlight the problematic code and to discuss the best way to resolve this.

We're seeing a double close panic on the `input` channel in `confluent kafka topic produce` (these issues can occur in both the cloud and on-premises versions as well as the local version). There are three places where we can close the `input` channel.

1. In the func returned by the `PrepareInputChannel` function:
```
func() {
	hasNext := scanner.Scan()
	if !hasNext {
		// Actual error.
		if scanner.Err() != nil {
			*scanErr = scanner.Err()
		}
		// Otherwise just EOF.
		close(input)
	} else {
		input <- scanner.Text()
	}
}
```
This is stored in the `scan` variable, and we run `go scan()` at the start. It is also run at the end of every loop over the `input` channel.

2.  In the goroutine that's blocked on the `signals` channel, waiting for a Ctrl-C input.
```
go func() {
	<-signals
	select {
	case <-input:
	default:
		close(input)
	}
}()
```

3. And finally, we close the input if `produce.Produce` runs into a specific error inside the loop over `input`:
```
if isProduceToCompactedTopicError {
	scanErr = err
	close(input)
	break
}
```

On Windows, hitting Ctrl-C not only unblocks the signals goroutine, but also the scan goroutine since Ctrl-C also acts like an end of line for the scanner. This was why we saw the previous panics on Windows; both goroutines would end up calling `close(input)`. 

On Unix, hitting Ctrl-C does not unblock the scanner, so it's much more rare to end up with a double close panic but it is still possible if someone times their Ctrl-C just as they're about to receive a `ProduceToCompactedTopicError`. This is of course also true for Windows.

The suggestion in this PR is to route every `close(input)` through a function with a select block to check if the `input` channel has already been closed, and a lock to prevent that check from happening multiple times at once. The `select` block alone is not enough: if two different goroutines call this function at the same time, both could determine that the `input` channel is still open, and subsequently both could call `close(input)`.

References
----------

Test & Review
-------------
still works on Windows

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/command_topic_produce.go
+++ b/internal/kafka/command_topic_produce.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/antihax/optional"
 	"github.com/spf13/cobra"
@@ -125,18 +126,14 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 	}
 
 	var scanErr error
-	input, scan := PrepareInputChannel(&scanErr)
+	input, scan, mutex := PrepareInputChannel(&scanErr)
 
 	// Trap SIGINT to trigger a shutdown.
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt)
 	go func() {
 		<-signals
-		select {
-		case <-input:
-		default:
-			close(input)
-		}
+		CloseChannel(input, mutex)
 	}()
 	// Prime reader
 	go scan()
@@ -156,7 +153,7 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 			isProduceToCompactedTopicError, err := errors.CatchProduceToCompactedTopicError(err, topic)
 			if isProduceToCompactedTopicError {
 				scanErr = err
-				close(input)
+				CloseChannel(input, mutex)
 				break
 			}
 			output.ErrPrintf(errors.FailedToProduceErrorMsg, message.TopicPartition.Offset, err)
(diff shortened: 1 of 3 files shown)


kevin-wu24:
I have questions about the first and third usages of `close(input)`: 

For the third usage, if we call `break`, why do we need to close the channel as well, since we will exit out of the for loop anyways? I think [open channels are garbage collected](https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open#:~:text=It's%20OK%20to%20leave%20a,that%20no%20more%20data%20follows.) so it would be fine to leave it open.

For the first usage of `close(input)`, can't we just remove this since `^C` will explicitly close the channel in the second usage?

sgagniere:
@kevin-wu24 That's a very good point on the third usage... closing that channel does seem redundant given that we can just break out of it. Removing it would eliminate one way of getting double close errors. And you're right about not needing to close channels; in fact I think there's a different channel in this file that's also pointlessly closed.

For the first usage, though, it's actually necessary for Unix. On Unix, Ctrl-D sends EOF to the scanner, and the first `close(input)` allows us to exit out of `produce`. It isn't an interrupt signal, so it doesn't unblock the signals channel.

kevin-wu24:
> For the first usage, though, it's actually necessary for Unix. On Unix, Ctrl-D sends EOF to the scanner, and the first `close(input)` allows us to exit out of `produce`. It isn't an interrupt signal, so it doesn't unblock the signals channel.

Is it possible to check if `data` is the EOF character within the `input` loop and then `break` if it is? If so, then we should be able to delete the first usage of `close(input)`.

sgagniere:
Closing this since @kevin-wu24 merged in the fix PR. Thanks!

PR #2200, opened 2023-08-16
[TRAFFIC-9067] - Add `confluent network delete` command

syoung-cflt:
Release Notes
-------------

New Features
* Add `confluent network delete` to delete a specific network.

Checklist
---------
   * yes: ok

What
----

Added the `confluent network delete` command

References
----------

Test & Review
-------------

Added integration tests.

Performed manual tests
1. Created a network in DEVEL
2. Successfully described the network
3. Removed the network
4. Described the network and expected an error / Checked DEVEL UI to make sure the network was deleted

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -19,6 +19,7 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
 	cmd.AddCommand(c.newDescribeCommand())
+	cmd.AddCommand(c.newDeleteCommand())
 
 	return cmd
 }
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -18,8 +18,8 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
 	}
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
-	cmd.AddCommand(c.newDescribeCommand())
 	cmd.AddCommand(c.newDeleteCommand())
+	cmd.AddCommand(c.newDescribeCommand())
 
 	return cmd
 }
--- a/internal/network/command_delete.go
+++ b/internal/network/command_delete.go
@@ -30,7 +30,7 @@ func (c *command) newDeleteCommand() *cobra.Command {
 	}
 
 	pcmd.AddForceFlag(cmd)
-	pcmd.AddClusterFlag(cmd, c.AuthenticatedCLICommand)
+	pcmd.AddContextFlag(cmd, c.CLICommand)
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 
 	return cmd
--- a/internal/network/command_delete.go
+++ b/internal/network/command_delete.go
@@ -1,0 +1,57 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/form"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/resource"
+)
+
+func (c *command) newDeleteCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete <id>",
+		Short: "Delete a network.",
+		Args:  cobra.ExactArgs(1),
+		// TODO: Implement autocompletion after List Network is implemented.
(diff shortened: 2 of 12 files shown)


internal/network/command_delete.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (c *command) delete(cmd *cobra.Command, args []string) error {
+	id := args[0]
+	environmentId, err := c.Context.EnvironmentId()
+	if err != nil {
+		return err
+	}
+
+	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmYesNoMsg, resource.Network, id)
```
sgagniere, line 46:
```suggestion
	_, err := c.V2Client.GetNetwork(environmentId, id)
	if err != nil {
		return err
	}

	promptMsg := fmt.Sprintf(errors.DeleteResourceConfirmYesNoMsg, resource.Network, id)
```

We generally want to verify that a resource exists before we ask the user to confirm its deletion (some resources are currently missing this check, but those will be fixed in a pending PR), which we can do by adding a describe call.

internal/network/command.go
```diff
@@ -19,6 +19,7 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
 	cmd.AddCommand(c.newDescribeCommand())
+	cmd.AddCommand(c.newDeleteCommand())
```
brianstrauch, line 22: Please alphabetize!
```suggestion
	
	cmd.AddCommand(c.newDeleteCommand())
	cmd.AddCommand(c.newDescribeCommand())
```

sgagniere (changes requested):
Looks good! But there's just one thing we should change here.

internal/network/command_delete.go
```diff
@@ -0,0 +1,57 @@
...
+		Use:   "delete <id>",
+		Short: "Delete a network.",
+		Args:  cobra.ExactArgs(1),
+		// TODO: Implement autocompletion after List Network is implemented.
+		// ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE: c.delete,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Delete Confluent network "n-abcde1".`,
+				Code: `confluent network delete n-abcde1`,
```
brianstrauch, line 27:
```suggestion
				Code: "confluent network delete n-abcde1",
```

internal/network/command_delete.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (c *command) newDeleteCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete <id>",
+		Short: "Delete a network.",
+		Args:  cobra.ExactArgs(1),
+		// TODO: Implement autocompletion after List Network is implemented.
+		// ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
+		RunE: c.delete,
+		Example: examples.BuildExampleString(
```
brianstrauch, line 24: Feel free to delete this example, since it doesn't use any flags and is just as helpful as the usage string, IMO.

test/fixtures/output/network/delete-network-not-exist.golden
```diff
@@ -0,0 +1 @@
+Error: The network n-invalid was not found.: 404 Not Found
```
brianstrauch, line 1: Can you update your API responses so they match Go error style (start with lowercase, no period at the end)?

If not, we plan to write a helper function in the CLI to clean up these error strings (you're not the only team who does this)

internal/network/command_delete.go
syoung-cflt: Thank you for reviewing the PR! I just added the `GetNetwork` call to verify that a resource exists.

internal/network/command.go
syoung-cflt: Fixed!

internal/network/command_delete.go
syoung-cflt: Removed the example as suggested below.

internal/network/command_delete.go
syoung-cflt: Makes sense! The example seems redundant. Thank you!

sgagniere (approved):
:+1:

test/fixtures/output/network/delete-network-not-exist.golden
brianstrauch: No worries if not, I think a lot of teams prefer uppercase with trailing period so it looks nice on the UI side.

brianstrauch (approved):

test/fixtures/output/network/delete-network-not-exist.golden
ejagade: we haven't been following strict guidelines in formatting the error messages in API responses.
I'm in favor of standardizing it but it could be a non-trivial effort to fix existing API responses. 
Let's take an AI here and check with the team.

ejagade (approved):

PR #2201, opened 2023-08-16
Formatting improvements

brianstrauch:
Checklist
---------
   * yes: ok

What
----
* Remove instances of `--flag=value` in examples (prefer `--flag value`)
* Add version to versioned package alias (Example: `mds` -> `mdsv1`)
* Put "next command" message at bottom of `confluent local kafka`, and print to stderr

Test & Review
-------------
Updated integration tests where necessary


sgagniere (approved):
lgtm

PR #2202, opened 2023-08-16
[CLI-2686] Deprecate "schema_registry_clusters" configuration field

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Thanks @MuweiHe for pointing out that as of #2134, we are essentially not using Schema Registry API keys in the CLI anymore. Rewrote `confluent kafka client-config create` to grab the Schema Registry cluster from the backend (this improves consistency) instead of the configuration file, and stopped marking Schema Registry API keys as "current" in `api-key list`. This allows us to deprecate the Schema Registry field in the configuration file!

Test & Review
-------------
Tests still pass

### DIFF
--- a/internal/kafka/command_clientconfig_create.go
+++ b/internal/kafka/command_clientconfig_create.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	srcmv2 "github.com/confluentinc/ccloud-sdk-go-v2/srcm/v2"
 	ckafka "github.com/confluentinc/confluent-kafka-go/kafka"
 	srsdk "github.com/confluentinc/schema-registry-sdk-go"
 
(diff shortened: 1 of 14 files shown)


internal/kafka/command_clientconfig_create.go
```diff
@@ -208,16 +208,8 @@ func (c *clientConfigCommand) setKafkaCluster(cmd *cobra.Command, configFile str
 }
 
 func (c *clientConfigCommand) setSchemaRegistryCluster(cmd *cobra.Command, configFile string) (string, error) {
```
MuweiHe, line 210: this function just calls c.getSchemaRegistryCluster(cmd) and there's no configFile involved. Can we just remove this function and replace it with c.getSchemaRegistryCluster(cmd)?

brianstrauch: Look further down?

MuweiHe: oh i think github cut the remaining - -

MuweiHe (approved):

PR #2203, opened 2023-08-16
[CLI-2642] Resolve race condition resulting in double channel close panic in `topic produce` commands

kevin-wu24:
Release Notes
-------------
Bug Fixes
- Resolved a close of closed channel panic in `topic produce`.

Checklist
---------
   * yes: ok

What
----
Building off of the ideas in https://github.com/confluentinc/cli/pull/2199, this race condition seems to have been resolved. 

Basically, sending an interrupt signal via `^C` now writes `\u0004` (Unicode encoding of `^D`) to the `input` channel, which can normally never receive this value, since writing `^D` to `stdIn` will just close the channel. 

Now, in the loop over `input`, we `break` if we get an error from the Scanner or if the value received from the `input` channel is `^D` (meaning a shutdown has been triggered with `^C`). This means the `input` channel is only closed once, when the user explicitly types `^D` into `stdIn`. 

References
----------
https://confluentinc.atlassian.net/browse/CLI-2642

Test & Review
-------------
Passes `make test` and manual testing. So far I haven't been able to recreate the `panic` and `topic produce` otherwise works as intended.

### DIFF
--- a/internal/kafka/command_topic.go
+++ b/internal/kafka/command_topic.go
@@ -17,6 +17,9 @@ import (
 
 const numPartitionsKey = "num.partitions"
 
+// EOF Unicode encoding for Ctrl+D (^D) character
+var EOF = "\u0004"
+
 type command struct {
 	*pcmd.AuthenticatedCLICommand
 	clientID string
--- a/internal/kafka/command_topic.go
+++ b/internal/kafka/command_topic.go
@@ -2,6 +2,9 @@ package kafka
 
 import (
 	"fmt"
+	"os"
+	"os/signal"
+	"runtime"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -13,13 +16,14 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/config"
 	"github.com/confluentinc/cli/v3/pkg/errors"
 	"github.com/confluentinc/cli/v3/pkg/log"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/serdes"
 )
 
-const (
-	// EOF Unicode encoding for Ctrl+D (^D) character
-	EOF              = "\u0004"
-	numPartitionsKey = "num.partitions"
-)
+// EOF Unicode encoding for Ctrl+D (^D) character
+const EOF = "\u0004"
+
+const numPartitionsKey = "num.partitions"
 
 type command struct {
 	*pcmd.AuthenticatedCLICommand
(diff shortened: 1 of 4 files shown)


internal/kafka/command_topic.go
```diff
@@ -15,7 +15,11 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/log"
 )
 
-const numPartitionsKey = "num.partitions"
+const (
+	// EOF Unicode encoding for Ctrl+D (^D) character
+	EOF              = "\u0004"
```
brianstrauch, line 20: Based on your description, it doesn't matter what this value is equal to. Let's just make it "EOF"?

internal/kafka/command_topic.go
```diff
@@ -15,7 +15,11 @@ import (
...
+	numPartitionsKey = "num.partitions"
```
brianstrauch, line 21: These constants are unrelated and shouldn't be grouped together.

internal/kafka/command_topic_produce.go
```diff
@@ -132,20 +132,21 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
...
-			close(input)
-		}
+		input <- EOF
 	}()
 	// Prime reader
 	go scan()
 
 	deliveryChan := make(chan ckafka.Event)
 	for data := range input {
 		if data == "" {
```
brianstrauch, line 142: `switch data`?

internal/kafka/command_topic_produce_onprem.go
```diff
@@ -171,15 +171,20 @@ func (c *command) produceOnPrem(cmd *cobra.Command, args []string) error {
...
 
 	deliveryChan := make(chan ckafka.Event)
 	for data := range input {
 		if data == "" {
+			if scanErr != nil {
+				break
+			}
 			go scan()
 			continue
+		} else if data == EOF {
```
brianstrauch, line 186: I wonder if there's a good way to combine the cloud/on-prem implementations? It's annoying that we have to keep updating both.

internal/kafka/command_topic.go
kevin-wu24: If I make it "EOF", typing "EOF" into `stdIn` will `break` out of the loop, since we're checking the value of `data`. I actually tried exactly this haha!

brianstrauch: Ah, got it. Does pressing Ctrl-C still print "^C" to the terminal?

kevin-wu24: yup!

internal/kafka/command_topic_produce_onprem.go
kevin-wu24: Yeah, I think that would be good, but it might be better to put in another PR, since there's other `kafka` and even `kafka topic` commands that whose cloud/on-prem implementations share a lot of code, and that would end up being a lot of code refactoring.

internal/kafka/command_topic_produce.go
kevin-wu24: I don't think we can `switch data` because we need to `break` out of the loop over `input`.

brianstrauch (approved):
nice job!!

PR #2209, opened 2023-08-18
Simplify KafkaREST setup for on-prem commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
I noticed that `getClusterIdForRestRequests` and `initKafkaRest` are always called together for (non-local) commands.

So I rolled the logic of `getClusterIdForRestRequests` into `initKafkaRest` to save an extra function call across all of the onprem commands.

This is also a small step towards unifying the cloud and onprem command implementations.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/kafka_rest_onprem.go
+++ b/internal/kafka/kafka_rest_onprem.go
@@ -12,21 +12,31 @@ import (
 
 	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
 	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/kafkarest"
 	"github.com/confluentinc/cli/v3/pkg/output"
 )
 
-func initKafkaRest(c *pcmd.AuthenticatedCLICommand, cmd *cobra.Command) (*kafkarestv3.APIClient, context.Context, error) {
+func initKafkaRest(c *pcmd.AuthenticatedCLICommand, cmd *cobra.Command) (*kafkarestv3.APIClient, context.Context, string, error) {
 	url, err := getKafkaRestUrl(cmd)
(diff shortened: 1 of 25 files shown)


internal/kafka/kafka_rest_onprem.go
```diff
@@ -12,21 +12,31 @@ import (
...
 )
 
-func initKafkaRest(c *pcmd.AuthenticatedCLICommand, cmd *cobra.Command) (*kafkarestv3.APIClient, context.Context, error) {
+func initKafkaRest(c *pcmd.AuthenticatedCLICommand, cmd *cobra.Command) (*kafkarestv3.APIClient, context.Context, string, error) {
 	url, err := getKafkaRestUrl(cmd)
 	if err != nil { // require the flag
-		return nil, nil, err
+		return nil, nil, "", err
 	}
 	kafkaREST, err := c.GetKafkaREST()
```
kevin-wu24, line 24: I think we might be able to unify everything under the `c.GetKafkaREST()` function? 

We can call the `kafka_rest_onprem` helpers to resolve the server URL with `c.command` before returning the `KafkaREST` struct in `kafka_rest.go` if we're on-prem. Finally, we can just add an on-prem case for getting the cluster ID. This way everything can be accessed from the same struct whether we're on-prem or cloud.

sgagniere: Yup, that's the goal. This is just to clean it up a bit in the meantime.

MuweiHe (approved):

PR #2210, opened 2023-08-21
Feat/kfs 1078 add service account flag

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Adds a use command to the `service-account` command and adds the service account flag to the `flink shell` command

### DIFF
--- a/internal/flink/command_shell.go
+++ b/internal/flink/command_shell.go
@@ -10,6 +10,7 @@ import (
 	client "github.com/confluentinc/cli/v3/pkg/flink/app"
 	"github.com/confluentinc/cli/v3/pkg/flink/test/mock"
 	"github.com/confluentinc/cli/v3/pkg/flink/types"
+	"github.com/confluentinc/cli/v3/pkg/output"
 	ppanic "github.com/confluentinc/cli/v3/pkg/panic-recovery"
 )
 
@@ -23,7 +24,8 @@ func (c *command) newShellCommand(cfg *config.Config, prerunner pcmd.PreRunner) 
 	}
 
 	c.addComputePoolFlag(cmd)
-	cmd.Flags().String("identity-pool", "", "Identity pool ID.")
+	cmd.Flags().String("identity-pool", "", "Identity pool ID (deprecated).")
+	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	cmd.Flags().String("database", "", "The database which will be used as default database. When using Kafka, this is the cluster display name.")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
@@ -117,10 +119,21 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
 		return err
 	}
 	if identityPool == "" {
-		if c.Context.GetCurrentIdentityPool() == "" {
-			return errors.NewErrorWithSuggestions("no identity pool set", "Set a persistent identity pool with `confluent iam pool use` or pass the `--identity-pool` flag.")
-		}
 		identityPool = c.Context.GetCurrentIdentityPool()
+	}
+
+	serviceAccount, err := cmd.Flags().GetString("service-account")
+	if err != nil {
+		return err
+	}
+	if serviceAccount == "" {
+		serviceAccount = c.Context.GetCurrentServiceAccount()
+	}
+
+	if serviceAccount == "" && identityPool == "" {
+		output.ErrPrintln("Warning: no service account provided. To ensure that your statements run continuously, " +
+			"switch to using a service account instead of your user identity by providing a service account with `confluent iam service-account use` or `--service-account`. " +
+			"Otherwise, statements will stop running after 4 hours.")
 	}
 
 	database, err := cmd.Flags().GetString("database")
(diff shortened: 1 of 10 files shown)


internal/flink/command_shell.go
```diff
@@ -23,7 +24,8 @@ func (c *command) newShellCommand(cfg *config.Config, prerunner pcmd.PreRunner)
 	}
 
 	c.addComputePoolFlag(cmd)
-	cmd.Flags().String("identity-pool", "", "Identity pool ID.")
+	cmd.Flags().String("identity-pool", "", "Identity pool ID (deprecated).")
```
brianstrauch, line 27: This is still EA, so can we just make a breaking change here? It would be odd if we removed the feature flag and this flag was marked as deprecated.

brianstrauch (changes requested):

internal/flink/command_shell.go
swist: This is a breaking change and identity pools will not be supported in OP. This is due to change of direction from a few months ago regarding how we do auth. We have to mark this as deprecated and remove it soon

brianstrauch: > We have to mark this as deprecated and remove it soon

:+1: As long as it's removed before GA!

internal/iam/command_service_account_use.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/resource"
+)
+
+func (c *serviceAccountCommand) newUseCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "use <id>",
+		Short:             "Choose a service account to be used in subsequent commands.",
+		Long:              "Choose a service account to be used in subsequent commands which support passing a service account with the `--service-account` flag.",
```
brianstrauch, line 16: There are a lot of commands that allow the `--service-account` flag. Have you validated that this works with commands other than the Flink commands?

YannickPferr: Yeah I did this at first, specifically for api-key `create` and `list`, but then I ran into the problem again that you can't `unset` a key and since service-account is an optional flag for those commands, I decided to remove it again. What do you think, should I still make it work with those commands?

brianstrauch: The long description seems to imply that `confluent iam service-account use` can be used in conjunction with any command that has `--service-account`, so let's either 1) support that or 2) remove this command.

My vote is for #2 based on:
> the problem again that you can't unset a key

YannickPferr: Alright, I'll just move it under `flink`, that way it won't affect the other commands but we can still use it for Flink specific ones.

YannickPferr: I've added it now on the other commands as well. I also implemented a `unset` command for service accounts.

internal/flink/command_shell.go
brianstrauch: Can you create and link a ticket here for removing this flag before GA (if you don't plan to remove it in this PR), so we don't forget?

internal/iam/command_service_account.go
```diff
@@ -36,6 +36,8 @@ func newServiceAccountCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd.AddCommand(c.newDescribeCommand())
 	cmd.AddCommand(c.newListCommand())
 	cmd.AddCommand(c.newUpdateCommand())
+	cmd.AddCommand(c.newUseCommand())
+	cmd.AddCommand(c.newUnsetCommand())
```
brianstrauch, line 40: Please alphabetize

internal/iam/command_service_account_unset.go
```diff
@@ -0,0 +1,41 @@
...
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/resource"
+)
+
+func (c *serviceAccountCommand) newUnsetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "unset",
+		Short:             "Unsets a service account that was set with `use`.",
```
brianstrauch, line 15:
```suggestion
		Short:            "Unset the current service account."
		Long:             "Unset the current service account that was set with `confluent iam service-account use`.",
```

internal/iam/command_service_account_unset.go
```diff
@@ -0,0 +1,41 @@
...
+
+	pcmd.AddContextFlag(cmd, c.CLICommand)
+
+	return cmd
+}
+
+func (c *serviceAccountCommand) unset(_ *cobra.Command, args []string) error {
+	serviceAccountToUnset := c.Context.GetCurrentServiceAccount()
+	if serviceAccountToUnset == "" {
+		output.Printf(errors.UnsetResourceNothingToUnsetMsg, resource.ServiceAccount)
```
brianstrauch, line 28: Thoughts on making this a no-op?

internal/flink/command_shell.go
YannickPferr: Added a todo and linked to the ticket. https://confluentinc.atlassian.net/browse/KFS-941

internal/iam/command_service_account.go
YannickPferr: Done!

internal/iam/command_service_account_unset.go
YannickPferr: Done!

internal/iam/command_service_account_unset.go
YannickPferr: Works for me too!

internal/flink/command_shell.go
```diff
@@ -117,12 +120,23 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
...
+	if err != nil {
+		return err
+	}
+	if serviceAccount == "" {
+		serviceAccount = c.Context.GetCurrentServiceAccount()
+	}
+
+	if serviceAccount == "" && identityPool == "" {
+		output.ErrPrintln("Warning: no service account provided. To ensure that your statements run continuously, " +
+			"switch to using a service account instead of your user identity by providing a service account with `confluent iam service-account use` or `--service-account`. " +
```
brianstrauch, line 136: Previous text was a bit redundant.
```suggestion
			"switch to using a service account instead of your user identity with `confluent iam service-account use` or `--service-account`. " +
```

brianstrauch (approved):

PR #2216, opened 2023-08-22
[CLI-2360] Revoking Auth0 refresh token on cloud logout

kevin-wu24:
Release Notes
-------------
Bug Fixes
- `confluent logout` revokes the refresh token when logging out of Confluent Cloud

Checklist
---------
   * yes: ok

What
----
Running `confluent logout` for cloud users now makes a request to `DELETE /api/sessions`.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2360
https://github.com/confluentinc/ccloud-sdk-go-v1-public/pull/24

Test & Review
-------------
Unit and integration tests still pass
Request returns a reply with code `200 OK`.


internal/command.go
```diff
@@ -121,8 +120,8 @@ func NewConfluentCommand(cfg *config.Config) *cobra.Command {
 	cmd.AddCommand(kafka.New(cfg, prerunner))
 	cmd.AddCommand(ksql.New(cfg, prerunner))
 	cmd.AddCommand(local.New(prerunner))
-	cmd.AddCommand(login.New(cfg, prerunner, ccloudClientFactory, mdsClientManager, netrcHandler, loginCredentialsManager, loginOrganizationManager, authTokenHandler))
-	cmd.AddCommand(logout.New(cfg, prerunner, netrcHandler))
+	cmd.AddCommand(login.NewLoginCommand(cfg, prerunner, ccloudClientFactory, mdsClientManager, netrcHandler, loginCredentialsManager, loginOrganizationManager, authTokenHandler))
+	cmd.AddCommand(login.NewLogoutCommand(cfg, prerunner, ccloudClientFactory, mdsClientManager, netrcHandler, loginCredentialsManager, loginOrganizationManager, authTokenHandler))
```
brianstrauch, line 124: If there's shared code, let's put it in `pkg`. Let's not put the `logout` command in the `login` package... that's just confusing.

internal/login/command_login_test.go
```diff
@@ -113,8 +112,7 @@ var (
 				return nil, nil
 			}
 		},
-		SetCloudClientFunc: func(arg0 *ccloudv1.Client) {
-		},
+		SetCloudClientFunc: func(arg0 *ccloudv1.Client) {},
```
brianstrauch, line 115:
```suggestion
		SetCloudClientFunc: func(_ *ccloudv1.Client) {},
```

pkg/auth/auth_token_handler.go
```diff
@@ -169,3 +170,18 @@ func login(client *ccloudv1.Client, req *ccloudv1.AuthenticateRequest) (*ccloudv
 		return client.Auth.Login(req)
 	}
 }
+
+func (a *AuthTokenHandlerImpl) RevokeRefreshToken(clientFactory CCloudClientFactory, url string, credentials *Credentials) error {
+	req := &ccloudv1.AuthenticateRequest{IdToken: credentials.AuthToken}
+	client := clientFactory.JwtHTTPClientFactory(context.Background(), credentials.AuthToken, url)
```
brianstrauch, line 176: Why can't we just use the default `c.Client` in login/command.go?

kevin-wu24: I mentioned this in slack, but that was the reason I was getting Error 401. I was using the `AnonHTTPClientFactory` before, which I assume is what you mean by default client. However, the backend handler [expects to have a userId in the header](https://github.com/confluentinc/cc-flow-service/blob/b16ae7af553340623a2e3c87daee0728028e34ea/api/v1/handler.go#L168), so using the `JwtHTTPClient` worked.

brianstrauch: Hmm, seems like this is due to `confluent logout` being an unauthenticated command. Maybe we should switch it to an authenticated command (we should be able to assume that the user is logged in before using it)

pkg/config/mock.go
```diff
@@ -20,7 +20,7 @@ var (
 	mockURL                = "http://test"
 	usernameCredentialName = fmt.Sprintf("username-%s-%s", mockEmail, mockURL)
 	apiKeyCredentialName   = fmt.Sprintf("api-key-%s", kafkaAPIKey)
-	mockContextName        = fmt.Sprintf("login-%s-%s", mockEmail, mockURL)
+	MockContextName        = fmt.Sprintf("login-%s-%s", mockEmail, mockURL)
```
MuweiHe, line 23: where else is this var being used?

internal/logout/command.go
```diff
@@ -15,46 +16,63 @@ import (
...
+		AuthenticatedCLICommand:  pcmd.NewAuthenticatedCLICommand(cmd, prerunner),
+		cfg:                      cfg,
+		ccloudClientFactory:      ccloudClientFactory,
+		netrcHandler:             netrcHandler,
+		loginCredentialsManager:  loginCredentialsManager,
+		loginOrganizationManager: loginOrganizationManager,
+		authTokenHandler:         authTokenHandler,
+	}
 	if cfg.IsCloudLogin() {
 		context = "Confluent Cloud"
```
MuweiHe, line 46:
```suggestion
		context = "Confluent Cloud"
		AuthenticatedCLICommand:  pcmd.NewAuthenticatedCLICommand(cmd, prerunner),
```
move it down for readibility

MuweiHe:
Have you manually tested, that IsSSO field in saved credentials is correct?

pkg/config/mock.go
kevin-wu24: In a `logout` unit test.

internal/logout/command.go
kevin-wu24: If we're not logged in, we still want to have this struct field set. I originally had the code this way, but it resulted in a panic rather than an error when the user was not logged in.

kevin-wu24:
> Have you manually tested, that IsSSO field in get ccloud credentials is always correct?

yup! It's the same code that used to exist in `login/command.go`, but I just moved it to `pkg` since it is shared.

internal/login/command_test.go
```diff
@@ -903,7 +902,7 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
 	if !isCloud {
 		mdsConfig := mdsv1.NewConfiguration()
 		mdsClient = mdsv1.NewAPIClient(mdsConfig)
-		mdsClient.TokensAndAuthenticationApi = &mdsmock.TokensAndAuthenticationApi{
+		mdsClient.TokensAndAuthenticationApi = &mdsMock.TokensAndAuthenticationApi{
```
brianstrauch, line 905: Package names should always be lowercase.
```suggestion
		mdsClient.TokensAndAuthenticationApi = &mdsmock.TokensAndAuthenticationApi{
```

internal/login/command_test.go
```diff
@@ -937,8 +936,38 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
...
+	if !isCloud {
+		mdsConfig := mdsv1.NewConfiguration()
+		mdsClient = mdsv1.NewAPIClient(mdsConfig)
+		mdsClient.TokensAndAuthenticationApi = &mdsMock.TokensAndAuthenticationApi{
+			GetTokenFunc: func(ctx context.Context) (mdsv1.AuthenticationResponse, *http.Response, error) {
+				return mdsv1.AuthenticationResponse{
+					AuthToken: testToken1,
+					TokenType: "JWT",
+					ExpiresIn: 100,
+				}, nil, nil
```
brianstrauch, line 952:
```suggestion
				res := mdsv1.AuthenticationResponse{
					AuthToken: testToken1,
					TokenType: "JWT",
					ExpiresIn: 100,
				}
				return res, nil, nil
```

pkg/auth/auth_token_handler.go
```diff
@@ -169,3 +170,11 @@ func login(client *ccloudv1.Client, req *ccloudv1.AuthenticateRequest) (*ccloudv
 		return client.Auth.Login(req)
 	}
 }
+
+func (a *AuthTokenHandlerImpl) RevokeRefreshToken(client *ccloudv1.Client, req *ccloudv1.AuthenticateRequest) (*ccloudv1.AuthenticateReply, error) {
```
brianstrauch, line 174: Does this function need to be inside `AuthTokenHandlerImpl`? (I don't see `a` being used anywhere here) It looks like you're introducing a dependency for no reason.

internal/login/command_test.go
```diff
@@ -937,8 +936,38 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
 	return loginCmd, cfg
 }
 
-func newLogoutCmd(cfg *config.Config, netrcHandler netrc.NetrcHandler) (*cobra.Command, *config.Config) {
-	logoutCmd := logout.New(cfg, climock.NewPreRunnerMock(nil, nil, nil, nil, cfg), netrcHandler)
+func newLogoutCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterface, isCloud bool, req *require.Assertions, netrcHandler netrc.NetrcHandler, authTokenHandler pauth.AuthTokenHandler, loginCredentialsManager pauth.LoginCredentialsManager, loginOrganizationManager pauth.LoginOrganizationManager, contextName string) (*cobra.Command, *config.Config) {
```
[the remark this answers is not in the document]
kevin-wu24: Yeah I forgot to remove some of these, as they aren't used anymore.

internal/login/command_test.go
```diff
@@ -811,10 +810,10 @@ func TestLoginWithExistingContext(t *testing.T) {
 		ctx.KafkaClusterContext.SetActiveKafkaCluster(kafkaCluster.ID)
 
 		// Executing logout
-		logoutCmd, _ := newLogoutCmd(cfg, mockNetrcHandler)
+		logoutCmd, cfg1 := newLogoutCmd(auth, userInterface, s.isCloud, req, mockNetrcHandler, AuthTokenHandler, mockLoginCredentialsManager, LoginOrganizationManager, ctx.Name)
```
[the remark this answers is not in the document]
kevin-wu24: This needs to be changed so that the `newLogoutCmd` doesn't generate a new config, but use the same `config` as above.

internal/logout/command.go
```diff
@@ -64,3 +78,18 @@ func (c *Command) logout(cmd *cobra.Command, _ []string) error {
...
+
+func (c *command) revokeCCloudRefreshToken() (*ccloudv1.AuthenticateReply, error) {
+	ctx := c.Config.Config.Context()
+	contextState := c.Config.Config.ContextStates[ctx.Name]
+	if err := contextState.DecryptContextStateAuthToken(ctx.Name); err != nil {
+		return nil, err
+	}
+
+	req := &ccloudv1.AuthenticateRequest{IdToken: contextState.AuthToken}
+	if sso.IsOkta(c.Client.BaseURL) {
```
brianstrauch, line 90: Can we get the URL from the configuration instead? This seems hacky

internal/logout/command_test.go
```diff
@@ -197,8 +216,34 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
 	return loginCmd, cfg
 }
 
-func newLogoutCmd(cfg *config.Config, netrcHandler netrc.NetrcHandler) (*cobra.Command, *config.Config) {
-	logoutCmd := New(cfg, climock.NewPreRunnerMock(nil, nil, nil, nil, cfg), netrcHandler)
+func newLogoutCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterface, isCloud bool, req *require.Assertions, netrcHandler netrc.NetrcHandler, authTokenHandler pauth.AuthTokenHandler, contextName string) (*cobra.Command, *config.Config) {
+	config.SetTempHomeDir()
+	cfg := config.AuthenticatedConfigMockWithContextName(contextName)
+	var mdsClient *mdsv1.APIClient
+	if !isCloud {
```
brianstrauch, line 223: This looks like duplicate code. Can we de-duplicate?

internal/login/command_test.go
```diff
@@ -937,9 +936,33 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
...
+	if !isCloud {
+		mdsConfig := mdsv1.NewConfiguration()
+		mdsClient = mdsv1.NewAPIClient(mdsConfig)
+		mdsClient.TokensAndAuthenticationApi = &mdsmock.TokensAndAuthenticationApi{
+			GetTokenFunc: func(_ context.Context) (mdsv1.AuthenticationResponse, *http.Response, error) {
+				return mockAuthResponse, nil, nil
+			},
+		}
+	}
+	ccloudClientFactory := &climock.CCloudClientFactory{
```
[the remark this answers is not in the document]
kevin-wu24: It looks like the `NewPreRunnerMock` function that takes in this variable runs whether or not `isCloud` is true.

internal/logout/command_test.go
kevin-wu24: Is there a place in the code base where shared unit test helpers already exist? That way we can just have one `newLoginCmd` and `newLogoutCmd` too.

internal/login/command_test.go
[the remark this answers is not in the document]
brianstrauch: Sounds like we should change that too?

internal/logout/command_test.go
brianstrauch: Not yet, but since these tests span two packages, I'd expect it to go somewhere in pkg/

kevin-wu24: I don't think I can put them in `pkg` because of import cycles, but I think the best place to put them would be `mock`.

kevin-wu24: I tried putting this in existing packags that made sense in `pkg/` (`auth` and `mock`) but both result in circular imports, so I had to create a new directory unfortunately.

internal/logout/command.go
```diff
@@ -64,3 +78,18 @@ func (c *Command) logout(cmd *cobra.Command, _ []string) error {
 	output.Println(errors.LoggedOutMsg)
 	return nil
 }
+
+func (c *command) revokeCCloudRefreshToken() (*ccloudv1.AuthenticateReply, error) {
+	ctx := c.Config.Config.Context()
```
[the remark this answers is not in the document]
kevin-wu24: Yeah we can just pass ctx into the function.

internal/logout/command_test.go
```diff
@@ -21,6 +18,7 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/config"
 	pmock "github.com/confluentinc/cli/v3/pkg/mock"
 	"github.com/confluentinc/cli/v3/pkg/netrc"
+	testhelp "github.com/confluentinc/cli/v3/pkg/test-helpers"
```
brianstrauch, line 21:
```suggestion
	testhelpers "github.com/confluentinc/cli/v3/pkg/test-helpers"
```

pkg/test-helpers/mock.go
```diff
@@ -0,0 +1,55 @@
...
+	var mdsClient *mdsv1.APIClient
+	mdsConfig := mdsv1.NewConfiguration()
+	mdsClient = mdsv1.NewAPIClient(mdsConfig)
+	mdsClient.TokensAndAuthenticationApi = &mdsmock.TokensAndAuthenticationApi{
+		GetTokenFunc: func(_ context.Context) (mdsv1.AuthenticationResponse, *http.Response, error) {
+			return mdsv1.AuthenticationResponse{
+				AuthToken: token,
+				TokenType: "JWT",
+				ExpiresIn: 100,
+			}, nil, nil
```
brianstrauch, line 51:
```suggestion
			res := mdsv1.AuthenticationResponse{
				AuthToken: token,
				TokenType: "JWT",
				ExpiresIn: 100,
			}
			return res, nil, nil
```

pkg/test-helpers/mock.go
```diff
@@ -0,0 +1,55 @@
...
+				User: userInterface,
+			}
+		},
+	}
+}
+
+func NewMdsClientMock(token string) *mdsv1.APIClient {
+	var mdsClient *mdsv1.APIClient
+	mdsConfig := mdsv1.NewConfiguration()
+	mdsClient = mdsv1.NewAPIClient(mdsConfig)
```
brianstrauch, line 44:
```suggestion
	mdsConfig := mdsv1.NewConfiguration()
	mdsClient := mdsv1.NewAPIClient(mdsConfig)
```

pkg/test-helpers/mock.go
```diff
@@ -0,0 +1,55 @@
+package test_helpers
```
brianstrauch, line 1: `test_helpers` is a really generic name and has the potential to get really big the more things we move into here. Why not put this in `pkg/mock`?

kevin-wu24: I get a circular import unfortunately.

brianstrauch: try to fix it another way?

kevin-wu24: I can move it to `mock`, just not `pkg/mock`.

internal/login/command_test.go
```diff
@@ -899,47 +902,42 @@ func TestValidateUrl(t *testing.T) {
...
-			}, Auth: auth, User: userInterface}
-		},
+		mdsClient = climock.NewMdsClientMock(testToken1)
+		prerunner = climock.NewPreRunnerMock(nil, nil, mdsClient, nil, cfg)
+	} else {
+		ccloudClientFactory = climock.NewCCloudClientFactoryMock(auth, userInterface, req)
+		prerunner = climock.NewPreRunnerMock(ccloudClientFactory.AnonHTTPClientFactory(ccloudURL), nil, nil, nil, cfg)
 	}
+
 	mdsClientManager := &climock.MDSClientManager{
```
brianstrauch, line 917: Likewise, this looks like it should only be declared inside the `if`

internal/login/command_test.go
```diff
@@ -899,47 +902,42 @@ func TestValidateUrl(t *testing.T) {
...
+
+	if !isCloud {
+		mdsClient = climock.NewMdsClientMock(testToken1)
+		prerunner = climock.NewPreRunnerMock(nil, nil, mdsClient, nil, cfg)
+	} else {
+		ccloudClientFactory = climock.NewCCloudClientFactoryMock(auth, userInterface, req)
+		prerunner = climock.NewPreRunnerMock(ccloudClientFactory.AnonHTTPClientFactory(ccloudURL), nil, nil, nil, cfg)
+	}
+	logoutCmd := logout.New(cfg, prerunner, netrcHandler, authTokenHandler)
+	return logoutCmd
```
brianstrauch, line 940:
```suggestion
	return logout.New(cfg, prerunner, netrcHandler, authTokenHandler)
```

internal/logout/command.go
```diff
@@ -15,46 +19,56 @@ import (
...
+	if ctx != nil {
 		username, err := c.netrcHandler.RemoveNetrcCredentials(c.cfg.IsCloudLogin(), c.Config.Config.Context().GetNetrcMachineName())
 		if err == nil {
 			log.CliLogger.Warnf(errors.RemoveNetrcCredentialsMsg, username, c.netrcHandler.GetFileName())
 		} else if !strings.Contains(err.Error(), "login credentials not found") && !strings.Contains(err.Error(), "keyword expected") {
 			// return err when other than NetrcCredentialsNotFoundErrorMsg or parsing error
 			return err
 		}
+
+		if isCCloud := ccloudv2.IsCCloudURL(ctx.Platform.Server, c.cfg.IsTest); isCCloud {
```
brianstrauch, line 67:
```suggestion
		if ccloudv2.IsCCloudURL(ctx.Platform.Server, c.cfg.IsTest) {
```

internal/logout/command_test.go
```diff
@@ -112,9 +109,10 @@ var (
 func TestLogout(t *testing.T) {
 	req := require.New(t)
 	clearCCloudDeprecatedEnvVar(req)
-	cfg := config.AuthenticatedCloudConfigMock()
+	cfg := config.AuthenticatedConfigMockWithContextName(config.MockContextName)
 	contextName := cfg.Context().Name
-	logoutCmd, cfg := newLogoutCmd(cfg, mockNetrcHandler)
+	// run login command
```
brianstrauch, line 114: This comment seems wrong

internal/login/command_test.go
```diff
@@ -904,21 +904,22 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
 	cfg := config.New()
 	var ccloudClientFactory *climock.CCloudClientFactory
 	var mdsClient *mdsv1.APIClient
```
brianstrauch, line 906: Should be defined inside the `if` since it's only used inside that scope

kevin-wu24: It's used outside of the `if` in line 923

brianstrauch: `mdsClientManager` is needed there, but `mdsClient` isn't

kevin-wu24: oh haha my bad, yeah.

pkg/cmd/prerunner.go
```diff
@@ -190,6 +190,10 @@ func (r *PreRun) Authenticated(command *AuthenticatedCLICommand) func(*cobra.Com
 		setContextErr := r.setAuthenticatedContext(command)
 		if setContextErr != nil {
 			if _, ok := setContextErr.(*errors.NotLoggedInError); ok {
+				if cmd.Use == "logout" {
```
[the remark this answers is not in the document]
kevin-wu24: I tried some other scenarios and it looks like other errors can surface. If the login token expires, `logout` throws an `expired token` error, and if you try logging out again after that you get the `not logged in` error again (which is because although our current context exists, but we're no longer considered logged into it). These errors result from the fact that we're trying to have `logout` work exactly the same as before when the user is not logged in, despite it now being an `authenticatedCLICommand`.

I think we should probably just wait until v4 to add this since it doesn't look like all the possible breaking changes from errors can be avoided without putting several of these hacky exceptions for the `logout` command in the prerunner.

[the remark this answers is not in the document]
brianstrauch: > I think we should probably just wait until v4 to add this since it doesn't look like all the possible breaking changes from errors can be avoided without putting several of these hacky exceptions for the logout command in the prerunner.

Labeling as v4!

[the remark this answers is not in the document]
kevin-wu24: Going to revert the last two commits since we want the breaking changes to be visible.

sgagniere:
I switched the target from main to v4 and worked out the merge conflicts, so I think this is ready for review again.

internal/login/command_test.go
```diff
@@ -847,44 +851,41 @@ func TestValidateUrl(t *testing.T) {
...
+func newLogoutCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterface, isCloud bool, req *require.Assertions, authTokenHandler pauth.AuthTokenHandler, cfg *config.Config) *cobra.Command {
+	var prerunner pcmd.PreRunner
+
+	if !isCloud {
+		mdsClient := climock.NewMdsClientMock(testToken1)
+		prerunner = climock.NewPreRunnerMock(nil, nil, mdsClient, nil, cfg)
+	} else {
+		ccloudClientFactory := climock.NewCCloudClientFactoryMock(auth, userInterface, req)
+		prerunner = climock.NewPreRunnerMock(ccloudClientFactory.AnonHTTPClientFactory(ccloudURL), nil, nil, nil, cfg)
+	}
```
brianstrauch, line 885: Switch these blocks and remove the `!`

brianstrauch (approved):

PR #2220, opened 2023-08-23
[CLI-2710] Update packaging during release

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Introduce a new target to update the packaging repo with the current CLI version after every release.

Test & Review
-------------
Tested with `DRY_RUN=true`.

### DIFF
--- a/mk-files/release.mk
+++ b/mk-files/release.mk
@@ -165,7 +165,7 @@ update-muckrake:
 	git clone git@github.com:confluentinc/muckrake.git $(MUCKRAKE) && \
 	cd $(MUCKRAKE) && \
 	git fetch --all && \
-	branch="cli-$${version}" && \
+	branch=bump-cli && \
 	base=$$(git branch --remote --format "%(refname:short)" | sed -n "s|^origin/\([1-9][0-9]*\.[0-9][0-9]*\.x\)$$|\1|p" | tail -1) && \
 	git checkout $$base && \
 	git checkout $$branch || git checkout -b $$branch && \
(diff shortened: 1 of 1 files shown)


mk-files/release.mk
```diff
@@ -179,3 +181,30 @@ update-muckrake:
...
+	version=$$(ls release-notes | $(SED) "s/.json$$//" | sort --version-sort | tail -1) && \
+	git clone git@github.com:confluentinc/packaging.git $(PACKAGING) && \
+	cd $(PACKAGING) && \
+	git fetch --all && \
+	branch="bump-cli" && \
+	base=$$(git branch --remote --format "%(refname:short)" | sed -n "s|^origin/\([1-9][0-9]*\.[0-9][0-9]*\.x\)$$|\1|p" | tail -1) && \
+	git checkout $$base && \
+	git checkout $$branch || git checkout -b $$branch && \
+	$(SED) -i "s|cli_BRANCH=\".*\"|cli_BRANCH=\"$${version}\"|" settings.sh && \
+	$(SED) -i "s|CLI_VERSION=.*|CLI_VERSION=$${version}|" release_testing/bin/smoke_test.sh && \
```
sgagniere, line 202: It looks like one of these should have a `v` and one should not. This code seems to be setting both to the same value, unless I'm reading it incorrectly.

brianstrauch: > It looks like one of these should have a v and one should not.

Correct.

sgagniere (approved):

brianstrauch:
Waiting until next release to merge, since we don't want to open a PR in packaging this week since we missed the CP 7.5 deadline.

PR #2224, opened 2023-08-26
Add opt-in colors to the CLI

brianstrauch:
Release Notes
-------------
New Features
- Optionally enable text highlighting with `confluent configuration update enable_colors true`

Checklist
---------
   * yes: ok

What
----
Styling for errors (red), code snippets (red with gray background), and links (blue with underline). Inference for dark and light terminals. Colors are not displayed by default can be enabled with `confluent configuration update enable_colors true`.

I considered adding highlighting for resources, i.e. "lkc-123456", but there are too many false positives at the moment (other text in quotes).

Screenshot 2023-08-25 at 5 10 44 PM

Screenshot 2023-08-25 at 5 11 09 PM

Screenshot 2023-08-25 at 6 39 56 PM

Test & Review
-------------
Manual verification


internal/audit-log/command_config_migrate.go
```diff
@@ -79,8 +79,8 @@ func (c *configCommand) migrate(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 	for _, warning := range warnings {
-		output.ErrPrintln(warning)
-		output.ErrPrintln()
+		color.ErrPrintln(c.Config.EnableColor, warning)
+		color.ErrPrintln(c.Config.EnableColor, "")
```
sgagniere, line 83: It looks like the `output` package is still around in this PR and being used in a few places; what's the general guideline for when to use `output` and when to use `color`? Or could we fully replace output with color?

internal/admin/command_payment_describe.go
```diff
@@ -31,20 +31,20 @@ func (c *command) describe(cmd *cobra.Command, _ []string) error {
 
 	marketplace := user.GetOrganization().GetMarketplace()
 	if marketplace.GetPartner() != ccloudv1.MarketplacePartner_UNKNOWN {
-		output.Printf("Organization is currently linked to %s Marketplace account.\n", marketplace.GetPartner())
+		color.Printf(c.Config.EnableColor, "Organization is currently linked to %s Marketplace account.\n", marketplace.GetPartner())
```
sgagniere, line 34: Is there any good way to hide this parameter? Like, what if we make a color or printer struct holding this bool, make the print functions into methods for that struct, and add it to the command structs, so that we could write something like `c.Color.Printf("Organization is...")`.

sgagniere (commented):
Nice; this will be a good quality of life upgrade. But I have some questions:

pkg/cmd/errors.go
```diff
@@ -0,0 +1,24 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/confluentinc/cli/v3/pkg/color"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+)
+
+func CheckErr(enableColor bool, msg interface{}) {
```
MuweiHe, line 10: This might be put together in pkg errors?

internal/audit-log/command_config_migrate.go
MuweiHe: +1

brianstrauch: Good idea, updated to fully replace color with output to avoid future confusion!

internal/admin/command_payment_describe.go
brianstrauch: Not all print statements have access to `c` at the moment, but I think they all should be rewritten to... I think after we do that then making this change would be a good idea. In the meantime, let's just have a single set of functions, `output.Print*()` as we discussed in the comment above

pkg/cmd/errors.go
brianstrauch: I put this function in this package since it (and other functions in this package) mirrors a function inside the `cobra` package.

internal/schema-registry/schema_utilities.go
```diff
@@ -64,7 +64,7 @@ func RegisterSchemaWithAuth(cmd *cobra.Command, schemaCfg *RegisterSchemaConfigs
 			return 0, err
 		}
 	} else {
-		output.Printf(errors.RegisteredSchemaMsg, response.Id)
+		output.Printf(false, errors.RegisteredSchemaMsg, response.Id)
```
sgagniere, line 67: I'm a bit unclear still on when we use `c.Config.EnableColor` and when we hardcode `false`.

sgagniere (commented):
A few more things:

internal/schema-registry/schema_utilities.go
brianstrauch: The majority of the time, `c` is not available and making it available would result in a huge rewrite. My plan is to gradually rewrite these one by one. For the time being, we use `false` which corresponds to no highlighting, even when colors are enabled.

sgagniere: Oh woops, you addressed this last time. Gotcha. Lgtm after the other comment.

internal/command.go
```diff
@@ -62,10 +62,11 @@ import (
...
 	cmd := &cobra.Command{
-		Use:     pversion.CLIName,
-		Short:   fmt.Sprintf("%s.", pversion.FullCLIName),
-		Long:    getLongDescription(cfg),
-		Version: cfg.Version.Version,
+		Use:           pversion.CLIName,
+		Short:         fmt.Sprintf("%s.", pversion.FullCLIName),
+		Long:          getLongDescription(cfg),
+		Version:       cfg.Version.Version,
+		SilenceErrors: true,
```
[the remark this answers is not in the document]
brianstrauch: No, this is because the by default, errors are printed by cobra in black and white. I was hoping I could disable this and print my own (colored) errors manually. Unfortunately it's proving to be more difficult than I thought...

sgagniere (approved):

PR #2227, opened 2023-08-30
[TRAFFIC-9317] - Add `confluent network peering list` command

syoung-cflt:
What
----

* Added `confluent network peering` to manage peering connections and `confluent network peering list` to display peering connections

References
----------

Test & Review
-------------

Added integration tests.

Performed manual tests

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -76,9 +76,8 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd.AddCommand(c.newDeleteCommand())
 	cmd.AddCommand(c.newDescribeCommand())
 	cmd.AddCommand(c.newListCommand())
-	cmd.AddCommand(c.newUpdateCommand())
-
 	cmd.AddCommand(newPeeringCommand(prerunner))
+	cmd.AddCommand(c.newUpdateCommand())
 
 	return cmd
 }
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -71,11 +71,14 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
 	}
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+
 	cmd.AddCommand(c.newCreateCommand())
 	cmd.AddCommand(c.newDeleteCommand())
 	cmd.AddCommand(c.newDescribeCommand())
 	cmd.AddCommand(c.newListCommand())
 	cmd.AddCommand(c.newUpdateCommand())
+
+	cmd.AddCommand(newPeeringCommand(prerunner))
 
 	return cmd
 }
--- a/internal/network/command_peering.go
+++ b/internal/network/command_peering.go
@@ -1,0 +1,53 @@
+package network
+
+import (
+	"github.com/spf13/cobra"
+
+	networkingv1 "github.com/confluentinc/ccloud-sdk-go-v2/networking/v1"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+)
+
+type peeringCommand struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+type peeringHumanOut struct {
+	Id        string `human:"ID"`
+	Name      string `human:"Name"`
+	NetworkId string `human:"Network ID"`
+	Cloud     string `human:"Cloud"`
+	Phase     string `human:"Phase"`
+}
+
+type peeringSerializedOut struct {
+	Id        string `serialized:"id"`
+	Name      string `serialized:"name"`
+	NetworkId string `serialized:"network_id"`
+	Cloud     string `serialized:"cloud"`
+	Phase     string `serialized:"phase"`
+}
+
+func newPeeringCommand(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "peering",
+		Short:       "Manage peering connections.",
+		Args:        cobra.NoArgs,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
+	}
+
+	c := &peeringCommand{AuthenticatedCLICommand: pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+
+	cmd.AddCommand(c.newPeeringListCommand())
+
+	return cmd
+}
+
+func (c *peeringCommand) getPeerings() ([]networkingv1.NetworkingV1Peering, error) {
+	environmentId, err := c.Context.EnvironmentId()
+	if err != nil {
+		return nil, err
+	}
+
+	return c.V2Client.ListPeerings(environmentId)
+}
--- a/internal/network/command_peering.go
+++ b/internal/network/command_peering.go
@@ -46,10 +46,9 @@ type peeringSerializedOut struct {
 
 func newPeeringCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:         "peering",
-		Short:       "Manage peering connections.",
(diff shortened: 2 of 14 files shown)


internal/network/command_peering.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	networkingv1 "github.com/confluentinc/ccloud-sdk-go-v2/networking/v1"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+)
+
+type peeringCommand struct {
```
syoung-cflt, line 11: I see two approaches in the code base for this.

[command_consumer.go in the stream-share package](https://github.com/confluentinc/cli/blob/main/internal/stream-share/command_consumer.go) is using `command` from [command.go](https://github.com/confluentinc/cli/blob/dc3ab0bcd3b7a2073c77acbe2f432a0856afc804/internal/stream-share/command.go#L9)

[command_cluster.go in the kafka package](https://github.com/confluentinc/cli/blob/dc3ab0bcd3b7a2073c77acbe2f432a0856afc804/internal/kafka/command_cluster.go#L29) defines its own `clusterCommand`.

Which one should we use?

internal/network/command_peering_list.go
```diff
@@ -0,0 +1,84 @@
...
+			return err
+		}
+
+		if output.GetFormat(cmd) == output.Human {
+			list.Add(&peeringHumanOut{
+				Id:        peering.GetId(),
+				Name:      peering.Spec.GetDisplayName(),
+				NetworkId: peering.Spec.Network.GetId(),
+				Cloud:     cloud,
+				Phase:     peering.Status.GetPhase(),
```
syoung-cflt, line 56: @ejagade @kabraham13 
For the `list` command, I think it makes sense to output only non cloud specific values, otherwise the output would take up too much horizontal screen space.

For the `describe` command, I am going to output the following cloud specific values
For `AWS` -  `account`, `vpc`, `routes`, `customer_region`
For `GCP` - `project`, `vpc_network`
For `Azure` - `tenant`, `vnet`, `customer_region`

What do you guys think?

internal/network/command_peering.go
MuweiHe: i think it's of better practice to use `peeringCommand` which is more specific. Usually we enforce using `<use>Command` when the struct get more complex. Good catch though.

syoung-cflt: Thank you! I will keep `peeringCommand` and follow this convention for the rest of our commands.

internal/network/command_peering_list.go
```diff
@@ -0,0 +1,84 @@
...
+			list.Add(&peeringSerializedOut{
+				Id:        peering.GetId(),
+				Name:      peering.Spec.GetDisplayName(),
+				NetworkId: peering.Spec.Network.GetId(),
+				Cloud:     cloud,
+				Phase:     peering.Status.GetPhase(),
+			})
+		}
+	}
+	list.Filter([]string{"Id", "Name", "NetworkId", "Cloud", "Phase"})
```
MuweiHe, line 68: can this be defined outside of function?

syoung-cflt: Are you referring to the string slice? It seems like this is the convention for most of the list commands, including the `confluent network list` we just implemented.

internal/network/command_peering.go
```diff
@@ -0,0 +1,69 @@
...
+	AzureVNet     string   `serialized:"azure_vnet,omitempty"`
+	AzureTenant   string   `serialized:"azure_tenant,omitempty"`
+}
+
+func newPeeringCommand(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "peering",
+		Short:       "Manage peering connections.",
+		Args:        cobra.NoArgs,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
brianstrauch, line 52: Annotations are inherited from parent commands, so this likely isn't needed.
```suggestion
```

internal/network/command.go
```diff
@@ -71,12 +71,15 @@ func New(prerunner pcmd.PreRunner) *cobra.Command {
...
 
 	c := &command{pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+
 	cmd.AddCommand(c.newCreateCommand())
 	cmd.AddCommand(c.newDeleteCommand())
 	cmd.AddCommand(c.newDescribeCommand())
 	cmd.AddCommand(c.newListCommand())
 	cmd.AddCommand(c.newUpdateCommand())
 
+	cmd.AddCommand(newPeeringCommand(prerunner))
```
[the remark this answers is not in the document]
syoung-cflt: sounds good, i thought it would look nicer if we separate the crudl command and the other subcommands 😄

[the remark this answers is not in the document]
syoung-cflt: Updated!

internal/network/command_peering_list.go
ejagade: One option I thought is - for each row print oneof the following, as a peering record can belong to only one cloud

oneof {
{awsAccount, awsVPC, region}, //aws
{gcpProject, vpc_network} //gcp
{azureTenant, azureVnet, region} //azure
}
this would add at most 3 columns to the output

But a customer can have with multiple peering records in more than 1 cloud. this would mean `list` command will contain 1 set of columns for each cloud

I'm ok with adding these in describe command

internal/network/command_peering_list.go
```diff
@@ -0,0 +1,84 @@
...
+	return list.Print()
+}
+
+func getCloud(peering networkingv1.NetworkingV1Peering) (string, error) {
+	cloud := peering.Spec.GetCloud()
+
+	if cloud.NetworkingV1AwsPeering != nil {
+		return "AWS", nil
```
ejagade, line 76: can we switch to string constants and make sure all the commands use the same constant. (helps with a consistent behavior while printing CSP names). they could be already one in CLI repo

ejagade (approved):

internal/network/command_peering_list.go
syoung-cflt: Yea, I don't see one in the CLI repo. I will create it in the `describe` PR that's coming up shortly

internal/network/command_peering_list.go
```diff
@@ -0,0 +1,84 @@
...
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+func (c *peeringCommand) newPeeringListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Display peering connections in the current environment.",
```
brianstrauch, line 18: Nit: They won't be in the current environment if you pass `--environment`.
```suggestion
		Short: "List peering connections.",
```

brianstrauch (approved):

internal/network/command_peering_list.go
syoung-cflt: I suppose we should change https://github.com/confluentinc/cli/blob/f94149e7a198c4ef3a19f4e3d2d4bd1e678d8b35/internal/network/command_list.go#L17 as well?

internal/network/command_peering_list.go
syoung-cflt: I will define it outside of the function if we end up reusing it for `describe`!

internal/network/command_peering_list.go
syoung-cflt: Updated. will merge after CI passes

PR #2230, opened 2023-08-31
Remove Link prefix validation

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Resolve an issue in `confluent kafka link delete` preventing users from deleting certain cluster links

Checklist
---------
   * yes: ok

What
----
Cluster links do not have a set prefix, so `ClusterLinkPrefix` causes the prefix validation to fail when deleting cluster links unless the link name happens to starts with `link-`.

This PR removes this variable from the resource/prefix maps, so that prefix validation will do nothing (as intended for resources with no prefix) when deleting cluster links.

References
----------

Test & Review
-------------
Added a new integration test case with a link that doesn't start with this prefix.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -46,7 +46,6 @@ const (
 )
 
 const (
-	ClusterLinkPrefix           = "link"
 	EnvironmentPrefix           = "env"
 	IdentityPoolPrefix          = "pool"
 	IdentityProviderPrefix      = "op"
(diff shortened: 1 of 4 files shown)


kpatelatwork (approved):
LGTM on a passing build

PR #2242, opened 2023-09-07
Resolve integer deserialization issue in `topic consume`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix an issue causing `confluent kafka topic consume` to deserialize integers to incorrect values when using `--key-format integer` or `value-format integer`
- Resolve a panic that could occur in `confluent kafka topic consume` when using `--key-format integer` and `--print-key`
- `--parse-key` must now be specified when using `confluent kafka topic produce --key-format`

Checklist
---------
   * yes: ok

What
----
Integers were being serialized with Little Endian but deserialized with Big Endian, resulting in wrong values.

While investigating I also discovered that you can end up with a panic if you consume any messages without an integer key after specifying `confluent kafka topic consume <topic-name> --key-format integer --print-key -b`, so this PR does two things to address this:
- Require that users use `--parse-key` if they provide `--key-format`
- Deserialize empty keys to the empty string for integers

References
----------

Test & Review
-------------
manual testing

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/command_topic_produce.go
+++ b/internal/kafka/command_topic_produce.go
@@ -89,6 +89,20 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	keyFormat, err := cmd.Flags().GetString("key-format")
+	if err != nil {
+		return err
+	}
+
+	parseKey, err := cmd.Flags().GetBool("parse-key")
+	if err != nil {
+		return err
+	}
+
+	if keyFormat != "string" && !parseKey {
+		return errors.New("`--parse-key` must be set when the key format is not \"string\"")
+	}
+
 	configFile, err := cmd.Flags().GetString("config-file")
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


MuweiHe:
be a bit more specific about `Resolve a panic that could occur in confluent kafka topic consume`? like what combination of flags are being used.

internal/kafka/command_topic_produce.go
```diff
@@ -89,6 +89,15 @@ func (c *command) produce(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	parseKey, err := cmd.Flags().GetBool("parse-key")
+	if err != nil {
+		return err
+	}
+
+	if cmd.Flags().Changed("key-format") && !parseKey {
+		return errors.New("`--parse-key` must be set when `key-format` is set")
```
MuweiHe, line 98: how about the other way, if parse-key is set?

sgagniere: That actually works, it defaults to assuming string keys. That's why I went with this instead of just marking them as required together.

MuweiHe (approved):

PR #2243, opened 2023-09-07
Temporarily use page size 99 for v2 api list calls

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix a bug in `confluent iam user invitation list` where only the first 100 results were listed

Checklist
---------
   * yes: ok

What
----

Backend api has a off-by-1 error and page size 100 doesn't work. temply use 99 for all calls to minimize the change. expect a backend fix next week so we can revert this change.

References
----------

Test & Review
-------------

tested with 110 invites and page size 99 returns all results.

Open Questions / Follow-ups
---------------------------


sgagniere (commented):
Let's just do this only for the invitations route, instead of across the board.

sgagniere (approved):

PR #2246, opened 2023-09-08
Switch remaining `confluent kafka mirror` commands to using the cloud version of the kafkarest sdk

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Switch `mirror describe | list` back to using the proper SDK, which has been updated to include the previously missing `LINK_PAUSED`, `LINK_FAILED`, and `SOURCE_UNAVAILABLE` statuses.

References
----------

Test & Review
-------------
Updated the list test to include the three missing statuses.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -190,3 +190,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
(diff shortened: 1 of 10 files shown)


brianstrauch (approved):
Thanks!!

brianstrauch:
Should we mention this in the "Bug Fixes" section of the release notes?

sgagniere:
@brianstrauch On the bug fixes: No, because this won't have any visible difference for users. The current on-prem SDK, which is also missing some statuses, doesn't do any checking and is able to handle these statuses anyway.

PR #2247, opened 2023-09-08
Remove suggestions to directly modify config.json

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
Since we now have a `confluent configuration update` command, we should no longer suggest direct modification of `config.json` in our suggestions.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


brianstrauch (approved):
thanks!!

PR #2250, opened 2023-09-11
Added commands and tests for custom connector plugin

bedeks:
Release Notes
-------------

New Features
- Added `confluent connect custom-plugin` commands

Bug Fixes
- Fix a bug where custom connectors no longer require "connector.class" in `confluent connect cluster create`

Checklist
---------
   * yes: ok

What
----
The changes are to support cli functionality for custom connector plugins. We have added crud functionality for the user to be able to manage their custom connector plugins.

References
----------
https://confluentinc.atlassian.net/browse/CPU-836

Test & Review
-------------
All paths were tested manually and have integration tests for

Open Questions / Follow-ups
---------------------------


brianstrauch:
Please format the "New Features" section of the release notes to use backticks for command names!

go.mod
```diff
@@ -112,6 +112,8 @@ require (
 	github.com/c-bata/go-prompt v0.2.6 // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/cloudflare/circl v1.3.3 // indirect
+	github.com/confluentinc/ccloud-sdk-go-v2/ccp v0.0.1 // indirect
+	github.com/confluentinc/ccloud-sdk-go-v2/connect-custom-plugin v0.0.1 // indirect
```
brianstrauch, line 116: What's the difference between these dependencies? Looks like you forgot to run `go mod tidy`, btw!

internal/cmd/connect/command.go
```diff
@@ -25,6 +25,7 @@ func New(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command {
 	cmd.AddCommand(newClusterCommand(cfg, prerunner))
 	cmd.AddCommand(newEventCommand(prerunner))
 	cmd.AddCommand(newPluginCommand(cfg, prerunner))
+	cmd.AddCommand(newCustomPluginCommand(cfg, prerunner))
```
brianstrauch, line 28: Please insert in alphabetical order!

internal/cmd/connect/command_custom_plugin.go
```diff
@@ -0,0 +1,44 @@
+package connect
+
+import (
+	v1 "github.com/confluentinc/cli/internal/pkg/config"
+	"github.com/spf13/cobra"
```
brianstrauch, line 5: Please run `make lint`

internal/cmd/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,83 @@
...
+	description, err := cmd.Flags().GetString("description")
+	documentationLink, err := cmd.Flags().GetString("documentation-link")
+	pluginFileName, err := cmd.Flags().GetString("plugin-file")
+	connectorClass, err := cmd.Flags().GetString("connector-class")
+	connectorType, err := cmd.Flags().GetString("connector-type")
+	sensitivePropertiesString, err := cmd.Flags().GetString("sensitive-properties")
+
+	extension := filepath.Ext(pluginFileName)[1:]
+	if extension != "zip" && extension != "jar" {
+		return errors.Errorf(errors.PluginInvalidExtensionErrorMsg)
```
sgagniere, line 59: We want to inline any error or suggestion message which is only used once, instead of adding it to the errors package.

internal/cmd/connect/command_custom_plugin.go
```diff
@@ -0,0 +1,44 @@
...
+	*pcmd.AuthenticatedCLICommand
+}
+
+type customPluginOut struct {
+	Id                        string   `human:"ID" serialized:"id"`
+	Name                      string   `human:"Name" serialized:"name"`
+	Description               string   `human:"Description" serialized:"description"`
+	ConnectorClass            string   `human:"Connector Class" serialized:"connector_class"`
+	ConnectorType             string   `human:"Connector Type" serialized:"connector_type"`
+	SensitiveConfigProperties []string `human:"Sensitive Config Properties" serialized:"sensitive_config_properties"`
```
brianstrauch, line 20: `confluent pipeline update` has a similar field called "Secret Names". Can we imitate that or come up with a shorter name?

internal/cmd/connect/command_custom_plugin.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if cfg.IsCloudLogin() {
+		c.AuthenticatedCLICommand = pcmd.NewAuthenticatedCLICommand(cmd, prerunner)
+		cmd.AddCommand(c.newCreateCommand())
+		cmd.AddCommand(c.newListCommand())
+		cmd.AddCommand(c.newDescribeCommand())
+		cmd.AddCommand(c.newDeleteCommand())
+		cmd.AddCommand(c.newUpdateCommand())
+	} else {
+		c.AuthenticatedCLICommand = pcmd.NewAuthenticatedWithMDSCLICommand(cmd, prerunner)
```
brianstrauch, line 40: Why aren't there any on-prem commands listed here?

internal/cmd/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,83 @@
...
+	cmd := &cobra.Command{
+		Use:         "create <name>",
+		Short:       "Create a custom connector plugin.",
+		Args:        cobra.ExactArgs(1),
+		RunE:        c.createCustomPlugin,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "Create a custom connector plugin ",
+				Code: "confluent connect custom-plugin create plugin_name --config-file config.json",
```
brianstrauch, line 28: Please take a look at other commands for examples of how to format the example field.
```suggestion
				Text: `Create custom connector plugin "my-plugin".`,
				Code: "confluent connect custom-plugin create my-plugin --config-file config.json",
```

internal/cmd/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,83 @@
...
+	cmd.Flags().String("documentation-link", "", "document link of plugin")
+	cmd.Flags().String("connector-class", "", "connector class of custom plugin")
+	cmd.Flags().String("connector-type", "", "connector type of custom plugin")
+	cmd.Flags().String("sensitive-properties", "", "sensitive properties config of custom plugin")
+
+	cobra.CheckErr(cmd.MarkFlagRequired("connector-class"))
+	cobra.CheckErr(cmd.MarkFlagRequired("connector-type"))
+
+	cobra.CheckErr(cmd.MarkFlagFilename("plugin-file", "zip", "jar"))
+	cobra.CheckErr(cmd.MarkFlagRequired("plugin-file"))
```
brianstrauch, line 44: Please list all required flags together, in the same order that they're defined in. Take a look at some of our other commands for examples.

internal/cmd/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,83 @@
...
+	return cmd
+}
+
+func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []string) error {
+	displayName := args[0]
+	description, err := cmd.Flags().GetString("description")
```
brianstrauch, line 50: Why are we initializing this error but ignoring it?

internal/cmd/connect/command_custom_plugin_delete.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/confluentinc/cli/internal/pkg/output"
+	"github.com/confluentinc/cli/internal/pkg/resource"
+	"github.com/spf13/cobra"
+)
+
+func (c *customPluginCommand) newDeleteCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "delete <id>",
+		Short:       "Delete a custom connector plugin.",
+		Args:        cobra.ExactArgs(1),
```
sgagniere, line 18: This command will need a rewrite to support multiple deletion after you merge main into this branch, unfortunately. There will be a `README.md` under `pkg/deletion` that explains what to do; you can also look at various other delete.go files for examples.

internal/cmd/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,83 @@
...
+		return errors.Errorf(errors.PluginInvalidExtensionErrorMsg)
+	}
+
+	resp, err := c.V2Client.GetPresignedUrl(extension)
+	if err != nil {
+		return err
+	}
+
+	err = uploadFile(resp.GetUploadUrl(), pluginFileName, resp.GetUploadFormData())
+	if err != nil {
```
brianstrauch, line 68: Reducing the scope of local variables can help prevent bugs
```suggestion
	if err := uploadFile(resp.GetUploadUrl(), pluginFileName, resp.GetUploadFormData()); err != nil {
```

internal/pkg/errors/error_message.go
```diff
@@ -61,6 +61,7 @@ const (
 	InvalidCloudErrorMsg               = "error defining plugin on given Kafka cluster"
 	InvalidCloudSuggestions            = "To list available connector plugin types, use `confluent connect plugin list`."
 	ConnectLogEventsNotEnabledErrorMsg = "Connect Log Events are not enabled for this organization"
+	PluginInvalidExtensionErrorMsg     = "Only ZIP/JAR plugin file is allowed"
```
sgagniere, line 64:
```suggestion
	PluginInvalidExtensionErrorMsg     = "only ZIP/JAR plugin file is allowed"
```
We only capitalize proper nouns in our error messages.

internal/cmd/connect/command_custom_plugin_delete.go
```diff
@@ -0,0 +1,52 @@
...
+		),
+	}
+
+	pcmd.AddForceFlag(cmd)
+	pcmd.AddContextFlag(cmd, c.CLICommand)
+
+	return cmd
+}
+
+func (c *customPluginCommand) delete(cmd *cobra.Command, args []string) error {
```
brianstrauch, line 35: Please take a look at other `delete` commands. I believe you need to define an "existence function" now. cc @sgagniere

sgagniere (commented):
Thanks! I've left some comments.

internal/cmd/connect/command_custom_plugin_describe.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/spf13/cobra"
+)
+
+func (c *customPluginCommand) newDescribeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "describe <id>",
+		Short:       "Describe a custom connector plugin.",
+		Args:        cobra.ExactArgs(1),
+		RunE:        c.describe,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
brianstrauch, line 16: This shouldn't be needed if the parent has the same annotation (I think that's what you should be doing...)

internal/cmd/connect/command_custom_plugin_describe.go
```diff
@@ -0,0 +1,51 @@
...
+	out := &customPluginOut{
+		Id:                        plugin.GetId(),
+		Name:                      plugin.GetDisplayName(),
+		Description:               plugin.GetDescription(),
+		ConnectorClass:            plugin.GetConnectorClass(),
+		ConnectorType:             plugin.GetConnectorType(),
+		SensitiveConfigProperties: plugin.GetSensitiveConfigProperties(),
+	}
+	if output.GetFormat(cmd) == output.Human {
+		output.Println("Plugin Details")
```
brianstrauch, line 45: Our other commands avoid printing strings like this.
```suggestion
```

internal/cmd/connect/command_custom_plugin_update.go
```diff
@@ -0,0 +1,48 @@
...
+		Args:        cobra.ExactArgs(1),
+		RunE:        c.update,
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
+	}
+	cmd.Flags().String("name", "", "name of plugin")
+	cmd.Flags().String("description", "", "description of plugin")
+	cmd.Flags().String("documentation-link", "", "document link of plugin")
+	cmd.Flags().String("sensitive-properties", "", "sensitive properties config of custom plugin")
+
+	pcmd.AddContextFlag(cmd, c.CLICommand)
```
brianstrauch, line 24: Please look at other commands for examples, but we could separate the flags with some whitespace here.
```suggestion

	cmd.Flags().String("name", "", "name of plugin")
	cmd.Flags().String("description", "", "description of plugin")
	cmd.Flags().String("documentation-link", "", "document link of plugin")
	cmd.Flags().String("sensitive-properties", "", "sensitive properties config of custom plugin")

	pcmd.AddContextFlag(cmd, c.CLICommand)
	
```

internal/cmd/connect/utils.go
```diff
@@ -20,9 +25,15 @@ func getConfig(cmd *cobra.Command) (*map[string]string, error) {
...
+	connectorType := options["confluent.connector.type"]
+	if connectorType != "" {
+		connectorType = "MANAGED"
+	}
+
 	_, nameExists := options["name"]
 	_, classExists := options["connector.class"]
-	if !nameExists || !classExists {
+
+	if connectorType != "CUSTOM" && (!nameExists || !classExists) {
```
brianstrauch, line 36: `connectorType != "CUSTOM"` will always evaluate to true. Based on the logic above, `connectorType` should either be `""` or `"Managed"`.

brianstrauch (changes requested):

internal/cmd/connect/command_custom_plugin_delete.go
sgagniere: Looks like we doubled up on the feedback in some parts; I mentioned the README above.

internal/cmd/connect/command_custom_plugin.go
bedeks: This feature is only in cloud currently

internal/cmd/connect/utils.go
bedeks: Good catch. Fixed it. Thanks!

internal/cmd/connect/command_custom_plugin.go
brianstrauch: What's the timeline for supporting this feature on-premises? We can probably just wait add the on-prem section until we support it.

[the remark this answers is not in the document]
bedeks: No plans for CP yet so changed line 27 to `pcmd.RequireNonAPIKeyCloudLogin`

internal/cmd/connect/command_custom_plugin_delete.go
bedeks: Rewrote this with the new delete

go.mod
bedeks: Ran go mod tidy in the end which removed the ccp dependency

internal/cmd/connect/command_custom_plugin.go
bedeks: Since connect has been calling it sensitive properties for connectors for a long time it might be best to keep the name consistent because of user behavior. I have removed config from the name but if you still think the name is too big I can check in with our pm.

internal/cmd/connect/command_custom_plugin.go
bedeks: We have not discussed the timeline for adding the feature on-premises. I do not think on-prem even makes sense for this product.

internal/cmd/connect/command_custom_plugin_create.go
bedeks: I had used this example for formatting https://github.com/confluentinc/cli/blob/main/internal/connect/command_cluster_create.go#L28

Is there a different example I should base my formatting of?

internal/cmd/connect/command_custom_plugin_describe.go
bedeks: I was actually following this https://github.com/confluentinc/cli/blob/main/internal/connect/command_cluster_describe.go#L98 

But I can remove it!

internal/connect/utils.go
```diff
@@ -71,3 +82,41 @@ func parseConfigFile(filename string) (map[string]string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	err = writer.Close()
+	if err != nil {
+		return err
+	}
+
+	_, err = sling.New().Base(url).Set("Content-Type", writer.FormDataContentType()).Post("").Body(&buffer).ReceiveSuccess(nil)
```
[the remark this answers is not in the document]
bedeks: This uploads it to a presigned s3 url which is unique for everytime someone requests an upload url so it can't be supported by a sdk.

internal/connect/command_custom_plugin_delete.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return err
+	}
+
+	existenceFunc := func(id string) bool {
+		_, ok := pluginIdToName[id]
+		return ok
+	}
+
+	if err := deletion.ValidateAndConfirmDeletion(cmd, args, existenceFunc, resource.CustomConnectorPlugin, pluginIdToName[args[0]]); err != nil {
```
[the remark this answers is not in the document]
sgagniere: This is the version which uses name validation for single arguments instead of a yes/no prompt, so only the first name is needed.

If multiple are provided, then the function won't actually use it.

We'll remove this and change everything to yes/no in v4.

internal/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,96 @@
...
+		return err
+	}
+	if connectorType, err = cmd.Flags().GetString("connector-type"); err != nil {
+		return err
+	}
+	if sensitivePropertiesString, err = cmd.Flags().GetString("sensitive-properties"); err != nil {
+		return err
+	}
+
+	extension := filepath.Ext(pluginFileName)[1:]
```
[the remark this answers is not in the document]
bedeks: Added a check

internal/connect/command_custom_plugin_delete.go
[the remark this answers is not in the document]
bedeks: I think that is just the name of the plugin. My understanding is if there are more than 1 to delete then it would anyway convert to a yes or no question instead of typing the name it is only valid if there is 1 plugin to delete. I had based it on this https://github.com/confluentinc/cli/blob/main/internal/connect/command_cluster_delete.go#L60

sgagniere (commented):
Just one more thing that I didn't notice the first time about the list sdk function.

And two small nitpicks.

sgagniere (approved):

internal/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	sensitivePropertiesString, err := cmd.Flags().GetString("sensitive-properties")
+	if err != nil {
+		return err
+	}
+
+	extension := filepath.Ext(pluginFileName)
+	if extension == "" {
+		return errors.Errorf("unable to read plugin file extension")
+	}
```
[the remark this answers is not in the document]
bedeks: Not sure if I follow - I return this error if I do not get a file extension. I do not know the file name so I would not know what the prefix would be. Filepath.Ext seems like the best way to get a file extension and if it is empty meaning it is not a valid filepath.

internal/connect/utils.go
```diff
@@ -71,3 +82,39 @@ func parseConfigFile(filename string) (map[string]string, error) {
...
+	}
+	defer file.Close()
+
+	part, err := writer.CreateFormFile("file", filepath.Base(filePath))
+	if err != nil {
+		return err
+	}
+	if _, err = io.Copy(part, file); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
bedeks: We're copying the file contents into the form for the api call to the presigned s3 url.

bedeks:
> @bedeks What you have listed under "bug fixes" sounds more like a feature to me. Was something actually broken?

Yeah this was discovered during preview and marked as a bug. It should have worked initially since connector class is optional on the apis but since cli was doing special validation I had to fix that validation to not happen for custom connectors. I had missed it when I initially implemented it.

internal/connect/command_custom_plugin.go
```diff
@@ -1,26 +1,40 @@
...
 }
 
-func newCustomPluginCommand(cfg *v1.Config, prerunner pcmd.PreRunner) *cobra.Command {
+type customPluginHumanOut struct {
+	Id                  string `human:"ID"`
+	Name                string `human:"Name"`
+	Description         string `human:"Description"`
+	ConnectorClass      string `human:"Connector Class"`
+	ConnectorType       string `human:"Connector Type"`
+	SensitiveProperties string `human:"Sensitive Properties"`
```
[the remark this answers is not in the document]
bedeks: Actually majority of use cases will have this set and we had a big incident related to users not using this so I think it would be nice for them to see that they did not set any sensitive properties that they should have

internal/connect/command_custom_plugin.go
```diff
@@ -29,14 +43,39 @@ func newCustomPluginCommand(cfg *v1.Config, prerunner pcmd.PreRunner) *cobra.Com
...
 
-	if cfg.IsCloudLogin() {
-		c.AuthenticatedCLICommand = pcmd.NewAuthenticatedCLICommand(cmd, prerunner)
-		cmd.AddCommand(c.newCreateCommand())
-		cmd.AddCommand(c.newListCommand())
-		cmd.AddCommand(c.newDescribeCommand())
-		cmd.AddCommand(c.newDeleteCommand())
-		cmd.AddCommand(c.newUpdateCommand())
-	}
+	c.AuthenticatedCLICommand = pcmd.NewAuthenticatedCLICommand(cmd, prerunner)
```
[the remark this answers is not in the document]
bedeks: Not sure if I follow is the suggestion to combine 44 and 46? c is being used again on lines 47- 51. I don't think it will be very readable!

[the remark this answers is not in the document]
bedeks: Oh I see what you mean now. Fixed it!

internal/connect/utils.go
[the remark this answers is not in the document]
bedeks: I haven not written a lot of go but this is what I find online:

io.Copy is a very efficient way to copy data from a Reader to a Writer in Go. It handles the reading and writing process, including buffering, in an optimized manner. 

When I use writer.CreateFormFile, I get back an io.Writer. This writer is specifically set up to write the file's contents into the correct place in the multipart form data, taking into account any necessary headers, boundaries, etc. When I pass this writer to io.Copy, the file's contents are written directly into the correct location in the multipart form data.

So, to clarify: we are directly writing the file's content into the multipart form. The io.Copy function is just an efficient helper to move data from the file (io.Reader) into the multipart form's appropriate section (io.Writer).

internal/connect/command_custom_plugin_create.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+		),
+	}
+
+	cmd.Flags().String("plugin-file", "", "ZIP/JAR custom plugin file.")
+	cmd.Flags().String("connector-class", "", "Connector class of custom plugin.")
+	cmd.Flags().String("connector-type", "", "Connector type of custom plugin.")
+	cmd.Flags().String("description", "", "Description of custom plugin.")
+	cmd.Flags().String("documentation-link", "", "Document link of custom plugin.")
+	cmd.Flags().String("sensitive-properties", "", "Sensitive properties of custom plugin.")
```
byang-confluent, line 39: should `sensitive-properties` be `list[string]` instead of `string`? how can use pass ["foo", "bar"] into it? can we update the example?

bedeks: Users should just provide comma separated strings of all the sensitive properties. I think storing it as a list is just an implementation detail they shouldn't need to care about.

byang-confluent: figure it out by reading the code :) can we add it example?

test/fixtures/output/connect/custom-plugin/update.golden
```diff
@@ -0,0 +1,5 @@
+Updated custom connector plugin "ccp-123456".
```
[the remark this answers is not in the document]
bedeks: Do we want to remove the update message? How will the customers know the update succeeded?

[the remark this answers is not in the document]
brianstrauch: They should see their changes in the output underneath, no?

internal/connect/command_custom_plugin.go
```diff
@@ -0,0 +1,80 @@
...
+	cmd.AddCommand(c.newListCommand())
+	cmd.AddCommand(c.newUpdateCommand())
+
+	return cmd
+}
+
+func printTable(cmd *cobra.Command, plugin connectcustompluginv1.ConnectV1CustomConnectorPlugin) error {
+	table := output.NewTable(cmd)
+	sensitiveProperties := plugin.GetSensitiveConfigProperties()
+	sort.Strings(sensitiveProperties)
```
brianstrauch, line 58: Can you sort in the backend?

pkg/ccloudv2/connect_custom_plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	return resp, errors.CatchCCloudV2Error(err, httpResp)
+}
+
+func (c *Client) CreateCustomPlugin(displayName, description, documentationLink, connectorClass, connectorType, uploadId string, sensitiveProperties []string) (connectcustompluginv1.ConnectV1CustomConnectorPlugin, error) {
+	createCustomPluginRequest := connectcustompluginv1.NewConnectV1CustomConnectorPlugin()
+	createCustomPluginRequest.SetDisplayName(displayName)
+	createCustomPluginRequest.SetDescription(description)
+	createCustomPluginRequest.SetDocumentationLink(documentationLink)
+	createCustomPluginRequest.SetConnectorClass(connectorClass)
+	createCustomPluginRequest.SetConnectorType(connectorType)
```
brianstrauch, line 39: Can you move this logic outside of this function? Please imitate the style of other similar functions in other files in this package. We'd like to generate this code in the near future and it would help if it was consistent!

pkg/ccloudv2/connect_custom_plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	httpResp, err := c.CcpClient.CustomConnectorPluginsConnectV1Api.DeleteConnectV1CustomConnectorPlugin(c.connectCustomPluginApiContext(), id).Execute()
+	return errors.CatchCCloudV2Error(err, httpResp)
+}
+
+func (c *Client) UpdateCustomPlugin(id, name, description, documentationLink string, sensitiveProperties []string) (connectcustompluginv1.ConnectV1CustomConnectorPlugin, error) {
+	updateCustomPluginRequest := connectcustompluginv1.NewConnectV1CustomConnectorPluginUpdate()
+	updateCustomPluginRequest.SetSensitiveConfigProperties(sensitiveProperties)
+	updateCustomPluginRequest.SetDisplayName(name)
+	updateCustomPluginRequest.SetDescription(description)
+	updateCustomPluginRequest.SetDocumentationLink(documentationLink)
```
brianstrauch, line 83: This should be moved outside of this function as well.

test/fixtures/output/connect/custom-plugin/update.golden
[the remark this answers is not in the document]
bedeks: They're just seeing the id and the name. If they had updated the description they wouldn't see it.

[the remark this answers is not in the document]
bedeks: I was thinking they would have to do describe if they want to see other fields. It might be better to remove the table actually to keep it consistent with other connect commands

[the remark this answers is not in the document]
brianstrauch: How about we add the connector description to the output table so they'll see it if it's updated?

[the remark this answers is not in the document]
bedeks: I can add all the fields since they can all be updated besides class and type but then it is not consistent with how connector update is. Is that okay given this command is under connect?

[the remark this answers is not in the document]
bedeks: @brianstrauch I removed the table in the latest pr to match connect if you think we should show all updatable properties I can put the table back. Let me know!

internal/connect/command_custom_plugin_create.go
```diff
@@ -87,7 +92,18 @@ func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []stri
...
+	createCustomPluginRequest := connectcustompluginv1.NewConnectV1CustomConnectorPlugin()
+	createCustomPluginRequest.SetDisplayName(displayName)
+	createCustomPluginRequest.SetDescription(description)
+	createCustomPluginRequest.SetDocumentationLink(documentationLink)
+	createCustomPluginRequest.SetConnectorClass(connectorClass)
+	createCustomPluginRequest.SetConnectorType(connectorType)
+	createCustomPluginRequest.SetSensitiveConfigProperties(sensitiveProperties)
+	createCustomPluginRequest.SetUploadSource(
+		connectcustompluginv1.ConnectV1UploadSourcePresignedUrlAsConnectV1CustomConnectorPluginUploadSourceOneOf(
+			connectcustompluginv1.NewConnectV1UploadSourcePresignedUrl("PRESIGNED_URL_LOCATION", resp.GetUploadId())))
```
brianstrauch, line 104: Can we create this with one initialization statement? Look at some other create/update commands for examples.

test/fixtures/output/connect/custom-plugin/update.golden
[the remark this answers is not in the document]
brianstrauch: Let's stay consistent here. Maybe in the next major version update we can switch to table output if that's more desirable.

brianstrauch (approved):

PR #2249, opened 2023-09-11
[TRAFFIC-9327] - Add `confluent network transit-gateway-attachment create` commands

syoung-cflt:
What
----

Note that this is merging into the feature branch https://github.com/confluentinc/cli/tree/traffic-8850-network-cli. We will merge commits into main from this feature branch after all of the commands for network are implemented
* Add `confluent network transit-gateway-attachment create`

References
----------

[CLI Commands Design](https://confluentinc.atlassian.net/wiki/spaces/PM/pages/2987394300/1-pager+--+Network+CLI)
[Implementation 1-Pager](https://confluentinc.atlassian.net/wiki/spaces/~712020463029eabbac401bbe1b033ce0d3a00e/pages/3136749631/Network+CLI)

Test & Review
-------------

Added integration tests.

Performed manual tests

Open Questions / Follow-ups
---------------------------

PR #2251, opened 2023-09-11
[TRAFFIC-9330][TRAFFIC-9331] - Add `confluent network private-link access list|describe` commands

syoung-cflt:
What
----

* Add `confluent network private-link` and an alias "pl"
* Add `confluent network private-link access`
* Add `confluent network private-link access list`
* Add `confluent network private-link access describe <id>`

References
----------

Test & Review
-------------

Added integration tests.

Performed manual tests

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/network/command_private_link.go
+++ b/internal/network/command_private_link.go
@@ -1,0 +1,20 @@
+package network
+
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+)
+
+func newPrivateLinkCommand(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "private-link",
+		Short:   "Manage private links.",
+		Aliases: []string{"pl"},
+		Args:    cobra.NoArgs,
+	}
+
+	cmd.AddCommand(newPrivateLinkAccessCommand(prerunner))
+
+	return cmd
+}
--- a/internal/network/command_private_link.go
+++ b/internal/network/command_private_link.go
@@ -2,18 +2,16 @@ package network
 
 import (
 	"github.com/spf13/cobra"
-
-	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
 )
 
-func newPrivateLinkCommand(prerunner pcmd.PreRunner) *cobra.Command {
+func (c *command) newPrivateLinkCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "private-link",
 		Short:   "Manage private links.",
 		Aliases: []string{"pl"},
 	}
 
-	cmd.AddCommand(newPrivateLinkAccessCommand(prerunner))
+	cmd.AddCommand(c.newPrivateLinkAccessCommand())
 
 	return cmd
 }
--- a/internal/network/command_private_link.go
+++ b/internal/network/command_private_link.go
@@ -11,7 +11,6 @@ func newPrivateLinkCommand(prerunner pcmd.PreRunner) *cobra.Command {
 		Use:     "private-link",
 		Short:   "Manage private links.",
 		Aliases: []string{"pl"},
-		Args:    cobra.NoArgs,
 	}
 
 	cmd.AddCommand(newPrivateLinkAccessCommand(prerunner))
--- a/internal/network/command_private_link_access.go
+++ b/internal/network/command_private_link_access.go
@@ -7,14 +7,9 @@ import (
 
 	networkingv1 "github.com/confluentinc/ccloud-sdk-go-v2/networking/v1"
 
-	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
 	"github.com/confluentinc/cli/v3/pkg/errors"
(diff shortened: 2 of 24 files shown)


internal/network/command_private_link.go
```diff
@@ -0,0 +1,20 @@
...
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+)
+
+func newPrivateLinkCommand(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "private-link",
+		Short:   "Manage private links.",
+		Aliases: []string{"pl"},
+		Args:    cobra.NoArgs,
```
brianstrauch, line 14: This is only needed for "leaf" commands
```suggestion
```

internal/network/command_private_link.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	cmd.AddCommand(newPrivateLinkAccessCommand(prerunner))
```
brianstrauch, line 17: Are we adding more subcommands here in the future?

internal/network/command_private_link_access.go
```diff
@@ -0,0 +1,130 @@
...
+	networkingv1 "github.com/confluentinc/ccloud-sdk-go-v2/networking/v1"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+type privateLinkAccessCommand struct {
+	*pcmd.AuthenticatedCLICommand
+}
```
brianstrauch, line 17: Any reason we aren't using the struct from the parent command?

internal/network/command_private_link_access.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *privateLinkAccessCommand) getCloud(access networkingv1.NetworkingV1PrivateLinkAccess) (string, error) {
+	cloud := access.Spec.GetCloud()
+
+	if cloud.NetworkingV1AwsPrivateLinkAccess != nil {
+		return CloudAws, nil
+	} else if cloud.NetworkingV1GcpPrivateServiceConnectAccess != nil {
+		return CloudGcp, nil
+	} else if cloud.NetworkingV1AzurePrivateLinkAccess != nil {
+		return CloudAzure, nil
+	}
```
brianstrauch, line 91: I think this code already exists? If not, seems like a good candidate for a unit-tested helper function.

internal/network/command_private_link_access.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return "", fmt.Errorf(errors.CorruptedNetworkResponseErrorMsg, "cloud")
+}
+
+func (c *privateLinkAccessCommand) printPrivateLinkAccessTable(cmd *cobra.Command, access networkingv1.NetworkingV1PrivateLinkAccess) error {
+	table := output.NewTable(cmd)
```
brianstrauch, line 97: Let's define this closer to where it's used?

internal/network/command_private_link.go
syoung-cflt: Yes, we will be adding `confluent network private-link attachment` and `confluent network private-link attachment connection`.

internal/network/command_private_link_access.go
syoung-cflt: That was for `peering`. Not sure if there's a more generic way to extract the cloud provider information.

internal/network/command_private_link_access.go
syoung-cflt: We will be using it for `update` and `create` as well. I thought it would make sense to define it here, just like what we have been doing for other subcommands. Should I move it to else where?

internal/network/command_private_link_access.go
brianstrauch: Ah, for future reference @MuweiHe we prefer defining as few structs as possible. If the parent command uses the same type of authentication as its children, the same struct can be used for all of them.

internal/network/command_private_link_access.go
brianstrauch: The typical pattern is to define it, then add to it, then print it. (No need to define it all the way at the top of the function if we only use it at the bottom)

syoung-cflt: Ah I thought you were referring to the `printPrivateLinkAccessTable` function.

I will move the `table` variable to the bottom of the function.

internal/network/command_private_link.go
syoung-cflt: Updated

internal/network/command_private_link_access.go
syoung-cflt: Thank you! Updated the private link access command to use the struct from the parent command.

I will update the rest in a separate PR after merging this PR and https://github.com/confluentinc/cli/pull/2249

internal/network/command_private_link_access.go
syoung-cflt: Updated!

brianstrauch (approved):

ejagade (approved):
lgtm

PR #2253, opened 2023-09-13
Don't send usage service requests for some hosts

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
These should silently fail anyway, but we should make sure we don't send pointless http requests for fedramp environments.

References
----------

Test & Review
-------------
TODO: test manually

Open Questions / Follow-ups
---------------------------


pkg/config/config_test.go
```diff
@@ -974,25 +974,23 @@ func TestConfig_IsCloud_False(t *testing.T) {
 	}
 }
 
-func TestConfig_IsCloud_NonGov(t *testing.T) {
+func TestConfig_GovHostname(t *testing.T) {
```
brianstrauch, line 977:
```suggestion
func TestGovHostname(t *testing.T) {
```

pkg/config/config_test.go
```diff
@@ -974,25 +974,23 @@ func TestConfig_IsCloud_False(t *testing.T) {
...
 	cfg := &Config{
 		Contexts: map[string]*Context{"context": {
-			PlatformName: "https://confluent.cloud",
+			PlatformName: "https://confluentgov-internal.com",
 		}},
 		CurrentContext: "context",
 	}
```
brianstrauch, line 984:
```suggestion
```

brianstrauch (approved):

PR #2264, opened 2023-09-15
Remove `--force` and add `--output` to `confluent kafka link describe`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Add the missing `--output` flag to `confluent kafka link describe` and remove the unused `--force` flag

Checklist
---------
   * yes: ok

What
----
I think this qualifies more as a bug fix than a breaking change, because the `--force` flag was never intended to be available for this command. Since `--output` is also missing, this was just a copy paste error.

References
----------

Test & Review
-------------
Added a new integration test for the output flag.

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/command_link_describe.go
+++ b/internal/kafka/command_link_describe.go
@@ -29,10 +29,10 @@ func (c *linkCommand) newDescribeCommand() *cobra.Command {
 		RunE:              c.describe,
 	}
 
-	pcmd.AddForceFlag(cmd)
 	pcmd.AddClusterFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddContextFlag(cmd, c.CLICommand)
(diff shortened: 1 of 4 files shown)


internal/kafka/command_link_describe.go
```diff
@@ -29,10 +29,10 @@ func (c *linkCommand) newDescribeCommand() *cobra.Command {
 		RunE:              c.describe,
 	}
 
-	pcmd.AddForceFlag(cmd)
```
brianstrauch, line 32: Huh, wonder why this was added in the first place.

brianstrauch (approved):

internal/kafka/command_link_describe.go
sgagniere: Copy paste from the wrong file.

PR #2267, opened 2023-09-15
Feat/KFS-1166 default and initialvalues

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

----------
We now have default and initial values. Initial values can be as all props set to nit. Default will always have a default value. Catalog and database are now initial values, thus can be reseted if necessary.
References

Test & Review
-------------

unit tests, snapshot tests

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/pkg/flink/internal/store/store_utils_test.go
+++ b/pkg/flink/internal/store/store_utils_test.go
@@ -130,6 +130,8 @@ func TestProcessResetStatement(t *testing.T) {
 		ServiceAccountId: "sa-123",
 	}
 	s := NewStore(client, nil, &appOptions, tokenRefreshFunc).(*Store)
+	s.Properties.Set(config.ConfigKeyLocalTimeZone, "London/GMT")
+
 	defaultSetOutput := createStatementResults([]string{"Key", "Value"}, [][]string{
 		{config.ConfigKeyLocalTimeZone, fmt.Sprintf("%s (default)", getLocalTimezone())},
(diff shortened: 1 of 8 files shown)


swist (approved):

pkg/flink/internal/store/store_utils_test.go
```diff
@@ -131,12 +131,20 @@ func TestProcessResetStatement(t *testing.T) {
 	}
 	s := NewStore(client, nil, &appOptions, tokenRefreshFunc).(*Store)
 	defaultSetOutput := createStatementResults([]string{"Key", "Value"}, [][]string{
-		{config.ConfigKeyCatalog, fmt.Sprintf("%s (default)", appOptions.EnvironmentName)},
```
twalthr, line 134: where do appOptions come from? In theory it is fine if we set this here, but they must be correct. and this is what caused issues.

Guttz: appOptions are defined according to what the user passes when starting the `flink shell`. 

Environment is a required parameter, so we'll always intialize the catalog for the user when starting the shell. But now he's able to reset it.

Database is not required, but I'll be initialized if the user passes the "--database" parameter to `flink shell`, or else it will be nil. It can also be reset now.

swist: appOptions is what you pass via command line flags. So if you want this to reset to null, we need to use them only once

twalthr: thanks for the explanation, let's give it a try :)

PR #2283, opened 2023-09-19
CLI-2675 check java version before starting confluent local services

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Verify Java version at the beginning when running `confluent local services start`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

https://confluentinc.atlassian.net/browse/CLI-2675
Check java version is valid before starting the services, so user doesn't need to wait until service is up.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/local/command_service.go
+++ b/internal/local/command_service.go
@@ -252,6 +252,10 @@ func (c *command) runServiceVersionCommand(cmd *cobra.Command, _ []string) error
 }
 
 func (c *command) startService(service, configFile string) error {
+	if err := c.checkJavaVersion(service); err != nil {
+		return err
+	}
+
 	isUp, err := c.isRunning(service)
 	if err != nil {
 		return err
@@ -260,7 +264,7 @@ func (c *command) startService(service, configFile string) error {
 		return c.printStatus(service)
 	}
 
-	if err := c.checkService(service); err != nil {
+	if err := c.checkOSVersion(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


sgagniere (approved):

internal/local/command_service.go
```diff
@@ -252,6 +252,10 @@ func (c *command) runServiceVersionCommand(cmd *cobra.Command, _ []string) error
 }
 
 func (c *command) startService(service, configFile string) error {
+	if err := c.checkJavaVersion(service); err != nil {
```
brianstrauch, line 255: Is there something within `isRunning()` that requires Java? Otherwise I don't see how this changes anything.

MuweiHe:
Java version check is actually done before starting the service already. closing the original ticket too.

internal/local/command_service.go
MuweiHe: if user already has service running,`isRunning` will return and it won't get to validate Java version. That brought some confusion in that original ticket like "If Java version was not valid then why were some services started"? checking java version first will avoid such confusion. 
It's up to us how we want users to perceive the message.

brianstrauch: Wait, how can `isRunning` be true if the user doesn't have Java installed? Did the user start `confluent local` and then remove their Java installation?

MuweiHe: 😕 this happened in that ticket. suppose a few services has been running before, and user changed java path / version and ran `start` again

brianstrauch (approved):

PR #2285, opened 2023-09-20
Fix/kfs 1190 fix user identity warning

YannickPferr:
Checklist
---------
   * yes: ok

What
----

The user identity warning was not displayed anymore, because the principal is now always set (either to the SA or the user ID). 
This is fixed now + the warning will only be shown for INSERT and STATEMENT SET statements, because the 4 hour time limit is only relevant for those.


swist (approved):

pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementPrintsNoWarningForStatementsOtherThanInsertOrStatementSet
```diff
@@ -0,0 +1,6 @@
+Statement name: test-statement
+Statement successfully submitted.
+Waiting for statement to be ready. Statement phase is PENDING.
+Statement phase is COMPLETED.
```
swist, line 4: The snapshot doesn't look right

swist (changes requested):
the snaphsot doesn't look right

swist (approved):
nvm misread the filename

PR #2286, opened 2023-09-20
Feat/KFS-810 client stop statement insteadof delete

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

statement will be stopped instead of deleted when closing the result's screen 

References
----------

Test & Review
-------------

unit tests

Open Questions / Follow-ups
---------------------------


swist (changes requested):
can you also update the readme to let people know that they need to get `go install go.uber.org/mock/mockgen`

YannickPferr:
I think we should do this in the panic recovery as well instead of deleting: 
here: https://github.com/confluentinc/cli/blob/b7855df6fbc1288143ea605617db41f5a38a22c4/pkg/flink/app/application.go#L118
slash here: https://github.com/confluentinc/cli/blob/main/pkg/flink/internal/controller/statement_controller.go#L171

Guttz:
> I think we should do this in the panic recovery as well instead of deleting: here:
> 
> https://github.com/confluentinc/cli/blob/b7855df6fbc1288143ea605617db41f5a38a22c4/pkg/flink/app/application.go#L118
> 
> 
> slash here: https://github.com/confluentinc/cli/blob/main/pkg/flink/internal/controller/statement_controller.go#L171

It's a good point and I thought about this and also about when the user manually cancel waiting for results but: 

1 - deleting is safer since it doesn't have to first fetch the statement. In other words, it doesn't depend on a successful GetStatement for it to happen
2 - We should align it Asher before changing this, it's ultimately a product decision

YannickPferr:
> > I think we should do this in the panic recovery as well instead of deleting: here:
> > https://github.com/confluentinc/cli/blob/b7855df6fbc1288143ea605617db41f5a38a22c4/pkg/flink/app/application.go#L118
> > 
> > slash here: https://github.com/confluentinc/cli/blob/main/pkg/flink/internal/controller/statement_controller.go#L171
> 
> It's a good point and I thought about this and also about when the user manually cancel waiting for results but:
> 
> 1 - deleting is safer since it doesn't have to first fetch the statement. In other words, it doesn't depend on a successful GetStatement for it to happen 2 - We should align it Asher before changing this, it's ultimately a product decision

For cancelling I agree, it is fine to Delete. But my point for panic is that this can happen after the user has already seen some results (e.g. a panic happens in the results view after some rows have already been displayed), and I don't think we should delete in that case.

But yeah, makes sense to align with Asher on this!

pkg/ccloudv2/flink_gateway.go
```diff
@@ -59,16 +60,23 @@ func (c *FlinkGatewayClient) flinkGatewayApiContext() context.Context {
...
-
 func (c *FlinkGatewayClient) GetStatement(environmentId, statementName, orgId string) (flinkgatewayv1beta1.SqlV1beta1Statement, error) {
 	resp, httpResp, err := c.StatementsSqlV1beta1Api.GetSqlv1beta1Statement(c.flinkGatewayApiContext(), orgId, environmentId, statementName).Execute()
 	return resp, flink.CatchError(err, httpResp)
 }
 
+func (c *FlinkGatewayClient) UpdateStatement(environmentId, statementName, orgId string, statement flinkgatewayv1beta1.SqlV1beta1Statement) error {
+	httpResp, err := c.StatementsSqlV1beta1Api.UpdateSqlv1beta1Statement(c.flinkGatewayApiContext(), orgId, environmentId, statementName).
+		SqlV1beta1Statement(statement).
+		Execute()
```
brianstrauch, line 71:
```suggestion
	httpResp, err := c.StatementsSqlV1beta1Api.UpdateSqlv1beta1Statement(c.flinkGatewayApiContext(), orgId, environmentId, statementName).SqlV1beta1Statement(statement).Execute()
```

pkg/flink/CONTRIBUTING.md
```diff
@@ -12,13 +12,20 @@ go tool cover -html=coverage.out -o coverage.html
...
+go install go.uber.org/mock/mockgen@latest
+```
+
+Generate mocks:
 ```
 go generate pkg/flink/test/mock/mock_generator.go
 ```
 
-More info on how to generate mocks: https://github.com/golang/mock
+If you need to generate mocks for a new interface, take a look at the mock_generator.go and change it accordingly.
```
brianstrauch, line 26: What's the reason for switching from the golang version to the uber version?

brianstrauch: I was hoping to standardize the golang version across the CLI

swist: golang one is no longer maintained, the go.uber.org fork is the one where the development is currently taking place

Guttz:
image

Tested this on devel yesterday and on stag today and it works. Merging it and you should easily reuse the flink gateway changes and mocks updates by rebasing on main :)

PR #2288, opened 2023-09-20
Remove trailing slashes from the config platform name

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
If users log in with a Confluent gov url with a trailing `/`, the CLI will fail to encrypt the auth refresh token. To determine if a token is an unencrypted Confluent gov refresh token, the code checks the config platform name against a list of hostnames without the trailing `/`. Because of this, the code fails to encrypt the token.

Since the unencrypted refresh token generally contains invalid base64 characters, running most commands after login will yield base64 data errors.

This PR trims `/` from the end of the platform name before it's added to the config.

References
----------

Test & Review
-------------
Manual testing with `confluent login --url confluent.cloud/`

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -109,7 +109,7 @@ func PersistCCloudCredentialsToConfig(config *config.Config, client *ccloudv1.Cl
 
 func addOrUpdateContext(cfg *config.Config, isCloud bool, credentials *Credentials, ctxName, url string, state *config.ContextState, caCertPath, orgResourceId string, save bool) error {
 	platform := &config.Platform{
-		Name:       strings.TrimPrefix(url, "https://"),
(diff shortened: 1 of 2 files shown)


brianstrauch:
Should we write release notes or does this only apply to FEDRAMP?

brianstrauch (approved):
Thanks!! Unit/integration tests?

sgagniere:
@brianstrauch this only applies to fedramp because we don't try to match the platform name during decryption for regular confluent environments.

I was actually going to ask about that: we don't do release notes for only fedramp things, right?

brianstrauch:
> we don't do release notes for only fedramp things, right?

For the time being, no because customers aren't using it as far as I know. Would be good to double check that though

PR #2289, opened 2023-09-21
fix: remove environment flag for compute pool use command

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Since `compute-pool use` will not switch environments for you, we ended up receiving some bug reports of people being confused why their use command didn't go through. By removing the flag we force users to manually switch, avoiding this confusion.

### DIFF
--- a/internal/flink/command_compute_pool_use.go
+++ b/internal/flink/command_compute_pool_use.go
@@ -19,8 +19,6 @@ func (c *command) newComputePoolUseCommand() *cobra.Command {
 		RunE:              c.computePoolUse,
 	}
 
-	pcmd.AddOutputFlag(cmd)
-
 	return cmd
 }
 
--- a/internal/flink/command_compute_pool_use.go
+++ b/internal/flink/command_compute_pool_use.go
@@ -19,7 +19,6 @@ func (c *command) newComputePoolUseCommand() *cobra.Command {
 		RunE:              c.computePoolUse,
 	}
 
-	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddOutputFlag(cmd)
 
 	return cmd
--- a/pkg/errors/error_message.go
+++ b/pkg/errors/error_message.go
@@ -200,7 +200,7 @@ const (
 	JavaExecNotFondErrorMsg         = "could not find java executable, please install java or set JAVA_HOME"
 	NothingToDestroyErrorMsg        = "nothing to destroy"
 	ComputePoolNotFoundErrorMsg     = `Flink compute pool "%s" not found or access forbidden.`
-	ComputePoolNotFoundSuggestions  = "List available Flink compute pools with `confluent flink compute-pool list`"
+	ComputePoolNotFoundSuggestions  = "List available Flink compute pools with `confluent flink compute-pool list` and make sure you have selected the compute pool's environment with `confluent environment use`"
 	FailedToReadPortsErrorMsg       = "failed to read local ports from config"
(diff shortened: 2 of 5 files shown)


internal/flink/command_compute_pool_use.go
```diff
@@ -19,7 +19,6 @@ func (c *command) newComputePoolUseCommand() *cobra.Command {
 		RunE:              c.computePoolUse,
 	}
 
-	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
```
swist, line 22: Can you add a check to verify that there is such a compute pool in this env and make it fail if not and print an error instructing people to do `environment use`

swist (changes requested):

internal/flink/command_compute_pool_use.go
YannickPferr: Currently it will tell you this when you try to select a pool from a different env:
```
Error: Flink compute pool "lfcp-123" not found or access forbidden.: Forbidden Access: 403 Forbidden

Suggestions:
    List available Flink compute pools with `confluent flink compute-pool list`
```
I'll add a suggestion to check if the user is in the correct env

YannickPferr: How's that?
```
Error: Flink compute pool "lfcp-devccnp8rgd" not found or access forbidden.: Forbidden Access: 403 Forbidden

Suggestions:
    List available Flink compute pools with `confluent flink compute-pool list` and make sure you have selected the compute pool's environment with `confluent environment use env-123`

```

swist: remove the placeholder env, otherwise good!

YannickPferr: alright, sounds good!

YannickPferr: Done

internal/flink/command_compute_pool_use.go
```diff
@@ -19,7 +19,6 @@ func (c *command) newComputePoolUseCommand() *cobra.Command {
...
 	pcmd.AddOutputFlag(cmd)
```
brianstrauch, line 22: Please also delete this flag... not sure why it was added since this command doesn't print any tables.
```suggestion
```

pkg/errors/error_message.go
```diff
@@ -199,8 +199,8 @@ const (
 	MacVersionErrorMsg              = "macOS version >= %s is required (detected: %s)"
 	JavaExecNotFondErrorMsg         = "could not find java executable, please install java or set JAVA_HOME"
 	NothingToDestroyErrorMsg        = "nothing to destroy"
-	ComputePoolNotFoundErrorMsg     = `Flink compute pool "%s" not found or access forbidden.`
-	ComputePoolNotFoundSuggestions  = "List available Flink compute pools with `confluent flink compute-pool list`"
+	ComputePoolNotFoundErrorMsg     = `Flink compute pool "%s" not found or access forbidden`
+	ComputePoolNotFoundSuggestions  = "List available Flink compute pools with `confluent flink compute-pool list` and make sure you have selected the compute pool's environment with `confluent environment use`"
```
brianstrauch, line 203:
```suggestion
	ComputePoolNotFoundSuggestions  = "List available Flink compute pools with `confluent flink compute-pool list`.\nMake sure you have selected the compute pool's environment with `confluent environment use`."
```

internal/flink/command_compute_pool_use.go
YannickPferr: Done!

pkg/errors/error_message.go
YannickPferr: Done!

swist (approved):

PR #2287, opened 2023-09-20
Feat/KFS-1189 multiple small changes

Guttz:
Release Notes
-------------

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok

What
----

Product asked for multiple small changes as part of the ongoing bugbash:

-Flag name from cfu to max-cfu
-Compute Pool Create description adjustment
-Default values for cfu adjusted

References
----------

Test & Review
-------------

golden files updated

Open Questions / Follow-ups
---------------------------

PR #2296, opened 2023-09-22
fix: change CFU suggestion from 2 to 5

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Updating the suggestion was missed when changing CFUs to be a multiple of 5.


Guttz (approved):

Guttz:
Small detail we forgot here https://github.com/confluentinc/cli/pull/2287

PR #2299, opened 2023-09-22
[TRAFFIC-9357][TRAFFIC-9358]: Add `confluent network private-link attachment connection list|describe` commands

syoung-cflt:
What
----

* Add `confluent network private-link attachment connection`
* Add `confluent network private-link attachment connection list`
* Add `confluent network private-link attachment connection describe`

References
----------

Test & Review
-------------

Added integration tests.

Performed manual tests

Open Questions / Follow-ups
---------------------------

Please see the questions below.

### DIFF
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -236,6 +236,6 @@ func addNetworkFlag(cmd *cobra.Command, c *pcmd.AuthenticatedCLICommand) {
 }
 
 func (c *command) addPrivateLinkAttachmentFlag(cmd *cobra.Command) {
-	cmd.Flags().String("attachment", "", "ID of the private link attachment.")
+	cmd.Flags().String("attachment", "", "Private link attachment ID.")
 	pcmd.RegisterFlagCompletionFunc(cmd, "attachment", c.validPrivateLinkAttachmentArgs)
 }
--- a/internal/network/command.go
+++ b/internal/network/command.go
@@ -234,3 +234,8 @@ func addNetworkFlag(cmd *cobra.Command, c *pcmd.AuthenticatedCLICommand) {
 		return autocompleteNetworks(c.V2Client, environmentId)
 	})
 }
+
+func (c *command) addPrivateLinkAttachmentFlag(cmd *cobra.Command) {
+	cmd.Flags().String("attachment", "", "ID of the private link attachment.")
+	pcmd.RegisterFlagCompletionFunc(cmd, "attachment", c.validPrivateLinkAttachmentArgs)
+}
--- a/internal/network/command_private_link_attachment_connection.go
+++ b/internal/network/command_private_link_attachment_connection.go
@@ -1,0 +1,64 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	networkingprivatelinkv1 "github.com/confluentinc/ccloud-sdk-go-v2/networking-privatelink/v1"
+
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+type privateLinkAttachmentConnectionOut struct {
+	Id                        string `human:"ID" serialized:"id"`
+	Name                      string `human:"Name" serialized:"name"`
+	Cloud                     string `human:"Cloud" serialized:"cloud"`
+	PrivateLinkAttachmentId   string `human:"Private Link Attachment ID" serialized:"private_link_attachment_id"`
+	Phase                     string `human:"Phase" serialized:"phase"`
+	AwsVpcEndpointId          string `human:"AWS VPC Endpoint ID,omitempty" serialized:"aws_vpc_endpoint_id,omitempty"`
+	AwsVpcEndpointServiceName string `human:"AWS VPC Endpoint Service Name,omitempty" serialized:"aws_vpc_endpoint_service_name,omitempty"`
+}
+
+func (c *command) newPrivateLinkAttachmentConnectionCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "connection",
+		Short: "Manage private link attachment connections.",
+	}
+
+	cmd.AddCommand(c.newPrivateLinkAttachmentConnectionDescribeCommand())
+	cmd.AddCommand(c.newPrivateLinkAttachmentConnectionListCommand())
+
+	return cmd
+}
+
+func printPrivateLinkAttachmentConnectionTable(cmd *cobra.Command, connection networkingprivatelinkv1.NetworkingV1PrivateLinkAttachmentConnection) error {
+	if connection.Spec == nil {
+		return fmt.Errorf(errors.CorruptedNetworkResponseErrorMsg, "spec")
+	}
+	if connection.Status == nil {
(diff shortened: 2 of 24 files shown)


internal/network/command_private_link_attachment_connection_list.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+func (c *command) newPrivateLinkAttachmentConnectionListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "list <private-link-attachment-id>",
```
syoung-cflt, line 15: Context: The [listNetworkingV1PrivateLinkAttachmentConnections](https://docs.confluent.io/cloud/current/api.html#tag/Private-Link-Attachment-Connections-(networkingv1)/operation/listNetworkingV1PrivateLinkAttachmentConnections) API endpoint requires `spec.private_link_attachment` in addition to `environment`. In other words, we can list out "private link attachment connections` only for a specific "private link attachment"

Q1. Does it make sense to use an argument here? Should I use a required flag `private-link-attachment` instead?
Q2. Is there a simple way to provide autocomplete for `private-link connection describe|update|delete` given that the list api endpoint requires this `private_link_attachment_id` param?

internal/network/command_private_link_attachment_connection.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/confluentinc/cli/v3/pkg/output"
+)
+
+type privateLinkAttachmentConnectionOut struct {
+	Id                        string `human:"ID" serialized:"id"`
+	Name                      string `human:"Name" serialized:"name"`
+	Cloud                     string `human:"Cloud" serialized:"cloud"`
+	PrivateLinkAttachmentId   string `human:"Private Link Attachment ID" serialized:"private_link_attachment_id"`
+	Phase                     string `human:"Phase" serialized:"phase"`
+	AwsVpcEndpointId          string `human:"AWS VPC Endpoint ID,omitempty" serialized:"aws_vpc_endpoint_id,omitempty"`
```
ejagade, line 20: I see that `vpc_endpoint_id` is an attribute in both `status` and `spec` objects - [Link](https://docs.confluent.io/cloud/current/api.html#tag/Private-Link-Attachment-Connections-(networkingv1)/The-Private-Link-Attachment-Connections-Model)

@kabraham13 do these refer to different cloud resources?

syoung-cflt: I see that they refer to the same cloud resource when I test it on devel but let's confirm with @kabraham13

kabraham13: It looks like it is the same. I am not sure why. Maybe @arunvijayk has more context to why its in both spots

syoung-cflt: I took a look at the `ToPrivateLinkAttachmentConnectionV2Proto` function in our repo, and the `VpcEndpointId` in `spec` and `status` are referring to the same `VpcEndpointId`

internal/network/command_private_link_attachment_connection_list.go
ejagade: 1) if it is a required attribute in the API spec, then let's add a a required flag private-link-attachment. 
But would this run into in the same known issue with openAPI generator that's related to MultipleSearchFilter usage
2) IMO, I don't think autocompletion helps in this particular case, since the LIST operation is supported with only a filter.

syoung-cflt: 1. This parameter actually uses `SearchFilter` instead of `MultipleSearchFilter` so it works right now. 
My question is more about the command structure. Currently, I have `confluent network private-link attachment connection list <platt-id>`. I wonder if it would make more sense to use `confluent network private-link attachment connection list --private-link-attachment platt-123456` instead.

sgagniere: Q1: both patterns show up in a few other commands. I'm not sure which one we should prefer, but it's probably something we can standardize in V4. Personally I think I might lean towards a required flag.

Q2: Usually we would just grab the `id` from a flag value (and autocomplete would either succeed if the user provided the flag before the arguments, or silently fail if they didn't), like we do for `kafka mirror` autocompletion. But the connection describe implementation doesn't seem to use any flag for the private link attachment id.

It seems a bit weird to me that the describe endpoint doesn't require the private link attachment id while the list endpoint does.

syoung-cflt: @sgagniere Thanks!
1. I am also leaning towards a required flag. I will update it.
2. Yeah, the describe endpoint does not take a private link attachment id so we won't have a flag for it. In this case, I assume that we cannot provide autocompletion for these three subcommands (`describe`, `update`, and `delete`.

sgagniere: Sounds good. Yeah I don't think we can do autocomplete. Which is ok.

I think if we wanted auto-complete in the future we would have to request that the endpoint be updated to accept an empty private link attachment id and return all connections in that case.

syoung-cflt: Sounds good and updated the `list` command to use a flag instead!

sgagniere (approved):

internal/network/command_private_link_attachment_connection.go
arunvijayk: Spec represents input from the customer - We need the customer to specify a VPC Endpoint ID while creating the PLATTC

Status represents information that we provide to the customer - The same vpcEndpointID that was specified by the customer when they created the PLATTC is returned here

ejagade (approved):

internal/network/command.go
```diff
@@ -234,3 +234,8 @@ func addNetworkFlag(cmd *cobra.Command, c *pcmd.AuthenticatedCLICommand) {
 		return autocompleteNetworks(c.V2Client, environmentId)
 	})
 }
+
+func (c *command) addPrivateLinkAttachmentFlag(cmd *cobra.Command) {
+	cmd.Flags().String("attachment", "", "ID of the private link attachment.")
```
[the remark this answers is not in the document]
syoung-cflt: Updated!

internal/network/command_private_link_attachment_connection.go
```diff
@@ -0,0 +1,64 @@
...
+		out.AwsVpcEndpointId = connection.Spec.Cloud.NetworkingV1AwsPrivateLinkAttachmentConnection.GetVpcEndpointId()
+	}
+
+	if connection.Status.Cloud != nil && connection.Status.Cloud.NetworkingV1AwsPrivateLinkAttachmentConnectionStatus != nil {
+		out.AwsVpcEndpointServiceName = connection.Status.Cloud.NetworkingV1AwsPrivateLinkAttachmentConnectionStatus.GetVpcEndpointServiceName()
+	}
+
+	table := output.NewTable(cmd)
+	table.Add(out)
+
```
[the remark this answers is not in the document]
syoung-cflt: Fixed!

brianstrauch (approved):

PR #2301, opened 2023-09-23
[CLI-2729] Check that the current environment is valid

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
A stale or malformed active environment ID in the config file will result in 403 errors for various commands. The `EnvironmentId()` method will also fail to return the `no environment found` error in this case, because that method only checks whether or not the active environment is empty.

`EnvironmentId()` will now call the describe endpoint on the active environment ID and return our active environment error message and suggestions if the API returns an error. This will also catch the previous case (empty id).

Adding this check broke all four of the KSQL unit tests. Instead of adding the org v2 mock, I just converted one into an integration test and removed the other three since the existing integration tests already cover the same cases.

References
----------
https://confluentinc.atlassian.net/browse/CLI-2729

Test & Review
-------------
Manually tested by making manual edits to the config file:
- Deleting the current environment
- Replacing the current environment with a bad id

Both give the following:
```
confluent environment list          
  Current |     ID     |    Name     
----------+------------+-------------
          | env-6305zj | ab          
          | env-7y5pjj | default     
          | env-ym1ww6 | no-sr-test  
```
```
confluent kafka cluster list        
Error: no environment found or selected

Suggestions:
    Set an active environment with `confluent environment use`.
    This issue may also occur if this user has no valid role bindings. Contact an Organization Admin to create a role binding for this user.
```

Open Questions / Follow-ups
---------------------------


pkg/ccloudv2/org.go
```diff
@@ -29,7 +29,8 @@ func (c *Client) CreateOrgEnvironment(environment orgv2.OrgV2Environment) (orgv2
 }
 
 func (c *Client) GetOrgEnvironment(envId string) (orgv2.OrgV2Environment, error) {
-	res, httpResp, err := c.OrgClient.EnvironmentsOrgV2Api.GetOrgV2Environment(c.orgApiContext(), envId).Execute()
+	req := c.OrgClient.EnvironmentsOrgV2Api.GetOrgV2Environment(c.orgApiContext(), envId)
+	res, httpResp, err := c.OrgClient.EnvironmentsOrgV2Api.GetOrgV2EnvironmentExecute(req) // calling Execute() instead will cause mock calls to fail
```
MuweiHe, line 33: Why would mock fail? Does the mock method not exist in the package?

sgagniere: Yup, the `Execute()` method isn't mocked.

Only the API methods themselves are mocked, but not the request struct methods.

brianstrauch, line 33: Instead of mocking the responses, can we use our integration test framework?

MuweiHe: oh right.. i think that's the default way we write those methods so think we can leave out this comment

pkg/dynamic-config/dynamic_context.go
```diff
@@ -172,11 +172,12 @@ func (d *DynamicContext) HasLogin() bool {
 }
 
 func (d *DynamicContext) EnvironmentId() (string, error) {
-	if id := d.GetCurrentEnvironment(); id != "" {
-		return id, nil
+	id := d.GetCurrentEnvironment()
+	if _, err := d.V2Client.GetOrgEnvironment(id); err != nil {
```
brianstrauch, line 176: This gets used in nearly every Cloud command? Are we sure we want to introduce an extra API call to every command?

sgagniere: I think it's worth it to prevent stale environment data from giving misleading errors. This particular bug sidesteps our suggestion to set an active environment, which adds to the confusion.

Alternatively, we could do this check if we run into a 403 from routes that take environments (and the user didn't use the environment flag).

brianstrauch: > Alternatively, we could do this check if we run into a 403 from routes that take environments

I'm conflicted on if this is the more correct approach here or not. This definitely falls in line with "optimize for 99% of cases"... I'm wondering if we could do this across the board, not just for environments but also for all of our other saved resources like Kafka clusters, etc.

sgagniere: Yeah I don't see any reason why we couldn't. In hindsight there's nothing stopping users from running into this problem with other saved resources, and I think we've only seen this with environments so far because of how people are using CLI, UI, terraform (Brett confirmed that terraform destroy can cause this too).

Just to clarify: if I address all saved resources, do you think it would be the better approach then?

brianstrauch: > if I address all saved resources, do you think it would be the better approach then?

I think so. I think the only resources that we're saving properly right now are environment and Flink compute pool.

brianstrauch: (see `EnvironmentContext`)

PR #2304, opened 2023-09-25
fix: truncate multi-line cells in the basic table properly

YannickPferr:
Checklist
---------
   * yes: ok

What
----

We did not account for multi line values in the basic table, which led to us truncating those values too early. This should fix this, by only taking too longest line into consideration, when calculating the cell width.


Guttz:
Example of how show jobs also look better now:

image

before: 
image

PR #2316, opened 2023-09-27
[TRAFFIC-9359][TRAFFIC-9360][TRAFFIC-9361]: Add `confluent network private-link attachment connection update|delete|create` commands

syoung-cflt:
What
----

* Add `confluent network private-link attachment connection update`
* Add `confluent network private-link attachment connection delete`
* Add `confluent network private-link attachment connection create`

References
----------

Test & Review
-------------

Added integration tests.

Performed manual tests

Open Questions / Follow-ups
---------------------------


internal/network/command_private_link_attachment_connection_delete.go
```diff
@@ -0,0 +1,62 @@
...
+
+	deleteFunc := func(id string) error {
+		if err := c.V2Client.DeletePrivateLinkAttachmentConnection(environmentId, id); err != nil {
+			return errors.Errorf(errors.DeleteResourceErrorMsg, resource.PrivateLinkAttachmentConnection, id, err)
+		}
+		return nil
+	}
+
+	deletedIds, err := deletion.DeleteWithoutMessage(args, deleteFunc)
+	deleteMsg := "Requested to delete %s %s.\n"
```
brianstrauch, line 54: If we get a 2xx from the backend, does that mean the resources were deleted or that they were requested to be deleted? Wondering if we can use stronger language here if there is a guarantee that they are deleted.

brianstrauch (approved):

internal/network/command_private_link_attachment_connection_delete.go
syoung-cflt: The [api doc](https://docs.confluent.io/cloud/current/api.html#tag/Private-Link-Attachment-Connections-(networkingv1)/operation/deleteNetworkingV1PrivateLinkAttachmentConnection) suggests that a 204 from the backend means "A Private Link Attachment Connection is being deleted." I suppose that these networking resources take some time to be deprovisioned so I follow the pattern in this guide [here](https://github.com/confluentinc/cli/blob/main/pkg/deletion/README.md).

We will have a bug bash to test out all of the network CLI commands before the release, and make sure that we provide all necessary information to the customers. I will note this down as an action item to check whether these networking resources can be deleted immediately so that we can improve the language being used here and other places if necessary.

ejagade: @syoung-cflt we typically use soft-deletion for these resources i.e users will get 204 response once it is deleted. the next "GET" request from the user will receive a 404
Internally, the resource is marked for deletion and removed asynchronously later. 

So having a stronger language here should be OK.

syoung-cflt: @ejagade Does this apply to all of the resources? I just tested the following case for peering. 
1. Create a peering (Phase: `PROVISIONING`)
2. List all peerings, the newly created peering is now in `PENDING_ACCEPT`
3. Delete the peering.
4. List all peerings, and get the peering. The state is `DEPROVISIONING` instead of a 404.

ejagade (approved):

internal/network/command_private_link_attachment_connection_delete.go
ejagade: Guess I'm wrong then. If the user can see the `DEPROVISIONING` status then my point above isn't valid.

syoung-cflt: I did see some resources that were deleted immediately. Perhaps we can have a stronger language for those resources to specify that they are in fact (soft) deleted. I will verify this when I test the CLI, and confirm with Kevin before making the changes.

internal/network/command_private_link_attachment_connection_create.go
```diff
@@ -0,0 +1,102 @@
...
+
+	createPrivateLinkAttachmentConnection := networkingprivatelinkv1.NetworkingV1PrivateLinkAttachmentConnection{
+		Spec: &networkingprivatelinkv1.NetworkingV1PrivateLinkAttachmentConnectionSpec{
+			DisplayName:           networkingprivatelinkv1.PtrString(args[0]),
+			Environment:           &networkingprivatelinkv1.ObjectReference{Id: environmentId},
+			PrivateLinkAttachment: &networkingprivatelinkv1.ObjectReference{Id: attachment},
+		},
+	}
+
+	switch cloud {
```
MuweiHe, line 72: is the value of `cloud` validated? If it's some invalid value, will it return an error?

internal/network/command_private_link_attachment_connection_update.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (c *command) newPrivateLinkAttachmentConnectionUpdateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update <id>",
+		Short: "Update an existing private link attachment connection.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.privateLinkAttachmentConnectionUpdate,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Update the name of private link attachment connection "plattc-123456".`,
```
MuweiHe, line 20: "plattc-123456" are the id's normally this length?

internal/network/command_private_link_attachment_connection_create.go
syoung-cflt: It's validated at the api level as `spec.cloud` will be missing if the value of `cloud` is not something that we expect. This is demonstrated and tested in https://github.com/confluentinc/cli/blob/ef124afee2ec0894cd6c97abb15a01fe10b14bc4/test/test-server/networking_handlers.go#L1171 and https://github.com/confluentinc/cli/blob/ef124afee2ec0894cd6c97abb15a01fe10b14bc4/test/network_test.go#L523

internal/network/command_private_link_attachment_connection_update.go
syoung-cflt: Yes!

internal/network/command_private_link_attachment_connection_delete.go
sgagniere: @syoung-cflt just to clarify, do you mean that there are currently resources which we say are deleted at the end of the command, but are actually soft deleted?

syoung-cflt: I think for the current implementation in the CLI for deleting networking resources, we have
`network` - "Deleted..."
`peering`, `transit-gateway-attachment`, `private-link-access`, `private-link-attachment`, `private-link-attachment-connection` - "Requested to delete..."

I think some of the resources with the "Requested to delete" message would actually be soft-deleted by the definition above - "i.e users will get 204 response once it is deleted. the next "GET" request from the user will receive a 404". In this case, I think we could use a stronger language (i.e. "Delete ..." instead of "Requested to delete ...").

@sgagniere Does that answer your question?

sgagniere: Yup, thanks.

sgagniere (approved):

PR #2317, opened 2023-09-27
Add `flink statement create --wait`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Add `--wait` flag to `confluent flink statement create` that blocks until the statement status is no longer "PENDING"

```
% confluent flink statement create --compute-pool lfcp-9w8jwy --sql "SELECT CURRENT_TIMESTAMP;" --wait
[WARN] No service account provided. To ensure that your statements run continuously, use a service account instead of your user identity with `confluent iam service-account use` or `--service-account`. Otherwise, statements will stop running after 4 hours.
+---------------+--------------------------------------+
| Creation Date | 2023-09-27 21:28:18 +0000 UTC        |
| Name          | f08be65a-4800-4ea8-b74d-a467c94e12c3 |
| Statement     | SELECT CURRENT_TIMESTAMP;            |
| Compute Pool  | lfcp-9w8jwy                          |
| Status        | RUNNING                              |
+---------------+--------------------------------------+
```

References
----------
https://docs.google.com/document/d/1dZ5pmVdvTYykJqJX2tAk_elxCITRuAt90x5tXSd8RDA/edit

Test & Review
-------------
Integration test and manual testing

### DIFF
--- a/internal/flink/command_statement_create.go
+++ b/internal/flink/command_statement_create.go
@@ -1,6 +1,9 @@
 package flink
 
 import (
+	"fmt"
+	"time"
+
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 
@@ -12,6 +15,7 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/flink"
 	"github.com/confluentinc/cli/v3/pkg/flink/config"
 	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/retry"
 )
 
 func (c *command) newStatementCreateCommand() *cobra.Command {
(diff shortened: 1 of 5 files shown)


internal/flink/command_statement_create.go
```diff
@@ -111,6 +116,28 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
...
+			}
+
+			if statement.Status.GetPhase() == "PENDING" {
+				return fmt.Errorf(`statement phase is "%s"`, statement.Status.GetPhase())
+			}
+
+			return nil
+		})
+		if err != nil {
+			return err
```
sgagniere, line 137: Will this line return the `retry failed` error message? If so, maybe we should replace it with something more specific to this scenario.

Perhaps also adding a suggestion to run `describe` to see the current status if we timeout here.

sgagniere (approved):
+1 w/ small question

internal/flink/command_statement_create.go
brianstrauch: it'll say "retry failed: statement phase is PENDING"

sgagniere (approved):

PR #2318, opened 2023-09-27
Remove single-use global string constants

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Cleaning up the global namespace, starting with string constants that are better suited to be defined in the function in which they are used. This makes it easier to find bugs and quicker to read code. I deleted about 20 strings from this list that weren't being used.

Test & Review
-------------
Tests still pass


sgagniere (approved):
Accidentally hit the button re-request a review from myself, lol

PR #2319, opened 2023-09-28
 Key format splitting in `confluent kafka topic produce`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Resolve an issue causing `confluent kafka topic produce` to fail when using schema-based key formats with some delimiter choices

Checklist
---------
   * yes: ok

What
----
The current mechanism for separating the key and the value is to split on the first instance of the delimiter. If the delimiter character appears inside the key when using avro, jsonschema, or protobuf for the key format, this splits the string along the wrong character and causes the key serialization to fail.

For example, `{"CustomerId": 1, "Name": "My Name"}:message` with the default delimiter `:` will incorrectly split to `{"CustomerId"` and `1, "Name": "My Name"}:message`.

This PR changes the splitting strategy for schema based key formats to split only if the string preceding the delimiter is a valid json string (whether or not that json string matches the avro/jsonschema/protobuf schema will be checked by the serializer).

References
----------

Test & Review
-------------
I've added a number of unit test cases for the new splitting function.
manually tested w/ different key and value formats

Open Questions / Follow-ups
---------------------------
This implementation might not be efficient enough. If that's the case, we can revisit the implementation. Hopefully, the test cases are extensive enough to prevent any new bugs if we do that.


internal/kafka/command_topic_produce.go
```diff
@@ -231,6 +232,31 @@ func serializeMessage(keyMetaInfo, valueMetaInfo []byte, data, delimiter string,
...
+		return strings.TrimSpace(dataSplit[0]), strings.TrimSpace(strings.Join(dataSplit[1:], delimiter)), nil
+	}
+
+	key := dataSplit[0]
+	if json.Valid([]byte(strings.TrimSpace(key))) {
+		return strings.TrimSpace(key), strings.TrimSpace(strings.Join(dataSplit[1:], delimiter)), nil
+	}
+
+	for i, substr := range dataSplit[1:] {
+		key += delimiter + substr
```
sgagniere, line 251: @brianstrauch I went with this instead of the bracket matching idea for now because that requires more care: bracket characters appearing inside quotations need to be ignored, which means quotation marks need to be matched, which itself means we need to be careful about escaped quotation marks and backslashes, etc.

internal/kafka/command_topic_produce.go
```diff
@@ -205,22 +206,22 @@ func GetProduceMessage(cmd *cobra.Command, keyMetaInfo, valueMetaInfo []byte, to
 }
 
 func serializeMessage(keyMetaInfo, valueMetaInfo []byte, data, delimiter string, parseKey bool, keySerializer, valueSerializer serdes.SerializationProvider) ([]byte, []byte, error) {
-	var serializedKey, val string
+	var serializedKey []byte
+	var val string
```
brianstrauch, line 210: This should let you delete the `else` branch
```suggestion
	val := data
```

internal/kafka/command_topic_produce.go
brianstrauch: Cool, I guess this is O(n)? Hopefully won't take too long for really big inputs

pkg/errors/error_message.go
```diff
@@ -162,7 +162,8 @@ const (
 	TopicExistsOnPremErrorMsg            = `topic "%s" already exists for the Kafka cluster`
 	TopicExistsOnPremSuggestions         = "To list topics for the cluster, use `confluent kafka topic list --url <url>`."
 	FailedToProduceErrorMsg              = "failed to produce offset %d: %s\n"
-	MissingKeyErrorMsg                   = "missing key in message"
+	MissingKeyOrValueErrorMsg            = "missing key or value in message"
+	MissingOrMalformedKeyErrorMsg        = "missing or malformed key in message"
```
brianstrauch, line 166: Looks like this is only used in one package? Can we define it there instead of here?

internal/kafka/command_topic_produce_test.go
```diff
@@ -0,0 +1,98 @@
...
+	// Missing key (non-schema key format)
+	testCases = []splitTest{
+		{Data: `{"CustomerId": 1, "Name": "My Name"}:message`, Delimiter: "|"},
+		{Data: `message`, Delimiter: ":"},
+	}
+
+	for _, testCase := range testCases {
+		_, _, err := getKeyAndValue(false, testCase.Data, testCase.Delimiter)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), errors.MissingKeyOrValueErrorMsg)
```
brianstrauch, line 96: Shouldn't this be equal?

pkg/serdes/avro_serialization_provider.go
```diff
@@ -34,6 +34,10 @@ func (a *AvroSerializationProvider) GetSchemaName() string {
 	return AvroSchemaBackendName
 }
 
+func (a *AvroSerializationProvider) SchemaBased() bool {
```
brianstrauch, line 37:
```suggestion
func (a *AvroSerializationProvider) IsSchemaBased() bool {
```

brianstrauch (commented):
Thanks for testing so thoroughly!

internal/kafka/command_topic_produce.go
sgagniere: Yeah I'm hoping that it's not too bad on average, since I don't expect users to be entering really complex keys. And I don't validate the formatting of the value/message, which I let the serializer do.

internal/kafka/command_topic_produce.go
```diff
@@ -231,6 +235,31 @@ func serializeMessage(keyMetaInfo, valueMetaInfo []byte, data, delimiter string,
...
+
+	key := dataSplit[0]
+	if json.Valid([]byte(strings.TrimSpace(key))) {
+		return strings.TrimSpace(key), strings.TrimSpace(strings.Join(dataSplit[1:], delimiter)), nil
+	}
+
+	for i, substr := range dataSplit[1:] {
+		key += delimiter + substr
+		if json.Valid([]byte(strings.TrimSpace(key))) {
+			return strings.TrimSpace(key), strings.TrimSpace(strings.Join(dataSplit[i+2:], delimiter)), nil
```
brianstrauch, line 256: Is it possible to get an out of bounds panic here if `dataSplit` is too short?

pkg/serdes/serdes.go
```diff
@@ -22,6 +22,7 @@ type SerializationProvider interface {
 	LoadSchema(string, map[string]string) error
 	Serialize(string) ([]byte, error)
 	GetSchemaName() string
+	IsSchemaBased() bool
```
brianstrauch, line 25: Is there any way to infer this from the other functions, for instance, `GetSchemaName() != ""`?

internal/kafka/command_topic_produce.go
sgagniere: No. As long as `i+2 <= len(dataSplit)`, `dataSplit[i+2:]` is valid.

Since `i` ranges from `0` to `len(dataSplit[1:]) - 1 = len(dataSplit) - 2`, `i+2` is within bounds.

If `dataSplit[1:]` is empty (which shouldn't happen because there's a check at the top enforcing `len(dataSplit) >= 2`), then this is still fine because the loop doesn't run at all.

brianstrauch (approved):

internal/kafka/command_topic_produce_test.go
```diff
@@ -0,0 +1,96 @@
...
+	Key       string
+	Value     string
+}
+
+func TestGetKeyAndValue(t *testing.T) {
+	testCases := []splitTest{
+		// Different delimiters
+		{Data: `{"CustomerId": 1, "Name": "My Name"}:message`, Delimiter: ":", Key: `{"CustomerId": 1, "Name": "My Name"}`, Value: "message"},
+		{Data: `{"CustomerId": 1, "Name": "My Name"};message`, Delimiter: ";", Key: `{"CustomerId": 1, "Name": "My Name"}`, Value: "message"},
+		{Data: `{"CustomerId": 1, "Name": "My Name"}"message`, Delimiter: "\"", Key: `{"CustomerId": 1, "Name": "My Name"}`, Value: "message"},
```
brianstrauch, line 21:
```suggestion
		{Data: `{"CustomerId": 1, "Name": "My Name"}"message`, Delimiter: `"`, Key: `{"CustomerId": 1, "Name": "My Name"}`, Value: "message"},
```

PR #2321, opened 2023-09-29
Remove `go test` time limit for integration tests

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----
The integration tests take longer to run on the semaphore darwin machine. If they time out but the Linux tests don't, then semaphore reports `none tests failed`, which is confusing.

The default `go test` timeout is `10m`. This PR changes it to `0`, which is no time out. There shouldn't be any danger of semaphore running indefinitely, since we still have the 1 hour overall execution time limit.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ ifdef CI
 	export GOCOVERDIR=test/coverage && \
 	if [ -d $${GOCOVERDIR} ]; then rm -r $${GOCOVERDIR}; fi && \
 	mkdir $${GOCOVERDIR} && \
-	gotestsum --junitfile integration-test-report.xml -- -v -race $$(go list ./... | grep github.com/confluentinc/cli/v3/test) && \
+	gotestsum --junitfile integration-test-report.xml -- -timeout 0 -v -race $$(go list ./... | grep github.com/confluentinc/cli/v3/test) && \
 	go tool covdata textfmt -i $${GOCOVERDIR} -o test/coverage.out
 else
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -106,7 +106,7 @@ ifdef CI
 	export GOCOVERDIR=test/coverage && \
 	if [ -d $${GOCOVERDIR} ]; then rm -r $${GOCOVERDIR}; fi && \
 	mkdir $${GOCOVERDIR} && \
-	gotestsum --junitfile integration-test-report.xml -- -v -race $$(go list ./... | grep github.com/confluentinc/cli/v3/test) && \
+	gotestsum --junitfile integration-test-report.xml -- -timeout 0 -v -race $$(go list ./... | grep github.com/confluentinc/cli/v3/test) && \
```
sgagniere, line 109: Alternatively, we could just raise this to something like 15 or 20 minutes.

brianstrauch: Thought I did this a while ago... thanks!

brianstrauch: Can we add it to the other calls to `go test` just in case?

sgagniere: Will do!

brianstrauch (approved):
Thanks!!

PR #2323, opened 2023-09-29
Remove single-use error constants

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Follow-up to #2318, this time with errors.

Test & Review
-------------
Tests still pass


pkg/dynamic-config/dynamic_context.go
```diff
@@ -204,7 +204,10 @@ func (d *DynamicContext) KeyAndSecretFlags(cmd *cobra.Command) (string, string,
 	}
 
 	if apiKey == "" && apiSecret != "" {
-		return "", "", errors.NewErrorWithSuggestions(errors.PassedSecretButNotKeyErrorMsg, errors.PassedSecretButNotKeySuggestions)
+		return "", "", errors.NewErrorWithSuggestions(
+			"no API key specified",
+			"Use the `--api-key` flag to specify an API key.",
+		)
```
sgagniere, line 210: Are we putting the error and suggestion on different lines for all or some of the calls to `NewErrorWithSuggestions`?

brianstrauch: Just the ones that are really long if we don't

sgagniere (approved):

PR #2336, opened 2023-10-06
Fix and test Flink autocomplete

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Certain Flink commands had been written to autocomplete with the wrong resource, or were broken (returned no results). Fixed and tested all instances, and added additional autocompletion for `--database`.

Test & Review
-------------
Integration tests, manual testing

### DIFF
--- a/internal/flink/command.go
+++ b/internal/flink/command.go
@@ -1,11 +1,13 @@
 package flink
 
 import (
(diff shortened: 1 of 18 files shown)


sgagniere (approved):
lgtm

PR #2337, opened 2023-10-09
feat: add panic recovery wrapper for goroutines

YannickPferr:
Checklist
---------
   * yes: ok

What
----

Adds a panic recovery wrapper that can be used when spawning goroutines. This wrapper ensures we recover from panics inside goroutines, so they don't crash our main routine.

### DIFF
--- a/pkg/flink/internal/utils/panic.go
+++ b/pkg/flink/internal/utils/panic.go
@@ -1,0 +1,20 @@
+package utils
+
+func WithPanicRecovery(fn func()) func() {
+	return WithCustomPanicRecovery(fn, func() {
+		OutputErr("Error: internal error occurred")
+	})
+}
+
+func WithCustomPanicRecovery(fn func(), customRecovery func()) func() {
+	return func() {
+		defer func() {
+			if r := recover(); r != nil {
+				if customRecovery != nil {
+					WithPanicRecovery(customRecovery)()
+				}
+			}
+		}()
+		fn()
+	}
+}
--- a/pkg/flink/internal/utils/panic_test.go
+++ b/pkg/flink/internal/utils/panic_test.go
@@ -1,9 +1,11 @@
 package utils
 
 import (
-	"github.com/bradleyjkemp/cupaloy"
-	"github.com/confluentinc/cli/v3/pkg/flink/test"
 	"testing"
+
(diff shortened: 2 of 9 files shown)


pkg/flink/internal/utils/panic.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+func WithPanicRecovery(fn func()) func() {
+	return func() {
+		defer func() {
+			if r := recover(); r != nil {
+				OutputErr("Error: internal error occurred")
+			}
+		}()
+
```
luca-filipponi, line 10: unucessary space?

pkg/flink/internal/utils/panic_test.go
```diff
@@ -0,0 +1,18 @@
...
+func TestPanicRecoveryWrapper(t *testing.T) {
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go WithPanicRecovery(func() {
+		defer wg.Done()
+		panic("a panic in a goroutine")
+	})()
+
+	wg.Wait()
```
luca-filipponi, line 17: Not sure if this is a good test, if you have a bug here the test will wait indefinitely? 

If here what are you testing is the panic recovery, which advantage is this grouting giving to you?

pkg/flink/internal/utils/panic.go
YannickPferr: true, will remove

pkg/flink/internal/utils/panic_test.go
YannickPferr: Good point, but doesn't the `defer wg.Done()` ensure it's always called? 🤔 
Anyways, I think you're right, it's probably enough to just test the panic recovery behavior without the goroutine. I wanted to prove it also works with goroutines, but that might be unnecessary.

luca-filipponi: ooh right, you had that in a defer. So i guess that's fine, you can leave it as it

YannickPferr: No, I think you had a good point. I've refactored it a bit and now I can also reuse this wrapper in application.go (where we already had a custom panic recovery )

Guttz (approved):
Cool PR! Also really easy to review ✅

PR #2341, opened 2023-10-10
[CLI-2825] Add `print-offset` in confluent kafka topic consume

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `--print-offset` to `confluent kafka topic consume` for printing partition number and offset of each message

Bug Fixes
- PLACEHOLDER

Checklist
---------
   * yes: ok

What
----

```
mhe@C02DV0KVMD6T cli % confluent kafka topic consume -b tester --print-offset --timestamp
Starting Kafka Consumer. Use Ctrl-C to exit.
Timestamp:1696896989560 Partition:3 Offset:0	6
Timestamp:1696896990093 Partition:3 Offset:1	7
```

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/command_topic_consume.go
+++ b/internal/kafka/command_topic_consume.go
@@ -44,7 +44,7 @@ func (c *command) newConsumeCommand() *cobra.Command {
 	pcmd.AddKeyFormatFlag(cmd)
 	pcmd.AddValueFormatFlag(cmd)
 	cmd.Flags().Bool("print-key", false, "Print key of the message.")
-	cmd.Flags().Bool("print-offset", false, "Print offset of the message.")
+	cmd.Flags().Bool("print-offset", false, "Print partition number and offset of the message.")
 	cmd.Flags().Bool("full-header", false, "Print complete content of message headers.")
 	cmd.Flags().String("delimiter", "\t", "The delimiter separating each key and value.")
 	cmd.Flags().Bool("timestamp", false, "Print message timestamp in milliseconds.")
--- a/internal/kafka/command_topic_consume.go
+++ b/internal/kafka/command_topic_consume.go
@@ -44,6 +44,7 @@ func (c *command) newConsumeCommand() *cobra.Command {
 	pcmd.AddKeyFormatFlag(cmd)
 	pcmd.AddValueFormatFlag(cmd)
 	cmd.Flags().Bool("print-key", false, "Print key of the message.")
+	cmd.Flags().Bool("print-offset", false, "Print offset of the message.")
 	cmd.Flags().Bool("full-header", false, "Print complete content of message headers.")
 	cmd.Flags().String("delimiter", "\t", "The delimiter separating each key and value.")
 	cmd.Flags().Bool("timestamp", false, "Print message timestamp in milliseconds.")
@@ -91,6 +92,11 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
 	}
 
 	printKey, err := cmd.Flags().GetBool("print-key")
+	if err != nil {
+		return err
+	}
+
+	printOffset, err := cmd.Flags().GetBool("print-offset")
 	if err != nil {
 		return err
 	}
@@ -206,11 +212,12 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
 		Out:         cmd.OutOrStdout(),
 		Subject:     subject,
 		Properties: ConsumerProperties{
-			PrintKey:   printKey,
-			FullHeader: fullHeader,
-			Timestamp:  timestamp,
-			Delimiter:  delimiter,
-			SchemaPath: schemaPath,
+			PrintKey:    printKey,
+			PrintOffset: printOffset,
+			FullHeader:  fullHeader,
+			Timestamp:   timestamp,
+			Delimiter:   delimiter,
+			SchemaPath:  schemaPath,
 		},
 	}
 	return RunConsumer(consumer, groupHandler)
--- a/internal/kafka/confluent_kafka.go
+++ b/internal/kafka/confluent_kafka.go
@@ -236,7 +236,7 @@ func consumeMessage(message *ckafka.Message, h *GroupHandler) error {
 		}
 	}
 
-	messageString, err := getMessageString(message, valueDeserializer, h.Properties.Timestamp, h.Properties.PrintOffset)
+	messageString, err := getMessageString(message, valueDeserializer, h.Properties)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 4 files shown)


internal/kafka/confluent_kafka.go
```diff
@@ -243,6 +244,12 @@ func consumeMessage(message *ckafka.Message, h *GroupHandler) error {
 		jsonMessage = fmt.Sprintf("Timestamp: %d\t%s", message.Timestamp.UnixMilli(), jsonMessage)
 	}
 
+	if h.Properties.PrintOffset {
+		partition := message.TopicPartition.Partition
+		offset := message.TopicPartition.Offset.String()
+		jsonMessage = fmt.Sprintf("Offset: [%d:%s]\t%s", partition, offset, jsonMessage)
```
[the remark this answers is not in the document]
MuweiHe: `"[Partition:Offset]: [%d:%s]\t%s"`?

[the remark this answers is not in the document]
brianstrauch: How about `[partition=%d,offset=%s]\t%s`? Unless there's a standard way that other Kafka clients use?

internal/kafka/command_topic_consume.go
```diff
@@ -44,6 +44,7 @@ func (c *command) newConsumeCommand() *cobra.Command {
 	pcmd.AddKeyFormatFlag(cmd)
 	pcmd.AddValueFormatFlag(cmd)
 	cmd.Flags().Bool("print-key", false, "Print key of the message.")
+	cmd.Flags().Bool("print-offset", false, "Print offset of the message.")
```
brianstrauch, line 47: This also prints the partition number, right?

internal/kafka/confluent_kafka.go
```diff
@@ -243,6 +244,10 @@ func consumeMessage(message *ckafka.Message, h *GroupHandler) error {
 		jsonMessage = fmt.Sprintf("Timestamp: %d\t%s", message.Timestamp.UnixMilli(), jsonMessage)
 	}
 
+	if h.Properties.PrintOffset {
+		jsonMessage = fmt.Sprintf("[Partition:Offset]: [%d:%s]\t%s", message.TopicPartition.Partition, message.TopicPartition.Offset.String(), jsonMessage)
```
brianstrauch, line 248: I assume we print this for messages other than JSON messages? Can we fix the variable name here?

internal/kafka/confluent_kafka.go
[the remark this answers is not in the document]
sgagniere: I think I like Muwei's version more because it's closer to the existing formatting for `Timestamp: %d\t%s`.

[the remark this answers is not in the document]
brianstrauch: If we print timestamps and offsets, wouldn't that look a little weird? What if we added all of the "add-ons" to that list, i.e. `[partition=%d,offset=%s,timestamp=%s]\t%s`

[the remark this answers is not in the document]
sgagniere: That could be good. Do we have any breaking change rules for `topic consume` formatting that might factor in to this?

internal/kafka/confluent_kafka.go
MuweiHe: json message is concatenated after partition:offset. suggestion for a var name?

internal/kafka/confluent_kafka.go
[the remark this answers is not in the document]
MuweiHe:
```
mhe@C02DV0KVMD6T cli % Confluent kafka topic consume -b tester --timestamp --print-offset
Starting Kafka Consumer. Use Ctrl-C to exit.
[Timestamp:1696896988982 Partition:1 Offset:0]	5
[Timestamp:1696896990766 Partition:1 Offset:1]	9
```

internal/kafka/command_topic_consume.go
MuweiHe: yes. updated the description

internal/kafka/confluent_kafka.go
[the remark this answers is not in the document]
brianstrauch: > Do we have any breaking change rules for topic consume formatting that might factor in to this?

Good idea, let's avoid a breaking change here. Not sure what the existing formatting looks like, but I assume it doesn't use `[]`. Maybe let's drop the `[]` when there's only one value to maintain compatibility?

internal/kafka/confluent_kafka.go
brianstrauch: What about `message`? It seems like the message could be either JSON or non-JSON here.

internal/kafka/confluent_kafka.go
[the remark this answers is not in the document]
sgagniere: That sounds good to me.

internal/kafka/confluent_kafka.go
MuweiHe: that's the "kafka.message". I changed this to `messageString`

internal/kafka/confluent_kafka.go
```diff
@@ -246,13 +246,16 @@ func consumeMessage(message *ckafka.Message, h *GroupHandler) error {
 	}
 
 	if h.Properties.PrintOffset {
-		info += fmt.Sprintf(" Partition:%d Offset:%s", message.TopicPartition.Partition, message.TopicPartition.Offset.String())
+		if len(info) > 0 {
+			info = fmt.Sprintf("%s ", info)
+		}
```
brianstrauch, line 251: Can we make a list of key:value pairs, and then use `strings.Join()` to combine them into a space-delimited string?

brianstrauch: Unit tests would be great for this, especially since we don't have integration tests here!

MuweiHe: added unit test!

sgagniere (approved):

internal/kafka/confluent_kafka.go
```diff
@@ -272,6 +257,26 @@ func consumeMessage(message *ckafka.Message, h *GroupHandler) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	var info []string
+	if timestamp {
+		info = append(info, fmt.Sprintf("Timestamp:%d", message.Timestamp.UnixMilli()))
+	}
+	if printOffset {
+		info = append(info, fmt.Sprintf("Partition:%d Offset:%s", message.TopicPartition.Partition, message.TopicPartition.Offset.String()))
```
brianstrauch, line 271: Append each individually here?

brianstrauch (approved):

internal/kafka/confluent_kafka.go
```diff
@@ -268,7 +268,8 @@ func getMessageString(message *ckafka.Message, valueDeserializer serdes.Deserial
 		info = append(info, fmt.Sprintf("Timestamp:%d", message.Timestamp.UnixMilli()))
 	}
 	if properties.PrintOffset {
-		info = append(info, fmt.Sprintf("Partition:%d Offset:%s", message.TopicPartition.Partition, message.TopicPartition.Offset.String()))
+		info = append(info, fmt.Sprintf("Partition:%d", message.TopicPartition.Partition))
+		info = append(info, fmt.Sprintf("Offset:%s", message.TopicPartition.Offset.String()))
```
brianstrauch, line 272: Untested, but I think we should be able to do this:
```suggestion
		info = append(info, fmt.Sprintf("Offset:%v", message.TopicPartition.Offset))
```

PR #2342, opened 2023-10-10
[CLI-2829] Remove direct dependency on `github.com/pkg/errors`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
The errors package is deprecated and can be replaced with standard library functions as of Go 2. Prefer `fmt.Errorf()` over `errors.*`.

Test & Review
-------------
Tests still pass

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -2,24 +2,22 @@ package errors
 
 import (
 	"fmt"
-
-	"github.com/pkg/errors"
 )
 
 func New(msg string) error {
-	return errors.New(msg)
+	return fmt.Errorf(msg)
 }
 
 func Wrap(err error, msg string) error {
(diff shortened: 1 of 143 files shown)


pkg/errors/errors.go
```diff
@@ -2,26 +2,8 @@ package errors
 
 import (
 	"fmt"
-
-	"github.com/pkg/errors"
 )
 
-func New(msg string) error {
```
sgagniere, line 9: What about keeping these wrappers but just changing the implementation to use `fmt.Errorf`?

brianstrauch: These wrappers shadow the old functions that were deprecated and that we've been asked to move away from. I think the Go stdlib functions are more idiomatic and explicit...

sgagniere (approved):

PR #2343, opened 2023-10-11
feat: support `confluent flink statement list --status {status}`

MarTango:
Release Notes
-------------
New Features
- Add `--status` flag to filter `confluent flink statement list` by status

Checklist
---------
   - [x] yes: ok
   - [ ] no: DO NOT MERGE until the required features are live in prod  

What
----

This commit enables the filtering of flink statement results by their status. Filtering is done on the client-side to keep things simple, but we may want to add filtering logic to the server-side if latency becomes an issue.

References
----------

Internal JIRA

Test & Review
-------------

Golden tests have been added/updated

Open Questions / Follow-ups
---------------------------


internal/flink/command_statement_list.go
```diff
@@ -35,13 +55,31 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
...
 	statements, err := client.ListAllStatements(environmentId, c.Context.GetCurrentOrganization(), c.Context.GetCurrentFlinkComputePool())
 	if err != nil {
 		return err
 	}
 
 	list := output.NewList(cmd)
 	for _, statement := range statements {
+		// filter out statements with a non-matching status
+		if status != "" && statement.Status.GetPhase() != status {
+			continue
```
luca-filipponi, line 80: `continue` can be harder to maintain, should flip the condition and ad only when you need to

internal/flink/command_statement_list.go
```diff
@@ -1,12 +1,27 @@
...
+	"strings"
+
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/errors"
 	"github.com/confluentinc/cli/v3/pkg/output"
 )
 
+var allowedStatuses = []string{
```
swist, line 15: This should come from an enum somewhere

internal/flink/command_statement_list.go
swist: or just use lo.Filter

swist (dismissed):

internal/flink/command_statement_list.go
```diff
@@ -74,12 +76,12 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
...
 
 	list := output.NewList(cmd)
-	for _, statement := range statements {
-		// filter out statements with a non-matching status
-		if status != "" && statement.Status.GetPhase() != status {
-			continue
-		}
 
+	statementsWithMatchingStatus := lo.Filter(statements, func(stmt v1beta1.SqlV1beta1Statement, _ int) bool {
+		return status == "" || stmt.Status.GetPhase() == status
```
brianstrauch, line 81: The check for `status == ""` should be moved outside of this function. No need to check it on every iteration.

internal/flink/command_statement_list.go
```diff
@@ -74,12 +76,12 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
 	}
 
 	list := output.NewList(cmd)
-	for _, statement := range statements {
-		// filter out statements with a non-matching status
-		if status != "" && statement.Status.GetPhase() != status {
-			continue
-		}
 
+	statementsWithMatchingStatus := lo.Filter(statements, func(stmt v1beta1.SqlV1beta1Statement, _ int) bool {
```
brianstrauch, line 80:
```suggestion
	statementsWithMatchingStatus := lo.Filter(statements, func(statement v1beta1.SqlV1beta1Statement, _ int) bool {
```

internal/flink/command_statement_list.go
brianstrauch: I wish our SDKs could generate lists of enums... since it's possible that this list gets out of sync with the backend, it seems like a good idea to only use this list for autocomplete, and not for erroring on invalid flag values.

test/fixtures/output/flink/statement/list-help.golden
```diff
@@ -10,6 +10,7 @@ Flags:
       --environment string    Environment ID.
       --context string        CLI context name.
   -o, --output string         Specify the output format as "human", "json", or "yaml". (default "human")
+      --status string         Filter the results by statement status
```
brianstrauch, line 13:
```suggestion
      --status string         Filter the results by statement status.
```

test/fixtures/output/flink/statement/list-completed.golden
```diff
@@ -0,0 +1,4 @@
+          Creation Date         |         Name         |     Statement      | Compute Pool |  Status   |       Status Detail         
+--------------------------------+----------------------+--------------------+--------------+-----------+-----------------------------
+  2022-01-01 00:00:00 +0000 UTC | 22222222-2222-2222-2 | CREATE TABLE test; | lfcp-123456  | COMPLETED | SQL statement is completed
```
brianstrauch, line 3: Out of curiosity: it looks like the backend doesn't care if a statement ends with a semi-colon. For the purposes of user-facing examples, should we recommend adding a semi-colon or not?
```suggestion
  2022-01-01 00:00:00 +0000 UTC | 22222222-2222-2222-2 | CREATE TABLE test; | lfcp-123456  | COMPLETED | SQL statement is completed  
```

brianstrauch:
FYI @MarTango, our release notes get autogenerated from the "Release Notes" section of the PR description. I cleaned yours up, but in the future try to avoid using different formatting than what already exists!

MarTango:
> it seems like a good idea to only use this list for autocomplete, and not for erroring on invalid flag values.

The issue I foresee is a user inputting an invalid status, and just getting a `None found.` 

How about logging a warning if the provided status is not recognised?

internal/flink/command_statement_list.go
```diff
@@ -1,12 +1,29 @@
...
 
 import (
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/samber/lo"
 	"github.com/spf13/cobra"
 
+	"github.com/confluentinc/ccloud-sdk-go-v2/flink-gateway/v1beta1"
```
brianstrauch, line 11: Run `make lint`!
```suggestion
	"github.com/confluentinc/ccloud-sdk-go-v2/flink-gateway/v1beta1"
	
```

internal/flink/command_statement_list.go
MarTango: > it seems like a good idea to only use this list for autocomplete, and not for erroring on invalid flag values.

The issue I foresee is a user inputting an invalid status, and just getting a None found.

How about logging a warning if the provided status is not recognised (instead of erroring)?

brianstrauch: :+1:

test/fixtures/output/flink/statement/list-completed.golden
swist: the backend doesn't care but the client recognises semicolons as statement terminator

internal/flink/command_statement_list.go
```diff
@@ -65,10 +64,7 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
 	status = strings.ToUpper(status)
 
 	if status != "" && !slices.Contains(allowedStatuses, status) {
-		return errors.NewErrorWithSuggestions(
-			"invalid value for flag --status",
-			fmt.Sprintf("Please select a value from the following: [%s]", strings.Join(allowedStatuses, ", ")),
-		)
+		log.CliLogger.Warnf("Unrecognized status '%s'. recognized statuses: [%s]", status, strings.Join(allowedStatuses, ", "))
```
brianstrauch, line 67:
```suggestion
		log.CliLogger.Warnf(`Invalid status "%s". Valid statuses are %s.`, status, utils.ArrayToCommaDelimitedString(allowedStatuses, "and"))
```

test/fixtures/output/flink/statement/list-completed.golden
brianstrauch: Sounds like we should recommend adding a semi-colon, in that case?

internal/flink/command_statement_list.go
```diff
@@ -1,12 +1,30 @@
...
+	"RUNNING",
+	"COMPLETED",
+	"DELETING",
+	"FAILING",
+	"FAILED",
+	"STOPPED",
+}
+
 func (c *command) newStatementListCommand() *cobra.Command {
 	cmd := &cobra.Command{
```
brianstrauch, line 29: Can you add the `Examples` field with one example using `--status`?

test/flink_test.go
```diff
@@ -65,6 +65,8 @@ func (s *CLITestSuite) TestFlinkStatement() {
 		{args: "flink statement delete my-statement --force --cloud aws --region eu-west-1", fixture: "flink/statement/delete.golden"},
 		{args: "flink statement list --cloud aws --region eu-west-1", fixture: "flink/statement/list.golden"},
 		{args: "flink statement list --cloud aws --region eu-west-1 -o yaml", fixture: "flink/statement/list-yaml.golden"},
+		{args: "flink statement list --cloud aws --region eu-west-1 --status COMPLETED", fixture: "flink/statement/list-completed.golden"},
+		{args: "flink statement list --cloud aws --region eu-west-1 --status PENDING", fixture: "flink/statement/list-pending.golden"},
```
brianstrauch, line 69: Let's test that the uppercase function works. This is also how we should recommend users to enter this in our examples.
```suggestion
		{args: "flink statement list --cloud aws --region eu-west-1 --status completed", fixture: "flink/statement/list-completed.golden"},
		{args: "flink statement list --cloud aws --region eu-west-1 --status pending", fixture: "flink/statement/list-pending.golden"},
```

internal/flink/command_statement_list.go
```diff
@@ -62,7 +67,7 @@ func (c *command) statementList(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	status = strings.ToUpper(status)
+	status = strings.ToLower(status)
```
brianstrauch, line 70: Interesting, most APIs at Confluent only accept uppercase constants.

test/fixtures/output/flink/statement/list-help.golden
```diff
@@ -3,6 +3,11 @@ List Flink SQL statements.
 Usage:
   confluent flink statement list [flags]
 
+Examples:
+List running statements.
+
+  $ confluent flink compute-pool list --status running
```
brianstrauch, line 9: Thanks!

brianstrauch (approved):

PR #2066, opened 2023-07-06
bugfix: do not require compute pool flag to list statements

swist:
Statements are environment scoped

Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
1. [CRUCIAL] Is the change for features that are already live in prod?
   * yes: ok
   * no: DO NOT MERGE until the required features are live in prod  

What
----

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------

PR #2344, opened 2023-10-11
Feat/statement list with cp or cloudregion

Guttz:
Release Notes
-------------

New Features
- `confluent flink statement list` now supports either only providing the `--compute-pool` flag or a `--cloud --region`
 pair.
 
Checklist
---------
   * yes: ok

What
----

Requiring the --cloud and --region flag even if you provide --compute-pool was confusing for users, since the --compute-pool exists inside one cloud and one region. Experience is more consistent now, since you list the statements for a compute-pool without having to also provide the region and cloud provider for it.

References
----------

Test & Review
-------------

golden tests

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/flink/command_statement_list.go
+++ b/internal/flink/command_statement_list.go
@@ -25,7 +25,13 @@ func (c *command) newStatementListCommand() *cobra.Command {
 }
 
 func (c *command) statementList(cmd *cobra.Command, args []string) error {
-	client, err := c.GetFlinkGatewayClient(false)
+	computePoolOnly := false
+
+	if c.Context.GetCurrentFlinkComputePool() != "" {
+		computePoolOnly = true
+	}
+
+	client, err := c.GetFlinkGatewayClient(computePoolOnly)
 	if err != nil {
 		return err
 	}
--- a/test/fixtures/output/flink/statement/list-compute-pool-yaml.golden
+++ b/test/fixtures/output/flink/statement/list-compute-pool-yaml.golden
@@ -1,0 +1,12 @@
+- creation_date: 2022-01-01T00:00:00Z
+  name: 22222222-2222-2222-2
+  statement: CREATE TABLE test;
+  compute_pool: lfcp-123456
+  status: COMPLETED
+  status_detail: SQL statement is completed
+- creation_date: 2023-01-01T00:00:00Z
+  name: 11111111-1111-1111-1
+  statement: CREATE TABLE test;
+  compute_pool: lfcp-123456
+  status: COMPLETED
+  status_detail: SQL statement is completed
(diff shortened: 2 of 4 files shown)


YannickPferr (approved):

brianstrauch (dismissed):
This seems to contradict https://github.com/confluentinc/cli/pull/2066, cc @swist. What's the reason for changing this back? Should we make this change to all subcommands of `confluent flink statement`?

Guttz:
> This seems to contradict #2066, cc @swist. What's the reason for changing this back? Should we make this change to all subcommands of `confluent flink statement`?

I don't think it contradicts it. --compute-pool is still optional and statements do not belong to a --compute-pool. 

We're just making the user life easier by allowing them to only provide the compute-pool and fetching the env/cloud from it instead of them having to provide all three flags when they want to fetch statements that were issued to one --compute-pool.

brianstrauch:
> > This seems to contradict #2066, cc @swist. What's the reason for changing this back? Should we make this change to all subcommands of `confluent flink statement`?
> 
> I don't think it contradicts it. --compute-pool is still optional and statements do not belong to a compute pool.
> 
> We're just making the user life easier by allowing them to only provide the compute-pool and fetching the env/cloud from it instead of them having to provide all three flags when they want to fetch statements that were issued to one --compute-pool.

Makes sense, what about the following?

> Should we make this change to all subcommands of confluent flink statement?

Guttz:
`statement list` is the only command that has the `--compute-pool` flag, apart from `statement create` where we already use the compute-pool for fetching the env/region by doing `c.GetFlinkGatewayClient(true)`. 

I think list is the only place we can do this small change to make the user's life easier - don't think we can add it to any other command..

internal/flink/command_statement_list.go
```diff
@@ -25,7 +25,13 @@ func (c *command) newStatementListCommand() *cobra.Command {
 }
 
 func (c *command) statementList(cmd *cobra.Command, args []string) error {
-	client, err := c.GetFlinkGatewayClient(false)
+	computePoolOnly := false
+
+	if c.Context.GetCurrentFlinkComputePool() != "" {
+		computePoolOnly = true
+	}
```
brianstrauch, line 32: This logic seems weird. What if all three flags, `--cloud`, `--region`, and `--compute-pool` are set? It looks like we'll use `--compute-pool` (which was originally designed to be a filter flag) and ignore `--cloud` and `--region`...

Guttz: You're right, we'll use the --compute-pool, but that's imo also better UX for the user because it prevents him getting empty results if he has a typo and starts wondering why (happened to our engineers yesterday and that's why I created the PR to make the UX for the command a bit better). 

If the user provides a compute-pool he wants to use for filtering and we didn't fetch the cloud/region from the compute pool, he *had* to also provide exactly the right provider/region or else he would get empty results and wonder why.

brianstrauch: Maybe we should be explicitly checking `--compute-pool` instead of using `c.Context.GetCurrentFlinkComputePool()`, then. If the user does `confluent flink compute-pool use` followed by `confluent flink statement list --cloud <> --region <>` it'll actually use the set compute pool and ignore the cloud and region that are passed in.

Guttz: Makes sense! Feel free to move on and do it 👍

PR #2346, opened 2023-10-11
[CLI-2832] Replace `gon` with `notarytool`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
It appears that there is no plan to update `gon` to replace `altool` with `notarytool`, which Apple requires that we switch to in 21 days: https://github.com/mitchellh/gon/issues/68. This PR migrates off of `gon` in favor of calling `codesign` and `notarytool` directly.

Test & Review
-------------
Manually tested release with `DRY_RUN=true`


.goreleaser.yml
```diff
@@ -57,8 +57,10 @@ builds:
     hooks:
       post:
         - cmd: make download-licenses
-        - cmd: bash scripts/gon_filepath_editor.sh {{ .Path }} amd64
-        - cmd: gon dist/gon_confluent_darwin_amd64.hcl
+        - cmd: 'codesign -s "Developer ID Application: Confluent, Inc." -v {{ .Path }} --options=runtime'
+          output: true
+        - cmd: zip "{{ .Path }}_signed.zip" {{ .Path }}
+        - cmd: xcrun notarytool submit "{{ .Path }}_signed.zip" --apple-id {{ .Env.AC_USERNAME }} --password {{ .Env.AC_PASSWORD }} --team-id RTSX8FNWR2 --wait
```
sgagniere, line 63: Is the team-id something that we're good to hardcode here?

brianstrauch: Yep, it's similar to hardcoding an email, but we all share it (decided this would be better than asking you all to add yet another environment variable to your zshrc)

PR #2347, opened 2023-10-11
Remove majority of references to ccloud-sdk-go-v2 client in `DynamicContext`

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Ran into a import cycle while trying to implement auth token refresh on HTTP retry. One solution is to remove ccloud-sdk-go-v2 from the config package, which also will help remove complexity from the config package in general. The only remaining step is to rewrite `FindKafkaCluster()` which unfortunately has lots of references throughout the entire codebase.

Test & Review
-------------
Tests still pass.


sgagniere (approved):
+1

PR #2348, opened 2023-10-12
fix: overlaying text during reverse search for multiline statements

YannickPferr:
Checklist
---------
   * yes: ok

What
----

We had a bug in reverse search when starting it from a multiline buffer. The buffer was not cleared correctly, leading to overlaying text:
image

This PR fixes that bug and also makes sure we clear the buffer before inserting the selected suggestion. Before, the suggestion was just inserted at the cursor position and the rest of what you had in the buffer was still there (which I don't think is desired). Now, after exiting the reverse search, your buffer will contain only what reverse search suggested.


Guttz (approved):
Lgtm! Also tested it and it works ✅

PR #2351, opened 2023-10-12
Use comma delimited string for human BYOK role output

brianstrauch:
Checklist
---------
   * yes: ok

What
----
For `confluent byok [create|describe|list]` commands, instead of `[a b c]`, print `a, b, c`. The same problem exists for at least 4 other commands, will create a few more follow-up PRs.

Test & Review
-------------
Update integration tests

### DIFF
--- a/internal/byok/command_create.go
+++ b/internal/byok/command_create.go
@@ -75,61 +75,58 @@ func (c *command) newCreateCommand() *cobra.Command {
 	return cmd
 }
 
-func (c *command) createAwsKeyRequest(keyArn string) *byokv1.ByokV1Key {
-	return &byokv1.ByokV1Key{Key: &byokv1.ByokV1KeyKeyOneOf{ByokV1AwsKey: &byokv1.ByokV1AwsKey{
+func (c *command) createAwsKeyRequest(keyArn string) byokv1.ByokV1Key {
+	return byokv1.ByokV1Key{Key: &byokv1.ByokV1KeyKeyOneOf{ByokV1AwsKey: &byokv1.ByokV1AwsKey{
 		KeyArn: keyArn,
 		Kind:   "AwsKey",
 	}}}
 }
 
-func (c *command) createAzureKeyRequest(cmd *cobra.Command, keyString string) (*byokv1.ByokV1Key, error) {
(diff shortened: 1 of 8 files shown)


internal/byok/command_create.go
```diff
@@ -172,22 +169,22 @@ func renderAzureEncryptionPolicy(key *byokv1.ByokV1Key) (string, error) {
...
-		fmt.Sprintf("    --role \"%s\" \\", keyVaultCryptoServiceEncryptionUser),
-		fmt.Sprintf("    --scope \"$(az keyvault show --name \"%s\" --query id --output tsv)\" \\", vaultName),
-		fmt.Sprintf("    --assignee-object-id \"%s\" \\", objectId),
-		"    --assignee-principal-type ServicePrincipal && \\",
-		"az role assignment create \\",
-		fmt.Sprintf("    --role \"%s\" \\", keyVaultReader),
-		fmt.Sprintf("    --scope \"$(az keyvault show --name \"%s\" --query id --output tsv)\" \\", vaultName),
-		fmt.Sprintf("    --assignee-object-id \"%s\" \\", objectId),
+		`az role assignment create \`,
+		fmt.Sprintf(`    --role "%s" \`, keyVaultCryptoServiceEncryptionUser),
```
MuweiHe, line 173: use \t ?

internal/byok/command_list.go
```diff
@@ -67,27 +65,28 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
...
-			Provider:  key.GetProvider(),
-			State:     key.GetState(),
-			CreatedAt: key.Metadata.CreatedAt.String(),
-			UpdatedAt: updatedAt,
-			DeletedAt: deletedAt,
-		})
 	}
 
+	// The API returns a list sorted by creation date already
+	list.Sort(false)
```
MuweiHe, line 88: this has to be explicitly set?

internal/byok/command_create.go
brianstrauch: I'd rather use 4 spaces, since different terminals display tabs differently.

internal/byok/command_list.go
brianstrauch: Yep, the default is true

MuweiHe (approved):

PR #2353, opened 2023-10-12
Use comma delimited string for human stream share output

brianstrauch:
Checklist
---------
   * yes: ok

What
----
Follow up to https://github.com/confluentinc/cli/pull/2351, this time with the "Resources" and "Network Zonal Subdomains" fields of the `confluent stream-share` commands.

Test & Review
-------------
Updated integration tests

### DIFF
--- a/internal/stream-share/command_consumer_redeem.go
+++ b/internal/stream-share/command_consumer_redeem.go
@@ -13,21 +13,38 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/output"
 )
 
-type redeemOut struct {
-	Id                         string   `human:"ID" serialized:"id"`
-	ApiKey                     string   `human:"API Key" serialized:"api_key"`
-	ApiSecret                  string   `human:"API Secret" serialized:"api_secret"`
-	KafkaBootstrapUrl          string   `human:"Kafka Bootstrap URL" serialized:"kafka_bootstrap_url"`
-	SchemaRegistryApiKey       string   `human:"Schema Registry API Key" serialized:"schema_registry_api_key"`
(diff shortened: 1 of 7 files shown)


internal/stream-share/command_consumer_redeem.go
```diff
@@ -13,21 +13,38 @@ import (
...
+	NetworkPrivateLinkData     string `human:"Network Private Link Data"`
+}
+
+type redeemSerializedOut struct {
+	Id                         string   `serialized:"id"`
+	ApiKey                     string   `serialized:"api_key"`
+	ApiSecret                  string   `serialized:"api_secret"`
+	KafkaBootstrapUrl          string   `serialized:"kafka_bootstrap_url"`
+	SchemaRegistryApiKey       string   `serialized:"schema_registry_api_key"`
+	SchemaRegistrySecret       string   `serialized:"schema_registry_secret"`
```
MuweiHe, line 39: api secret?

brianstrauch: Thoughts on renaming this field to "API secret"? @confluentinc/data-exchange

sgagniere (approved):
Would we be able to roll this into the table printing code in the future?

edit: I mean processing a string array into a comma delimited list for human output instead of creating two output structs

brianstrauch:
> Would we be able to roll this into the table printing code in the future?
> 
> edit: I mean processing a string array into a comma delimited list for human output instead of creating two output structs

As long as we're guaranteed not to have exceptions... there might be a field in the future that we want to delimit with newlines instead of commas

PR #2354, opened 2023-10-13
Inspect code with Goland static analysis tools

brianstrauch:
Checklist
---------
   * yes: ok

What
----
* Deleted dead public functions and constants
* Code style suggestions (SNAKE_CASE to CamelCase, etc.)
* Remove redundancies and standardize function headers

Test & Review
-------------
Tests still pass


Guttz (approved):

internal/flink/command.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/config"
 )
 
+const serviceAccountWarning = "[WARN] No service account provided. To ensure that your statements run continuously, " +
```
MuweiHe, line 13: we may want to make those style more consistent. in your previous pr it was using `[WARNING]`

brianstrauch: Good point, changing the other PR to use `[WARN]`! (The goal is to match the style of the logging output)

internal/kafka/command_topic_update.go
```diff
@@ -19,10 +19,7 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/types"
 )
 
-const (
-	updateTopicConfigRestMsg    = "Updated the following configuration values for topic \"%s\"%s:\n"
-	readOnlyConfigNotUpdatedMsg = "(read-only configs were not updated)"
-)
+const updateTopicConfigRestMsg = "Updated the following configuration values for topic \"%s\"%s:\n"
```
MuweiHe, line 22: is this message being used as error or suggestion? it's capitalized.

brianstrauch: Looks like it's only used in one place though... updating!

MuweiHe (approved):
Thank you! :)

PR #2360, opened 2023-10-17
fix: put --fake-gateway behind local const

Guttz:
Release Notes
-------------

Checklist
---------
   * yes: ok

What
----

We've hidden a debugging mode we use behind a local const instead of cfg.isTest so it's not shown in the docs.

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


YannickPferr (approved):

Guttz:
Merging this so docs team can regenerate the docs without the fake-gateway flag

PR #2361, opened 2023-10-18
feat: add retry when GW returns 409 for GetStatementResults

YannickPferr:
Checklist
---------
   - [x] yes: ok

What
----

The CLI recently introduced a change to improve latency by not waiting for the statement phase to be in the RUNNING state as designed initially but rather to try and optimistically fetch results as soon as the status.result_schema is available. 
This has resulted in a race with 500s returned earlier(replaced by 409 now) as the backend was not ready to return results. 

This PR introduces a simple retry for this case. We retry 10 times in 100ms interval (so we wait for 1s in total) and stop the retry when the GetStatementResults endpoint either returns success or an error that is not 409.

### DIFF
--- a/pkg/flink/internal/store/store.go
+++ b/pkg/flink/internal/store/store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


swist (approved):

pkg/flink/internal/store/store.go
```diff
@@ -168,6 +169,23 @@ func (s *Store) FetchStatementResults(statement types.ProcessedStatement) (*type
 	return &statement, nil
 }
 
+// TODO: remove this when backend fixes latency problem/the RUNNING state is set with the ResultsSchema
```
Guttz, line 172: Maybe add the related jira ticket we have?

```suggestion
// TODO: remove this when backend fixes latency problem/the RUNNING state is set with the ResultsSchema
// Related ticket: https://confluentinc.atlassian.net/browse/KFS-1211
```

Guttz (approved):

pkg/flink/internal/store/store.go
YannickPferr: Done!

PR #2363, opened 2023-10-18
feat: display warning for incompatible flink shell versions

YannickPferr:
Checklist
---------
   - [x] yes: ok

What
----

Due to a recent breaking change in the flink gateway (scheduled to be released on Oct 23th, 2023) users with CLI versions v3.36.0, v3.37.0 and v3.38.0 will not be able to submit SELECT statements anymore. We want to show these users a warning to let them know they should update their CLI.


YannickPferr:
This doesn't make sense, since only users with the latest version will have that piece of code (rendering it useless).
Also, the CLI already displays a warning if there is a more recent version.

PR #2365, opened 2023-10-20
[CLI-2823] Add chocolatey to goreleaser config

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------
   - [ ] yes: ok

What
----
Add a new Windows build (w/ updates disabled) for Chocolate.
Add the chocolatey block to the goreleaser file.

References
----------
https://goreleaser.com/customization/chocolatey/

Test & Review
-------------
TODO: test locally

Open Questions / Follow-ups
---------------------------


.goreleaser.yml
```diff
@@ -178,6 +178,33 @@ builds:
...
+      - CGO_ENABLED=1
+      - CC=x86_64-w64-mingw32-gcc
+      - CXX=x86_64-w64-mingw32-g++
+      - CGO_LDFLAGS=${CGO_LDFLAGS} -static
+    goos:
+      - windows
+    goarch:
+      - amd64
+    hooks:
+      # TODO: We might have to convert some of these to global before/after hooks since we'll now have multiple Windows binaries w/ different ldflags
```
brianstrauch, line 200: Agreed, let's try to avoid running `az login` more than once. When this is automated in Semaphore it won't result in an annoying pop-up, but might result in a longer release time.

brianstrauch: (and I guess we only need to download the signing secret once, too)

.goreleaser.yml
```diff
@@ -284,6 +321,7 @@ blobs:
     - archive
     - archive-alpine
     - archive-homebrew
+    - archive-chocolatey
```
brianstrauch, line 324: Nit: alphabetize

.goreleaser.yml
```diff
@@ -308,4 +346,27 @@ brews:
...
+    copyright: 2023 Confluent Inc.
+    license_url: https://github.com/confluentinc/cli/blob/main/LICENSE
+    require_license_acceptance: false
+    project_source_url: https://github.com/confluentinc/cli
+    bug_tracker_url: https://github.com/confluentinc/cli/issues
+    summary: CLI for Confluent Cloud and Confluent Platform
+    description: CLI for Confluent Cloud and Confluent Platform
+    release_notes: https://docs.confluent.io/confluent-cli/current/release-notes.html
+    # TODO
+    api_key: "{{ .Env.CHOCOLATEY_API_KEY }}"
```
brianstrauch, line 370: Should we be reading this directly from `vault` instead?

.goreleaser.yml
```diff
@@ -178,6 +178,33 @@ builds:
       post:
         - cmd: ./lib/osslsigncode sign -n "Confluent CLI" -i "https://confluent.io" -pkcs12 CLIEVCodeSigningCertificate2.pfx -in {{ .Path }} -out {{ .Path }}
         - cmd: rm CLIEVCodeSigningCertificate2.pfx
+  - id: confluent-windows-amd64-chocolatey
```
brianstrauch, line 181: We might want to come up with a more generic name, i.e. `confluent-windows-amd64-noupdate` in case we want to support `scoop` (or other Windows package managers) in the future. We should probably consider doing the same for the linux/darwin builds too.

.goreleaser.yml
sgagniere: Aside from just the annoyance of two popups, I actually discovered a pretty annoying thing when I was testing this: if you log in w/ one of the tabs and close out the other, goreleaser just blocks on that pre-hook.

Converting it into a global `before` hook worked perfectly, though (I'll update the remote branch soon).

For the downloading, we have some options. What I tried is downloading to two separate files so that the removal post hook doesn't remove the file before the other build's post hook can sign with it. This works, although it's a bit clunky carrying two copies of the signing file with different names.

The other option is to download the file in the global `before` hook and remove it in the global `after` hook. The documentation for the after hook has weird wording which is why I didn't try it first.

.goreleaser.yml
sgagniere: Yeah if we do end up using the chocolatey repo, that's what we'll do. This is just a "don't forget this" placeholder.

docker/Dockerfile_windows_amd64
```diff
@@ -0,0 +1,15 @@
+FROM --platform=linux/amd64 ubuntu:jammy-20231211.1
```
brianstrauch, line 1: I think we have other Dockerfiles like this one. Let's make sure they're all using the same ubuntu version?

scripts/build_windows.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+function cleanup {
+  shred --force --remove --zero --iterations=10 CLIEVCodeSigningCertificate2.pfx
```
brianstrauch, line 4: Probably not the end of the world if we don't do this since the CI machines get reset between runs.

sgagniere: True, but I'm writing this so far on the assumption that we might be releasing before we fully switch to the CI.

docker/Dockerfile_windows_amd64
sgagniere: Can't, unfortunately.

The other Ubuntu image doesn't have a new enough version of `mingw-w64` to be able to cross compile for Windows.

This Ubuntu image has `glibc` version 2.35, which is too high for cross compiling Linux arm64 without a breaking change.

PR #2367, opened 2023-10-20
[CLI-2831] Allow api-key authenticated produce and consume in cloud

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Allow `confluent kafka topic list/produce/consume` in Confluent Cloud without login, when using a context of Kafka cluster bootstrap, Kafka API key

Bug Fixes
- PLACEHOLDER

Checklist
---------
   - [ ] yes: ok

What
----

Create a pair of kafka cluster api key, and discover the endpoint of the cluster. Create a new context:
```
confluent context create test --bootstrap <cluster-endpoint> --api-key <kafka-api-key> --api-secret <kafka-api-secret>
```
use that context, you can then list topics:
```
confluent kafka topic list
```
then produce and consume. if use schema, provide schema endpoint and api key.
```
confluent kafka topic produce hello --value-format avro --schema ~/test.json --schema-registry-endpoint https://psrc-mw731.us-east-2.aws.confluent.cloud --schema-registry-api-key xxx --schema-registry-api-secret xxx
```

References
----------

Test & Review
-------------

manual tests

Open Questions / Follow-ups
---------------------------

### DIFF
--- a/internal/kafka/metrics_utils.go
+++ b/internal/kafka/metrics_utils.go
@@ -81,7 +81,7 @@ func (c *clusterCommand) validateClusterLoad(clusterId string, isLatestMetric bo
 
 	maxClusterLoad := maxApiDataValue(clusterLoadResponse.FlatQueryResponse.GetData())
 	if maxClusterLoad.Value >= 0.7 {
-		return fmt.Errorf("Cluster Load was %f percent at %s.\nRecommended cluster load should be less than 70 percent", maxClusterLoad.Value*100, maxClusterLoad.Timestamp.In(time.Local))
+		return fmt.Errorf("cluster Load was %f percent at %s.\nRecommended cluster load should be less than 70 percent", maxClusterLoad.Value*100, maxClusterLoad.Timestamp.In(time.Local))
 	}
 
 	return nil
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -20,13 +20,15 @@ var serializationFormats = []string{"string", "avro", "integer", "jsonschema", "
 func AddApiKeyFlag(cmd *cobra.Command, c *AuthenticatedCLICommand) {
 	cmd.Flags().String("api-key", "", "API key.")
 
-	RegisterFlagCompletionFunc(cmd, "api-key", func(cmd *cobra.Command, args []string) []string {
-		if err := c.PersistentPreRunE(cmd, args); err != nil {
-			return nil
-		}
+	if c.V2Client != nil {
+		RegisterFlagCompletionFunc(cmd, "api-key", func(cmd *cobra.Command, args []string) []string {
+			if err := c.PersistentPreRunE(cmd, args); err != nil {
+				return nil
+			}
 
-		return AutocompleteApiKeys(c.V2Client)
-	})
+			return AutocompleteApiKeys(c.V2Client)
+		})
+	}
 }
 
 func AddApiSecretFlag(cmd *cobra.Command) {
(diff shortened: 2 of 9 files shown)


pkg/cmd/prerunner.go
```diff
@@ -164,6 +165,27 @@ func IsFlagRequired(flag *pflag.Flag) bool {
...
+// Authenticated provides PreRun operations for commands that require a logged-in Confluent Cloud user.
+func (r *PreRun) AuthenticatedWithAPIKey(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := r.Anonymous(command.CLICommand, true)(cmd, args); err != nil {
+			return err
+		}
+
+		ctx := command.Config.Context()
+		if ctx == nil {
+			return errors.NewErrorWithSuggestions("no context found", "Create a context with `confluent context create --bootstrap <bootstrap-server> --api-key <api-key> --api-secret <api-secret>`.")
```
sgagniere, line 177: I think this should be an example in the produce/consume commands.

The schema registry version in the PR description should probably be an example, too.

internal/kafka/metrics_utils.go
```diff
@@ -81,7 +81,7 @@ func (c *clusterCommand) validateClusterLoad(clusterId string, isLatestMetric bo
 
 	maxClusterLoad := maxApiDataValue(clusterLoadResponse.FlatQueryResponse.GetData())
 	if maxClusterLoad.Value >= 0.7 {
-		return fmt.Errorf("Cluster Load was %f percent at %s.\nRecommended cluster load should be less than 70 percent", maxClusterLoad.Value*100, maxClusterLoad.Timestamp.In(time.Local))
+		return fmt.Errorf("cluster Load was %f percent at %s.\nRecommended cluster load should be less than 70 percent", maxClusterLoad.Value*100, maxClusterLoad.Timestamp.In(time.Local))
```
brianstrauch, line 84:
```suggestion
		return fmt.Errorf("cluster load was %f percent at %s, but recommended load is less than 70 percent", maxClusterLoad.Value*100, maxClusterLoad.Timestamp.In(time.Local))
```

pkg/cmd/flags.go
```diff
@@ -96,17 +98,20 @@ func AddCloudFlag(cmd *cobra.Command) {
...
-			return nil
-		}
 
-		environmentId, err := c.Context.EnvironmentId()
-		if err != nil {
-			return nil
-		}
-		return AutocompleteClusters(environmentId, c.V2Client)
-	})
+	if c.V2Client != nil {
```
brianstrauch, line 102: Isn't it possible for this to be set in `PersistentPreRunE()`? If so, maybe we should wrap lines 108-112 with this `if`.

pkg/cmd/prerunner.go
```diff
@@ -39,6 +39,7 @@ var wrongLoginCommandsMap = map[string]string{
 type PreRunner interface {
 	Anonymous(command *CLICommand, willAuthenticate bool) func(*cobra.Command, []string) error
 	Authenticated(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error
+	AuthenticatedWithAPIKey(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error
```
brianstrauch, line 42: Can we combine this with `Authenticated()` so the logic to determine if the function should be authenticated with a token or API key lives here instead of duplicated across multiple command declarations?

brianstrauch (changes requested):

PR #2368, opened 2023-10-20
[FCP-1283] fix: bug where expired token was used on redirect

YannickPferr:
Checklist
---------
   - [x] yes: ok 

What
----

PR #2231 introduced client redirection, which also sets the Authorization header when redirects happen. The problem is that the token used for the header, is actually the local variable that's passed to the GatewayClient constructor. This will lead to the token never being refreshed and thus to 401s after the initial token has expired. 

Instead we need to either use the token inside the GW client struct `client.AuthToken`, which gets refreshed by the flink shell or copy over the authorization header from the previous request.

### DIFF
--- a/pkg/ccloudv2/flink_gateway.go
+++ b/pkg/ccloudv2/flink_gateway.go
@@ -28,10 +28,6 @@ type FlinkGatewayClient struct {
 }
 
 func NewFlinkGatewayClient(url, userAgent string, unsafeTrace bool, authToken string) *FlinkGatewayClient {
-	client := &FlinkGatewayClient{
-		AuthToken: authToken,
-	}
-
 	cfg := flinkgatewayv1beta1.NewConfiguration()
 	cfg.Debug = unsafeTrace
 	cfg.HTTPClient = NewRetryableHttpClientWithRedirect(unsafeTrace,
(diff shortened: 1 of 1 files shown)


pkg/ccloudv2/flink_gateway.go
```diff
@@ -43,17 +47,16 @@ func NewFlinkGatewayClient(url, userAgent string, unsafeTrace bool, authToken st
 			// header is copied only when the hostname is exactly the same or an exact subdomain of the hostname
 			// Ideally, we should check the status code returned in via[0].Response. However, via[0].Response is nil in underlying
 			// retryable http implementation
-			req.Header.Add("Authorization", "Bearer "+authToken)
+			// To ensure we have a fresh token, we need to use client.AuthToken and NOT the local authToken
+			req.Header.Add("Authorization", "Bearer "+client.AuthToken)
```
luca-filipponi, line 51: maybe you should just use what you have in the request: https://github.com/confluentinc/cc-system-tests-flink/blob/22abfe822f45ad1a4b29aa720cc57e10e8f74191/internal/utils/auth.go#L89 so you don't care really of what it's in the context

YannickPferr: Yeah I think it will work too, let me try

pkg/ccloudv2/flink_gateway.go
```diff
@@ -37,13 +37,15 @@ func NewFlinkGatewayClient(url, userAgent string, unsafeTrace bool, authToken st
...
 				return fmt.Errorf("stopped after 5 redirects")
 			}
-			log.CliLogger.Debugf("Following redirect with authorization to %s", req.URL)
-			// Customize the redirect to add authorization header on 307 Redirect.
-			// This is required as the Location header returned by the gateway may not be an exact subdomain and Authorization
-			// header is copied only when the hostname is exactly the same or an exact subdomain of the hostname
-			// Ideally, we should check the status code returned in via[0].Response. However, via[0].Response is nil in underlying
-			// retryable http implementation
-			req.Header.Add("Authorization", "Bearer "+authToken)
+			if len(via) > 0 {
```
bideyio, line 40: `len(via)` will at least be 1 since the redirect happens after at least 1 request was made. The `if` check maybe redundant. lmk if my understanding is incorrect.

pkg/ccloudv2/flink_gateway.go
bideyio: IMO, using client.AuthToken is better since the bearer token may once again be expired by the time the request reaches the gateway authentication module. By always using the refreshed token, we never have to worry about a redirect request with a stale token.

YannickPferr: That's what I thought at first as well, but I'm not sure that's true anymore. We refresh the token right before we call any of the functions inside the Gateway client, but in case of a redirect I don't think we get another chance at refreshing. It basically goes like this:
```
call GetStatement() with a fresh token -> call to GW -> redirect back to CLI -> call to GW again
```
Since everything is handled by the HttpClient after submitting the initial request, I don't think we have control over refreshing again.

Anyways, I'm indifferent which approach we use, both should work.

pkg/ccloudv2/flink_gateway.go
YannickPferr: Yeah I think you're right, I just left it there to be safe

pkg/ccloudv2/flink_gateway.go
YannickPferr: Yes that makes sense, and no we don't use a goroutine to refresh the token, we just synchronously check expiry and refresh if needed before every request.

So I'll stick with extracting the token from the previous request! 👍

luca-filipponi (approved):

bideyio (approved):

pkg/ccloudv2/flink_gateway.go
luca-filipponi: Yea, tbh this is bit tricky.
In my head, I see this request as an atomic request to gateway, even if you have a redirect in the middle, this is why even if expires in the middle is not a problem.

This is the same for a request without redirect, by the time you reach gateway the token could expire (the network trip will always take a few milliseconds, much more if people use cli from another far region).
I don't think we worry at all about this scenario (this could happen for every request from CLI or cloud client), only solution is to have a retry (not sure if there is already, yannick would know better)

YannickPferr: For now considering the token as invalid 1min before it expires, seems to work fine. We might have to tweak this in the future, or add a retry for 401s if all else fails.

PR #2375, opened 2023-10-27
Abstract away data field from SDK list functions

brianstrauch:
Checklist
---------
   - [x] yes: ok
   - [ ] no: DO NOT MERGE until the required features are live in prod  

What
----
Code is less readable if we need to access the data field when listing / updating data from API calls. Preferring `[]` data structures over `*List` data structures

Test & Review
-------------
Tests still pass

### DIFF
--- a/internal/kafka/command_partition_reassignment_list.go
+++ b/internal/kafka/command_partition_reassignment_list.go
@@ -62,7 +62,7 @@ func (c *partitionCommand) reassignmentList(cmd *cobra.Command, args []string) e
 	if err != nil {
 		return err
 	}
-	var reassignmentListResp kafkarestv3.ReassignmentDataList
+	var reassignments kafkarestv3.ReassignmentDataList
 	var resp *http.Response
 	if len(args) > 0 {
 		partitionId, err := partitionIdFromArg(args)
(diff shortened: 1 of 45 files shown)


internal/kafka/command_partition_reassignment_list.go
```diff
@@ -78,19 +78,19 @@ func (c *partitionCommand) reassignmentList(cmd *cobra.Command, args []string) e
...
-		reassignmentListResp, resp, err = restClient.PartitionApi.ClustersClusterIdTopicsPartitionsReassignmentGet(restContext, clusterId)
+		reassignments, resp, err = restClient.PartitionApi.ClustersClusterIdTopicsPartitionsReassignmentGet(restContext, clusterId)
 	}
 	if err != nil {
 		return kafkarest.NewError(restClient.GetConfig().BasePath, err, resp)
 	}
 
 	list := output.NewList(cmd)
-	for _, reassignment := range reassignmentListResp.Data {
+	for _, reassignment := range reassignments.Data {
```
sgagniere, line 93: Looks like we'll need to write wrapper functions for this sdk if we want to remove these later, right?

sgagniere (approved):

internal/kafka/command_partition_reassignment_list.go
brianstrauch: Yep!

PR #2389, opened 2023-11-01
[CLI-2831] Authenticate cloud produce and consume with kafka api key

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Users can run Cloud `confluent kafka topic produce/consume <topic>` when not logged in, by passing in `--kafka-bootstrap`, `--api-key` and `--api-secret`

Bug Fixes
- PLACEHOLDER

Checklist
---------
   - [x] yes: ok

What
----

```
confluent kafka topic produce <topic> --api-key xx --api-secret xx--bootstrap SASL_SSL://pkc-pgq85.us-west-2.aws.confluent.cloud:9092 /
--value-format avro --schema ~/test.json --schema-registry-endpoint https://psrc-mw731.us-east-2.aws.confluent.cloud/
--schema-registry-api-key xx --schema-registry-api-secret xxx
Successfully registered schema with ID "100001".
Starting Kafka Producer. Use Ctrl-C or Ctrl-D to exit.
{"id":"one","amount":1.2}


confluent kafka topic consume -b <topic> --api-key xx --api-secret xx--bootstrap SASL_SSL://pkc-pgq85.us-west-2.aws.confluent.cloud:9092 /
--value-format avro --schema-registry-endpoint https://psrc-mw731.us-east-2.aws.confluent.cloud/
--schema-registry-api-key xx --schema-registry-api-secret xxx
{"id":"one","amount":1.2}
```

References
----------

Test & Review
-------------

Open Questions / Follow-ups
---------------------------


internal/kafka/command_topic.go
```diff
@@ -38,29 +40,30 @@ func newTopicCommand(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Comman
 
 	c := &command{clientID: cfg.Version.ClientID}
 
-	if cfg.IsCloudLogin() {
+	if cfg.CheckIsCloudLoginOrOnPremLogin() != nil && !cfg.IsTest {
```
brianstrauch, line 43: Why is `cfg.IsTest` here? This seems like a hacky way to make integration tests pass.

internal/kafka/command_topic_consume.go
```diff
@@ -212,12 +234,135 @@ func (c *command) consume(cmd *cobra.Command, args []string) error {
...
+
+	var srClient *schemaregistry.Client
+	if slices.Contains(serdes.SchemaBasedFormats, valueFormat) {
+		srClient, err = c.GetSchemaRegistryClient(cmd)
+		if err != nil {
+			return err
+		}
+	}
+
+	dir, err := sr.CreateTempDir()
```
brianstrauch, line 346: We should not be importing other `internal` packages. That's a surefire way to get import cycles

internal/kafka/command_topic_produce.go
```diff
@@ -33,36 +34,40 @@ func (c *command) newProduceCommand() *cobra.Command {
...
 	pcmd.AddValueFormatFlag(cmd)
-	cmd.Flags().String("key-references", "", "The path to the message key schema references file.")
 	cmd.Flags().String("references", "", "The path to the message value schema references file.")
 	cmd.Flags().Bool("parse-key", false, "Parse key from the message.")
 	cmd.Flags().String("delimiter", ":", "The delimiter separating each key and value.")
 	cmd.Flags().StringSlice("config", nil, `A comma-separated list of configuration overrides ("key=value") for the producer client.`)
 	pcmd.AddProducerConfigFileFlag(cmd)
 	cmd.Flags().String("schema-registry-endpoint", "", "Endpoint for Schema Registry cluster.")
+	pcmd.AddOutputFlag(cmd) // Deprecated
+	cobra.CheckErr(cmd.Flags().MarkHidden("output"))
```
brianstrauch, line 50: Can we leave these at the bottom? They're going to be removed soon

internal/kafka/command_topic_produce.go
```diff
@@ -33,36 +34,40 @@ func (c *command) newProduceCommand() *cobra.Command {
...
+
+	// cloud-only flag
```
brianstrauch, line 52:
```suggestion
	// cloud-only flags
```

internal/kafka/command_topic_produce.go
```diff
@@ -33,36 +34,40 @@ func (c *command) newProduceCommand() *cobra.Command {
...
+	cmd.Flags().String("kafka-bootstrap", "", "Bootstrap URL for Kafka cluster.")
```
brianstrauch, line 53: Should the description mention that this is only for Cloud Kafka clusters?

brianstrauch: Also, it looks like we're going to have both `--bootstrap` and `--kafka-bootstrap`. Can we mark `--kafka-bootstrap` as hidden and deprecate it? (... and open a ticket to remove it in V4)

internal/kafka/command_topic_produce.go
```diff
@@ -74,6 +79,22 @@ func (c *command) newProduceCommand() *cobra.Command {
 }
 
 func (c *command) produce(cmd *cobra.Command, args []string) error {
+	if c.Context == nil || c.Context.State == nil {
+		cobra.CheckErr(cmd.MarkFlagRequired("kafka-bootstrap"))
```
brianstrauch, line 83: This function can only be called when the command is initialized, right?

internal/kafka/command_topic.go
```diff
@@ -142,6 +145,40 @@ func (c *command) provisioningClusterCheck(lkc string) error {
...
+	}
+
+	platform := &config.Platform{
+		Server: kafkaBootstrap,
+		Name:   strings.TrimPrefix(kafkaBootstrap, "https://"),
+	}
+
+	kafkaClusterCfg := &config.KafkaClusterConfig{
+		ID:        "anonymous-id",
+		Name:      "anonymous-cluster",
```
brianstrauch, line 161: Why do these need to be set? Is there any way we can avoid that?

internal/kafka/command_topic.go
```diff
@@ -142,6 +145,40 @@ func (c *command) provisioningClusterCheck(lkc string) error {
...
+		Bootstrap: kafkaBootstrap,
+		APIKeys:   map[string]*config.APIKeyPair{},
```
brianstrauch, line 163: Our code should be able to safely handle the case where this is `nil`, IMO.

pkg/cmd/authenticated_cli_command.go
```diff
@@ -175,14 +175,50 @@ func (c *AuthenticatedCLICommand) GetMetricsClient() (*ccloudv2.MetricsClient, e
 	return c.metricsClient, nil
 }
 
+func (c *AuthenticatedCLICommand) GetSchemaRegistryClientByFlags(cmd *cobra.Command, unsafeTrace bool) error {
```
brianstrauch, line 178: The regular `GetSchemaRegistryClient()` function should be able to read flags, right?

internal/kafka/command_topic.go
MuweiHe: no it gave me a npe later

internal/kafka/command_topic.go
MuweiHe: i followed the standard of creating a context for anonymous cluster. May wanna refactor them together when we talk about removing `context create`

internal/kafka/command_topic.go
MuweiHe: many of our integration tests are not actually logged in, then they're routed into this branch. This `isTest` doesn't affect non-test use. But i'll look if i can fix the tests login directly

internal/kafka/command_topic_produce.go
MuweiHe: agree with updating the description. those 2 bootstraps are totally different, why are we removing `kafka-bootstrap`?

internal/kafka/command_topic_produce.go
MuweiHe: If it's not logged in, even though the cmd is initialized, the context is still nil and empty...

internal/kafka/command_topic.go
brianstrauch: Can we prevent it from panicking later?

pkg/cmd/authenticated_cli_command.go
MuweiHe: Yes and no, i need to add a different branch where it's neither cloud or on-prem login. And this function is just i refactored some reusable code.

internal/kafka/command_topic.go
brianstrauch: My worry is that when we remove `context create`, this code will still exist and will prevent us from being able to easily refactor it

internal/kafka/command_topic_produce.go
brianstrauch: Even though they're totally different, a user might be confused about what the difference is. Instead we should just use `--bootstrap` for everything so the user doesn't have to think about which one to use

MuweiHe: hmm `bootstrap` is part of standard on-prem authentication flags. i hesitate on mix-using them

internal/kafka/command_topic.go
MuweiHe: it still needs to be initialized. the other option is to add a `cluster.APIKeys = map[]{}` later

internal/kafka/command_topic_produce.go
brianstrauch: Right, but this function is called during runtime, at which point the command has already been parsed to determine if required flags are present or missing.

internal/kafka/command_topic.go
MuweiHe: we can remove `name`, but `id` is still required. i can write a comment here.

internal/kafka/command_topic_produce.go
brianstrauch: What's wrong with using the same flag for both cloud and on-prem? There's already a lot of flags that are used for both.

internal/kafka/command_topic.go
brianstrauch: Any way we can get around `id` and other fields being required?

pkg/cmd/authenticated_cli_command.go
brianstrauch: > i need to add a different branch where it's neither cloud or on-prem login

This sounds good, we should add it to `GetSchemaRegistryClient()` so there's only one function!

internal/kafka/command_topic_produce.go
MuweiHe: They have totally different meanings and uses and that's ok?

internal/kafka/command_topic.go
MuweiHe: I don't think so. It'll being referenced as the key in the map

pkg/cmd/authenticated_cli_command.go
MuweiHe: But that's good chunk of reusable code so i took it out

internal/kafka/command_topic_produce.go
MuweiHe: Was thinking about "url". Agree we can use bootstrap for both cloud and onprem

test/fixtures/output/kafka/topic/consume-help-onprem.golden
```diff
@@ -10,6 +10,10 @@ Consume items from topic "my-topic" and press "Ctrl-C" to exit.
 
   $ confluent kafka topic consume my-topic --from-beginning
 
+Consume from a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.
+
+  $ confluent kafka topic consume my_topic --api-key <api-key> --api-secret <api-secret> --bootstrap <Kafka-cluster-endpoint> --value-format avro --schema-registry-endpoint <schema-registry-endpoint> --schema-registry-api-key <schema-registry-api-key> --schema-registry-api-secret <schema-registry-api-secret>
```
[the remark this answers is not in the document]
MuweiHe: just updated!

internal/kafka/command_topic_consume.go
```diff
@@ -34,7 +34,7 @@ func (c *command) newConsumeCommand() *cobra.Command {
 			},
 			examples.Example{
 				Text: `Consume from a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.`,
-				Code: "confluent kafka topic consume my_topic --api-key <api-key> --api-secret <api-secret> --bootstrap <Kafka-cluster-endpoint> --value-format avro --schema-registry-endpoint <schema-registry-endpoint> --schema-registry-api-key <schema-registry-api-key> --schema-registry-api-secret <schema-registry-api-secret>",
+				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap SASL_SSL://pkc-123:9092 --value-format avro --schema-registry-endpoint https://psrc-456 --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
```
brianstrauch, line 37: Nit: In reality, I think PKCs have 6 characters. How about `pkc-123456` instead?

brianstrauch: Also, I believe cloud API keys are 16 characters long, and only contain uppercase letters and digits

brianstrauch (changes requested):

internal/kafka/command_topic_consume.go
MuweiHe: Secrets are even longer that will make this example message really really long.. 
i can update the pkc example

MuweiHe: updated!

internal/kafka/command_topic_consume.go
```diff
@@ -50,8 +53,10 @@ func (c *command) newConsumeCommand() *cobra.Command {
 	cmd.Flags().Bool("timestamp", false, "Print message timestamp in milliseconds.")
 	cmd.Flags().StringSlice("config", nil, `A comma-separated list of configuration overrides ("key=value") for the consumer client.`)
 	pcmd.AddConsumerConfigFileFlag(cmd)
-	cmd.Flags().String("schema-registry-context", "", "The Schema Registry context under which to look up schema ID.")
 	cmd.Flags().String("schema-registry-endpoint", "", "Endpoint for Schema Registry cluster.")
+
+	// cloud-only flags
```
sgagniere, line 58: In some other commands we use `cfg.IsCloudLogin()` to determine which flags to show, is that something we could do here as well?

brianstrauch: We want this command to be able to be run without requiring login, for some reason... it was a customer request.

internal/kafka/command_topic_produce.go
```diff
@@ -74,6 +87,31 @@ func (c *command) newProduceCommand() *cobra.Command {
...
+
+		if err := c.prepareAnonymousContext(cmd); err != nil {
+			return err
+		}
+		return c.produceCloud(cmd, args)
+	} else if c.Context.Config.IsCloudLogin() {
+		return c.produceCloud(cmd, args)
+	} else {
+		if !cmd.Flags().Changed("bootstrap") {
+			return fmt.Errorf(errors.RequiredFLagNotSetErrorMsg, "bootstrap")
```
brianstrauch, line 103:
```suggestion
			return fmt.Errorf(errors.RequiredFlagNotSetErrorMsg, "bootstrap")
```

internal/kafka/command_topic.go
```diff
@@ -41,26 +44,25 @@ func newTopicCommand(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Comman
...
 		cmd.AddCommand(c.newCreateCommandOnPrem())
 		cmd.AddCommand(c.newDeleteCommandOnPrem())
 		cmd.AddCommand(c.newDescribeCommandOnPrem())
 		cmd.AddCommand(c.newListCommandOnPrem())
-		cmd.AddCommand(c.newProduceCommandOnPrem())
 		cmd.AddCommand(c.newUpdateCommandOnPrem())
 	}
 
+	cmd.AddCommand(c.newConsumeCommand())
+	cmd.AddCommand(c.newProduceCommand())
```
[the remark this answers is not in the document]
MuweiHe: the command needs to be initialized first and then it can add commands

internal/kafka/command_topic_consume.go
sgagniere: Ah, right.

MuweiHe: not really. The only other command that shows both flags is `login`, but you haven't logged in yet so no

internal/kafka/command_topic.go
```diff
@@ -142,6 +144,39 @@ func (c *command) provisioningClusterCheck(lkc string) error {
...
+		return err
+	}
+
+	platform := &config.Platform{
+		Server: kafkaBootstrap,
+		Name:   strings.TrimPrefix(kafkaBootstrap, "https://"),
+	}
+
+	kafkaClusterCfg := &config.KafkaClusterConfig{
+		ID:        "anonymous-id", // TODO: remove in V4
```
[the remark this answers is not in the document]
MuweiHe: https://confluentinc.atlassian.net/browse/CLI-2887

internal/kafka/command_topic.go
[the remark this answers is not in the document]
brianstrauch: Hmm, I didn't think that was the case. We're just initializing fields of `command` in the branch.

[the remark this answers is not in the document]
MuweiHe: no i've tried it. It'll be npe

internal/kafka/command_topic.go
```diff
@@ -145,19 +145,19 @@ func (c *command) provisioningClusterCheck(lkc string) error {
...
-		Server: kafkaBootstrap,
-		Name:   strings.TrimPrefix(kafkaBootstrap, "https://"),
+		Server: bootstrap,
+		Name:   strings.TrimPrefix(bootstrap, "https://"),
 	}
 
 	kafkaClusterCfg := &config.KafkaClusterConfig{
-		ID:        "anonymous-id", // TODO: remove in V4
-		Bootstrap: kafkaBootstrap,
+		ID:        "anonymous-id", // TODO: CLI-2887
```
brianstrauch, line 159:
```suggestion
		ID:        "anonymous-id",
```

pkg/errors/error_message.go
```diff
@@ -23,7 +23,7 @@ const (
 	UnknownConnectorIdErrorMsg        = `unknown connector ID "%s"`
 	InvalidApiKeyErrorMsg             = `invalid API key "%s" for resource "%s"`
 	InvalidApiKeySuggestions          = "To list API keys that belong to resource \"%[1]s\", use `confluent api-key list --resource %[1]s`.\nTo create new API key for resource \"%[1]s\", use `confluent api-key create --resource %[1]s`."
-	RequiredFLagNotSetErrorMsg        = `required flag(s) "%s" not set`
+	RequiredFlagNotSetErrorMsg        = `required flag(s) "%s" not set`
```
brianstrauch, line 26: Looks like only one flag is ever passed, and flags should be formatted with backticks so they get colored properly
```suggestion
	RequiredFlagNotSetErrorMsg        = "required flag `--%s` not set"
```

internal/kafka/command_topic_consume.go
```diff
@@ -28,15 +27,19 @@ func (c *command) newConsumeCommand() *cobra.Command {
...
 				Code: "confluent kafka topic consume my-topic --from-beginning",
 			},
+			examples.Example{
+				Text: `Consume from a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.`,
+				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap SASL_SSL://pkc-abc12:9092 --value-format avro --schema-registry-endpoint https://psrc-ab123 --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
+			},
 		),
 	}
 
+	cmd.Flags().String("bootstrap", "", `Comma-separated list of broker hosts, each formatted as "host" or "host:port". For Confluent Cloud, it's Kafka cluster endpoint.`)
```
brianstrauch, line 42: I think you want a `StringSlice` here, so users can optionally pass multiple instances of `--bootstrap`

brianstrauch: > Comma-separated list of broker hosts

This is not the case for Confluent Cloud. Can we rewrite this to something like "Kafka cluster endpoint (Confluent Cloud) or list of broker hosts (Confluent Platform)"?

internal/kafka/command_topic_produce.go
```diff
@@ -33,36 +35,47 @@ func (c *command) newProduceCommand() *cobra.Command {
 		Args:              cobra.ExactArgs(1),
 		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validArgs),
 		RunE:              c.produce,
-		Annotations:       map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Produce to a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.`,
```
brianstrauch, line 40:
```suggestion
				Text: `Produce to topic "my_topic" in Confluent Cloud with a Confluent Cloud API key.`,
```

internal/kafka/command_topic_consume.go
MuweiHe: as it's not the case for cloud, i think it's better to leave it a string

MuweiHe: `Kafka cluster endpoint (Confluent Cloud); or comma-separated list of broker hosts (Confluent Platform), each formatted as "host" or "host:port".`

MuweiHe: how does it look

internal/kafka/command_topic_consume.go
```diff
@@ -28,15 +27,19 @@ func (c *command) newConsumeCommand() *cobra.Command {
...
 				Code: "confluent kafka topic consume my-topic --from-beginning",
 			},
+			examples.Example{
+				Text: `Consume from a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.`,
+				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap SASL_SSL://pkc-abc12:9092 --value-format avro --schema-registry-endpoint https://psrc-ab123 --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
+			},
 		),
 	}
 
+	cmd.Flags().String("bootstrap", "", `Kafka cluster endpoint (Confluent Cloud); or comma-separated list of broker hosts (Confluent Platform), each formatted as "host" or "host:port".`)
```
brianstrauch, line 42:
```suggestion
	cmd.Flags().String("bootstrap", "", `Kafka cluster endpoint (Confluent Cloud) or a comma-separated list of broker hosts, each formatted as "host" or "host:port" (Confluent Platform).`)
```

internal/kafka/command_topic_consume.go
```diff
@@ -69,6 +79,27 @@ func (c *command) newConsumeCommand() *cobra.Command {
 }
 
 func (c *command) consume(cmd *cobra.Command, args []string) error {
+	if c.Context == nil || c.Context.State == nil {
```
brianstrauch, line 82:
```suggestion
	if c.Context.GetState() == nil {
```

brianstrauch (commented):
Can you show some output for these commands in the PR description, since we still don't have tests?

MuweiHe:
> Can you show some output for these commands in the PR description, since we still don't have tests?

It's just like a normal produce and consume! i added a some example

internal/kafka/command_topic_consume.go
```diff
@@ -28,15 +27,19 @@ func (c *command) newConsumeCommand() *cobra.Command {
...
 		RunE:              c.consume,
-		Annotations:       map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
 		Example: examples.BuildExampleString(
 			examples.Example{
 				Text: `Consume items from topic "my-topic" and press "Ctrl-C" to exit.`,
 				Code: "confluent kafka topic consume my-topic --from-beginning",
 			},
+			examples.Example{
+				Text: `Consume from a cloud Kafka topic named "my_topic" without logging in to Confluent Cloud.`,
+				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap SASL_SSL://pkc-abc12:9092 --value-format avro --schema-registry-endpoint https://psrc-ab123 --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
```
brianstrauch, line 37:
```suggestion
				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap pkc-abc12:9092 --value-format avro --schema-registry-endpoint https://psrc-ab123 --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
```

brianstrauch, line 37:
```suggestion
				Code: "confluent kafka topic consume my_topic --api-key my-key --api-secret my-secret --bootstrap SASL_SSL://pkc-12345.us-west-2.aws.confluent.cloud:9092 --value-format avro --schema-registry-endpoint https://psrc-12345.us-west-2.aws.confluent.cloud --schema-registry-api-key my-sr-key --schema-registry-api-secret my-sr-secret",
```

pkg/cmd/authenticated_cli_command.go
```diff
@@ -175,14 +175,50 @@ func (c *AuthenticatedCLICommand) GetMetricsClient() (*ccloudv2.MetricsClient, e
...
+
 func (c *AuthenticatedCLICommand) GetSchemaRegistryClient(cmd *cobra.Command) (*schemaregistry.Client, error) {
 	if c.schemaRegistryClient == nil {
 		unsafeTrace, err := cmd.Flags().GetBool("unsafe-trace")
 		if err != nil {
 			return nil, err
 		}
 
-		if c.Config.IsCloudLogin() {
+		if c.Config.CheckIsCloudLoginOrOnPremLogin() != nil {
```
brianstrauch, line 217:
```suggestion
		if err := c.Config.CheckIsCloudLoginOrOnPremLogin(); err != nil {
```

pkg/cmd/authenticated_cli_command.go
brianstrauch, line 178:
```suggestion
func (c *AuthenticatedCLICommand) getSchemaRegistryClientByFlags(cmd *cobra.Command, unsafeTrace bool) error {
```

brianstrauch (approved):

brianstrauch:
Was auto-merge turned on? I had one comment left @MuweiHe

PR #1606, opened 2022-12-15
CIRE-STORAGE: Add BYOK API support to cli

thllwg:
Release Notes
-------------
New Features
- Add ``confluent byok`` commands for customer-managed keys (Bring Your Own Key) in Confluent Cloud

What
----
This PR adds support for the new Bring-Your-Own-Key (BYOK) API as defined in the [minispec](https://github.com/confluentinc/api/blob/master/byok/minispec.yaml) and implemented in [Cire-storage-bucket-service](https://github.com/confluentinc/CIRE-Storage-bucket-service).

The PR introduces commands to create, read, list and delete customer-managed keys, as well as accompanying tests. Additionally, the PR adds a `byok` flag to the `kafka cluster create` command. In contrast to the currently used `encryption-id` flag, users are required to `byok register <key>` their self-managed key with confluent cloud prior to encrypted cluster creation. 

The BYOK API will go GA for Azure and AWS now, with GCP support following later this year.

Test & Review
-------------
As of now, the byok api is imported from the [ccloud-sdk-go-internal](https://github.com/confluentinc/cli/blob/fa331345361d2617d1d9c82b416437c376da37ea/go.mod#L141). Once released, the replace instruction will be removed.
For similar reasons, the `kafka cluster create` already contains code to accept a `byok` flag, but does not process the flag yet. The `cmk` public API does not yet support byok, whereas the internal API introduces breaking changes. Please review the PR with that in mind.

PR #2404, opened 2023-11-09
[STORAGE-434] Add GCP support for BYOK clusters (DRAFT: API & SDK RELEASE PENDING)

thllwg:
Release Notes
-------------

Breaking Changes
- The flag `-encryption-key` of `confluent kafka cluster create` has been removed. Use `-cck-id` to pass the id of a customer-managed key to create BYOK clusters in Confluent Cloud instead.

New Features
- Adds GCP support for `confluent byok` (Bring Your Own Key) commands for customer-managed keys in Confluent Cloud

Checklist
---------

What
----
This PR 
- adds the capability to manage GCP BYOK objects via the CLI
- adds the capability to create BYOK clusters on GCP using the same provisioning flow as for AWS and Azure
- removes support for the legacy `encryption-key`.

Context: In [1606](https://github.com/confluentinc/cli/pull/1606), we added CLI support for the BYOK API for AWS and Azure. For GCP, the CLI remained on the legacy provisioning flow, using the `encryption-key` flag to create BYOK cluster for GCP. There is work being done to migrate the GCP bucket creation and BYOK flow from the old bucket service to the new bucket service API. Consequently, CLI support is required.

References
----------
[STORAGE-434](https://confluentinc.atlassian.net/browse/STORAGE-434)


brianstrauch:
> The -encryption-key flag in the confluent kafka cluster create command has been removed.

Breaking changes are only allowed in major releases. Our next major release (v4) will be in Q1. Are you ok with waiting that long?

internal/byok/command_create.go
```diff
@@ -43,15 +41,62 @@ var encryptionKeyPolicyAws = template.Must(template.New("encryptionKeyPolicyAws"
...
+	key      string
+	group    string
+}
+
+func (g gcpPolicyMetadata) renderPolicy() string {
+	return fmt.Sprintf(`
+gcloud iam roles create %[1]s \
+	--project=%[2]s \
+	--description="Grant necessary permissions for Confluent to access KMS key" \
+	--permissions=cloudkms.cryptoKeyVersions.useToDecrypt,cloudkms.cryptoKeyVersions.useToEncrypt,cloudkms.cryptoKeys.get  && \
```
brianstrauch, line 65:
```suggestion
	--permissions=cloudkms.cryptoKeyVersions.useToDecrypt,cloudkms.cryptoKeyVersions.useToEncrypt,cloudkms.cryptoKeys.get && \
```

internal/byok/command_create.go
```diff
@@ -43,15 +41,62 @@ var encryptionKeyPolicyAws = template.Must(template.New("encryptionKeyPolicyAws"
...
+gcloud kms keys add-iam-policy-binding %[3]s \
+	--project=%[2]s \
+	--keyring='%[4]s' \
```
brianstrauch, line 68: We normallyuse " instead of ' in suggested commands. Can we use double quotes here or could there be special characters in these values?

internal/byok/command_create.go
```diff
@@ -43,15 +41,62 @@ var encryptionKeyPolicyAws = template.Must(template.New("encryptionKeyPolicyAws"
...
+
+func newGcpPolicyMetadata(project, location, keyRing, key, group string) gcpPolicyMetadata {
+	return gcpPolicyMetadata{
+		project:  project,
+		location: location,
+		keyRing:  keyRing,
+		key:      key,
+		group:    group,
+	}
+}
```
brianstrauch, line 82: Turning this into a function just makes this less readable since you can't tell what all the field names are, IMO

internal/byok/command_create.go
```diff
@@ -138,12 +197,18 @@ func isAWSKey(key string) bool {
 	return keyArn.Service == "kms" && strings.HasPrefix(keyArn.Resource, "key/")
 }
 
+func isGCPKey(key string) bool {
```
brianstrauch, line 200:
```suggestion
func isGcpKey(key string) bool {
```

internal/byok/command_create.go
```diff
@@ -184,12 +249,24 @@ func renderAzureEncryptionPolicy(key byokv1.ByokV1Key) (string, error) {
 	return strings.Join(az, "\n"), nil
 }
 
+func renderGCPEncryptionPolicy(key byokv1.ByokV1Key) string {
+	// No need to do a sanity check for this key as it is handled by the BYOK API validation
+	// assumed to be a valid key ID
+	splitKeyId := strings.Split(key.Key.ByokV1GcpKey.KeyId, "/")
+	project, location, keyRing, keyName, group := splitKeyId[1], splitKeyId[3], splitKeyId[5], splitKeyId[7], key.Key.ByokV1GcpKey.GetSecurityGroup()
```
brianstrauch, line 256:
```suggestion
	project, location, keyRing, keyName := splitKeyId[1], splitKeyId[3], splitKeyId[5], splitKeyId[7]
	group := key.Key.ByokV1GcpKey.GetSecurityGroup()
```

internal/byok/command_create.go
```diff
@@ -184,12 +249,24 @@ func renderAzureEncryptionPolicy(key byokv1.ByokV1Key) (string, error) {
...
+}
+
 func getPostCreateStepInstruction(key byokv1.ByokV1Key) string {
 	switch {
 	case key.Key.ByokV1AwsKey != nil:
 		return `Copy and append these permissions into the key policy "Statements" field of the ARN in your AWS key management system to authorize access for your Confluent Cloud cluster.`
 	case key.Key.ByokV1AzureKey != nil:
 		return "To ensure the key vault has the correct role assignments, please run the following azure-cli command (certified for azure-cli v2.45):"
+	case key.Key.ByokV1GcpKey != nil:
+		return "To ensure the key has the correct role assignments, please run the following gcloud-cli command:"
```
brianstrauch, line 269:
```suggestion
		return "To ensure the key has the correct role assignments, please run the following `gcloud` CLI command:"
```

internal/byok/command_create.go
thllwg: Is there a particular reason we don't want to stick with golang naming [conventions](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) here?

internal/byok/command_create.go
thllwg: I'd like to keep it somewhat consistent with the above instruction for Azure.
What do you think about just using `Azure CLI` and `Google Cloud CLI`?

thllwg:
> > The -encryption-key flag in the confluent kafka cluster create command has been removed.

> Breaking changes are only allowed in major releases. Our next major release (v4) will be in Q1. Are you ok with waiting that long?

The `encryption-key` flag cannot be used to provision BYOK clusters on AWS and Azure since v3.3.0, a minor release (https://docs.confluent.io/confluent-cli/current/release-notes.html#confluent-cli-v3-3-0-release-notes). Why is it different with GCP? Or are you suggesting we just need to keep the flag around and to return a message pointing towards the `cck-id` flag til Q1?

sarguru:
> > > The -encryption-key flag in the confluent kafka cluster create command has been removed.
> 
> > Breaking changes are only allowed in major releases. Our next major release (v4) will be in Q1. Are you ok with waiting that long?
> 
> The `encryption-key` flag cannot be used to provision BYOK clusters on AWS and Azure since v3.3.0, a minor release (https://docs.confluent.io/confluent-cli/current/release-notes.html#confluent-cli-v3-3-0-release-notes). Why is it different with GCP? Or are you suggesting we just need to keep the flag around and to return a message pointing towards the `cck-id` flag til Q1?

+1 This flag wouldn't work for clusters provisioned with the new flow too. The actual usage of this endpoint is fairly low and the tech debt required to support this flow is significantly high so we will really like to drop this the way we did in v-3-3.

internal/byok/command_create.go
brianstrauch: Cool, didn't know this was a thing! Carry on...

internal/byok/command_create.go
brianstrauch: > What do you think about just using Azure CLI and Google Cloud CLI?

Sure, that works for me. I just want to avoid `gcloud-cli` since that's not a real command.

thllwg: Regarding consistency between the Azure and the GCP case: Currently, we advise that the Azure command is 
> (certified for azure-cli v2.45)
I'm in favor of getting rid of this statement. If we want to keep some information on what version we ran this statement against, could we perhaps move the information to the BYOK documentation? What do you think?

brianstrauch: In retrospect, we could probably change this to say "latest version of v2", but I'm hesitant to remove this entirely, since Microsoft could choose to remove or rename this command in v3, and I don't want to have to wait for a customer to tell us that it's broken to fix it.

internal/byok/command_create.go
brianstrauch: Looks like we agreed to lowercase this after all in a later comment
```suggestion
func isGcpKey(key string) bool {
```

internal/kafka/command_cluster_create.go
```diff
@@ -257,13 +187,8 @@ func setCmkClusterConfig(typeString string, cku int32, encryptionKeyID string) *
...
 		return &cmkv2.CmkV2ClusterSpecConfigOneOf{CmkV2Enterprise: &cmkv2.CmkV2Enterprise{Kind: "Enterprise"}}
 	case skuDedicated:
-		var encryptionPtr *string
-		if encryptionKeyID != "" {
-			encryptionPtr = &encryptionKeyID
-		}
-
 		return &cmkv2.CmkV2ClusterSpecConfigOneOf{
-			CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: cku, EncryptionKey: encryptionPtr},
+			CmkV2Dedicated: &cmkv2.CmkV2Dedicated{Kind: "Dedicated", Cku: cku},
```
brianstrauch, line 191:
```suggestion
			CmkV2Dedicated: &cmkv2.CmkV2Dedicated{
				Kind: "Dedicated",
				Cku: cku,
			},
```

internal/byok/command_create.go
```diff
@@ -43,15 +41,52 @@ var encryptionKeyPolicyAws = template.Must(template.New("encryptionKeyPolicyAws"
...
 const (
+	failedToRenderKeyPolicyErrorMsg = "BYOK error: failed to render key policy"
+
 	keyVaultCryptoServiceEncryptionUser = "e147488a-f6f5-4113-8e2d-b22465e65bf6"
 	keyVaultReader                      = "21090545-7ca7-4776-b22c-e363652d74d2"
+
+	defaultGcpRoleName = "custom_kms_role"
 )
 
+type gcpPolicyMetadata struct {
```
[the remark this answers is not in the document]
sarguru: Fixed in https://github.com/confluentinc/cli/pull/2404/commits/707e31dd6e8265b4d7fad3bdbf7cca74485024d8

PR #2405, opened 2023-11-09
Fix description of `--deleted` in SR commands

brianstrauch:
Checklist
---------

What
----
The `--deleted` flag does not filter deleted resources, it includes them in the existing list.

Test & Review
-------------
Updated integration tests


sgagniere (approved):
lgtm

PR #2406, opened 2023-11-09
Fix `kafka topic produce` panic

brianstrauch:
Release Notes
-------------
Bug Fixes
- Fix panic in `confluent kafka topic produce`

Checklist
---------

What
----
If context is `nil`, accessing context state will panic.

Test & Review
-------------
Manual verification


sgagniere (approved):

pkg/cmd/authenticated_cli_command.go
```diff
@@ -227,7 +227,7 @@ func (c *AuthenticatedCLICommand) GetSchemaRegistryClient(cmd *cobra.Command) (*
 			configuration.Debug = unsafeTrace
 			configuration.HTTPClient = ccloudv2.NewRetryableHttpClient(nil, unsafeTrace)
 
-			if c.Context.GetState() != nil {
+			if c.Context != nil && c.Context.GetState() != nil {
```
sgagniere, line 230: I don't think we need this change. `c.Context.GetState()` only returns `c.Context.State` if `c.Context != nil`, which seems to make checking `c.Context` here redundant. Same w/ line 287

brianstrauch: Actually we do, this is a sneaky bug... `c.Context` is a dynamic context, which contains a regular context. So it's possible that `c.Context` is not `nil`, but the regular context inside of `c.Context` is. Then `c.Context.GetState()` panics.

brianstrauch: The long term fix for this will be to use a single struct for the context, instead of this nested context approach. I removed the API client from the dynamic context struct recently, so we're one step closer to making that change!

sgagniere: Oh wow.

Can you add a comment on one or both of these lines about that?

sgagniere (dismissed):

sgagniere (approved):

PR #2408, opened 2023-11-09
fix error where confluent local couldn't work with newer docker desktop versions

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Resolved the `no such host` error when running `confluent local kafka start`. Confluent Local is now working with the latest Docker Desktop version 4.25.0

Checklist
---------

What
----

Something changed in the way the docker client looks up host. `localhost` the string doesn't work as it failed to map it to `0.0.0.0`. instead, explicitly declare it as `0.0.0.0` will resolve the issue.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/local/command_kafka_start.go
+++ b/internal/local/command_kafka_start.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker (versions greater than v4.25.0)."
+	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker."
 	confluentBrokerPrefix     = "confluent-local-broker-%d"
 	controllerVoterPrefix     = "%d@confluent-local-broker-%d:%s"
 	bootstrapServerPrefix     = "confluent-local-broker-%d:%s"
--- a/internal/local/command_kafka_start.go
+++ b/internal/local/command_kafka_start.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker."
+	dockerWorkingVersionMsg   = "This command has been verified to work with Docker v4.25.0. Verify that you are not using a newer, unverified version of Docker."
 	confluentBrokerPrefix     = "confluent-local-broker-%d"
 	controllerVoterPrefix     = "%d@confluent-local-broker-%d:%s"
 	bootstrapServerPrefix     = "confluent-local-broker-%d:%s"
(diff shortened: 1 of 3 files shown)


pkg/cmd/prerunner.go
```diff
@@ -793,12 +793,14 @@ func (r *PreRun) shouldCheckForUpdates(cmd *cobra.Command) bool {
 func warnIfConfluentLocal(cmd *cobra.Command) {
 	if strings.HasPrefix(cmd.CommandPath(), "confluent local kafka start") {
 		output.ErrPrintln(false, "The local commands are intended for a single-node development environment only, NOT for production usage. See more: https://docs.confluent.io/current/cli/index.html")
+		output.ErrPrintln(false, "The local commands have been verified to work with Docker Desktop version as late as 4.25.0.")
```
brianstrauch, line 796: I think we should only show this if there's a Docker-related error.

MuweiHe: updated!

internal/local/command_kafka_start.go
```diff
@@ -78,6 +78,7 @@ func (c *command) kafkaStart(cmd *cobra.Command, _ []string) error {
 
 	containers, err := dockerClient.ContainerList(context.Background(), types.ContainerListOptions{All: true})
 	if err != nil {
+		output.ErrPrintln(false, "The local commands have been verified to work with Docker Desktop version as late as 4.25.0.")
```
sgagniere: We should probably make this a string so that we don't have to modify all instances of it whenever the newest verified version needs changing.

brianstrauch: How about including this as a suggestion? "Verify that you are not using an unverified version of Docker (versions greater than v4.25.0)."

brianstrauch: Also, please make a constant instead of copying and pasting code!

brianstrauch (changes requested):

internal/local/command_kafka_start.go
MuweiHe: yes i added a commit to constant it!
will update the string now

internal/local/command_kafka_start.go
```diff
@@ -29,6 +29,7 @@ import (
 )
 
 const (
+	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker (versions greater than v4.25.0)."
```
sgagniere, line 32:
```suggestion
	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker."
```

The first part of the string already says which versions are verified, so I think we can drop the last bit.

sgagniere (dismissed):

internal/local/command_kafka_start.go
```diff
@@ -29,6 +29,7 @@ import (
 )
 
 const (
+	dockerWorkingVersionMsg   = "The local commands have been verified to work with Docker Desktop version as late as 4.25.0. Verify that you are not using an unverified version of Docker."
```
brianstrauch, line 32:
```suggestion
	dockerWorkingVersionMsg   = "This command has been verified to work with Docker v4.25.0. Verify that you are not using a newer, unverified version of Docker."
```

internal/local/command_kafka_start.go
```diff
@@ -78,6 +79,7 @@ func (c *command) kafkaStart(cmd *cobra.Command, _ []string) error {
 
 	containers, err := dockerClient.ContainerList(context.Background(), types.ContainerListOptions{All: true})
 	if err != nil {
+		output.ErrPrintln(false, dockerWorkingVersionMsg)
```
brianstrauch, line 82: Let's use a suggestion here instead of printing to stderr.

sgagniere (approved):

brianstrauch (approved):

PR #2424, opened 2023-11-16
Bump version of `github.com/ory/dockertest` to `3.10.0`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Running `trivy filesystem --scanners vuln .` shows three CVEs related to `github.com/opencontainers/runc`.

This is an indirect dependency of `ory/dockertest`, so this is resolved by bumping the version from `v3.3.5+incompatible` to `v3.10.0`.

There is still one CVE remaining for `github.com/docker/distribution`. This is also an indirect dependency, for `docker/docker` (which is now `moby/moby`). The `docker/distribution` version should be updated to the required version in the next release of `docker/docker`, so we're waiting on that to fix the remaining CVE.

Also bumped the version of `bradleyjkemp/cupaloy` to `v2.8.0` and `stripe/stripe-go` to `v76.4.0` to remove the `+incompatible` in `go.mod`.

References
----------

Test & Review
-------------
The three CVEs are gone when running `trivy filesystem --scanners vuln .` again
Tests pass (but I'm on Darwin, which skips the relevant test... the pipeline will run this one on Linux)


go.mod
```diff
@@ -99,6 +99,12 @@ require (
 	pgregory.net/rapid v1.1.0
 )
 
+require (
+	github.com/cenkalti/backoff/v4 v4.1.3 // indirect
+	github.com/docker/cli v20.10.17+incompatible // indirect
+	github.com/mitchellh/mapstructure v1.4.1 // indirect
+)
+
 require (
```
brianstrauch, line 108: Apply the following, then run `go mod tidy` to sort
```suggestion
```

go.mod
```diff
@@ -71,7 +71,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/opencontainers/image-spec v1.0.2
-	github.com/ory/dockertest v3.3.5+incompatible
+	github.com/ory/dockertest/v3 v3.10.0
```
brianstrauch, line 74: Nice! Any other dependencies with version v2 or above that we can be switched to this format?

sgagniere: `cupaloy` (for Flink) and `stripe-go` (for payment).

I'm inclined to leave `stripe-go` alone for now because I don't want to make any changes to the payment code unless we need to.

I'll go ahead and update `cupaloy` and run the vulnerability scan again though.

brianstrauch: > I'm inclined to leave stripe-go alone for now because I don't want to make any changes to the payment code unless we need to.

I say do it... the code itself shouldn't be any different, just the format of the dependency

brianstrauch (approved):

brianstrauch (approved):

PR #2432, opened 2023-11-20
Remove `AuthenticatedState()`

brianstrauch:
Checklist
---------

What
----
This function is no longer needed. If the user is not logged in, the prerunner will catch it and return an error, so a separate error-returning function is not needed. We can also simplify the function signature of `GetDataplaneToken()` and prepare to move away from the dynamic context struct. This helps us switch to standardized API clients more easily.

Test & Review
-------------
All tests still pass


sgagniere (approved):
lgtm

sgagniere (approved):
+1

PR #2439, opened 2023-11-27
fix: update sql keywords list to fix missing syntax highlights

YannickPferr:
Checklist
---------

What
----

Updated the Flink SQL keyword list to fix some syntax highlighting issues like this:
image
METADATA and VIRTUAL should be highlighted

References
----------

Test & Review
-------------


Guttz (approved):

Guttz:
We now have a more up-to-date list of keywords compatible with our sql version and that's why the update.

pkg/flink/config/sql_keywords.go
```diff
@@ -4,565 +4,508 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/types"
 )
 
-// These are all Flink SQL reserved keywords extracted from https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/overview/
-
 var SQLKeywords = types.NewSet(
```
brianstrauch, line 7: This will be replaced by the language server in the future, right?

brianstrauch (approved):

pkg/flink/config/sql_keywords.go
YannickPferr: Yes, we're already looking for ways to do syntax highlighting with the language service!

PR #2448, opened 2023-11-27
CIAM-3912: Increase CLI error clarity

AydenAdair:
Checklist
---------

What
----
When updating an IP filter or IP group in a way which would remove all associated IP groups or CIDR blocks, respectively, it gives a more clear error about what is going wrong.

Resource group flags were given a default value, but couldn't be utilized because it was listed as "required"

References
----------
[Bug bash](https://confluentinc.atlassian.net/wiki/spaces/SECENG/pages/3240723144/IP+Filtering+CLI+Bug+Bash)
[Ticket ](https://confluentinc.atlassian.net/browse/CIAM-3912)

Test & Review
-------------
`make lint`
`make test`
Local testing after `make build`

### DIFF
--- a/internal/iam/command_ipfilter_update.go
+++ b/internal/iam/command_ipfilter_update.go
@@ -140,6 +140,11 @@ func (c *ipFilterCommand) update(cmd *cobra.Command, args []string) error {
 		IpGroupIdObjects[i] = iamv2.GlobalObjectReference{Id: ipGroupId}
 	}
 
+	if len(IpGroupIdObjects) == 0 {
+		return errors.NewErrorWithSuggestions("Cannot remove all IP groups from IP filter",
+			"Please double check the IP filter you are updating. Use `confluent iam ip-filter describe <ip-filter>` to see the IP groups associated with an IP filter.")
+	}
+
 	updateIpFilter.IpGroups = &IpGroupIdObjects
 
 	filter, err := c.V2Client.UpdateIamIpFilter(updateIpFilter, currentIpFilterId)
--- a/internal/iam/command_ipfilter_update.go
+++ b/internal/iam/command_ipfilter_update.go
@@ -1,6 +1,7 @@
 package iam
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


internal/iam/command_ipfilter_update.go
```diff
@@ -140,6 +140,11 @@ func (c *ipFilterCommand) update(cmd *cobra.Command, args []string) error {
 		IpGroupIdObjects[i] = iamv2.GlobalObjectReference{Id: ipGroupId}
 	}
 
+	if len(IpGroupIdObjects) == 0 {
+		return errors.NewErrorWithSuggestions("Cannot remove all IP groups from IP filter",
+			"Please double check the IP filter you are updating. Use `confluent iam ip-filter describe <ip-filter>` to see the IP groups associated with an IP filter.")
```
brianstrauch, line 145: Use a format string here to print the IP filter ID instead of `<ip-filter>`?

test/iam_test.go
```diff
@@ -347,6 +347,7 @@ func (s *CLITestSuite) TestIamIpGroup() {
 func (s *CLITestSuite) TestIamIpFilter() {
 	tests := []CLITest{
 		{args: `iam ip-filter create "demo-ip-filter" --resource-group "management" --ip-groups "ipg-3g5jw,ipg-wjnde"`, fixture: "iam/ip-filter/create.golden"},
+		{args: `iam ip-filter create "demo-ip-filter" --ip-groups "ipg-3g5jw,ipg-wjnde"`, fixture: "iam/ip-filter/create.golden"},
```
brianstrauch, line 350:
```suggestion
		{args: "iam ip-filter create demo-ip-filter --ip-groups ipg-3g5jw,ipg-wjnde", fixture: "iam/ip-filter/create.golden"},
```

brianstrauch (approved):

brianstrauch:
Deleted release notes as this hasn't been released yet (these are not bug fixes since the "bugs" have not been released yet)
```
- More clear error messages when removing the last resource for an IP filter or IP group
- Remove `Resource Group` as a required flag, since it has a default value of `management`
```

PR #2450, opened 2023-11-29
Fix human list output

brianstrauch:
Checklist
---------

What
----
Various fixes for human list output.
1. Only numbers should be right-aligned
2. Tables should not contain []

Test & Review
-------------
Updated integration tests


internal/iam/command_user_invitation_list.go
```diff
@@ -34,11 +34,6 @@ func (c invitationCommand) listInvitations(cmd *cobra.Command, _ []string) error
 		return err
 	}
 
-	if len(invitations) == 0 {
```
MuweiHe, line 37: so it will print nothing if no invitations were found? Or it should be a `None found`

internal/kafka/command_partition_reassignment_list.go
```diff
@@ -14,12 +16,20 @@ import (
...
+	PartitionId      int32  `human:"Partition ID"`
+	AddingReplicas   string `human:"Adding Replicas"`
+	RemovingReplicas string `human:"Removing Replicas"`
+}
+
+type getReassignmentOutSerialized struct {
+	ClusterId        string  `serialized:"cluster_id"`
+	TopicName        string  `serialized:"topic_name"`
+	PartitionId      int32   `serialized:"partition_id"`
+	AddingReplicas   []int32 `serialized:"adding_replicas"`
```
MuweiHe, line 31: can we combine those 2 structs?

internal/kafka/command_partition_reassignment_list.go
```diff
@@ -91,13 +101,31 @@ func (c *partitionCommand) reassignmentList(cmd *cobra.Command, args []string) e
...
-			PartitionId:      reassignment.PartitionId,
-			AddingReplicas:   reassignment.AddingReplicas,
-			RemovingReplicas: reassignment.RemovingReplicas,
-		})
+		if output.GetFormat(cmd) == output.Human {
+			list.Add(&getReassignmentOutHuman{
+				ClusterId:        reassignment.ClusterId,
+				TopicName:        reassignment.TopicName,
+				PartitionId:      reassignment.PartitionId,
+				AddingReplicas:   join(reassignment.AddingReplicas),
```
MuweiHe, line 109: how does it print other []string? can we just do the join in the `output` package

test/fixtures/output/service-quota/1.golden
```diff
@@ -1,6 +1,6 @@
   Quota Code |  Name   |     Scope     | Applied Limit | Usage | Organization | Environment | Kafka Cluster | Network | User  
 -------------+---------+---------------+---------------+-------+--------------+-------------+---------------+---------+-------
-  quota_a    | Quota A | kafka_cluster |            15 |    10 | org-123      | env-1       | lkc-1         |         |
```
sgagniere, line 3: `Applied Limit` is `int32` in the `quotaValue` struct, so shouldn't it also be left-aligned here? Same w/ `Partition`, `Partition Mirror Lag`, etc. in the mirror commands.

test/fixtures/output/service-quota/1.golden
```diff
@@ -1,6 +1,6 @@
   Quota Code |  Name   |     Scope     | Applied Limit | Usage | Organization | Environment | Kafka Cluster | Network | User  
 -------------+---------+---------------+---------------+-------+--------------+-------------+---------------+---------+-------
-  quota_a    | Quota A | kafka_cluster |            15 |    10 | org-123      | env-1       | lkc-1         |         |       
-  quota_a    | Qutoa A | kafka_cluster |            16 |    11 | org-123      | env-2       | lkc-2         |         |       
+  quota_a    | Quota A | kafka_cluster |            15 | 10    | org-123      | env-1       | lkc-1         |         |
```
sgagniere, line 3: Reposting because I highlighted the wrong line last time.

`Applied Limit` is `int32` in the `quotaValue` struct, so shouldn't it also be left-aligned here? Same w/ `Partition`, `Partition Mirror Lag`, etc. in the mirror commands.

pkg/output/table.go
```diff
@@ -155,23 +155,37 @@ func (t *Table) printCore(writer io.Writer, auto bool) error {
...
+		w.SetBorder(false)
+
+		n := reflect.TypeOf(t.objects[0]).Elem().NumField()
+
 		var header []string
-		for i := 0; i < reflect.TypeOf(t.objects[0]).Elem().NumField(); i++ {
-			tag := strings.Split(reflect.TypeOf(t.objects[0]).Elem().Field(i).Tag.Get(t.format.String()), ",")
+		var alignment []int
+		for i := 0; i < n; i++ {
+			field := reflect.TypeOf(t.objects[0]).Elem().Field(i)
```
MuweiHe, line 169: will this cause a npe? `objects[0]`

internal/iam/command_user_invitation_list.go
brianstrauch: It'll print "None found." instead of "No invitations found." This is what all of our other commands say!

test/fixtures/output/service-quota/1.golden
brianstrauch: Integers are intentionally right-aligned. "Usage" is a string so it is left-aligned (they wanted to be able to set it to `""` when the value was 0 since 0 would convey a different meaning here)

pkg/output/table.go
brianstrauch: No we check the size of the slice above!

test/fixtures/output/service-quota/1.golden
sgagniere: Ohh, got it.

MuweiHe (approved):

PR #2458, opened 2023-11-29
Update goreleaser version and change build names in goreleaser files

sgagniere:
…for builds (but not archive)

Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Some preparation for adding rpm/deb support. Two things:
- Update the goreleaser version. The current version has an rpm naming bug that's fixed in the newer versions.
- Change the build id suffix from `-homebrew` to `-disableupdates` for the Darwin and Linux builds. The Linux binaries will be used to create the deb/rpm packages, so not just homebrew. Changing the suffix for the Darwin builds is simply for consistency.

I did not change the `archive-homebrew` archive id since that is still only used for Homebrew.

References
----------

Test & Review
-------------
Since this change also changes the prebuilt path for the linux binaries, I ran `scripts/build_linux.sh` and checked that the prebuilt path matches the updated pattern in this PR.

### DIFF
--- a/.goreleaser-linux-amd64.yml
+++ b/.goreleaser-linux-amd64.yml
@@ -18,7 +18,7 @@ builds:
       - linux
     goarch:
       - amd64
-  - id: confluent-linux-amd64-homebrew
+  - id: confluent-linux-amd64-disableupdates
     binary: confluent
     main: cmd/confluent/main.go
     flags:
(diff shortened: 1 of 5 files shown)


brianstrauch:
> I did not change the archive-homebrew archive id since that is still only used for Homebrew.

We're going to change this in the future though, right? Might as well do it now?

brianstrauch (approved):

PR #2481, opened 2023-12-11
feat: add environment id to output of compute pool commands

YannickPferr:
Release Notes
-------------

New Features
- [Early Access] Add "Environment" field to the output of the `confluent flink compute-pool` commands

Checklist
---------

What
----

Added the environment id to the output of compute pool commands

References
----------

Test & Review
-------------

### DIFF
--- a/internal/flink/command_compute_pool_create.go
+++ b/internal/flink/command_compute_pool_create.go
@@ -79,13 +79,12 @@ func (c *command) computePoolCreate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	computePoolEnv := computePool.Spec.GetEnvironment()
 	table := output.NewTable(cmd)
 	table.Add(&computePoolOut{
 		IsCurrent:   computePool.GetId() == c.Context.GetCurrentFlinkComputePool(),
 		Id:          computePool.GetId(),
 		Name:        computePool.Spec.GetDisplayName(),
-		Environment: computePoolEnv.GetId(),
+		Environment: computePool.Spec.Environment.GetId(),
 		CurrentCfu:  computePool.Status.GetCurrentCfu(),
 		MaxCfu:      computePool.Spec.GetMaxCfu(),
 		Cloud:       computePool.Spec.GetCloud(),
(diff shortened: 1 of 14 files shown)


internal/flink/command_compute_pool_create.go
```diff
@@ -79,16 +79,18 @@ func (c *command) computePoolCreate(cmd *cobra.Command, args []string) error {
...
-		Name:       computePool.Spec.GetDisplayName(),
-		CurrentCfu: computePool.Status.GetCurrentCfu(),
-		MaxCfu:     computePool.Spec.GetMaxCfu(),
-		Cloud:      computePool.Spec.GetCloud(),
-		Region:     computePool.Spec.GetRegion(),
-		Status:     computePool.Status.GetPhase(),
+		IsCurrent:   computePool.GetId() == c.Context.GetCurrentFlinkComputePool(),
+		Id:          computePool.GetId(),
+		Name:        computePool.Spec.GetDisplayName(),
+		Environment: computePoolEnv.GetId(),
```
brianstrauch, line 88:
```suggestion
		Environment: computePool.Spec.Environment.GetId(),
```

test/test-server/fcpm_handlers.go
```diff
@@ -25,6 +25,7 @@ func handleFcpmComputePools(t *testing.T) http.HandlerFunc {
 					MaxCfu:      flinkv2.PtrInt32(1),
 					Region:      flinkv2.PtrString("us-west-1"),
 					Cloud:       flinkv2.PtrString("AWS"),
+					Environment: flinkv2.NewGlobalObjectReference("env-123", "-", "-"),
```
brianstrauch, line 28: Let's construct this with a struct definition and not a function? I'm not sure what the last two parameters are for by looking at this (although I get that you're trying to leave them out)

internal/flink/command_compute_pool_create.go
YannickPferr: I was doing this to avoid nil pointers in case Environment was nil (even though that shouldn't actually happen)

test/test-server/fcpm_handlers.go
YannickPferr: Sure!

internal/flink/command_compute_pool_create.go
brianstrauch: Yeah, shouldn't happen. We're planning on updating our SDK soon so we can handle scenarios where a breaking change to the API doesn't break the CLI

brianstrauch (approved):

PR #2487, opened 2023-12-11
feat: add AltEnter shortcut as command termination trigger

YannickPferr:
Release Notes
-------------

New Features
- [Early Access] Alt + Enter can now be used to trigger statement submission, without having to add a semicolon at the end of the statement
- [Early Access] Alt + Backspace can now be used to delete the previous word

Checklist
---------

What
----

- Alt + Enter can now be used to trigger statement submission, without having to add a semicolon at the end of the statement 
- Alt + Backspace can now be used to delete the previous word

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/internal/controller/input_controller.go
+++ b/pkg/flink/internal/controller/input_controller.go
@@ -161,6 +161,10 @@ func (c *InputController) Prompt() prompt.IPrompt {
 			ASCIICode: []byte{0x1b, 0x66},
 			Fn:        prompt.GoRightWord,
 		}),
+		prompt.OptionAddASCIICodeBind(prompt.ASCIICodeBind{
+			ASCIICode: []byte{0x1b, 0x7F},
+			Fn:        prompt.DeleteWord,
+		}),
 		prompt.OptionPrefixTextColor(prompt.Yellow),
 		prompt.OptionPreviewSuggestionTextColor(prompt.Blue),
 		prompt.OptionSelectedSuggestionBGColor(prompt.LightGray),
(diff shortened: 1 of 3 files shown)


pkg/flink/internal/controller/input_controller.go
```diff
@@ -161,6 +161,10 @@ func (c *InputController) Prompt() prompt.IPrompt {
 			ASCIICode: []byte{0x1b, 0x66},
 			Fn:        prompt.GoRightWord,
 		}),
+		prompt.OptionAddASCIICodeBind(prompt.ASCIICodeBind{
+			ASCIICode: []byte{0x1b, 0x7F},
```
brianstrauch, line 165: Can you leave a comment explaining what 0x1b and 0x7F mean? It would be nice not to need to consult an ASCII chart in the future

pkg/flink/internal/controller/input_controller.go
```diff
@@ -169,14 +173,10 @@ func (c *InputController) Prompt() prompt.IPrompt {
...
-			// We add exit here because we also want to exit without the need of adding semicolon, which is the default flow for all statements
-			if text == "exit" {
-				return true
-			}
-			if text == "" || !strings.HasSuffix(text, ";") {
+			if text == "" {
 				return false
 			}
-			return true
+			return text == "exit" || strings.HasSuffix(text, ";") || lastKeyStroke == prompt.AltEnter
```
brianstrauch, line 179: This would let us delete the `if` above... just a suggestion!
```suggestion
			return text == "exit" || strings.HasSuffix(text, ";") || text != "" && lastKeyStroke == prompt.AltEnter
```

pkg/flink/internal/controller/input_controller.go
YannickPferr: Yep, will add some explanation to all those bindings!

pkg/flink/internal/controller/input_controller.go
brianstrauch: In boolean logic, and always comes before or regardless of language! Happy to leave things as is for extra clarity though

brianstrauch (approved):

pkg/flink/internal/controller/input_controller.go
YannickPferr: Sounds good, I think having the extra if makes it tiny bit more easier to understand

PR #2493, opened 2023-12-12
Remove caching for CI

brianstrauch:
What
----
After switching to on-prem Semaphore, adding caching actually slows down CI by 4 seconds (and 20 seconds for darwin). This is because it takes too long to upload/download the cache.


sgagniere (approved):
well that's unfortunate

PR #2520, opened 2024-01-04
feat: add statement name to interactive results view

YannickPferr:
Release Notes
-------------

New Features
- [Early Access] Add statement name to title of interactive results view in `confluent flink shell`

Checklist
---------

What
----

Adds the statement name to the title of the interactive results view:

table mode:
image

changelog mode:
image

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/internal/controller/interactive_output_controller.go
+++ b/pkg/flink/internal/controller/interactive_output_controller.go
@@ -193,16 +193,16 @@ func (t *InteractiveOutputController) handleKeyUpOrDownPress(event *tcell.EventK
 }
 
 func (t *InteractiveOutputController) getTableTitle() string {
-	mode := "changelog mode"
+	mode := "Changelog mode"
 	if t.resultFetcher.IsTableMode() {
-		mode = "table mode"
+		mode = "Table mode"
 	}
 
 	if t.debug {
 		return fmt.Sprintf(
-			" Statement: %s (%s) | last page size: %d | current cache size: %d/%d | table size: %d ",
-			t.resultFetcher.GetStatement().StatementName,
+			" %s (%s) | last page size: %d | current cache size: %d/%d | table size: %d ",
 			mode,
+			t.resultFetcher.GetStatement().StatementName,
 			t.resultFetcher.GetStatement().GetPageSize(),
 			t.resultFetcher.GetMaterializedStatementResults().GetChangelogSize(),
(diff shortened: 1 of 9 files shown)


Guttz (commented):
What do you think about this format: "Table Mode (test-statement)"

Also, do you have to delete the tests?

YannickPferr:
> What do you think about this format: "Table Mode (test-statement)"

Not really a strong opinion which one should come first 🤷 I can change it if you think that's better? 

> Also, do you have to delete the tests?

These tests were not useful anymore. I think we forgot to remove them when we moved the statement status (RUNNING, COMPLETED, FAILED) from the title to the footer row

Guttz (approved):

Guttz (approved):

brianstrauch:
Please update the release notes to refer to `confluent flink shell`. Otherwise customers will not know what command this change is for.

pkg/flink/internal/controller/interactive_output_controller.go
```diff
@@ -200,14 +200,15 @@ func (t *InteractiveOutputController) getTableTitle() string {
...
+			t.resultFetcher.GetStatement().StatementName,
 			t.resultFetcher.GetStatement().GetPageSize(),
 			t.resultFetcher.GetMaterializedStatementResults().GetChangelogSize(),
 			t.resultFetcher.GetMaterializedStatementResults().GetMaxResults(),
 			t.resultFetcher.GetMaterializedStatementResults().GetTableSize(),
 		)
 	}
 
-	return fmt.Sprintf(" %s ", mode)
+	return fmt.Sprintf(" %s (%s) ", mode, t.resultFetcher.GetStatement().StatementName)
```
brianstrauch, line 213:
```suggestion
	return fmt.Sprintf(" %s (%s) ", mode, t.resultFetcher.Statement.GetStatementName())
```

YannickPferr: I can't access it like that, because `statement` is private and a shared resource. The getter function is used to synchronize access to it to avoid races:
https://github.com/confluentinc/cli/blob/6a3583314693136dc7a7e1324e2a75b3412e6a79/pkg/flink/internal/results/result_fetcher.go#L144-L149

YannickPferr:
> Please update the release notes to refer to `confluent flink shell`. Otherwise customers will not know what command this change is for.

Oh yeah totally forgot that, changed it now!

brianstrauch (approved):

PR #2529, opened 2024-01-08
Revert "chore: update repo by service bot"

brianstrauch:
Reverts confluentinc/cli#2523


sgagniere (approved):
Should we just remove gitignore from our service yaml file?

brianstrauch:
> Should we just remove gitignore from our service yaml file?

It would be nice to have it enabled so we can ignore various IDE files and the like...

PR #2540, opened 2024-01-10
Use resource ID in `ksql cluster configure-acls`

brianstrauch:
Release Notes
-------------
New Features
- Use service account resource IDs instead of integer IDs in `confluent ksql cluster configure-acls`

Checklist
---------

What
----
Integer IDs have been deprecated for a while now across Confluent Cloud, so these should be transitioned too. Awaiting sign-off from someone on the KSQL team. Maybe @stevenpyzhang can +1?

Test & Review
-------------
Updated integration tests


stevenpyzhang (approved):
I think this should be ok since both int id and resource id are supported 
https://docs.confluent.io/cloud/current/access-management/access-control/acl.html#create-acl

PR #2542, opened 2024-01-11
Add actions workflow to close inactive issues.

MuweiHe:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----

Add actions workflow to close inactive issues. the workflow runs daily. We can manually label issues "stale" of no activity, and this job will close it after 14 days of being labeled.

References
----------

https://docs.github.com/en/actions/quickstart

Test & Review
-------------


brianstrauch:
I'm not sure I agree with this change. Every issue that's open right now should not be closed since it represents a concrete new feature we can add or bug we can fix. Maybe we can manually mark issues as "stale" if we don't get a reply and this job can close those issues after a month.

MuweiHe:
> I'm not sure I agree with this change. Every issue that's open right now should not be closed since it represents a concrete new feature we can add or bug we can fix. Maybe we can manually mark issues as "stale" if we don't get a reply after a while and this job can close those issues after a month.

@brianstrauch reasonable. Updated the pr!

brianstrauch:
Does the "stale" tag automatically get removed when someone comments?

MuweiHe:
> Does the "stale" tag automatically get removed when someone comments?

@brianstrauch yes, you can either comment or manually remove the label.

.github/workflows/inactive-issues.yaml
```diff
@@ -0,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
```
brianstrauch, line 4: Does this mean every day?

.github/workflows/inactive-issues.yaml
```diff
@@ -0,0 +1,22 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: -1
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
```
brianstrauch, line 18: Will this appear when I add the "stale" label manually?

.github/workflows/inactive-issues.yaml
```diff
@@ -0,0 +1,22 @@
...
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: -1
+          days-before-issue-close: 14
```
brianstrauch, line 16: How about a month? Seems a bit more generous than just 14 days.
```suggestion
          days-before-issue-close: 30
```

.github/workflows/inactive-issues.yaml
MuweiHe: yes

.github/workflows/inactive-issues.yaml
MuweiHe: Unsure 😕

MuweiHe: unsure..  leaning to the answer no

brianstrauch: I can see it being slightly annoying if we had to manually write this every time (it would be good to let the user know their issue has become stale / how much time they have to respond)

brianstrauch (approved):

PR #2544, opened 2024-01-12
Enable CGO for darwin/amd64

brianstrauch:
Checklist
---------

What
----
`CGO_ENABLED=1` is needed to build darwin/amd64 from darwin/arm64. This PR adds it to the global env list of all goreleaser files for simplicity now that all os/arch pairs use it.

Test & Review
-------------
Tested during this week's release and was able to build for all os/arch pairs from darwin/arm64


sgagniere (dismissed):
+1

sgagniere (approved):
+1

PR #2556, opened 2024-01-17
chore: remove language service flag

Guttz:
Release Notes
-------------

Checklist
---------

What
----

The language service is now enabled by default when running confluent flink shell.

References
----------

Test & Review
-------------


sgagniere:
Since this is for a feature that isn't GA yet, is it fine to introduce this change without a release note? The reason is that our release pipeline will attempt to release v4.0 if the breaking changes field isn't empty.

Guttz:
Sure, we almost only internal users. I've updated description. And actually, feel free to update it yourself if you want to change something.

sgagniere (approved):

brianstrauch (approved):

PR #2565, opened 2024-01-23
[CLI-2705] Switch all delete/undelete confirmation prompts to yes/no

sgagniere:
Release Notes
-------------

Breaking Changes
- All confirmation prompts for resource `delete` and `undelete` commands are now yes/no prompts.

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Remove the confirmation prompts which requests that users type in a resource name. They are now yes/no prompts so that the all delete and undelete commands have the same behavior across all resources.

References
----------

Test & Review
-------------
Updated integration tests.


internal/api-key/command_delete.go
```diff
@@ -32,7 +32,7 @@ func (c *command) delete(cmd *cobra.Command, args []string) error {
 		return err == nil
 	}
 
-	if err := deletion.ValidateAndConfirmDeletionYesNo(cmd, args, existenceFunc, resource.ApiKey); err != nil {
+	if err := deletion.ValidateAndConfirmDeletion(cmd, args, existenceFunc, resource.ApiKey); err != nil {
```
brianstrauch, line 35: Since the package is called "deletion" already, I'm ok removing the second "deletion"
```suggestion
	if err := deletion.ValidateAndConfirm(cmd, args, existenceFunc, resource.ApiKey); err != nil {
```

brianstrauch (approved):
Much better UX, thanks :)

internal/api-key/command_delete.go
sgagniere: I think we can leave this one because we also have `ValidateAndConfirmUndeletion`.

brianstrauch (approved):

PR #2568, opened 2024-01-23
[CLI-2992] Return an error instead of a panic if integer & double keys are switched in `topic produce` and `topic consume`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix panic in `confluent kafka topic consume --key-format double` when consuming a message with an integer key

Checklist
---------

What
----
Producing a message with an integer key and then deserializing it with the double deserializer produces a panic because `binary.LittleEndian.Uint64` checks that the 8th index of the key byte slice is set, but integer keys have a byte slice of length 4.

Also, producing a message with a double key and then using the integer deserializer produces garbage.

So in both cases, they now return an error if the length is incorrect.

References
----------

Test & Review
-------------
Manual testing

### DIFF
--- a/pkg/serdes/double_deserialization_provider.go
+++ b/pkg/serdes/double_deserialization_provider.go
@@ -13,6 +13,10 @@ func (DoubleDeserializationProvider) LoadSchema(_ string, _ map[string]string) e
 }
 
 func (DoubleDeserializationProvider) Deserialize(data []byte) (string, error) {
+	if len(data) == 0 {
+		return "", nil
+	}
+
 	if len(data) != 8 {
 		return "", fmt.Errorf("the double key is invalid")
 	}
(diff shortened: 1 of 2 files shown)


pkg/serdes/double_deserialization_provider.go
```diff
@@ -13,8 +13,8 @@ func (DoubleDeserializationProvider) LoadSchema(_ string, _ map[string]string) e
 }
 
 func (DoubleDeserializationProvider) Deserialize(data []byte) (string, error) {
-	if len(data) == 0 {
-		return "", nil
+	if len(data) != 8 {
+		return "", fmt.Errorf("the double key is invalid")
```
brianstrauch, line 17: This is a breaking change. Let's keep the original `if` statement and put it above this one.

brianstrauch (changes requested):

brianstrauch:
> Fix panic in confluent kafka topic consume --key-format [integer | double]

Fix panic in `confluent kafka topic consume --key-format` when passing an incorrectly sized integer or double

sgagniere:
@brianstrauch How about

Fix panic in `confluent kafka topic consume --key-format double` when consuming a message with an integer key

brianstrauch (approved):
Thanks!

PR #2577, opened 2024-01-26
Update `release-docs` to work with patch releases

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
For patch versions, `release-docs` was still attempting to push directly to protected branches.

Instead of attempting to push to the `-post` branch, this PR just opens a PR to pull the new `command-reference` and `release-notes.rst` from the `.x` branch into the `-post` branch.

I didn't merge `.x` into `-post` to avoid picking up incorrect changes to `settings.sh` and `conf.py`. These files will be updated in the `-post` branch anyway by `release-docs-staging`

References
----------

Test & Review
-------------


mk-files/docs.mk
```diff
@@ -68,10 +65,8 @@ release-docs:
 		git checkout master && \
 		git checkout -b $(STAGING_BRANCH) && \
 		$(call dry-run,git push -u origin $(STAGING_BRANCH)); \
-	fi && \
-	git checkout -B $(CURRENT_SHORT_MINOR_VERSION)-post origin/$(STAGING_BRANCH) && \
-	$(call dry-run,git push -u origin $(CURRENT_SHORT_MINOR_VERSION)-post) && \
-	if [[ $(CLEAN_VERSION) == *.0 ]]; then \
+		git checkout -B $(CURRENT_SHORT_MINOR_VERSION)-post origin/$(STAGING_BRANCH) && \
```
brianstrauch, line 68: Do we still need the capital "-B" here, or would lowercase "-b" work fine? (I've always been confused by this)

mk-files/docs.mk
```diff
@@ -81,6 +76,17 @@ release-docs:
 		git commit -am "[ci skip] chore: update settings.sh and conf.py due to $(CLEAN_VERSION) release" && \
 		$(call dry-run,git push origin release-docs-$(CLEAN_VERSION)) && \
 		$(call dry-run,gh pr create --base master --title "Release docs for v$(CLEAN_VERSION)" --body ""); \
+	else \
+		git checkout -b release-docs-$(STAGING_BRANCH_DASH) origin/$(STAGING_BRANCH) && \
+		git checkout origin/master -- command-reference release-notes.rst && \
```
brianstrauch, line 81: Can we just `cp` these over, like we do the first time they're added?

mk-files/docs.mk
sgagniere: Lowercase `-b` should work here. For non-patch releases (this case here), these branches don't exist yet anyway so there shouldn't be a difference between `-b/-B`.

mk-files/docs.mk
sgagniere: I think I was going to try this based on this comment, but on second thought I think this is simpler. We would have to repeat the process of cloning CLI to a temp folder and generating the command reference. Since the first docs step already does this and updates origin master, we can just pull it from there.

brianstrauch: No worries, we're going to switch to a docs team supported pipeline soon anyway!

brianstrauch (approved):

PR #2596, opened 2024-02-08
[KFS-1607] removed flink-region flag, since is not possible anymore t…

luca-filipponi:
…o grant FlinkDeveloper role to a compute pool

Release Notes
-------------
Bug Fixes
- Remove broken `--flink-region` flag from `confluent iam rbac role-binding` commands

Checklist
---------

What
----

References
----------

Test & Review
-------------


YannickPferr (approved):

brianstrauch (commented):
Interesting that this wasn't put behind a feature flag since Flink is still in EA. Removing it would be a breaking change, so we'll have to wait until our v4 release.

luca-filipponi:
>Interesting that this wasn't put behind a feature flag since Flink is still in EA. Removing it would be a breaking change, so we'll have to wait until our v4 release.

The backend is going to change without a feature flag before GA, so this command won't work anymore soon. Will check if they can put this behind a feature flag

brianstrauch:
> Will check if they can put this behind a feature flag.

Thanks!

> When V4 release is going to happen?

Early Q2

brianstrauch (approved):

PR #2598, opened 2024-02-08
[CLI-3010] chore: upgrade go-prompt to v0.2.26

Guttz:
Release Notes
-------------

Checklist
---------
- [X] Leave this box unchecked if features are not yet available in production

What
----

Upgrade go-prompt's version, which includes a fix to avoid panics caused by multibyte(special characters) when using `confluent flink shell`.

References
----------

Test & Review
-------------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/confluentinc/confluent-kafka-go v1.9.3-RC3
 	github.com/confluentinc/go-editor v0.11.0
 	github.com/confluentinc/go-netrc v0.0.0-20220321173724-4d50f36ff450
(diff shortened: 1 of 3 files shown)


brianstrauch:
Looks like the mocks are failing to build? @Guttz

sgagniere (approved):

brianstrauch:
Make sure you write release notes in the future @Guttz!

Guttz:
> Make sure you write release notes in the future @Guttz!

I somehow thought we shouldn't write them for flink yet since wa haven't released GA. If that's not the case, glad to do it!

PR #2409, opened 2023-11-09
DGS-9026: Update environment create/update commands to include stream governance package

GunalKupta:
Release Notes
-------------

New Features
- `--governance-package` flag on environment create/update commands
- environment create, update, list, and describe commands return a new field "Stream Governance Package"

Checklist
---------
- [ ] Leave this box unchecked if features are not yet available in production

What
----

As part of Always-On Governance, a stream governance package will be enabled in every Confluent environment by default. This PR allows users to specify which package they want enabled using the CLI (default is `ESSENTIALS`).

References
----------

[DGS-9026](https://confluentinc.atlassian.net/browse/DGS-9026)
[Always-On Governance - Environment](https://confluentinc.atlassian.net/wiki/spaces/DG/pages/3120630044/1-Pager+Always+On+Stream+Governance+Phase+2#Environments-(org%2Fv2))

Test & Review
-------------

[DGS-9026]: https://confluentinc.atlassian.net/browse/DGS-9026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2600, opened 2024-02-09
DGS-10013: Do not provide default SG package to new environments

GunalKupta:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Defaulting the stream governance package to Essentials for new environments causes a potentially breaking change for users who automate environment and SR cluster creation. This breaking change should only be included in v4 deployment and not the current minor release.

References
----------
https://confluentinc.atlassian.net/browse/DGS-10013
https://confluentinc.atlassian.net/wiki/x/twI3xQ

### DIFF
--- a/internal/environment/command_create.go
+++ b/internal/environment/command_create.go
@@ -21,7 +21,7 @@ func (c *command) newCreateCommand() *cobra.Command {
 		RunE:  c.create,
 	}
 
-	c.addStreamGovernancePackageFlag(cmd, "essentials")
+	c.addStreamGovernancePackageFlag(cmd, "")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 	pcmd.AddOutputFlag(cmd)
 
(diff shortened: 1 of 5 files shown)


sgagniere (dismissed):
+1

internal/environment/command_create.go
```diff
@@ -21,7 +21,7 @@ func (c *command) newCreateCommand() *cobra.Command {
 		RunE:  c.create,
 	}
 
-	c.addStreamGovernancePackageFlag(cmd, "essentials")
+	c.addStreamGovernancePackageFlag(cmd, "")
```
brianstrauch, line 24: This is a breaking change... unless the backend treats `""` as `"essentials"`?

brianstrauch: (No need to wait for v4 if the above is true)

sgagniere: I think the idea of this PR is that introducing this itself was a breaking change, so reversing that should just be a bug fix, right?

brianstrauch: Wait, we accidentally merged a breaking change and this reverts it?

GunalKupta: Correct. The changes were made in https://github.com/confluentinc/cli/pull/2409. The "breaking" part of the changes are explained [here](https://confluentinc.atlassian.net/wiki/x/twI3xQ)

brianstrauch (approved):

akhileshm1 (approved):

PR #2602, opened 2024-02-09
[RCCA-17513] Fix group mapping errors in `confluent iam rbac role-binding` commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix a bug preventing role bindings with an SSO group mapping principal from displaying in `confluent iam rbac role-binding list`
- Fix a bug causing an incorrect error message to be displayed in `confluent iam rbac role-binding create` and `confluent iam rbac role-binding delete` when the principal is an SSO group mapping

Checklist
---------

What
----
We weren't handling group mappings in `conflulent iam rbac role-binding` commands, so `list` omitted role bindings for group mappings, and `create|delete` treated group mapping IDs as user IDs and subsequently produced an error while trying to display the result, even though the underlying create/delete succeeded.

References
----------

Test & Review
-------------
I added a group mapping role binding to the test backend and updated the tests.
I also switched `pool-` to `group-` throughout the tests.

### DIFF
--- a/internal/iam/command_rbac_role_binding.go
+++ b/internal/iam/command_rbac_role_binding.go
@@ -55,8 +55,6 @@ var (
 
 	literalPatternType  = "LITERAL"
 	prefixedPatternType = "PREFIXED"
-
-	nonUserPrincipalTypes = []string{presource.ServiceAccount, presource.IdentityPool, presource.SsoGroupMapping}
 )
 
 type roleBindingOptions struct {
@@ -401,6 +399,8 @@ func (c *roleBindingCommand) displayCCloudCreateAndDeleteOutput(cmd *cobra.Comma
 
 	userId := strings.TrimPrefix(roleBinding.GetPrincipal(), "User:")
 	principalType := presource.LookupType(userId)
+
+	nonUserPrincipalTypes := []string{presource.ServiceAccount, presource.IdentityPool, presource.SsoGroupMapping}
 
 	var fields []string
 	if slices.Contains(nonUserPrincipalTypes, principalType) {
--- a/internal/iam/command_rbac_role_binding.go
+++ b/internal/iam/command_rbac_role_binding.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -54,6 +55,8 @@ var (
 
 	literalPatternType  = "LITERAL"
 	prefixedPatternType = "PREFIXED"
+
+	nonUserPrincipalTypes = []string{presource.ServiceAccount, presource.IdentityPool, presource.SsoGroupMapping}
 )
 
 type roleBindingOptions struct {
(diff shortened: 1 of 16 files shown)


internal/iam/command_rbac_role_binding.go
```diff
@@ -400,7 +400,7 @@ func (c *roleBindingCommand) displayCCloudCreateAndDeleteOutput(cmd *cobra.Comma
 	principalType := presource.LookupType(userId)
 
 	var fields []string
-	if principalType == presource.ServiceAccount || principalType == presource.IdentityPool {
+	if principalType == presource.ServiceAccount || principalType == presource.IdentityPool || principalType == presource.SsoGroupMapping {
```
brianstrauch, line 403: https://pkg.go.dev/slices#Contains

internal/iam/command_rbac_role_binding_list.go
```diff
@@ -491,6 +510,10 @@ func (c *roleBindingCommand) ccloudListRolePrincipals(cmd *cobra.Command, listRo
 			row.Name = name
 			list.Add(row)
 		}
+		if name, ok := groupMappingToNameMap[principal]; ok {
```
brianstrauch, line 513: Unfortunate that we need to keep adding to this `for` loop every time we introduce a new resource.

test/fixtures/output/iam/rbac/role-binding/list-user-orgadmin-cloud.golden
```diff
@@ -1,5 +1,6 @@
...
-------------------+-----------------+-----------------------
-  User:pool-12345 | identity-pool   |                       
-  User:sa-12345   | service-account |                       
-  User:u-11aaa    |                 | u-11aaa@confluent.io  
+     Principal    |         Name          |        Email          
+------------------+-----------------------+-----------------------
+  User:group-abc  | another-group-mapping |                       
+  User:pool-12345 | identity-pool         |                       
+  User:sa-12345   | service-account       |                       
+  User:u-11aaa    |                       | u-11aaa@confluent.io
```
brianstrauch, line 6: What about combining the "Name" and "Email" column in v4?

sgagniere: That's an option. The other option is that we can just display the user's name here; I'm not sure why we currently aren't.

brianstrauch: Oh, I don't think that field was added until a year or so ago. We should definitely add that, but removing the email might cause problems.

internal/iam/command_rbac_role_binding.go
```diff
@@ -54,6 +55,8 @@ var (
 
 	literalPatternType  = "LITERAL"
 	prefixedPatternType = "PREFIXED"
+
+	nonUserPrincipalTypes = []string{presource.ServiceAccount, presource.IdentityPool, presource.SsoGroupMapping}
```
brianstrauch, line 59: Let's define this in the function it's used in? No need for it to be global in this package

brianstrauch (approved):

PR #2603, opened 2024-02-09
Fix `confluent local` test

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
More kafka topic configs are being returned in `confluent local kafka topic describe`, so this PR updates the golden file to fix the test.

References
----------

Test & Review
-------------

### DIFF
--- a/test/fixtures/output/local/kafka/topic/describe.golden
+++ b/test/fixtures/output/local/kafka/topic/describe.golden
@@ -19,16 +19,21 @@ Configuration
   follower.replication.throttled.replicas |                      
   index.interval.bytes                    | 4096                 
   leader.replication.throttled.replicas   |                      
+  local.retention.bytes                   | -2                   
+  local.retention.ms                      | -2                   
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
I don't remember this test failing in the past. What changed?

sgagniere:
@brianstrauch A new version of the confluent-local image, I think. I now have two versions (one dangling) after running the local tests locally.

brianstrauch:
The CLI should not be using the "latest" tag of that image, if that's what you mean. We should validate that each image works with our test suite before updating it.

javabrett:
> The CLI should not be using the "latest" tag of that image, if that's what you mean. We should validate that each image works with our test suite before updating it.

However now we are pinned to a 2-versions-old version, since there's no release process to trigger review/test/bump-version for this.

PR #2623, opened 2024-02-15
KGLOBAL-4436: Different Config Parsing for Cluster Link

chernyih:
Release Notes
-------------

Bug Fixes
- Fix a bug causing `confluent kafka link create`, `confluent kafka link configuration update`, and `confluent kafka mirror create` to fail when the cluster link configuration "ssl.truststore.certificates" contains an explicit "\n" character

Checklist
---------

What
----
Change cluster link and mirror topic operations to use a different config parsing that is more compatible with Java property format https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html. More specifically, explicit \n in the value is treated as \\\n by current CLI parser while the Java property parser treats it as \n.

Since the new parser has a different behavior and not exactly the same as Java property parser, i only change cluster link and mirror topic to use the new parser to reduce the blast radius.

References
----------
https://confluentinc.atlassian.net/browse/KGLOBAL-4436

Test & Review
-------------
Added various test cases. Manually tested create cluster link, update cluster link config that caused issues to customer.

### DIFF
--- a/internal/kafka/command_link_configuration_update.go
+++ b/internal/kafka/command_link_configuration_update.go
@@ -58,7 +58,7 @@ func (c *linkCommand) configurationUpdate(cmd *cobra.Command, args []string) err
 		config = []string{configFile}
 	}
 
-	configMap, err := properties.GetMap(config)
+	configMap, err := properties.GetMapWithJavaPropertyParsing(config)
 	if err != nil {
 		return err
 	}
--- a/internal/kafka/command_link_configuration_update.go
+++ b/internal/kafka/command_link_configuration_update.go
@@ -58,7 +58,7 @@ func (c *linkCommand) configurationUpdate(cmd *cobra.Command, args []string) err
 		config = []string{configFile}
 	}
 
-	configMap, err := properties.GetMapWithJavaPropertyParsing(config)
+	configMap, err := properties.GetMap(config)
 	if err != nil {
 		return err
 	}
--- a/pkg/properties/properties.go
+++ b/pkg/properties/properties.go
@@ -3,6 +3,7 @@ package properties
 import (
 	"bytes"
 	"fmt"
+	"github.com/confluentinc/properties"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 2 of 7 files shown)


internal/kafka/command_link_configuration_update.go
```diff
@@ -58,7 +58,7 @@ func (c *linkCommand) configurationUpdate(cmd *cobra.Command, args []string) err
 		config = []string{configFile}
 	}
 
-	configMap, err := properties.GetMap(config)
+	configMap, err := properties.GetMapWithJavaPropertyParsing(config)
```
brianstrauch, line 61: I don't think we need a separate function here. Why wouldn't we want this bug fix in the existing function?

brianstrauch (changes requested):

internal/kafka/command_link_configuration_update.go
chernyih: There is a difference in behavior. The main one the new parser treats the first empty space as key value separator. For example, "key val" equals to "key=val". In the current parser, this is an error. Also there might be behavior that i don't know.
Therefore, i want to reduce the blast radius. If things are ok with CL for sometime, we can switch others to use it.

chernyih: In the latest patch, I replaced the default parser with the new parser. Let's see how it goes. If no tests failing, we can try using it. If found regression, we can either revert the change or fix it.

pkg/properties/properties_test.go
```diff
@@ -1,41 +1,12 @@
...
 
-func TestParseLines_Empty(t *testing.T) {
-	lines := parseLines("")
-	require.Empty(t, lines)
-}
-
-func TestParseLines_Comment(t *testing.T) {
-	lines := parseLines("#key=val")
-	require.Empty(t, lines)
-}
```
brianstrauch, line 17: Even though we're getting rid of this function can we rewrite this test case? This is something we still want to support.

pkg/properties/properties.go
```diff
@@ -45,42 +51,14 @@ func ConfigSliceToMap(configs []string) (map[string]string, error) {
...
-		if line != "" && line[0] != '#' {
-			lines = append(lines, line)
-		}
-	}
-
-	return lines
-}
-
 // ConfigFlagToMap reads key=values pairs from the --config flag and supports configuration values containing commas.
 func ConfigFlagToMap(configs []string) (map[string]string, error) {
```
brianstrauch, line 55: Can we still parse `--config key1=value,with,commas,key2=another,value,with,commas` after this change? I don't think we can.

brianstrauch (changes requested):

pkg/properties/properties.go
```diff
@@ -3,14 +3,21 @@ package properties
...
+
 	"github.com/confluentinc/cli/v3/pkg/utils"
 )
 
 // GetMap reads newline-separated configuration files or comma-separated lists of key=value pairs, and supports configuration values containing commas.
+// One known difference compared to Java Property format is it treats the first space as the key value separator.
+// For examples:
+// key is equal to key=
+// key val is equal to key=val
+// key val1 val2 is equal to key=val1 val2
```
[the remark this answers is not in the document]
chernyih: Yup

pkg/properties/properties_test.go
chernyih: In new PR https://github.com/confluentinc/cli/pull/2627, this function is not removed

chernyih:
After thinking more about it, i think it is safer to just replace those explicit special characters than changing the entire parser. New PR: https://github.com/confluentinc/cli/pull/2627

chernyih:
Closing this because going with https://github.com/confluentinc/cli/pull/2627 because the new parser is not compatible with what we are doing currently.

PR #2625, opened 2024-02-16
Feat/KFS-1598 KFS-1599 escaped use statement

Guttz:
Release Notes
-------------

New Features
- Support escaped characters for catalogs and databases set with "USE" in `confluent flink shell`

Checklist
---------

What
----

As Specification for SQL Clients: USE statements in `confluent flink shell` now support escaped catalog and database using backticks.

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/internal/store/store_test.go
+++ b/pkg/flink/internal/store/store_test.go
@@ -627,6 +627,30 @@ func (s *StoreTestSuite) TestParseUseStatementCatalogDatabasePath() {
 	catalog, database, _ = parseUseStatement("USE `my_catalog`.`my_database`")
 	assert.Equal(s.T(), "my_catalog", catalog)
 	assert.Equal(s.T(), "my_database", database)
+
+	catalog, database, _ = parseUseStatement("USE `my catalog`.database")
+	assert.Equal(s.T(), "my catalog", catalog)
+	assert.Equal(s.T(), "database", database)
+
+	catalog, database, _ = parseUseStatement("USE cat.`my database`")
+	assert.Equal(s.T(), "cat", catalog)
+	assert.Equal(s.T(), "my database", database)
+
+	catalog, database, _ = parseUseStatement("USE `my catalog`   .   `my database`")
+	assert.Equal(s.T(), "my catalog", catalog)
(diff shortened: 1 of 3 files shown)


brianstrauch:
> USE statements in confluent flink shell now support escaped catalog and database using backticks

How about:
```
Support escaped characters for catalogs and databases selected with the "USE" statement in `confluent flink shell`.
```

pkg/flink/internal/store/store_test.go
```diff
@@ -522,47 +522,153 @@ func (s *StoreTestSuite) TestParseSetStatementError() {
 }
 
 func (s *StoreTestSuite) TestParseUseStatement() {
-	key, value, _ := parseUseStatement("USE CATALOG c;")
-	assert.Equal(s.T(), flinkconfig.KeyCatalog, key)
-	assert.Equal(s.T(), "c", value)
+	catalog, database, _ := parseUseStatement("USE CATALOG c;")
```
brianstrauch, line 525: I like the original API better. We should return a type and name `("CATALOG", "c")` to make things easier if we decide to add more use-able types in the future

Guttz:
> > USE statements in confluent flink shell now support escaped catalog and database using backticks
> 
> How about:
> 
> ```
> Support escaped characters for catalogs and databases selected with the "USE" statement in `confluent flink shell`.
> ```

Updated it and just changed select to set :)

pkg/flink/internal/store/store_test.go
Guttz: Me too... However, use can now return both catalog and database, and it couldn't in the previous version. That means, I had to either:

1 - Return two strings
2 - Return an array with two positions
3 - Return an object

Since SQL is really stable and very rarely changes, this function will probably return two values for a long time. My thought was that 2 and 3 would require too much boil code for a function that only returns two values. I'd be ok with all 3 but that's why I personally prefer 1.

brianstrauch: > use can now return both catalog and database

Ah, didn't realize that :+1:

brianstrauch (approved):

YannickPferr (approved):

PR #2626, opened 2024-02-16
DGS-10095: Add unit test for environment update package downgrade

GunalKupta:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Adds a unit test for downgrading SG package using environment update command.

References
----------

https://confluentinc.atlassian.net/browse/DGS-10095

Test & Review
-------------


akhileshm1 (dismissed):

test/environment_test.go
```diff
@@ -63,6 +63,17 @@ func (s *CLITestSuite) TestEnvironmentUse() {
 	}
 }
 
+func (s *CLITestSuite) TestEnvironmentNoPackage() {
```
brianstrauch, line 66:
```suggestion
func (s *CLITestSuite) TestEnvironmentUpdate_NoPackage() {
```

brianstrauch: Why is this called "NoPackage"? It looks like the package is "essentials".

test/environment_test.go
```diff
@@ -63,6 +63,17 @@ func (s *CLITestSuite) TestEnvironmentUse() {
 	}
 }
 
+func (s *CLITestSuite) TestEnvironmentUpdate_PackageDowngrade() {
+	tests := []CLITest{
+		{args: "environment update env-595 --governance-package essentials", fixture: "environment/sg-downgrade-fail.golden", exitCode: 1},
```
brianstrauch, line 68: Should we also test the case where the downgrade is successful?

GunalKupta: Downgrade should never be successful :)

GunalKupta: Should I add a comment to clarify that?

test/test-server/org_handlers.go
```diff
@@ -56,8 +58,17 @@ func handleOrgEnvironment(t *testing.T) http.HandlerFunc {
 				StreamGovernanceConfig: &orgv2.OrgV2StreamGovernanceConfig{Package: orgv2.PtrString("ESSENTIALS")}}
 			err := json.NewDecoder(r.Body).Decode(req)
 			require.NoError(t, err)
-			req.Id = orgv2.PtrString(id)
+			if id == "env-595" && req.StreamGovernanceConfig.GetPackage() == "ESSENTIALS" {
+				w.WriteHeader(http.StatusForbidden)
+				resp := errors.ErrorResponseBody{Errors: []errors.ErrorDetail{
+					{Detail: "SR package downgrade is not a supported operation"}},
```
brianstrauch, line 64:
```suggestion
					{Detail: "Schema Registry package downgrade is not a supported operation"}},
```

test/environment_test.go
brianstrauch: You could update the description of the `--governance-package` flag to say that!

brianstrauch, line 68:
```suggestion
		{args: "environment update env-595 --governance-package essentials", fixture: "environment/upgrade-governance-package-downgrade-fail.golden", exitCode: 1},
```

test/test-server/org_handlers.go
GunalKupta: I'm gonna reject this one because it already mimics the exact response from the org service

brianstrauch: Can we change the backend error to say "Schema Registry"? We try to avoid abbreviations in any customer-facing text.

internal/environment/command_create.go
```diff
@@ -65,5 +65,5 @@ func (c *command) create(cmd *cobra.Command, args []string) error {
 
 func (c *command) addStreamGovernancePackageFlag(cmd *cobra.Command, defaultValue string) {
 	values := utils.ArrayToCommaDelimitedString([]string{"essentials", "advanced"}, "or")
-	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf("Specify the Stream Governance package as %s.", values))
+	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf("Specify the Stream Governance package as %s. Note: downgrading the package from \"advanced\" to \"essentials\" is forbidden.", values))
```
brianstrauch, line 68:
```suggestion
	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf(`Specify the Stream Governance package as %s. Downgrading the package from "advanced" to "essentials" is not allowed.`, values))
```

brianstrauch (dismissed):

internal/environment/command_create.go
brianstrauch: Wait, shouldn't the message about downgrading only be included in `confluent environment update`?

GunalKupta: I think it would be helpful for users to know before they create an environment that they can't downgrade the package, right? Users will be less likely to make an irreversible decision.

brianstrauch: That's a good point. Let's merge this!

brianstrauch (dismissed):

internal/environment/command_create.go
```diff
@@ -65,5 +65,5 @@ func (c *command) create(cmd *cobra.Command, args []string) error {
 
 func (c *command) addStreamGovernancePackageFlag(cmd *cobra.Command, defaultValue string) {
 	values := utils.ArrayToCommaDelimitedString([]string{"essentials", "advanced"}, "or")
-	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf("Specify the Stream Governance package as %s.", values))
+	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf(`Specify the Stream Governance package as %s. Downgrading the package from "advanced" to "essentials" is not allowed.`, values))
```
akhileshm1, line 68: i think this is a bit nuanced. we only don't allow downgrades if the SR cluster is already provisioned. we can probably update it to:
`Downgrading the package from "advanced" to "essentials" is not allowed once the schema registry cluster is provisioned.`
or remove it altogether. will defer to @brianstrauch to decide which is better.

akhileshm1 (approved):

akhileshm1 (dismissed):

internal/environment/command_create.go
```diff
@@ -65,5 +65,5 @@ func (c *command) create(cmd *cobra.Command, args []string) error {
 
 func (c *command) addStreamGovernancePackageFlag(cmd *cobra.Command, defaultValue string) {
 	values := utils.ArrayToCommaDelimitedString([]string{"essentials", "advanced"}, "or")
-	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf("Specify the Stream Governance package as %s.", values))
+	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf(`Specify the Stream Governance package as %s. Downgrading the package from "advanced" to "essentials" is not allowed once the schema registry cluster is provisioned.`, values))
```
brianstrauch, line 68:
```suggestion
	cmd.Flags().String("governance-package", defaultValue, fmt.Sprintf(`Specify the Stream Governance package as %s. Downgrading the package from "advanced" to "essentials" is not allowed once the Schema Registry cluster is provisioned.`, values))
```

brianstrauch: So if a user runs `confluent environment create test --governance-package advanced` followed by `confluent environment update --governance-package essentials` fast enough, it'll work, as long as the Schema Registry cluster isn't provisioned?

GunalKupta: SR cluster doesn't get provisioned until after Kafka cluster is created, so they would just need to run `confluent environment update test --governance-package essentials` before running `confluent kafka cluster create cluster_0`

brianstrauch (approved):

PR #2627, opened 2024-02-17
KGLOBAL-4436: Replace special characters when parsing the configurations

chernyih:
Release Notes
-------------
Bug Fixes
- Fix a bug causing `confluent kafka link create` and `confluent kafka link configuration update` to fail when the cluster link configuration "ssl.truststore.certificates" contains an explicit "\n" character

Checklist
---------

What
----
Replace explicit special characters (newline, carriage return and tab) in config value as specified in https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html. More specifically, explicit \n in the value is treated as \\n by current CLI parser while the Java property parser treats it as \n.

References
----------
https://confluentinc.atlassian.net/browse/KGLOBAL-4436

Test & Review
-------------
Tested on real cluster. Added unit test.

### DIFF
--- a/pkg/properties/properties.go
+++ b/pkg/properties/properties.go
@@ -39,7 +39,7 @@ func ConfigSliceToMap(configs []string) (map[string]string, error) {
 			return nil, fmt.Errorf(`failed to parse "key=value" pattern from configuration: %s`, config)
 		}
 
-		m[x[0]] = x[1]
+		m[x[0]] = replaceSpecialCharacters(x[1])
 	}
 
 	return m, nil
@@ -69,7 +69,7 @@ func ConfigFlagToMap(configs []string) (map[string]string, error) {
 		if strings.Contains(configs[i], "=") {
 			x := strings.SplitN(configs[i], "=", 2)
 			if _, ok := m[x[0]]; !ok {
-				m[x[0]] = x[1]
+				m[x[0]] = replaceSpecialCharacters(x[1])
 			}
 		} else {
 			if i-1 >= 0 {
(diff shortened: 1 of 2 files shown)


pkg/properties/properties.go
```diff
@@ -96,3 +96,10 @@ func CreateKeyValuePairs(m map[string]string) string {
 	}
 	return b.String()
 }
+
+func replaceSpecialCharacters(val string) string {
+	// Replace \\n, \\r and \\t with newline, carriage return and tab characters as specified in
+	// https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html.
+	return strings.ReplaceAll(strings.ReplaceAll(
+		strings.ReplaceAll(val, "\\n", "\n"), "\\r", "\r"), "\\t", "\t")
```
brianstrauch, line 104: From the link:
> The backslash character must be escaped as a double backslash.

Seems like we can simplify to the following (and move this out of its own function since it's a 1-liner)

```suggestion
	return strings.ReplaceAll(val, "\\", "\")
```

chernyih: @brianstrauch Thanks for the review. "\" is not a valid string and will cause compilation error. It is an escape character that changes the meaning of the next character.

Do you want still want me to inline the following?
return strings.ReplaceAll(strings.ReplaceAll(
		strings.ReplaceAll(val, "\\n", "\n"), "\\r", "\r"), "\\t", "\t")

brianstrauch: Oops, updated the suggestion above.

chernyih: strings.ReplaceAll(val, `\\`, `\`) is equivalent to strings.ReplaceAll(val, "\\\\", "\\") which wouldn't help in converting "\\n" to "\n"

brianstrauch (approved):

chernyih:
@brianstrauch Thanks for the review. Merged to master

PR #2633, opened 2024-02-19
fix: make colors more visible in interactive output

YannickPferr:
Release Notes
-------------

Bug Fixes
- Improve text visibility in interactive output of `confluent flink shell`

Checklist
---------

What
----

Switch from dark cyan color name to hex code, since the color is different on Windows and Mac

Windows before:
image

Windows after:
image

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/color/color.go
+++ b/pkg/color/color.go
@@ -12,3 +12,5 @@ const AccentColor = color.FgCyan
 
 const ErrorColor = color.FgHiRed
 const WarnColor = color.FgHiYellow
+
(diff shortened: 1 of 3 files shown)


Guttz (approved):

Guttz:
Obs.: this is only applied to tview - the interactive output mode and that's why the change is so small. When using go-prompt (for input), we can use the colors defined as variables by go-prompt

brianstrauch (approved):

PR #2648, opened 2024-02-22
Add confluent/api to CLI's CODEOWNERS

linouk23:
### What
This PR adds API team to CLI repo's CODEOWNERS.


brianstrauch (commented):
What if we added everyone in developer interfaces to the @confluentinc/cli handle instead?

linouk23:
That sounds like a great idea too 👍

PR #2652, opened 2024-02-24
[CLI-2816] On-prem OAuth SSO support

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Support on-premises SSO authentication with `confluent login`
- Add a new on-premises `confluent iam user describe` to display username and authentication token

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Add support for on-premises SSO authentication when `oidc.login.device.1.enabled` is enabled.

When both SSO and LDAP are enabled (which is possible for on-prem), we should prefer SSO. So the current order of precedence is `prompt flag > environment variables (SSO > LDAP) > LDAP (keychain > config > netrc) > SSO > LDAP (prompt)`.

References
----------

Test & Review
-------------
Manual testing


internal/login/command.go
```diff
@@ -300,16 +284,44 @@ func (c *command) loginMDS(cmd *cobra.Command, url string) error {
...
-		return caCertPath, err
+// Current functionality:
+// empty ca-cert-path is equivalent to not using ca-cert-path flag
+// if users want to login with ca-cert-path they must explicitly use the flag every time they login
+//
+// For legacy users:
+// if ca-cert-path flag is not used, then return caCertPath value stored in config for the login context
+// if user passes empty string for ca-cert-path flag then reset the ca-cert-path value in config for the context
+// (only for legacy contexts is it still possible for the context name without ca-cert-path to have ca-cert-path)
+func (c *command) getCACertPath(cmd *cobra.Command, username, url string) (string, bool, error) {
```
brianstrauch, line 295:
```suggestion
func (c *command) getCaCertPath(cmd *cobra.Command, username, url string) (string, bool, error) {
```

internal/login/command.go
```diff
@@ -300,16 +284,44 @@ func (c *command) loginMDS(cmd *cobra.Command, url string) error {
...
+	caCertPath, err := cmd.Flags().GetString("ca-cert-path")
+	if err != nil {
+		return "", false, err
+	}
+
+	if caCertPath == "" {
+		caCertPath = pauth.GetEnvWithFallback(pauth.ConfluentPlatformCACertPath, pauth.DeprecatedConfluentPlatformCACertPath)
```
brianstrauch, line 302: Let's make sure we're removing this in v4

internal/login/command.go
```diff
@@ -300,16 +284,44 @@ func (c *command) loginMDS(cmd *cobra.Command, url string) error {
...
+	}
+
+	var isLegacyContext bool
+	if caCertPath == "" {
+		contextName := pauth.GenerateContextName(username, url, "")
+		caCertPath = c.checkLegacyContextCACertPath(cmd, contextName)
```
brianstrauch, line 308: This also seems like something we could remove in v4

internal/login/command.go
```diff
@@ -300,16 +284,44 @@ func (c *command) loginMDS(cmd *cobra.Command, url string) error {
...
+		}
+	}
+
+	return caCertPath, isLegacyContext, nil
 }
 
-// Order of precedence: env vars > netrc > prompt
+// Order of precedence: env vars > sso > ldap (keychain > config > netrc > prompt)
 // i.e. if login credentials found in env vars then acquire token using env vars and skip checking for credentials else where
+// SSO and LDAP (basic auth) can be enabled simultaneously; in this case we should prefer SSO (unless --prompt is used)
```
brianstrauch, line 324: This might not be true, pending discussion with CP team

pkg/auth/auth.go
```diff
@@ -82,6 +93,25 @@ func PersistConfluentLoginToConfig(cfg *config.Config, credentials *Credentials,
 	return addOrUpdateContext(cfg, false, credentials, ctxName, url, state, caCertPath, "", save)
 }
 
+func getSsoSub(authToken string) (string, error) {
```
brianstrauch, line 96: We may want to use this for non-SSO logins in the future (mostly just to simplify the code)
```suggestion
func getSub(authToken string) (string, error) {
```

pkg/auth/auth.go
```diff
@@ -82,6 +93,25 @@ func PersistConfluentLoginToConfig(cfg *config.Config, credentials *Credentials,
...
+	token, err := jwt.ParseSigned(authToken)
+	if err != nil {
+		return "", err
+	}
+
+	var claims map[string]any
+	if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
+		return "", err
+	}
```
brianstrauch, line 105: I think this code already exists somewhere else, maybe we can write a function that takes the authToken and a field like "sub" and spits out the value.

pkg/auth/auth.go
```diff
@@ -82,6 +93,25 @@ func PersistConfluentLoginToConfig(cfg *config.Config, credentials *Credentials,
...
+
+	sub, ok := claims["sub"].(string)
+	if ok {
+		return sub, nil
+	}
+
+	return "", nil
```
brianstrauch, line 112: I feel like it should definitely be an error if "sub" is not found in the token

pkg/auth/login_credentials_manager.go
```diff
@@ -235,6 +238,38 @@ func (h *LoginCredentialsManagerImpl) GetPrerunCredentialsFromConfig(cfg *config
...
+			if err != nil {
+				return nil, err
+			}
+
+			return &Credentials{
+				Username:  sub,
+				IsSSO:     isSSO,
+				AuthToken: ctx.GetAuthToken(),
+			}, nil
+		}
```
brianstrauch, line 267:
```suggestion
		if h.isOnPremSSOUser(url, caCertPath, unsafeTrace) {
			sub, err := getSsoSub(ctx.GetAuthToken())
			if err != nil {
				return nil, err
			}

			credentials := &Credentials{
				Username:  sub,
				IsSSO:     true,
				AuthToken: ctx.GetAuthToken(),
			}
			return credentials, nil
		}
```

pkg/auth/login_credentials_manager.go
```diff
@@ -260,6 +295,19 @@ func (h *LoginCredentialsManagerImpl) getNetrcMachine(filterParams netrc.NetrcMa
 	return netrcMachine, err
 }
 
+func (h *LoginCredentialsManagerImpl) GetOnPremSsoCredentials(url, caCertPath string, unsafeTrace bool) func() (*Credentials, error) {
+	return func() (*Credentials, error) {
+		isSSO := h.isOnPremSSOUser(url, caCertPath, unsafeTrace)
+		if isSSO {
```
brianstrauch, line 301: Do we necessarily need this? I think if you return the credential struct and SSO is false, it doesn't count that as a "complete" credential set

pkg/auth/login_credentials_manager.go
```diff
@@ -323,6 +371,20 @@ func (h *LoginCredentialsManagerImpl) isSSOUser(email, organizationId string) bo
...
+	client, err := clientManager.GetMDSClient(url, caCertPath, unsafeTrace)
+	if err != nil {
+		return false
+	}
+
+	featuresInfo, _, err := client.MetadataServiceOperationsApi.Features(context.Background())
+	if err != nil {
+		return false
+	}
+	return featuresInfo.Features["oidc.login.device.1.enabled"]
```
brianstrauch, line 385: Let's check if this exists so we don't panic if it doesn't?
```suggestion
	enabled, ok := featuresInfo.Features["oidc.login.device.1.enabled"]
```

internal/login/command.go
sgagniere: Yup, there's v4 tickets for this and the deprecated environment variables.

pkg/auth/login_credentials_manager.go
sgagniere: It doesn't panic if the key doesn't exist, it just returns the zero value (so `false` in this case since `Features` is `map[string]bool`). Tested in go playground.

pkg/auth/login_credentials_manager.go
sgagniere: Good point

pkg/auth/auth.go
sgagniere: Yeah it was in `pkg/cmd/token_validator.go`. I moved it to a new package to avoid an import cycle between cmd and auth.

pkg/auth/login_credentials_manager.go
brianstrauch: TIL!

brianstrauch:
> Support on-premises SSO authentication with `confluent login`

Would it make sense to say `confluent login --url <sso-enabled-mds-url>` in this release note?

internal/command.go
```diff
@@ -79,7 +80,7 @@ func NewConfluentCommand(cfg *config.Config) *cobra.Command {
 	authTokenHandler := pauth.NewAuthTokenHandler()
 	ccloudClientFactory := pauth.NewCCloudClientFactory(cfg.Version.UserAgent)
 	flagResolver := &pcmd.FlagResolverImpl{Prompt: form.NewPrompt(), Out: os.Stdout}
-	jwtValidator := pcmd.NewJWTValidator()
+	jwtValidator := jwt.NewJWTValidator()
```
brianstrauch, line 83: Nice change! Always good to avoid writing the package name in public functions of a package
```suggestion
	jwtValidator := jwt.NewValidator()
```

internal/flink/command_shell.go
```diff
@@ -39,7 +40,7 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	return cmd
 }
 
-func (c *command) authenticated(authenticated func(*cobra.Command, []string) error, cmd *cobra.Command, jwtValidator pcmd.JWTValidator) func() error {
+func (c *command) authenticated(authenticated func(*cobra.Command, []string) error, cmd *cobra.Command, jwtValidator jwt.JWTValidator) func() error {
```
brianstrauch, line 43:
```suggestion
func (c *command) authenticated(authenticated func(*cobra.Command, []string) error, cmd *cobra.Command, jwtValidator jwt.Validator) func() error {
```

pkg/auth/auth.go
```diff
@@ -73,13 +73,17 @@ func PersistLogout(config *config.Config) error {
...
+		subClaim, err := jwt.GetClaim(token, "sub")
 		if err != nil {
 			return err
 		}
-		if sub != "" {
-			credentials.Username = sub
+
+		sub, ok := subClaim.(string)
+		if !ok {
+			return fmt.Errorf("malformed token: exp claim has the wrong type")
```
brianstrauch, line 83:
```suggestion
			return fmt.Errorf("malformed token: sub claim has the wrong type")
```

pkg/auth/auth.go
```diff
@@ -73,13 +73,17 @@ func PersistLogout(config *config.Config) error {
...
 	if credentials.IsSSO {
-		sub, err := getSsoSub(token)
+		subClaim, err := jwt.GetClaim(token, "sub")
 		if err != nil {
 			return err
 		}
-		if sub != "" {
-			credentials.Username = sub
+
+		sub, ok := subClaim.(string)
```
brianstrauch, line 81: Isn't every claim a string? I think we can move this line inside of `GetClaim()`

pkg/jwt/token_validator.go
```diff
@@ -32,19 +32,14 @@ func NewJWTValidator() *JWTValidatorImpl {
 // Validate returns an error if the JWT in the specified context is invalid.
 // The JWT is invalid if it's not parsable or expired.
 func (v *JWTValidatorImpl) Validate(context *config.Context) error {
```
brianstrauch, line 34:
```suggestion
func (v *ValidatorImpl) Validate(context *config.Context) error {
```

pkg/auth/auth.go
sgagniere: Nope, `exp` is a float64. That's why I put the responsibility for the type assertion outside of the GetClaim function.

test/test-server/backend.go
```diff
@@ -102,3 +102,11 @@ func (b *TestBackend) GetKafkaRestUrl() string {
...
 	return b.mds.URL
 }
+
+func EnableMdsSso() {
+	isSsoEnabled = true
+}
+
+func DisableMdsSso() {
+	isSsoEnabled = false
+}
```
brianstrauch, line 112: Just make this variable public?

brianstrauch: I wonder if we could use this pattern elsewhere. We have a weird process for enabling/disabling audit logs

sgagniere: Good point; I'm including the test-server package anyway to access these functions so I'll delete them and make the var public.

I don't see any reason why we couldn't do this for audit logs too.

Side note on audit logs: I initially tried to set this up so that it would close and restart the service with a different value for the sso switch like the current audit log pattern, but for some reason I couldn't figure out, the server from `httptest.NewServer` refuses any connections if it's closed and restarted. It's the only testing backend that isn't started with `newTestCloudServer`, and none of those seem to be affected.

pkg/auth/auth_token_handler.go
```diff
@@ -141,13 +145,74 @@ func (a *AuthTokenHandlerImpl) refreshCCloudSSOToken(client *ccloudv1.Client, re
 	return res.GetToken(), refreshToken, err
 }
 
-func (a *AuthTokenHandlerImpl) GetConfluentToken(mdsClient *mdsv1.APIClient, credentials *Credentials) (string, error) {
+func (a *AuthTokenHandlerImpl) GetConfluentToken(mdsClient *mdsv1.APIClient, credentials *Credentials, noBrowser bool) (string, string, error) {
 	ctx := utils.GetContext()
```
brianstrauch, line 149: Didn't know we had this function... maybe we should try to deprecate/remove it? The standard approach to getting the context is from the (un)authenticated command struct

pkg/auth/auth_token_handler.go
```diff
@@ -141,13 +145,74 @@ func (a *AuthTokenHandlerImpl) refreshCCloudSSOToken(client *ccloudv1.Client, re
...
+			refreshToken = checkAuthResp.RefreshToken
+			return nil
+		})
+		if err != nil {
+			return "", "", err
+		}
+
+		return authToken, refreshToken, nil
+	} else {
+		basicContext := context.WithValue(ctx, mdsv1.ContextBasicAuth, mdsv1.BasicAuth{UserName: credentials.Username, Password: credentials.Password})
```
brianstrauch, line 197: Let's construct `basicAuth` outside of this function call. This line is a bit too long

pkg/auth/login_credentials_manager.go
```diff
@@ -260,6 +301,17 @@ func (h *LoginCredentialsManagerImpl) getNetrcMachine(filterParams netrc.NetrcMa
 	return netrcMachine, err
 }
 
+func (h *LoginCredentialsManagerImpl) GetOnPremSsoCredentials(url, caCertPath string, unsafeTrace bool) func() (*Credentials, error) {
+	return func() (*Credentials, error) {
+		// For on-prem SSO logins, the sub claim of the Confluent Token is used in place of the Username
+		// A placeholder is used here since we don't have the token yet
+		return &Credentials{
+			Username: "placeholder",
```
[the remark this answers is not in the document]
sgagniere: Empty username will cause `credentials.IsFullSet()` to be false. This value only exists to bypass that check, and it gets replaced with the actual sub before being written to the config.

pkg/auth/login_credentials_manager.go
```diff
@@ -323,6 +375,20 @@ func (h *LoginCredentialsManagerImpl) isSSOUser(email, organizationId string) bo
 	return err == nil && res.GetIsSso()
 }
 
+func (h *LoginCredentialsManagerImpl) isOnPremSSOUser(url, caCertPath string, unsafeTrace bool) bool {
+	clientManager := &MDSClientManagerImpl{}
```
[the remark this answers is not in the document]
sgagniere: Sort of. I'll have to look into this again because I don't remember the specifics, but I did this to avoid using the mocked version for one reason or another.

pkg/auth/login_credentials_manager.go
```diff
@@ -235,6 +239,39 @@ func (h *LoginCredentialsManagerImpl) GetPrerunCredentialsFromConfig(cfg *config
...
+		ctx := cfg.Context()
+		if ctx == nil {
+			return nil, nil
+		}
+
+		url := ctx.GetPlatform().GetServer()
+		caCertPath := ctx.GetPlatform().GetCaCertPath()
+
+		if h.isOnPremSSOUser(url, caCertPath, unsafeTrace) {
+			subClaim, err := jwt.GetClaim(ctx.GetAuthToken(), "sub")
```
brianstrauch, line 253: Write a comment about why this is being used as the username?

pkg/auth/login_credentials_manager.go
```diff
@@ -75,6 +77,8 @@ type LoginCredentialsManager interface {
 	GetCredentialsFromConfig(*config.Config, netrc.NetrcMachineParams) func() (*Credentials, error)
 	GetCredentialsFromKeychain(bool, string, string) func() (*Credentials, error)
 	GetCredentialsFromNetrc(netrc.NetrcMachineParams) func() (*Credentials, error)
+	GetOnPremSsoCredentials(url, caCertPath string, unsafeTrace bool) func() (*Credentials, error)
+	GetOnPremSsoCredentialsFromConfig(*config.Config, bool) func() (*Credentials, error)
```
brianstrauch, line 81: I think these functions should be able to be combined into one, right?

sgagniere: Probably, yeah, I'll give it a try.

sgagniere: Not sure how to do this neatly, actually.

The issue is that when you first do `confluent login --url ...`, the context is empty so you need to pass in the url and ca-cert-path values from the flag. But afterwards, when using the refresh token, I need to use the url and ca-cert-path from the Context (because some on-prem commands don't support those flags).

So a combined function needs to pass in url, ca-cert-path, and config/context (which I also need for auth and refresh token in addition to url and ca-cert-path) to account for both cases.

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/confluentinc/go-prompt v0.2.31
 	github.com/confluentinc/go-ps1 v1.0.2
 	github.com/confluentinc/kafka-rest-sdk-go/kafkarestv3 v0.3.18
-	github.com/confluentinc/mds-sdk-go-public/mdsv1 v0.0.0-20230117192233-7e6d894d74a9
+	github.com/confluentinc/mds-sdk-go-public/mdsv1 v0.0.0-20240404070529-d5081dacb556
```
sgagniere, line 47: I'll update this when the mds-sdk-go-public PR is merged.

pkg/auth/auth.go
```diff
@@ -28,6 +28,7 @@ const (
 	ConfluentPlatformPassword    = "CONFLUENT_PLATFORM_PASSWORD"
 	ConfluentPlatformMDSURL      = "CONFLUENT_PLATFORM_MDS_URL"
 	ConfluentPlatformCACertPath  = "CONFLUENT_PLATFORM_CA_CERT_PATH"
+	ConfluentPlatformSSO         = "CONFLUENT_PLATFORM_SSO"
```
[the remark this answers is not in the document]
brianstrauch: Let's make sure we add an example of `CONFLUENT_PLATFORM_SSO=true` so users know to pass "true" here instead of something like "1" which is also common

pkg/auth/auth_token_handler.go
brianstrauch: Oh, meaning from the stdlib "context" package?

brianstrauch: Huh, I'm looking at the implementation of this. Why do we need to print DNS/TLS information here?

sgagniere: Hmm. I'm not sure. Maybe we can replace both calls to this w/ background context and remove it like you suggested at the start.

brianstrauch (approved):

PR #2660, opened 2024-02-27
test: add integration test for flink shell

YannickPferr:
Release Notes
-------------

Checklist
---------

What
----

Added integration tests for the flink shell. I tried a bunch of different methods, but because go-prompt is interacting with `/dev/tty` directly, it is quite hard to submit commands programatically to the shell. T. work around this, we now create a file that will then be used as the input stream for go-prompt.

References
----------

Test & Review
-------------


.semaphore/semaphore.yml
```diff
@@ -25,6 +25,7 @@ blocks:
             - make generate-packaging-patch
             - diff -w -u <(git cat-file --filters HEAD:debian/patches/standard_build_layout.patch | awk "{if (NR>3) {print}}") <(cat debian/patches/standard_build_layout.patch | awk "{if (NR>3) {print}}")
             - make lint
+            - go install github.com/charmbracelet/vhs@latest
```
brianstrauch, line 28: :heart:

Guttz (approved):

test/fixtures/input/flink/flinkshell-commands
```diff
@@ -0,0 +1,11 @@
+use catalog default;
```
brianstrauch, line 1: Can integration tests test only one (or if necessary, two) statement at a time. From previous experience, once you start chaining multiple statements/commands in each test it gets harder to debug.

test/fixtures/output/flink/shell/shell.golden
```diff
@@ -0,0 +1,232 @@
...
+TEXT OUTPUT 
+========================================================================== 
+
+Welcome! 
+To exit, press Ctrl-Q or type "exit". 
+
+[Ctrl-Q] Quit [Ctrl-S] Toggle Smart Completion 
+ 
+Smart Completion is now disabled                                                                    
+> > > use catalog default;> use catalog default;> use catalog default;
```
brianstrauch, line 11: Looks like we're submitting multiple statements in parallel, without waiting for the previous one to execute?

test/fixtures/output/flink/shell/shell.golden
```diff
@@ -0,0 +1,232 @@
...
++------------------------+---------------------+
+|          Key           |        Value        |
++------------------------+---------------------+
+| client.service-account | <unset> (default)   |
+| sql.local-time-zone    | GMT+00:00 (default) |
++------------------------+---------------------+
+ > exit> exit> exit
+]2;
+
+TEXT + ANSI ESCAPE SEQUENCES OUTPUT
```
brianstrauch, line 118: Is this really that useful? Personally, I just find it really confusing. If we make a PR that changes the output colors, wouldn't it be easier to manually verify the output?

test/flink_test.go
```diff
@@ -120,3 +142,139 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+
+	var outputBuffer bytes.Buffer
+	cmd.Stdout = &outputBuffer
+	cmd.Stderr = &outputBuffer
+
+	err = cmd.Start()
+	require.NoError(s.T(), err, outputBuffer.String())
+
+	// wait for flink shell to be ready
+	time.Sleep(5 * time.Second)
```
brianstrauch, line 179: Is there a better way to verify this?

test/flink_test.go
```diff
@@ -120,3 +142,139 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
 		s.runIntegrationTest(test)
 	}
 }
+
+func (s *CLITestSuite) TestFlinkShell() {
```
brianstrauch, line 146: Is the idea to replace all existing `flink shell` integration tests with this style of testing?

test/fixtures/output/flink/shell/shell.golden
YannickPferr: No I think this just comes from @Guttz removing the ansi sequences to make the golden file more readable

test/fixtures/output/flink/shell/shell.golden
YannickPferr: hmm yeah I kinda agree, I was also considering disabling colors since it will make the golden file easier to read

test/flink_test.go
YannickPferr: hmm maybe.. Perhaps we could wait for the `Welcome` text? 🤔

test/flink_test.go
YannickPferr: I think it still makes sense to keep our existing more unit style tests since they are fast, easier to implement and enable us to test certain edge cases more easily. This is just to also have a test for actually starting up the binary and executing some commands without mocking everything.

Currently we're also missing a mock backend and just test local commands, I was planning to add a mock backend in the future to fake for example a `select` statement

test/fixtures/input/flink/flinkshell-commands
YannickPferr: yeah we can split it up a bit, I was thinking that it is faster this way, since we don't have to do the setup all over again.

test/fixtures/output/flink/shell/shell.golden
brianstrauch: Any way we can clean up the formatting here?

YannickPferr: I'll disable the colors and then there should be no need for removing the ansi sequences, which will in turn make the golden file more beautiful to look at 😄

test/flink_test.go
brianstrauch: > Currently we're also missing a mock backend and just test local commands, I was planning to add a mock backend in the future to fake for example a select statement

This would be great. The end goal would be for our existing integration tests and the Flink integration tests to use the same framework... although the DevProd team has been trying to switch us over to contract tests lately.

YannickPferr:
Also just noticed the race in CI, will fix this and then try to address the rest of the things tomorrow

test/flink_test.go
```diff
@@ -120,3 +152,214 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+	tests := []FlinkShellTest{
```
Guttz, line 157: nit: wdyt of listing tests from folder to exclude one step when adding test?


```
    files, err := os.ReadDir('input/flink/shell')
    if err != nil {
        fmt.Println("Error reading directory:", err)
        return
    }

    for _, file := range files {
         testName := file.Name()
         s.runFlinkShellTest({{inputFile: testName, goldenFile: testName + ".golden" }})
    }
```

test/flink_test.go
```diff
@@ -120,3 +152,214 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+		s.runFlinkShellTest(test)
+	}
+}
+
+func (s *CLITestSuite) runFlinkShellTest(flinkShellTest FlinkShellTest) {
+	s.T().Run(flinkShellTest.inputFile, func(t *testing.T) {
+		s.login(t)
+
+		// Create a file for go-prompt to use as the input stream
+		stdin, err := os.Create("input.txt")
```
Guttz, line 175:
```suggestion
		stdin, err := os.Create("flink_shell_input_stream.txt")
```
maybe reduce the chance that the CI starts to fail because someone created a input.txt file here? lol

test/flink_test.go
```diff
@@ -120,3 +152,214 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+		// Register stdout scanner
+		pipe, err := cmd.StdoutPipe()
+		require.NoError(t, err)
+		defer pipe.Close()
+		stdoutScanner := bufio.NewScanner(pipe)
+
+		// Start command
+		err = cmd.Start()
+		require.NoError(t, err)
+		go killCommandIfTimeoutExceeded(flinkShellTest.timeout, cmd)
```
Guttz, line 197: nice

test/flink_test.go
```diff
@@ -120,3 +152,214 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+	time.Sleep(timeout)
+	if err := cmd.Process.Kill(); err != nil {
+		fmt.Printf("failed to kill command %v/n", err)
+	}
+}
+
+func waitForFlinkShellToBeReady(scanner *bufio.Scanner, output *strings.Builder) {
+	for scanner.Scan() {
+		line := scanner.Text()
+		output.WriteString(line + "\n")
```
Guttz, line 293: do we have to execute this for every scan before checking if strings.Contains(line, "[Ctrl-Q]..?

Guttz (commented):
Beautiful 💯

test/flink_test.go
brianstrauch: Or better yet, we could just use a `[]string` instead of a file for the input? A good test will have a short script of statements so this should be fine.

test/flink_test.go
```diff
@@ -120,3 +152,214 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+					return err
+				}
+				break
+			}
+		}
+	}
+	return nil
+}
+
+func removeANSIEscapeSequences(input string) string {
```
brianstrauch, line 341:
```suggestion
func removeAnsiEscapeSequences(input string) string {
```

YannickPferr: renamed!

test/flink_test.go
YannickPferr: yeah sounds good, I wanted to make use of the input folder 😄 Using a string array now instead!

test/flink_test.go
```diff
@@ -155,11 +155,48 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+			name:       "TestUseDatabase",
+			goldenFile: "use-database.golden",
+			commands: []string{
+				"use db1;",
+				"set;",
+			},
+		},
+		{
+			name:       "TestSetSingleKey",
+			goldenFile: "set-key.golden",
```
brianstrauch, line 176: Another great improvement might be to let the golden files describe the test, instead of `name`. This'll make it much quicker to write new tests.
```suggestion
			goldenFile: "set-single-key.golden",
```

test/flink_test.go
```diff
@@ -120,3 +152,244 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+				"set;",
+			},
+		},
+		{
+			name:       "TestResetAllKeys",
+			goldenFile: "reset-all.golden",
+			commands: []string{
+				"set 'cli.a-key'='a value';",
+				"set 'cli.another-key'='another value';",
+				"reset;", "set;",
```
brianstrauch, line 197:
```suggestion
				"reset;",
				"set;",
```

test/flink_test.go
```diff
@@ -120,3 +152,244 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+	input = stripCursorUp.ReplaceAllString(input, "")
+
+	return strings.TrimSpace(input)
+}
+
+func submitStatement(stdin *os.File) error {
+	// We need to wait before and after writing a new line, so go-prompt receives the ENTER as an individual key press
+	time.Sleep(50 * time.Millisecond)
+	_, err := stdin.WriteString("\n")
+	time.Sleep(50 * time.Millisecond)
```
brianstrauch, line 393: I see a few `Sleep()` functions hidden throughout these tests. How long do these tests take end-to-end? Can we decrease this to 10ms? Can we scan for specific text to know when we can write a newline?

test/flink_test.go
```diff
@@ -120,3 +152,244 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+	return err
```
brianstrauch, line 394: If there was an error, wouldn't we want to return it right away instead of waiting 50ms?

test/flink_test.go
YannickPferr: hmm let me see, maybe we can wait for the `Statement successfully submitted.` text 🤔

test/flink_test.go
Guttz: Ah, I understand! You do this to print whatever we have as welcome message/headers when starting the CLI.

test/flink_test.go
YannickPferr: okay that works, added a helper that waits for a particular line to appear and then returns the output recorded up until that point

test/flink_test.go
YannickPferr: should be resolved as well with the latest commit

test/flink_test.go
YannickPferr: done

test/flink_test.go
YannickPferr: fixed

test/flink_test.go
YannickPferr: > How long do these tests take end-to-end?

oh and the tests take about 7s to run locally atm:
```
--- PASS: TestCLI (6.73s)
    --- PASS: TestCLI/TestFlinkShell (4.34s)
        --- PASS: TestCLI/TestFlinkShell/use-catalog (1.67s)
        --- PASS: TestCLI/TestFlinkShell/use-database (0.81s)
        --- PASS: TestCLI/TestFlinkShell/set-single-key (0.54s)
        --- PASS: TestCLI/TestFlinkShell/reset-single-key (0.61s)
        --- PASS: TestCLI/TestFlinkShell/reset-all-keys (0.73s)

```

brianstrauch: Huh, I wonder if that's including the time to build the binary? Let's see where else we can speed up these tests... 1s per integration test is not going to scale well

Guttz: I think we could create another ticket on how to increase the performance here, wdyt @brianstrauch? Yannick has been working on this over a week, and the whole ticket already required much more time than we planned

Guttz: Also, I don't think we're planning to add 100s of integration tests now. However, if we start adding more we'll prioritize increasing the performance first

YannickPferr: on CI the test takes about 2s, which seems reasonable to me 🤔  
image

test/flink_test.go
```diff
@@ -120,3 +151,233 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+		s.login(t)
+
+		// Create a file for go-prompt to use as the input stream
+		stdin, err := os.Create("flink_shell_input_stream.txt")
+		require.NoError(t, err, "error creating file")
+		defer cleanupInputFile(stdin)
+
+		envVars, err := setEnvVars(stdin.Name())
+		require.NoError(t, err, "error setting env vars")
+		defer cleanupEnvVars(envVars)
```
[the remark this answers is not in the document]
YannickPferr: yeah let me do that

test/flink_test.go
```diff
@@ -120,3 +151,233 @@ func (s *CLITestSuite) TestFlink_Autocomplete() {
...
+			}),
+		)
+		assert.NoError(t, snapshotConfig.SnapshotWithName(flinkShellTest.goldenFile, output.String()),
+			fmt.Sprintf("full output was %s", output.String()))
+	})
+}
+
+func (s *CLITestSuite) login(t *testing.T) {
+	loginString := fmt.Sprintf("login --url %s", s.TestBackend.GetCloudUrl())
+	env := []string{pauth.ConfluentCloudEmail + "=fake@user.com", pauth.ConfluentCloudPassword + "=pass1"}
```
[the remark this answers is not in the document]
YannickPferr: I could also hardcode them, but the runCommand func needs an array and does some things with it: https://github.com/confluentinc/cli/blob/6d0ca9fad8d79cf4664dd8ef9880adc2aeb9838f/test/cli_test.go#L225 

Maybe I could pass an empty array here?

test/flink_test.go
```diff
@@ -1,5 +1,37 @@
...
+const (
+	flinkShellInputStreamFile     = "flink_shell_input_stream.txt"
+	flinkShellFixtureOutputFolder = "test/fixtures/output/flink/shell"
+	timezoneEnvVar                = "TZ"
+)
+
+type flinkShellTest struct {
+	commands   []string
+	goldenFile string
+	timeout    time.Duration
```
brianstrauch, line 32: This looks unused. Can we leave it out until we need it?

test/flink_test.go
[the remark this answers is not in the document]
brianstrauch: Oops, I didn't see that. It looks like we're setting these environment variables twice, so let's change that to only be once.

brianstrauch (approved):

Guttz (approved):

PR #2662, opened 2024-02-28
fix: wait for statements properly

YannickPferr:
Checklist
---------

What
----

- We should use the "listening for execution errors" flow for all statements that are RUNNING but don't have results
- We should also still check the statement prefix in case sqlTraits are not available

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementWaitsForCompletedState
+++ b/pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementWaitsForCompletedState
@@ -1,0 +1,6 @@
+Statement successfully submitted.
+Waiting for statement to be ready. Statement phase is PENDING.
+Statement phase is RUNNING.
+Listening for execution errors. Press Enter to detach.
+Statement phase is COMPLETED.
+
--- a/pkg/flink/internal/controller/statement_controller_test.go
+++ b/pkg/flink/internal/controller/statement_controller_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stretchr/testify/suite"
 	"go.uber.org/mock/gomock"
 
-	flinkgatewayv1 "github.com/confluentinc/ccloud-sdk-go-v2/flink-gateway/v1"
 	"github.com/confluentinc/go-prompt"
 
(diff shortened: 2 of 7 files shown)


pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementWaitsForCompletedState
```diff
@@ -0,0 +1,6 @@
+Statement successfully submitted.
+Waiting for statement to be ready. Statement phase is PENDING.
+Statement phase is RUNNING.
+Listening for execution errors. Press Enter to detach.
```
Guttz, line 4: Should we update this to `Getting your statement ready. You can press enter to detach.`?

pkg/flink/internal/controller/statement_controller_test.go
```diff
@@ -231,17 +227,73 @@ func (s *StatementControllerTestSuite) TestExecuteStatementPrintsNoWarningForSta
 	cupaloy.SnapshotT(s.T(), stdout)
 }
 
+func (s *StatementControllerTestSuite) TestExecuteStatementWaitsForCompletedState() {
+	statementToExecute := "select 1;"
```
Guttz, line 231: maybe it makes sense to replicate these tests for insert into and create table? 

We know that the behavior should be exactly the same, but just as a guardrail to make sure the test fails if it changes for different types of queries

Guttz (dismissed):

pkg/flink/internal/controller/statement_controller_test.go
YannickPferr: I see what you mean but it won't really help, since the logic is not influenced by the statements text, but only by what's inside the backend response. So the test would look identical

YannickPferr: I can change the query text to `insert into` though, that should be less confusing

pkg/flink/internal/controller/.snapshots/TestStatementControllerTestSuite-TestExecuteStatementWaitsForCompletedState
YannickPferr: I would rather leave it as is for now, and then we can adjust it later wenn we might have to support this flow for select statements

PR #2667, opened 2024-03-01
KFS-1706 Add missing keyboard shortcuts to flink shell

MarTango:
Release Notes
-------------

New Features
- Flink shell supports new keybindings
  - Ctrl-D deletes characters and only closes if the buffer is empty
  - Alt-D deletes words
  - Alt-L lowercases words
  - Alt-U uppercases words

Checklist
---------
- [X] Leave this box unchecked if features are not yet available in production

What
----

https://github.com/confluentinc/cli/assets/13372091/e0629d3f-6802-4155-bc0d-199ec4818871

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/internal/controller/input_controller.go
+++ b/pkg/flink/internal/controller/input_controller.go
@@ -222,16 +222,34 @@ func getUnixBindings() []prompt.Option {
 	return []prompt.Option{
 		prompt.OptionAddASCIICodeBind(
 			prompt.ASCIICodeBind{
-				ASCIICode: []byte{0x1b, 0x62}, // Alt/Option + Arrow Left
+				ASCIICode: []byte{0x1b, 0x62}, // Alt/Option + Arrow Left (ESC + b)
 				Fn:        prompt.GoLeftWord,
 			},
 			prompt.ASCIICodeBind{
-				ASCIICode: []byte{0x1b, 0x66}, // Alt/Option + Arrow Right
+				ASCIICode: []byte{0x1b, 0x66}, // Alt/Option + Arrow Right (ESC + f)
 				Fn:        prompt.GoRightWord,
 			},
 			prompt.ASCIICodeBind{
 				ASCIICode: []byte{0x1b, 0x7F}, // Alt/Option + Backspace
 				Fn:        prompt.DeleteWord,
+			},
+			prompt.ASCIICodeBind{
+				ASCIICode: []byte{0x1b, 0x64}, // ForwardDeleteWord (ESC + d)
+				Fn: func(buf *prompt.Buffer) {
+					buf.Delete(buf.Document().FindEndOfCurrentWordWithSpace())
+				},
+			},
+			prompt.ASCIICodeBind{
+				ASCIICode: []byte{0x1b, 0x75}, // UpCaseWord (Alt/Option + u)
+				Fn: func(buf *prompt.Buffer) {
+					buf.InsertText(strings.ToUpper(buf.Document().GetWordAfterCursorWithSpace()), true, true)
+				},
+			},
+			prompt.ASCIICodeBind{
+				ASCIICode: []byte{0x1b, 0x6c}, // DownCaseWord (Alt/Option + l)
+				Fn: func(buf *prompt.Buffer) {
+					buf.InsertText(strings.ToLower(buf.Document().GetWordAfterCursorWithSpace()), true, true)
+				},
 			},
 		),
 	}
--- a/pkg/flink/internal/controller/input_controller.go
+++ b/pkg/flink/internal/controller/input_controller.go
@@ -188,16 +188,6 @@ func (c *InputController) getKeyBindings() []prompt.Option {
 	return append(
 		[]prompt.Option{
 			prompt.OptionAddKeyBind(prompt.KeyBind{
-				Key: prompt.ControlD,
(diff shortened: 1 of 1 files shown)


Guttz:
Nice! Which tool did you use for the recording?

pkg/flink/internal/controller/input_controller.go
```diff
@@ -190,7 +190,11 @@ func (c *InputController) getKeyBindings() []prompt.Option {
 			prompt.OptionAddKeyBind(prompt.KeyBind{
 				Key: prompt.ControlD,
 				Fn: func(b *prompt.Buffer) {
-					c.shouldExit = true
+					if b.Text() != "" {
+						b.Delete(1)
```
Guttz, line 194: lol, for some reason this is deleting two characters on native mac terminal

Guttz: https://github.com/confluentinc/cli/assets/11739405/caf7cb26-0f71-4a83-9c9e-f7d4f63154d7

Guttz: My first suspicion is that the keybind is somehow weirdly getting triggered twice

pkg/flink/internal/controller/input_controller.go
```diff
@@ -218,17 +222,35 @@ func getUnixBindings() []prompt.Option {
...
 				Fn:        prompt.DeleteWord,
 			},
+			prompt.ASCIICodeBind{
+				ASCIICode: []byte{0x1b, 0x64}, // ForwardDeleteWord (ESC + d)
+				Fn: func(buf *prompt.Buffer) {
+					buf.Delete(buf.Document().FindEndOfCurrentWordWithSpace())
+				},
+			},
+			prompt.ASCIICodeBind{
+				ASCIICode: []byte{0x1b, 0x75}, // UpCaseWord (Alt/Option + u)
```
Guttz, line 243: For this, the Opt as Metakey has to be enabled, right? We could add this to the comment?

Do you maybe know what's the behavior on windows? (if the user has to enable something for this to work)

pkg/flink/internal/controller/input_controller.go
```diff
@@ -218,17 +222,35 @@ func getUnixBindings() []prompt.Option {
 	return []prompt.Option{
 		prompt.OptionAddASCIICodeBind(
 			prompt.ASCIICodeBind{
-				ASCIICode: []byte{0x1b, 0x62}, // Alt/Option + Arrow Left
+				ASCIICode: []byte{0x1b, 0x62}, // Alt/Option + Arrow Left (ESC + b)
```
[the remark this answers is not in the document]
MarTango: I can't find any official documentation, but from my experience, typically (although this no longer seems to be the default in terminals) Alt + key being pressed in terminals gets mapped to the input `ESC, key` ( `^[` / `0x1b` is the [`ESC` character](https://en.wikipedia.org/wiki/Escape_character#:~:text=The%20ASCII%20%22escape%22%20character%20(,control%20sequence%20or%20escape%20sequence.)).

Somehow with default terminal settings `Alt + Left` actually results in the same input as my `Alt + b` (`^[b`). Maybe because arrow keys were introduced to the keyboard after `esc` and `b` keys already existed? 🤷 . iterm defaults do a completely weird thing where they completely remap it to another key sequence

pkg/flink/internal/controller/input_controller.go
MarTango: Ohh. We don't need to have this keybinding at all since it's already covered by emacsKeyBindings. And somehow it still has the behaviour of ending the session when the prompt is empty (but I don't currently see the function that handles that)

MarTango: ah nvm found it - it's in prompt.go

pkg/flink/internal/controller/input_controller.go
MarTango: I rewrote the comment to use "ESC" instead of "Alt". I think that's the most platform-generic way of specifying the key.

> Do you maybe know what's the behavior on windows? (if the user has to enable something for this to work)

I would assume that none of the shortcuts that use Alt work by default on windows. I'm not sure though and can't really test it.

pkg/flink/internal/controller/input_controller.go
MarTango: I don't know why it was working for me though.. but it seems to be working consistently now

Guttz: it does!

pkg/flink/internal/controller/input_controller.go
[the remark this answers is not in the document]
Guttz: I see... As a non-expert in that area, I think this is hard to understand from just reading the comment

pkg/flink/internal/controller/input_controller.go
```diff
@@ -218,17 +212,35 @@ func getUnixBindings() []prompt.Option {
...
```
Guttz, line 215:
```suggestion
				ASCIICode: []byte{0x1b, 0x62}, // Alt/Option + Arrow Left (legacy ESC + b)
```
Do you think adding this makes sense? And then updating all comments to follow the same format

Guttz (approved):

Guttz (approved):

PR #2677, opened 2024-03-05
[CLI-3075] Prevent panic in `confluent context update`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix a panic in `confluent context update --kafka-cluster`

Checklist
---------

What
----
The context update command sends a `nil` client to `FindCluster`, which causes a panic if the cluster isn't found in the user's Kafka Cluster context.

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/kafka/cluster.go
+++ b/pkg/kafka/cluster.go
@@ -51,7 +51,7 @@ func FindCluster(client *ccloudv2.Client, ctx *config.Context, clusterId string)
 	}
 
 	if client == nil {
-		return nil, fmt.Errorf(errors.KafkaClusterNotFoundErrorMsg, clusterId)
+		return nil, errors.NewErrorWithSuggestions(fmt.Sprintf(errors.KafkaClusterNotFoundErrorMsg, clusterId), "You can set the active Kafka cluster with `confluent kafka cluster use`.")
 	}
 
 	cluster, httpResp, err := client.DescribeKafkaCluster(clusterId, environmentId)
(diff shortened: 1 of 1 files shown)


pkg/kafka/cluster.go
```diff
@@ -50,6 +50,10 @@ func FindCluster(client *ccloudv2.Client, ctx *config.Context, clusterId string)
 		return nil, err
 	}
 
+	if client == nil {
+		return nil, fmt.Errorf(errors.KafkaClusterNotFoundErrorMsg, clusterId)
```
brianstrauch, line 54: The panic report says `Flags: [name kafka-cluster]`, so it seems wrong to say "kafka cluster not found" when it's being passed. How would a user fix this if they ran into this error?

sgagniere: The error says that the cluster ID they passed in wasn't found.

The fix is odd. The command works for any cluster that they've previously set with `confluent kafka cluster use`, so the fix is to run that command.

We could probably initialize a client if the user is logged in instead of just returning an error, but if the user isn't logged in then there's no fixing this.

brianstrauch: > The command works for any cluster that they've previously set with confluent kafka cluster use, so the fix is to run that command.

Let's add this as a suggestion?

brianstrauch (approved):

brianstrauch:
Should the release notes mention `--kafka-cluster`?

PR #2683, opened 2024-03-11
SCP-1879-update-confluent-cli-to-create-azure-plugin-plugin-and-list-with-cloud-filter

NicoleYWDu:
[SCP-1879] update confluent cli to create azure plugin plugin and list with cloud filter

related API change is here: https://confluentinc.atlassian.net/browse/SCP-1871

Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- create custom plugin --cloud [aws, gcp, azure] (optional) default is aws
- list custom plugin --cloud [aws, gcp, azure] (optional) default is aws

  - The default value:aws is set in [cc-connect-plugin-service](https://github.com/confluentinc/cc-connect-plugin-service)
  - Related implementation pr: 
https://confluentinc.atlassian.net/browse/SCP-109
https://confluentinc.atlassian.net/browse/SCP-1872

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----

References
----------

Test & Review
-------------


internal/connect/command_custom_plugin_create.go
```diff
@@ -66,6 +68,10 @@ func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []stri
 	if err != nil {
 		return err
 	}
+	cloud, err := cmd.Flags().GetString("cloud")
+	if err != nil {
```
byang-confluent, line 72: this introduces breaking change, we should avoid. 

cloud is optional in api level. we should carried over same behavior  to the cli as well. aka, allow empty cloud attribute when create plugin.

brianstrauch: It also looks like it's optional here, no? Not sure what the breaking change is.

internal/connect/command_custom_plugin_create.go
```diff
@@ -83,9 +89,11 @@ func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []stri
 	if extension != "zip" && extension != "jar" {
 		return fmt.Errorf(`only file extensions ".jar" and ".zip" are allowed`)
 	}
+	cloudProvider := strings.ToLower(cloud)
 
 	request := *connectcustompluginv1.NewConnectV1PresignedUrlRequest()
 	request.SetContentFormat(extension)
+	request.SetCloud(cloudProvider)
```
brianstrauch, line 96: Should we be setting this value if it is `""`?

test/test-server/connect_handler.go
```diff
@@ -298,6 +303,7 @@ func handleCustomPluginUploadUrl(t *testing.T) http.HandlerFunc {
 		if r.Method == http.MethodPost {
 			uploadUrl := connectcustompluginv1.ConnectV1PresignedUrl{
 				ContentFormat: PtrString("ZIP"),
+				Cloud:         PtrString("aws"),
 				UploadId:      PtrString("e53bb2e8-8de3-49fa-9fb1-4e3fd9a16b66"),
```
brianstrauch, line 307: Can we use the built-in function from the SDK, instead of defining our own function?
```suggestion
				ContentFormat: connectcustompluginv1.PtrString("ZIP"),
				Cloud:         connectcustompluginv1.PtrString("aws"),
				UploadId:      connectcustompluginv1.PtrString("e53bb2e8-8de3-49fa-9fb1-4e3fd9a16b66"),
```

internal/connect/command_custom_plugin_create.go
byang-confluent: yes, it is optional when create plugin too.

NicoleYWDu: We want the cloud input be optional. My understanding is `cobra.CheckErr(cmd.MarkFlagRequired("connector-type"))` will check the flag is required or not. So for cloud we don't check it.  This looks like it's optional here

internal/connect/command_custom_plugin_create.go
NicoleYWDu: Since we will check here: https://github.com/confluentinc/cc-connect-plugin-service/blob/master/api/v2/middleware.go#L42. It will be ok if the input is ""

internal/connect/command_custom_plugin_list.go
```diff
@@ -22,28 +23,35 @@ func (c *customPluginCommand) newListCommand() *cobra.Command {
 		Example: examples.BuildExampleString(
 			examples.Example{
 				Text: "List custom connector plugins in the org",
-				Code: "confluent connect custom-plugin list",
+				Code: "confluent connect custom-plugin list --cloud aws",
```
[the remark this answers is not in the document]
NicoleYWDu: Right now the doc for list is like this: https://github.com/confluentinc/cli/blob/6600807cfbbf587e5a61ea4206c7c3594645864b/test/fixtures/output/connect/custom-plugin/list-help.golden
Would you mean to have `OPTIONAL: Filter plugins by cloud provider.` there?
For create part, https://github.com/confluentinc/cli/blob/6600807cfbbf587e5a61ea4206c7c3594645864b/test/fixtures/output/connect/custom-plugin/create-help.golden. For the required part we have format like this ` REQUIRED: ZIP/JAR custom plugin file.`  And for option part just remove REQUIRED: .

internal/connect/command_custom_plugin_create.go
brianstrauch: Great, can we combine the three lines above into one? I know this isn't part of your PR, but we don't recommend using `NewConnectV1PresignedUrlRequest()`. Instead, you should just initialize `request` as a single struct.

test/test-server/connect_handler.go
```diff
@@ -297,9 +302,10 @@ func handleCustomPluginUploadUrl(t *testing.T) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodPost {
 			uploadUrl := connectcustompluginv1.ConnectV1PresignedUrl{
-				ContentFormat: PtrString("ZIP"),
-				UploadId:      PtrString("e53bb2e8-8de3-49fa-9fb1-4e3fd9a16b66"),
-				UploadUrl:     PtrString(fmt.Sprintf("http://%s/connect/v1/dummy-presigned-url", TestV2CloudUrl.Host)),
+				ContentFormat: connectcustompluginv1.PtrString("ZIP"),
+				Cloud:         connectcustompluginv1.PtrString("aws"),
+				UploadId:      connectcustompluginv1.PtrString("e53bb2e8-8de3-49fa-9fb1-4e3fd9a16b66"),
+				UploadUrl:     connectcustompluginv1.PtrString(fmt.Sprintf("http://%s/connect/v1/dummy-presigned-url", TestV2CloudUrl.Host)),
```
brianstrauch, line 308: The URL host is included in `TestV2CloudUrl`
```suggestion
				UploadUrl:     connectcustompluginv1.PtrString(fmt.Sprintf("%s/connect/v1/dummy-presigned-url", TestV2CloudUrl.String())),
```

brianstrauch: I think the following should work as well
```suggestion
				UploadUrl:     connectcustompluginv1.PtrString(fmt.Sprintf("%v/connect/v1/dummy-presigned-url", TestV2CloudUrl)),
```

internal/connect/command_custom_plugin_create.go
```diff
@@ -91,9 +91,10 @@ func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []stri
 	}
 	cloudProvider := strings.ToLower(cloud)
```
brianstrauch, line 92:
```suggestion
	cloud = strings.ToLower(cloud)
```

internal/connect/command_custom_plugin_list.go
```diff
@@ -22,28 +23,35 @@ func (c *customPluginCommand) newListCommand() *cobra.Command {
...
 			},
 		),
 	}
 
+	cmd.Flags().String("cloud", "", "Filter plugins by cloud provider.")
```
brianstrauch, line 31: Can you add autocompletion? There is already a function you might be interested in using called `pcmd.AddCloudFlag()`

internal/connect/command_custom_plugin_create.go
```diff
@@ -66,6 +68,10 @@ func (c *customPluginCommand) createCustomPlugin(cmd *cobra.Command, args []stri
 	if err != nil {
 		return err
 	}
+	cloud, err := cmd.Flags().GetString("cloud")
+	if err != nil {
+		return err
+	}
```
brianstrauch, line 74:
```suggestion

	cloud, err := cmd.Flags().GetString("cloud")
	if err != nil {
		return err
	}
	
```

internal/connect/command_custom_plugin_create.go
```diff
@@ -39,7 +39,7 @@ func (c *customPluginCommand) newCreateCommand() *cobra.Command {
 	cmd.Flags().String("plugin-file", "", "ZIP/JAR custom plugin file.")
 	cmd.Flags().String("connector-class", "", "Connector class of custom plugin.")
 	cmd.Flags().String("connector-type", "", "Connector type of custom plugin.")
-	cmd.Flags().String("cloud", "", "Cloud provider of custom plugin.")
+	cmd.Flags().String("cloud", "", "Cloud provider of custom plugin. Supported value [\"aws\",\"azure\"] (default \"aws\").")
```
brianstrauch, line 42: Use `pcmd.AddCloudFlag()` here, too?

NicoleYWDu: We do not support gcp right now, so might need to create a new function for this.
May I ask how can I get the completion suggestion while using this function?
I guess it is when you start typing the command in the terminal: `connect custom-plugin list --cloud` and hit the TAB key. it will show the suggestion. 
I make build use the new cli to test,  but not get the sugguestion.

NicoleYWDu: For create cmd, we want to let user know we have aws as the default value.
For list cmd, the default value should be "".
Would you think we need to have two autocompletion funcs?

brianstrauch: > I guess it is when you start typing the command in the terminal and hit the TAB key

Yep, that's correct. Make sure you follow the steps in `confluent completion -h` too.

> Would you think we need to have two autocompletion funcs?

Sounds like it, since they both autocomplete a different set of values.

brianstrauch: Also- please add an integration test for the autocompletion function in the `test/` directory, since this autocompletion is non-trivial

internal/connect/command_custom_plugin_create.go
```diff
@@ -39,10 +39,10 @@ func (c *customPluginCommand) newCreateCommand() *cobra.Command {
 	cmd.Flags().String("plugin-file", "", "ZIP/JAR custom plugin file.")
 	cmd.Flags().String("connector-class", "", "Connector class of custom plugin.")
 	cmd.Flags().String("connector-type", "", "Connector type of custom plugin.")
-	cmd.Flags().String("cloud", "", "Cloud provider of custom plugin. Supported value [\"aws\",\"azure\"] (default \"aws\").")
 	cmd.Flags().String("description", "", "Description of custom plugin.")
 	cmd.Flags().String("documentation-link", "", "Document link of custom plugin.")
 	cmd.Flags().StringSlice("sensitive-properties", nil, "A comma-separated list of sensitive property names.")
+	pcmd.AddPluginCloudFlag(cmd)
```
brianstrauch, line 45: Will this flag be used for connect and flink?

pkg/cmd/flags.go
```diff
@@ -94,6 +94,11 @@ func AddCloudFlag(cmd *cobra.Command) {
 	RegisterFlagCompletionFunc(cmd, "cloud", func(_ *cobra.Command, _ []string) []string { return kafka.Clouds })
 }
 
+func AddPluginCloudFlag(cmd *cobra.Command) {
+	cmd.Flags().String("cloud", "", fmt.Sprintf("Cloud provider of custom plugin. Supported value %s. (default \"aws\").", utils.ArrayToCommaDelimitedString([]string{"aws", "azure"}, "or")))
```
brianstrauch, line 98:
```suggestion
	cmd.Flags().String("cloud", "aws", fmt.Sprintf("Set cloud provider of custom plugin as %s.", utils.ArrayToCommaDelimitedString([]string{"aws", "azure"}, "or")))
```

brianstrauch: Nit: Define the slice as a variable so we only need to edit the list of accepted cloud providers once

brianstrauch: Just curious, what's the reason for defaulting to AWS instead of making this a required field? Don't we try to treat all cloud providers equally as a company?

NicoleYWDu: while we create api change and talk to api team, they recommend to use as optional. If use as optional this will not be a breaking change. 
Our team is working on supporting azure Q1 (not announce now but at the end of Q1), and will start to work on supporting gcp in Q2. So we  not want to give user the gcp option now.

internal/connect/command_custom_plugin_create.go
NicoleYWDu: We do not support flink artifact cloud attribute now. 
This is more like a temporary flag, cuz we will also support gcp in the future.

pkg/kafka/utils.go
```diff
@@ -1,7 +1,8 @@
 package kafka
 
 var (
-	Clouds         = []string{"aws", "azure", "gcp"}
-	Availabilities = []string{"single-zone", "multi-zone"}
-	Types          = []string{"basic", "standard", "enterprise", "dedicated"}
+	Clouds            = []string{"aws", "azure", "gcp"}
+	Availabilities    = []string{"single-zone", "multi-zone"}
+	Types             = []string{"basic", "standard", "enterprise", "dedicated"}
+	BYOCSupportClouds = []string{"aws", "azure"}
```
brianstrauch, line 7:
```suggestion
	ByocSupportedClouds = []string{"aws", "azure"}
```

brianstrauch: What does BYOC stand for? If "C" stands for connect, maybe this belongs in the "connect" package instead of the "kafka" package?

internal/connect/utils.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
+var (
+	BYOCSupportClouds = []string{"aws", "azure"}
```
brianstrauch: Please define this in `pkg` instead of `internal`, this is a standard practice in Go to avoid import cycles

internal/connect/command_custom_plugin_create.go
brianstrauch: Let's define this function in the current package then. No need to define it in `pkg/cmd` if it's not going to be used globally.

pkg/utils/utils.go
```diff
@@ -183,3 +183,49 @@ func UploadFile(url, filePath string, formFields map[string]any) error {
...
 }
+
+func UploadFileToAzureBlob(url, filePath, contentFormat string) error {
+	fileInfo, err := os.Stat(filePath)
+	if err != nil {
+		return err
+	}
+
+	if fileInfo.Size() > maxFileSize {
+		return fmt.Errorf("file size %d exceeds the 1GB limit", fileInfo.Size())
```
brianstrauch, line 194: There is a lot of duplication between this function and `UploadFile()` above, and it's clear that you just copied and pasted it. Can we de-duplicate these functions in any way or at the very least define this shared string as a constant so it only needs to be updated in one place if we decide to change the file size limit in the future?

NicoleYWDu: The header and body input are all different for these two funcs. It is hard to decouple these two. In the future I guess we will have func to upload to gcp, which might have a diff request body format and header content too.

internal/connect/command_custom_plugin_create.go
```diff
@@ -42,7 +43,8 @@ func (c *customPluginCommand) newCreateCommand() *cobra.Command {
 	cmd.Flags().String("description", "", "Description of custom plugin.")
 	cmd.Flags().String("documentation-link", "", "Document link of custom plugin.")
 	cmd.Flags().StringSlice("sensitive-properties", nil, "A comma-separated list of sensitive property names.")
-	pcmd.AddPluginCloudFlag(cmd)
+	cmd.Flags().String("cloud", "aws", fmt.Sprintf("Set cloud provider of custom plugin as %s.", utils.ArrayToCommaDelimitedString(ccloudv2.ByocSupportClouds, "or")))
+	pcmd.RegisterFlagCompletionFunc(cmd, "cloud", func(_ *cobra.Command, _ []string) []string { return ccloudv2.ByocSupportClouds })
```
brianstrauch, line 47: Please define this in a function, i.e. `c.addCloudFlag()`

pkg/utils/utils.go
```diff
@@ -16,7 +16,10 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
-const maxFileSize = 1024 * 1024 * 1024 // 1GB
+const (
+	maxFileSize     = 1024 * 1024 * 1024 // 1GB
+	maxFileSizeInGB = float64(maxFileSize) / float64(1024*1024*1024)
```
brianstrauch, line 21: This seems a little silly
```suggestion
	maxFileSizeGb = 1
```

internal/connect/command_custom_plugin_create.go
```diff
@@ -43,8 +43,9 @@ func (c *customPluginCommand) newCreateCommand() *cobra.Command {
 	cmd.Flags().String("description", "", "Description of custom plugin.")
 	cmd.Flags().String("documentation-link", "", "Document link of custom plugin.")
 	cmd.Flags().StringSlice("sensitive-properties", nil, "A comma-separated list of sensitive property names.")
-	cmd.Flags().String("cloud", "aws", fmt.Sprintf("Set cloud provider of custom plugin as %s.", utils.ArrayToCommaDelimitedString(ccloudv2.ByocSupportClouds, "or")))
-	pcmd.RegisterFlagCompletionFunc(cmd, "cloud", func(_ *cobra.Command, _ []string) []string { return ccloudv2.ByocSupportClouds })
+	//cmd.Flags().String("cloud", "aws", fmt.Sprintf("Set cloud provider of custom plugin as %s.", utils.ArrayToCommaDelimitedString(ccloudv2.ByocSupportClouds, "or")))
+	//pcmd.RegisterFlagCompletionFunc(cmd, "cloud", func(_ *cobra.Command, _ []string) []string { return ccloudv2.ByocSupportClouds })
```
brianstrauch, line 47:
```suggestion
```

brianstrauch (approved):

brianstrauch (approved):

PR #2697, opened 2024-03-26
Feat/kfs 987 flinkcc 456 use upsert columns

YannickPferr:
Release Notes
-------------

New Features
- `confluent flink shell` uses the `upsertColumns` property now when its available, which makes results in table mode more stable

Checklist
---------

What
----

When upsertColumns are availabe, we can find previously inserted rows by their upsert keys and update them in place instead of deleting the whole row when a -U comes in and appending at the end. 

References
----------

Test & Review
-------------

https://github.com/confluentinc/cli/assets/37211050/2d376876-7410-4c0c-9e42-19cdcc40c730

### DIFF
--- a/pkg/flink/types/materialized_results.go
+++ b/pkg/flink/types/materialized_results.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/confluentinc/cli/v3/pkg/flink/internal/utils"
+	"github.com/confluentinc/cli/v3/pkg/log"
 )
 
 type MaterializedStatementResultsIterator struct {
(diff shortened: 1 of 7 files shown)


pkg/flink/types/materialized_results.go
```diff
@@ -158,6 +154,48 @@ func (s *MaterializedStatementResults) Append(rows ...StatementResultRow) bool {
 	return allValuesInserted
 }
 
+func (s *MaterializedStatementResults) processRowWithUpsertColumns(row StatementResultRow, rowKey string) {
```
Guttz, line 157: Can you refactor this so it's easier to understand where each flink operation is being handled? (+I, +U, -U, -D). Either some ifs/switch case or simply some comments

YannickPferr: done

Guttz (approved):

Guttz (approved):

PR #2699, opened 2024-03-28
[skip secret scan] [TRAFFIC-8850] Confluent CLI Support for Networking - Network Link Commands

ay-zhang:
Release Notes
-------------

New Features

- Add `confluent network link` commands for managing network link resources in Confluent Cloud
- Add `confluent network region` command to list cloud provider regions for networking

Checklist
---------

What
----

The PR introduces the `confluent network link` commands for managing network link resources in Confluent Cloud. Note that each sub-command has been reviewed in their corresponding PR.

This includes CRUDL operations for the following resources:
- network link service
- network link endpoint
- network link service association (list and describe only)

References
----------

Test & Review
-------------

- Each sub-command has been tested in their corresponding PR.


internal/network/command_network_link.go
```diff
@@ -0,0 +1,18 @@
+package network
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (c *command) newNetworkLinkCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "network-link",
```
brianstrauch, line 9: I think we already discussed this, but do you remember why we chose `confluent network-link` instead of `confluent network`?

brianstrauch: We may want to consider doing the latter. Recently we chose to introduce `confluent kafka consumer group` instead of `confluent kafka consumer-group`, so maybe we should follow the same pattern here.

ay-zhang: This decision predates my time but I think `confluent network link` should be fine, @kabraham13 thoughts on this naming?

kabraham13: Old convo was here - https://confluentinc.atlassian.net/wiki/spaces/PM/pages/2987394300/1-pager+--+Network+CLI?focusedCommentId=3137868981

I am okay with `confluent network link ...` as it was my original choice

ay-zhang: Perfect, will update these commands. I assume this applies to `private-link` and `transit-gateway-attachment` as well?

kabraham13: @ay-zhang we should keep those as is. `private-link` is a specific thing/construct so i think its okay. Same with `transit-gateway-attachment` its a very specific thing

ay-zhang:
Secret scan is failing due to vendored dependencies from [use customized sdk to make build](https://github.com/confluentinc/cli/pull/2699/commits/3df31dab3af9527012634762bafe080ad5a750c6) which was reverted.

ay-zhang:
@brianstrauch It seems like Flink Shell tests have started failing due to incorrect local-time-zone value, is this expected?

internal/network/command.go
```diff
@@ -321,8 +325,8 @@ func autocompleteNetworks(client *ccloudv2.Client, environmentId string) []strin
 	return suggestions
 }
 
-func getNetworks(client *ccloudv2.Client, environmentId string) ([]networkingv1.NetworkingV1Network, error) {
-	return client.ListNetworks(environmentId)
+func getNetworks(client *ccloudv2.Client, environmentId string, name, cloud, region, cidr, phase, connectionType []string) ([]networkingv1.NetworkingV1Network, error) {
+	return client.ListNetworks(environmentId, name, cloud, region, cidr, phase, connectionType)
```
brianstrauch, line 329: What's the reason for writing a wrapper function with the exact same signature?

brianstrauch (approved):

PR #2703, opened 2024-04-03
[CLI-3048] Remove login requirement for `confluent connect plugin install`

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- `confluent connect plugin install` no longer requires users to log in

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Not all on-premises users have MDS set up. This PR removes the login requirement so all on-premises users can use this command.

References
----------

Test & Review
-------------
Manual testing: checked that this command can be used when logged out and when logged in to on-prem.

### DIFF
--- a/internal/connect/command_plugin_install.go
+++ b/internal/connect/command_plugin_install.go
@@ -62,9 +62,7 @@ func newInstallCommand(prerunner pcmd.PreRunner) *cobra.Command {
 
 	cobra.CheckErr(cmd.MarkFlagDirname("plugin-directory"))
 
-	c := &pluginInstallCommand{
-		CLICommand: pcmd.NewAnonymousCLICommand(cmd, prerunner),
-	}
+	c := &pluginInstallCommand{pcmd.NewAnonymousCLICommand(cmd, prerunner)}
 	cmd.RunE = c.install
 
 	return cmd
--- a/internal/connect/command_plugin_install.go
+++ b/internal/connect/command_plugin_install.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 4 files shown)


internal/connect/command_plugin_install.go
```diff
@@ -58,10 +62,19 @@ func (c *pluginCommand) newInstallCommand() *cobra.Command {
 
 	cobra.CheckErr(cmd.MarkFlagDirname("plugin-directory"))
 
+	c := &pluginInstallCommand{
+		CLICommand: pcmd.NewAnonymousCLICommand(cmd, prerunner),
+	}
```
brianstrauch, line 67:
```suggestion
	c := &pluginInstallCommand{pcmd.NewAnonymousCLICommand(cmd, prerunner)}
```

internal/connect/command_plugin_install.go
```diff
@@ -58,10 +62,19 @@ func (c *pluginCommand) newInstallCommand() *cobra.Command {
...
+	cmd.RunE = c.install
+
 	return cmd
 }
 
-func (c *pluginCommand) install(cmd *cobra.Command, args []string) error {
+func (c *pluginInstallCommand) install(cmd *cobra.Command, args []string) error {
+	if runtime.GOOS == "windows" {
+		return fmt.Errorf("this command is not available on Windows")
+	}
```
brianstrauch, line 76: I guess this is better than hiding the command outright on Windows. Any other commands that could take advantage of this pattern? The new `confluent local` commands maybe?

brianstrauch (approved):
Thanks so much!

internal/connect/command_plugin_install.go
sgagniere: Hmm, maybe we should.

I don't think there's anything actually preventing windows users from using the local commands, but it is kind of awkward to allow users to demo something that isn't supported on their platform.

brianstrauch (approved):

internal/connect/command_plugin_install.go
brianstrauch: Oh wait, it should be supported since we're using Docker. Never mind!

PR #2712, opened 2024-04-15
feat: add function to parse app options from slices

YannickPferr:
Checklist
---------

What
----

Adds a function to parse app options from slices using reflections. This will enable us to overwrite the ApplicationOptions with special config flags that will only be available behind a feature flag. We can then use this to run the CLI against a local k8s instance of the flink system

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/types/application_options.go
+++ b/pkg/flink/types/application_options.go
@@ -1,7 +1,13 @@
 package types
 
 import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+
 	"github.com/confluentinc/cli/v3/pkg/config"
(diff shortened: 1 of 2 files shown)


pkg/flink/types/application_options.go
```diff
@@ -15,9 +21,98 @@ type ApplicationOptions struct {
...
+				configKey,
+			)
+		}
+	}
+
+	return appOptions, nil
+}
+
+func getAvailableConfigFields() []string {
+	appOptionsType := reflect.TypeOf(ApplicationOptions{})
```
Guttz, line 84: hmm nice

Guttz (approved):

PR #2713, opened 2024-04-15
[KFS-1863] add local mode to flink shell

YannickPferr:
Checklist
---------

What
----

This adds two flags behind a feature flag which will only be enabled for internal orgs. With these flags we can overwrite the ApplicationOptions the shell uses. This enables us to run the flink shell with a local instance of the Flink system.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/flink/command_shell.go
+++ b/internal/flink/command_shell.go
@@ -73,13 +73,19 @@ func (c *command) authenticated(authenticated func(*cobra.Command, []string) err
 
 func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Command) error {
 	if featureflags.Manager.BoolVariation("cli.flink.internal", c.Context, config.CliLaunchDarklyClient, true, false) {
+		// get config keys and values from flags
 		configKeys, err := cmd.Flags().GetStringSlice("config-key")
 		if err != nil {
 			return err
 		}
-		// if config-keys were passed, we should enter local mode
-		if len(configKeys) > 0 {
-			return c.startWithLocalMode(cmd)
+		configValues, err := cmd.Flags().GetStringSlice("config-value")
+		if err != nil {
+			return err
+		}
+
+		// if configs were passed, we should enter local mode
+		if len(configKeys) > 0 && len(configValues) > 0 {
+			return c.startWithLocalMode(configKeys, configValues)
 		}
 	}
 
@@ -171,17 +177,7 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
 	return client.StartApp(flinkGatewayClient, c.authenticated(prerunner.Authenticated(c.AuthenticatedCLICommand), cmd, jwtValidator), opts, reportUsage(cmd, c.Config, unsafeTrace))
 }
 
-func (c *command) startWithLocalMode(cmd *cobra.Command) error {
-	// get config keys and values from flags
-	configKeys, err := cmd.Flags().GetStringSlice("config-key")
-	if err != nil {
-		return err
-	}
-	configValues, err := cmd.Flags().GetStringSlice("config-value")
-	if err != nil {
-		return err
-	}
-
+func (c *command) startWithLocalMode(configKeys, configValues []string) error {
 	// parse app options from given flags
 	appOptions, err := types.ParseApplicationOptionsFromSlices(configKeys, configValues)
 	if err != nil {
--- a/internal/flink/command_shell.go
+++ b/internal/flink/command_shell.go
@@ -160,7 +160,7 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
 		ComputePoolId:    computePool,
 		ServiceAccountId: serviceAccount,
 		Verbose:          verbose > 0,
-		LSPBaseURL:       lspBaseUrl,
+		LSPBaseUrl:       lspBaseUrl,
 	}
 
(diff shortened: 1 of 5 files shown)


internal/flink/command_shell.go
```diff
@@ -36,6 +33,11 @@ func (c *command) newShellCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 
+	if featureflags.Manager.BoolVariation("cli.flink.internal", c.Context, config.CliLaunchDarklyClient, true, false) {
+		cmd.Flags().StringSlice("config-key", []string{}, "app option keys for local mode")
+		cmd.Flags().StringSlice("config-value", []string{}, "app option values for local mode")
```
brianstrauch, line 38: https://github.com/spf13/pflag/issues/129 :(

internal/flink/command_shell.go
```diff
@@ -70,14 +72,10 @@ func (c *command) authenticated(authenticated func(*cobra.Command, []string) err
...
-			func() error { return nil },
-			types.ApplicationOptions{
-				Context:   c.Context,
-				UserAgent: c.Version.UserAgent,
-			}, func() {})
+	// if config-keys were passed, we should enter local mode
+	configKeys, _ := cmd.Flags().GetStringSlice("config-key")
+	if len(configKeys) > 0 {
+		return c.startWithLocalMode(cmd)
 	}
```
brianstrauch, line 79: Instead of ignoring this error, consider wrapping this block with the feature flag?

internal/flink/command_shell.go
```diff
@@ -162,12 +160,40 @@ func (c *command) startFlinkSqlClient(prerunner pcmd.PreRunner, cmd *cobra.Comma
 		ComputePoolId:    computePool,
 		ServiceAccountId: serviceAccount,
 		Verbose:          verbose > 0,
-		LSPBaseUrl:       lspBaseUrl,
+		LSPBaseURL:       lspBaseUrl,
```
brianstrauch, line 163:
```suggestion
		LspBaseUrl:       lspBaseUrl,
```

pkg/flink/app/application.go
```diff
@@ -63,7 +63,7 @@ func StartApp(gatewayClient ccloudv2.GatewayClientInterface, tokenRefreshFunc fu
 	resultFetcher := results.NewResultFetcher(dataStore)
 
 	// Instantiate lsp
```
brianstrauch, line 65:
```suggestion
	// Instantiate LSP
```

pkg/flink/app/application.go
```diff
@@ -63,7 +63,7 @@ func StartApp(gatewayClient ccloudv2.GatewayClientInterface, tokenRefreshFunc fu
...
-	lspClient := lsp.NewLSPClientWS(getAuthToken, appOptions.GetLSPBaseUrl(), appOptions.GetOrganizationId(), appOptions.GetEnvironmentId())
+	lspClient := lsp.NewLSPClientWS(getAuthToken, appOptions.GetLSPBaseURL(), appOptions.GetOrganizationId(), appOptions.GetEnvironmentId())
```
brianstrauch, line 66: Functions of a package do not generally contain the name of the package... it's a bit redundant. Example: `http.NewClient()` instead of `http.NewHttpClient()`
```suggestion
	lspClient := lsp.NewClientWs(getAuthToken, appOptions.GetLSPBaseURL(), appOptions.GetOrganizationId(), appOptions.GetEnvironmentId())
```

internal/flink/command_shell.go
YannickPferr: yeah was looking for something like that, would be way simpler 😢

brianstrauch: The linter should catch this, but I'm not sure if it works for flags behind feature flags...
```suggestion
		cmd.Flags().StringSlice("config-key", []string{}, "App option keys for local mode.")
		cmd.Flags().StringSlice("config-value", []string{}, "App option values for local mode.")
```

internal/flink/command_shell.go
```diff
@@ -70,14 +72,15 @@ func (c *command) authenticated(authenticated func(*cobra.Command, []string) err
...
-			}, func() {})
+	if featureflags.Manager.BoolVariation("cli.flink.internal", c.Context, config.CliLaunchDarklyClient, true, false) {
+		configKeys, err := cmd.Flags().GetStringSlice("config-key")
+		if err != nil {
+			return err
+		}
+		// if config-keys were passed, we should enter local mode
+		if len(configKeys) > 0 {
+			return c.startWithLocalMode(cmd)
+		}
```
brianstrauch, line 83: To make this code easier to unit-test:
1. Get both flag values
2. Enter the function if both are non-empty, and pass in the values themselves, not `cmd`

pkg/flink/lsp/lsp_completer_ws.go
```diff
@@ -80,7 +80,7 @@ func (w *WebsocketLSPClient) refreshWebsocketConnection() {
 	}
 }
 
-func NewLSPClientWS(getAuthToken func() string, baseUrl, organizationId, environmentId string) LspInterface {
+func NewClientWs(getAuthToken func() string, baseUrl, organizationId, environmentId string) LspInterface {
```
brianstrauch, line 83: How about `NewWebsocketClient()` for people who are unfamiliar with what "WS" means?

YannickPferr: that's better 👍

internal/flink/command_shell.go
YannickPferr: yeah makes sense!

brianstrauch (approved):

MarTango (approved):

PR #2733, opened 2024-04-24
feat/KFS-1891 KFS-1881 warning and new retention for compute pool deletion msg

Guttz:
Release Notes
-------------

New Features
- Added warning with statement retention time for `confluent flink compute-pool delete`

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/internal/flink/command_compute_pool_delete.go
+++ b/internal/flink/command_compute_pool_delete.go
@@ -65,20 +65,16 @@ func (c *command) computePoolDelete(cmd *cobra.Command, args []string) error {
 }
 
 func confirmDeletionString(name, id string) string {
-	deletionString := fmt.Sprintf("Are you sure you want to delete the compute pool \"%s\"?"+
+	return fmt.Sprintf("Are you sure you want to delete the compute pool \"%s\"?"+
 		" All statements leveraging the compute pool will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
 		"After that, they will be permanently deleted. \n"+
 		"To confirm, type \"%s\". To cancel, press Ctrl-C", id, name)
-
-	return deletionString
 }
 
 func confirmMultipleDeletionString(idList []string) string {
-	deletionString := fmt.Sprintf("Are you sure you want to delete compute pools %s?"+
+	return fmt.Sprintf("Are you sure you want to delete compute pools %s?"+
 		" All statements leveraging the compute pools will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
 		"After that, they will be permanently deleted. \n", utils.ArrayToCommaDelimitedString(idList, "and"))
-
-	return deletionString
 }
 
 func validateAndConfirmComputePoolDeletion(cmd *cobra.Command, args []string, checkExistence func(string) bool, resourceType, name string) error {
--- a/internal/flink/command_compute_pool_delete.go
+++ b/internal/flink/command_compute_pool_delete.go
@@ -1,6 +1,8 @@
 package flink
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


internal/flink/command_compute_pool_delete.go
```diff
@@ -46,7 +49,7 @@ func (c *command) computePoolDelete(cmd *cobra.Command, args []string) error {
 		return err == nil
 	}
 
-	if err := deletion.ValidateAndConfirmDeletion(cmd, args, existenceFunc, resource.FlinkComputePool, computePool.Spec.GetDisplayName()); err != nil {
+	if err := validateAndConfirmComputePoolDeletion(cmd, args, existenceFunc, resource.FlinkComputePool, computePool.Spec.GetDisplayName()); err != nil {
```
sgagniere, line 52: To be more consistent w/ other CLI resources that have extra delete information, can we add the extra information (about 30 days in list history & permanent deletion) as a message after the delete operation? Instead of giving the information before in this new validation function.

Like in kafka/command_broker_delete.go which calls `deletion.DeleteWithoutMessage` and then outputs a custom post-delete message.

Guttz: This is an irreversible operation and this could lead to people stopping production flink jobs by accident. That's why I think it's actually better to include the info in the warning before the user deletion the pool it.

Also, the feature was designed in a product one pager, and it would be up to product to decide @kasher-confluent

kasher-confluent: I do agree with @Guttz. Deleting a compute pool can have consequences on the customer's workload. Thus, it is important to inform them about what is expected to happen when they delete a compute pool. We do the same for the UI as well. Meaning, show them the below popup before deletion to confirm. 

Screenshot 2024-04-25 at 07 47 48

Would still recommend showing the confirmation prompt before deletion.

sgagniere: Ok, this seems fine then.

sgagniere (approved):

internal/flink/command_compute_pool_delete.go
```diff
@@ -61,6 +64,44 @@ func (c *command) computePoolDelete(cmd *cobra.Command, args []string) error {
 	return errs.ErrorOrNil()
 }
 
+func confirmDeletionString(name, id string) string {
+	deletionString := fmt.Sprintf("Are you sure you want to delete the compute pool \"%s\"?"+
+		" All statements leveraging the compute pool will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
+		"After that, they will be permanently deleted. \n"+
+		"To confirm, type \"%s\". To cancel, press Ctrl-C", id, name)
+
+	return deletionString
```
brianstrauch, line 73:
```suggestion
	return fmt.Sprintf(`Are you sure you want to delete the compute pool "%s"?`+
		" All statements leveraging the compute pool will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
		"After that, they will be permanently deleted.\n"+
		`To confirm, type "%s". To cancel, press Ctrl-C`, id, name)
```

internal/flink/command_compute_pool_delete.go
```diff
@@ -61,6 +64,44 @@ func (c *command) computePoolDelete(cmd *cobra.Command, args []string) error {
...
+}
+
+func confirmMultipleDeletionString(idList []string) string {
+	deletionString := fmt.Sprintf("Are you sure you want to delete compute pools %s?"+
+		" All statements leveraging the compute pools will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
+		"After that, they will be permanently deleted. \n", utils.ArrayToCommaDelimitedString(idList, "and"))
+
+	return deletionString
```
brianstrauch, line 81:
```suggestion
	return fmt.Sprintf("Are you sure you want to delete compute pools %s?"+
		" All statements leveraging the compute pools will be STOPPED immediately and be available for 30 days in the statement list history.\n"+
		"After that, they will be permanently deleted.\n", utils.ArrayToCommaDelimitedString(idList, "and"))
```

sgagniere (approved):
lgtm

PR #2739, opened 2024-04-30
[CLI-3094] Update CLI error message for Group Mapping quota limit

tmalikconfluent:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Removed the : 402 Payment Required part of the error message when limit of number of group mappings allowed is exceeded

References
----------
https://confluentinc.atlassian.net/browse/CLI-3094

Test & Review
-------------
Wrote an integration test iam_test.go to test the change

### DIFF
--- a/pkg/ccloudv2/sso.go
+++ b/pkg/ccloudv2/sso.go
@@ -23,7 +23,7 @@ func (c *Client) ssoApiContext() context.Context {
 
 func (c *Client) CreateGroupMapping(groupMapping ssov2.IamV2SsoGroupMapping) (ssov2.IamV2SsoGroupMapping, error) {
 	resp, httpResp, err := c.SsoClient.GroupMappingsIamV2SsoApi.CreateIamV2SsoGroupMapping(c.ssoApiContext()).IamV2SsoGroupMapping(groupMapping).Execute()
-	return resp, errors.CatchPaymentRequiredError(err, httpResp)
+	return resp, errors.CatchCCloudV2Error(err, httpResp)
 }
 
 func (c *Client) DeleteGroupMapping(id string) error {
--- a/pkg/ccloudv2/sso.go
+++ b/pkg/ccloudv2/sso.go
@@ -2,11 +2,9 @@ package ccloudv2
 
 import (
 	"context"
-	"net/http"
-
 	ssov2 "github.com/confluentinc/ccloud-sdk-go-v2/sso/v2"
-
 	"github.com/confluentinc/cli/v3/pkg/errors"
+	"net/http"
 )
 
 func newSsoClient(httpClient *http.Client, url, userAgent string, unsafeTrace bool) *ssov2.APIClient {
@@ -25,7 +23,7 @@ func (c *Client) ssoApiContext() context.Context {
 
 func (c *Client) CreateGroupMapping(groupMapping ssov2.IamV2SsoGroupMapping) (ssov2.IamV2SsoGroupMapping, error) {
 	resp, httpResp, err := c.SsoClient.GroupMappingsIamV2SsoApi.CreateIamV2SsoGroupMapping(c.ssoApiContext()).IamV2SsoGroupMapping(groupMapping).Execute()
-	return resp, errors.CatchCCloudV2Error(err, httpResp)
+	return resp, errors.CatchPaymentRequiredError(err, httpResp)
 }
 
 func (c *Client) DeleteGroupMapping(id string) error {
--- a/pkg/ccloudv2/sso.go
+++ b/pkg/ccloudv2/sso.go
@@ -2,9 +2,11 @@ package ccloudv2
 
 import (
 	"context"
-	ssov2 "github.com/confluentinc/ccloud-sdk-go-v2/sso/v2"
-	"github.com/confluentinc/cli/v3/pkg/errors"
 	"net/http"
+
+	ssov2 "github.com/confluentinc/ccloud-sdk-go-v2/sso/v2"
+
+	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
 func newSsoClient(httpClient *http.Client, url, userAgent string, unsafeTrace bool) *ssov2.APIClient {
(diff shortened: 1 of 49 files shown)


pkg/errors/catcher.go
```diff
@@ -303,6 +303,24 @@ func CatchServiceNameInUseError(err error, r *http.Response, serviceName string)
...
+	}
+
+	if httpResp == nil {
+		return err
+	}
+
+	errorPayment := CatchCCloudV2Error(err, httpResp)
+	if strings.Contains(errorPayment.Error(), "402 Payment Required") {
+		message := strings.TrimSpace(errorPayment.Error()[0 : len(errorPayment.Error())-len(": 402 Payment Required")])
+		return fmt.Errorf("%s", message)
```
brianstrauch, line 318:
```suggestion
		return errors.New(message)
```

pkg/errors/catcher.go
```diff
@@ -303,6 +303,24 @@ func CatchServiceNameInUseError(err error, r *http.Response, serviceName string)
...
+		return nil
+	}
+
+	if httpResp == nil {
+		return err
+	}
+
+	errorPayment := CatchCCloudV2Error(err, httpResp)
+	if strings.Contains(errorPayment.Error(), "402 Payment Required") {
+		message := strings.TrimSpace(errorPayment.Error()[0 : len(errorPayment.Error())-len(": 402 Payment Required")])
```
brianstrauch, line 317: Try https://pkg.go.dev/strings#TrimSuffix?

brianstrauch (commented):
Nice job! Left a few comments.

Some context: We typically suggest that users re-run a command with `--unsafe-trace` if they're interested in seeing a detailed HTTP trace which includes the status code, so it's good that we're getting rid of the status code here. :smile:

I'm not sure where "402 Payment Required" is being appended in the first place... either in the backend or in the CLI. If it's being appended in the CLI, we should remove that line of code since it could be doing the same thing for commands other than this one, too.

sgagniere:
@brianstrauch @tmalikconfluent 

re: 402 being appended: Our ccloudv2 error handler is actually appending the status code to the message, so it's a widespread thing. Do we want to make this change across the board? That would probably simplify the PR too.

brianstrauch:
> Do we want to make this change across the board? That would probably simplify the PR too.

:100:

pkg/errors/catcher.go
tmalikconfluent: had a look at this and used it here. It was a good idea to use this! But this entire part of code will not be needed anymore i guess

pkg/ccloudv2/sso.go
```diff
@@ -2,11 +2,9 @@ package ccloudv2
 
 import (
 	"context"
-	"net/http"
-
 	ssov2 "github.com/confluentinc/ccloud-sdk-go-v2/sso/v2"
-
 	"github.com/confluentinc/cli/v3/pkg/errors"
+	"net/http"
```
sgagniere, line 7: We probably want to revert this to maintain our usual style of having line-separated blocks for 1) go standard library, 2) non-Confluent third party dependencies, 3) Confluent dependencies, 4) CLI pkgs.

tmalikconfluent: okay, i didnt make this change manually, it happened automatically so i thought it was what was preferred, but will revert it back now.

sgagniere (approved):
lgtm

pkg/errors/catcher.go
```diff
@@ -178,14 +179,14 @@ func CatchCCloudV2Error(err error, r *http.Response) error {
...
 	if resBody.Message != "" {
 		message := strings.TrimRight(resBody.Message, "\n")
-		return fmt.Errorf("%s: %w", message, err)
+		return errors.New(message)
 	}
 
 	if resBody.Error.Message != "" {
 		message := strings.TrimFunc(resBody.Error.Message, func(c rune) bool {
 			return c == '.' || c == '\n'
 		})
```
brianstrauch, line 188: Unrelated to the PR, but https://pkg.go.dev/strings#TrimRight would be more concise! It takes a "cutset" which we could set to ".\n"

test/fixtures/output/iam/rbac/role-binding/list-invalid-role-error-type-1-cloud.golden
```diff
@@ -1 +1 @@
-Error: Invalid role name : InvalidOrgAdmin: 400 Bad Request
+Error: Invalid role name : InvalidOrgAdmin
```
brianstrauch, line 1: Unrelated, but can we remove this space?
```suggestion
Error: Invalid role name: InvalidOrgAdmin
```

test/fixtures/output/network/link/service/describe-invalid.golden
```diff
@@ -1 +1 @@
-Error: The network link service nls-invalid was not found.: 404 Not Found
+Error: The network link service nls-invalid was not found.
```
brianstrauch, line 1: Trailing periods should be stripped to align with our other errors and to prevent ".:" if more than one error is chained. Can we also use the `strings.TrimRight()` function on this part of the error to trim off the "." too?

brianstrauch (commented):
Thank you!!

test/fixtures/output/iam/rbac/role-binding/list-invalid-role-error-type-1-cloud.golden
tmalikconfluent: removed

test/fixtures/output/network/link/service/describe-invalid.golden
tmalikconfluent: removed this particular one, do we need to make any other changes in the code for this PR related to trailing periods?

brianstrauch: Yeah, it would be great if we could remove them all by adding a period to the "cutset" here: https://github.com/confluentinc/cli/pull/2739/files#diff-0dc1a72e07c38cc95af9014c65d3d2a71fa1dd7a7911e309dcc57be4f9371043R181

brianstrauch (approved):
Amazing work!!

PR #2749, opened 2024-05-06
CLI-3124: Print role bindings with deleted resources

brianstrauch:
Checklist
---------

What
----
It's possible for the role binding API to return a specific role binding even though its resource has been deleted. We want to align with the UI and print the role binding in that case.

Test & Review
-------------
Added integration test, will manually test before merging.


sgagniere (approved):
Thanks!

PR #2751, opened 2024-05-07
[CLI-2340] Remove .netrc support

sgagniere:
Release Notes
-------------

Breaking Changes
- Saved credentials will no longer be 

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
.netrc support for saved credentials was deprecated a while ago. This PR removes it.

References
----------

Test & Review
-------------


brianstrauch:
> Saved credentials will no longer be read from the ".netrc" file on Unix systems

Let's explain that they'll be read from the configuration file or macOS keychain so users don't get worried the `--save` functionality isn't going away.

go.mod
```diff
@@ -40,7 +40,6 @@ require (
 	github.com/confluentinc/ccloud-sdk-go-v2/stream-designer v0.3.0
 	github.com/confluentinc/confluent-kafka-go v1.9.3-RC3
 	github.com/confluentinc/go-editor v0.11.0
-	github.com/confluentinc/go-netrc v0.0.0-20220321173724-4d50f36ff450
```
brianstrauch, line 43: Let's archive https://github.com/confluentinc/go-netrc after the v4 release?

pkg/keychain/keychain_darwin.go
```diff
@@ -9,16 +9,44 @@ import (
...
-	"github.com/confluentinc/cli/v3/pkg/netrc"
 )
 
+type Machine struct {
+	Name     string
+	User     string
+	Password string
+}
+
+type MachineParams struct {
```
brianstrauch, line 20: I'm not sure this belongs here. I think this struct is used by the configuration file *and* the keychain, so putting it in the `keychain` package might be misleading?

brianstrauch: > //go:build darwin

Also, be careful, this file only gets built on macOS.

internal/login/command.go
```diff
@@ -201,7 +198,7 @@ func (c *command) printRemainingFreeCredit(client *ccloudv1.Client, currentOrg *
 	}
 }
 
-// Order of precedence: env vars > config file > netrc file > prompt
+// Order of precedence: env vars > config file > prompt
```
brianstrauch, line 201:
```suggestion
// Order of precedence: environment variables > configuration file / macOS keychain > prompt
```

internal/login/command_test.go
```diff
@@ -931,13 +861,13 @@ func newLoginCmd(auth *ccloudv1mock.Auth, userInterface *ccloudv1mock.UserInterf
...
 	prerunner := climock.NewPreRunnerMock(ccloudClientFactory.AnonHTTPClientFactory(ccloudURL), nil, mdsClient, nil, cfg)
-	loginCmd := New(cfg, prerunner, ccloudClientFactory, mdsClientManager, netrcHandler, loginCredentialsManager, loginOrganizationManager, authTokenHandler)
+	loginCmd := New(cfg, prerunner, ccloudClientFactory, mdsClientManager, loginCredentialsManager, loginOrganizationManager, authTokenHandler)
 	loginCmd.Flags().Bool("unsafe-trace", false, "")
 	return loginCmd, cfg
 }
 
-func newLogoutCmd(cfg *config.Config, netrcHandler netrc.NetrcHandler) (*cobra.Command, *config.Config) {
-	logoutCmd := logout.New(cfg, climock.NewPreRunnerMock(nil, nil, nil, nil, cfg), netrcHandler)
+func newLogoutCmd(cfg *config.Config) (*cobra.Command, *config.Config) {
```
brianstrauch, line 869: I wonder if we can retire this function. Doesn't seem to be doing much... and strange that it's returning `cfg`.

pkg/auth/login_credentials_manager_test.go
```diff
@@ -302,76 +220,24 @@ func (suite *LoginCredentialsManagerTestSuite) TestGetConfluentPrerunCredentials
...
-		},
-		GetFileNameFunc: func() string {
-			return netrcFileName
-		},
-	}
 	cmd := &cobra.Command{}
 	cmd.Flags().Bool("save", false, "test")
-	// No credentials in env var and netrc so should look for prompt
-	loginCredentialsManager := NewLoginCredentialsManager(noCredentialsNetrcHandler, suite.prompt, suite.ccloudClient)
+	// No credentials in env var so should look for prompt
```
brianstrauch, line 226: Unrelated, but can we specify which "env var" we're talking about in this comment?

pkg/keychain/keychain_darwin.go
```diff
@@ -9,16 +9,44 @@ import (
...
+type credentialType int
+
+const (
+	mdsUsernamePassword credentialType = iota
+	ccloudUsernamePassword
+)
+
+func (c credentialType) String() string {
+	credTypes := [...]string{mdsUsernamePasswordString, ccloudUsernamePasswordString}
+	return credTypes[c]
```
brianstrauch, line 45: Does this work?
```suggestion
	return []string{mdsUsernamePasswordString, ccloudUsernamePasswordString}[c]
```

pkg/keychain/keychain_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"regexp"
+	"strings"
+	"testing"
+)
+
+func TestGetMachineNameRegex(t *testing.T) {
+	url := "https://confluent.cloud"
+	ccloudCtxName := "login-csreesangkom@confleunt.io-https://confluent.cloud"
+	confluentCtxName := "login-csreesangkom@confluent.io-http://localhost:8090"
+	specialCharsCtxName := `login-csreesangkom+adoooo+\/@-+\^${}[]().*+?|<>-&@confleunt.io-https://confluent.cloud`
```
brianstrauch, line 14: Is this supposed to say "confleunt"?

pkg/keychain/keychain_darwin.go
sgagniere: Good catch, thanks.

pkg/keychain/keychain_test.go
sgagniere: Probably not, looks like a typo to me.

pkg/keychain/keychain_darwin.go
sgagniere: It should.

pkg/config/context.go
```diff
@@ -27,6 +27,7 @@ type Context struct {
 	FeatureFlags        *FeatureFlags                  `json:"feature_flags,omitempty"`
 
 	// Deprecated
+	NetrcMachineName       string                            `json:"netrc_machine_name"`
```
brianstrauch, line 30: What's the reason for keeping this?

brianstrauch (approved):

brianstrauch:
> Saved credentials will no longer be read from the .netrc file on Unix systems

I think we had Windows support for this too... maybe we can drop the part about "Unix systems" just in case.

PR #2753, opened 2024-05-08
[CLI-3126] Fix panic in Connect offset commands

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Prevent `confluent connect offset delete`, `confluent connect offset update`, and `confluent connect offset status describe` from crashing

Checklist
---------

What
----
`IsSet()` in ccloud-sdk-go-v2 doesn't guarantee that a nullable field is actually non-nil, so we need to check directly that it isn't nil.

References
----------

Test & Review
-------------


brianstrauch (approved):
Thanks! Are there any other usages of `IsSet()` through the CLI that should be changed?

sgagniere:
@brianstrauch There are other usages from the schema registry SDK, but I need to check if that's the same situation or not. I guess that can be a followup item.

PR #2754, opened 2024-05-08
Updated Connect create cluster command to include offsets

nmala01:
Release Notes
-------------
New Features
- Support passing offsets to `confluent connect cluster create` as part of the configuration file

What
----
Updated Connect create cluster command to include offsets

References
----------

Test & Review
-------------
Updated unit tests to include offsets

### DIFF
--- a/internal/connect/command_cluster_create.go
+++ b/internal/connect/command_cluster_create.go
@@ -85,7 +85,7 @@ func (c *clusterCommand) create(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	userConfigs, offsets, err := getConfig(cmd)
+	userConfigs, offsets, err := getConfig(cmd, false)
 	if err != nil {
 		return err
 	}
--- a/internal/connect/command_cluster_create.go
+++ b/internal/connect/command_cluster_create.go
@@ -26,7 +26,7 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
 		Example: examples.BuildExampleString(
 			examples.Example{
-				Text: "The configuration file contains configs including offsets for the connector.",
+				Text: "Create a configuration file with connector configs and offsets.",
 				Code: `{
   "name": "MyGcsLogsBucketConnector",
   "config": {
@@ -84,7 +84,7 @@ func (c *clusterCommand) create(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	userConfigs, offsets, err := getConfig(cmd, false)
+	userConfigs, offsets, err := getConfigAndOffsets(cmd, false)
 	if err != nil {
 		return err
 	}
--- a/internal/connect/command_cluster_create.go
+++ b/internal/connect/command_cluster_create.go
@@ -90,9 +90,9 @@ func (c *clusterCommand) create(cmd *cobra.Command, _ []string) error {
 	}
 
 	connectConfig := connectv1.InlineObject{
-		Name:    connectv1.PtrString((*userConfigs)["name"]),
-		Config:  userConfigs,
-		Offsets: offsets,
+		Name:    connectv1.PtrString((userConfigs)["name"]),
+		Config:  &userConfigs,
+		Offsets: &offsets,
 	}
 
 	environmentId, err := c.Context.EnvironmentId()
--- a/internal/connect/command_cluster_create.go
+++ b/internal/connect/command_cluster_create.go
@@ -26,12 +26,43 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
 		Example: examples.BuildExampleString(
 			examples.Example{
+				Text: "The configuration file contains configs including offsets for the connector.",
+				Code: `{
+  "name": "az-connector",
+  "config": {
+    "name": "az-connector",
+    "azblob.account.name": "azsink",
+    "azblob.account.key": "key",
+    "azblob.container.name": "azsink",
+    "data.format": "JSON",
+    "kafka.api.key": "key",
+    "kafka.api.secret": "key",
+    "tasks.max": "1",
+    "time.interval": "HOURLY",
+    "topics": "apples",
+    "connector.class": "AzureBlobSink"
+  },
+  "offsets": [
+	{
+	  "partition": {
+		"kafka_partition": 0,
+		"kafka_topic": "topic_A"
+	  },
+	  "offset": {
+		"kafka_offset": 1000
+	  }
+	}
+  ]
+}`,
+			},
+			examples.Example{
 				Text: "Create a connector in the current or specified Kafka cluster context.",
 				Code: "confluent connect cluster create --config-file config.json",
 			},
 			examples.Example{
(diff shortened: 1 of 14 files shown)


internal/connect/command_cluster_create.go
```diff
@@ -25,13 +25,44 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 		RunE:        c.create,
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
 		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "The configuration file contains configs including offsets for the connector.",
+				Code: `{
```
naveenmall11, line 30: Code inside example is different from other examples, is it expected here?

test/fixtures/input/connect/config.yaml
```diff
@@ -9,5 +9,21 @@
   "tasks.max": "1",
   "time.interval": "HOURLY",
   "topics": "apples",
-  "connector.class": "AzureBlobSink"
+  "connector.class": "AzureBlobSink",
```
naveenmall11, line 12: This seems to be input file for CreateORUpdateConnector and offsets field is not supported in this API. Can you please check where is this file being used?

internal/connect/command_cluster_create.go
nmala01: Yes, kept in consistent with update offset.

test/fixtures/input/connect/config.yaml
nmala01: for CreateOrUpdate, we are not passing the offset.

naveenmall11 (approved):
LGTM!!

brianstrauch:
This looks like a user-facing change, so can you fill in the release notes section?

internal/connect/command_cluster_create.go
```diff
@@ -25,6 +25,36 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 		RunE:        c.create,
 		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
 		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: "The configuration file contains configs including offsets for the connector.",
```
brianstrauch, line 29:
```suggestion
				Text: "Create a configuration file with connector offsets.",
```

internal/connect/command_cluster_create.go
```diff
@@ -25,6 +25,36 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
...
+				Code: `{
+  "name": "MyGcsLogsBucketConnector",
```
brianstrauch, line 31: What are GCS logs? Can we update the text of this example to explain what they are?

internal/connect/utils.go
```diff
@@ -10,15 +10,15 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
-func getConfig(cmd *cobra.Command) (*map[string]string, error) {
+func getConfig(cmd *cobra.Command, isUpdate bool) (*map[string]string, *[]map[string]any, error) {
```
brianstrauch, line 13: It looks like this function is doing two separate things. Can we split this into two functions: `getConfig()` and `getOffsets()`?

internal/connect/utils.go
```diff
@@ -30,51 +30,63 @@ func getConfig(cmd *cobra.Command) (*map[string]string, error) {
...
 				kvPairs[configKey] = value
 			}
+		} else if key == "offsets" {
+			if isUpdate {
+				return nil, nil, fmt.Errorf(`offsets are not allowed in configuration file for update`)
+			}
+			var request *[]map[string]any
+			valBytes, err := json.Marshal(val)
+			if err != nil {
+				return nil, nil, fmt.Errorf(`error while marshallling offsets :- value for the configuration "offsets" is malformed`)
```
brianstrauch, line 79: What does ":-" mean?

test/fixtures/output/connect/cluster/update-error.golden
```diff
@@ -0,0 +1 @@
+Error: unable to read configuration file "test/fixtures/input/connect/update-config-malformed.json": offsets are not allowed in configuration file for update
```
brianstrauch, line 1:
```suggestion
Error: unable to read configuration file "test/fixtures/input/connect/update-config-malformed.json": offsets are not allowed in updated configuration file
```

internal/connect/utils.go
```diff
@@ -30,51 +30,63 @@ func getConfig(cmd *cobra.Command) (*map[string]string, error) {
...
+			}
+			if err := json.Unmarshal(valBytes, &request); err != nil {
+				return nil, nil, fmt.Errorf(`error while unmarshallling offsets :- value for the configuration "offsets" is malformed`)
```
sgagniere, line 82:
```suggestion
				return nil, nil, fmt.Errorf(`error while marshalling offsets :- value for the configuration "offsets" is malformed`)
			}
			if err := json.Unmarshal(valBytes, &request); err != nil {
				return nil, nil, fmt.Errorf(`error while unmarshalling offsets :- value for the configuration "offsets" is malformed`)
```

internal/connect/command_cluster_create.go
nmala01: Its just a example of connector type.  Doesnt have any specific relation here.

internal/connect/command_cluster_create.go
nmala01: Not sure if `Create a configuration file with connector offsets.`  makes sense, we are inputting config file which contains user configs and offsets.

internal/connect/utils.go
nmala01: both is read from same file, i can rename the method.

test/fixtures/output/connect/cluster/update-error.golden
nmala01: May be  i can do `offsets are not allowed in configuration file for update command`

internal/connect/command_cluster_create.go
nmala01: ok

test/fixtures/output/connect/cluster/update-error.golden
brianstrauch: > offsets are not allowed in configuration file for `confluent connect cluster update`

internal/connect/utils.go
nmala01: Done

internal/connect/utils.go
nmala01: Done

internal/connect/utils.go
nmala01: removed

internal/connect/utils.go
```diff
@@ -10,15 +10,15 @@ import (
 	"github.com/confluentinc/cli/v3/pkg/errors"
 )
 
-func getConfig(cmd *cobra.Command) (*map[string]string, error) {
+func getConfigAndOffsets(cmd *cobra.Command, isUpdate bool) (*map[string]string, *[]map[string]any, error) {
```
brianstrauch, line 13: Maps and slices are reference types, so they are always passed by reference. You don't need a pointer.
```suggestion
func getConfigAndOffsets(cmd *cobra.Command, isUpdate bool) (map[string]string, []map[string]any, error) {
```

sgagniere (approved):

linouk23 (approved):

PR #2761, opened 2024-05-13
[CLI-2908] Add plugin uninstall command

tmalikconfluent:
Release Notes
-------------

Breaking Changes
- The "name" field in in the `confluent plugin search` has been renamed to "id"; a new "name" field has been added in its place

New Features
- Added `confluent plugin uninstall` command for Confluent Platform

Checklist
---------

What
----
Added plugin uninstall command to remove the plugins that are no longer needed for a user. Also added the ID for a particular plugin in the command plugin list (which can be used as the flag in uninstall) and changed the one of the column names in plugin search for the sake of consistency with list command

References
----------
https://confluentinc.atlassian.net/browse/CLI-2908

Test & Review
-------------
Created a new function in the TestPluginUninstall in the plugin_test.go class to test the unistall functionality added. This involved creating a dummy plugin, uninstalling it and checking that the plugin list command gives the expected output after the uninstall.
For manual testing, added one of the plugins using the install. Verified it using plugin list command. Then uninstalled it using the plugin uninstall command and again verified the results using the plugin list command.

### DIFF
--- a/internal/plugin/command_install.go
+++ b/internal/plugin/command_install.go
@@ -56,7 +56,7 @@ func (c *command) install(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	output.Printf(c.Config.EnableColor, "Installed plugin `%s`.\n", plugin.ToCommandName(manifest.Name))
+	output.Printf(c.Config.EnableColor, "Installed plugin `%s`.\n", plugin.ToCommandName(manifest.Id))
 
 	return nil
 }
@@ -86,7 +86,7 @@ func getPluginManifest(pluginName, dir string) (*Manifest, error) {
 		if err := yaml.Unmarshal(manifestFile, manifest); err != nil {
 			return nil, err
 		}
-		manifest.Name = file.Name()
+		manifest.Id = file.Name()
 
 		return manifest, nil
 	}
@@ -103,16 +103,16 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
 	var pluginInstaller plugin.PluginInstaller
 	switch language {
 	case "go":
-		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Name}
+		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Id}
 	case "python":
 		pluginInstaller = &plugin.PythonPluginInstaller{
-			Name:          manifest.Name,
+			Name:          manifest.Id,
 			RepositoryDir: repositoryDir,
 			InstallDir:    installDir,
 		}
 	case "bash":
 		pluginInstaller = &plugin.BashPluginInstaller{
-			Name:          manifest.Name,
+			Name:          manifest.Id,
 			RepositoryDir: repositoryDir,
 			InstallDir:    installDir,
 		}
--- a/internal/plugin/command_install.go
+++ b/internal/plugin/command_install.go
@@ -103,16 +103,16 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
 	var pluginInstaller plugin.PluginInstaller
 	switch language {
 	case "go":
-		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Id}
+		pluginInstaller = &plugin.GoPluginInstaller{Id: manifest.Id}
 	case "python":
 		pluginInstaller = &plugin.PythonPluginInstaller{
-			Name:          manifest.Id,
+			Id:            manifest.Id,
 			RepositoryDir: repositoryDir,
 			InstallDir:    installDir,
 		}
 	case "bash":
 		pluginInstaller = &plugin.BashPluginInstaller{
-			Name:          manifest.Id,
+			Id:            manifest.Id,
 			RepositoryDir: repositoryDir,
 			InstallDir:    installDir,
 		}
--- a/internal/plugin/command_list.go
+++ b/internal/plugin/command_list.go
@@ -15,8 +15,8 @@ import (
 )
 
 type out struct {
-	PluginName string `human:"Plugin Name" serialized:"plugin_name"`
-	PluginId   string `human:"Plugin Id" serialized:"plugin_id"`
+	PluginName string `human:"Name" serialized:"plugin_name"`
+	PluginId   string `human:"ID" serialized:"plugin_id"`
 	FilePath   string `human:"File Path" serialized:"file_path"`
 }
 
@@ -52,15 +52,15 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
 
 	list := output.NewList(cmd)
 	var overshadowedPlugins, nameConflictPlugins []*out
-	for name, paths := range pluginMap {
+	for id, paths := range pluginMap {
 		path := paths[0]
 		if home != "" && strings.HasPrefix(path, home) {
 			path = filepath.Join("~", strings.TrimPrefix(path, home))
 		}
 
 		pluginInfo := &out{
-			PluginName: strings.ReplaceAll(strings.ReplaceAll(name, "-", " "), "_", "-"),
-			PluginId:   name,
+			PluginName: strings.ReplaceAll(strings.ReplaceAll(id, "-", " "), "_", "-"),
+			PluginId:   id,
 			FilePath:   path,
 		}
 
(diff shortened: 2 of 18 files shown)


brianstrauch:
> Added Plugin uninstall command

Make sure you:
1. Start this release note with a dash "-" (bullet point)
2. Format the full command name with backticks "`" so it looks like `confluent plugin install`
3. Specify that this command is only for Confluent Platform (I'm pretty sure this is not compatible with Confluent Cloud)

sgagniere:
@brianstrauch The `plugin` command is for both Cloud and Platform (the `connect plugin install` is Platform only though)

brianstrauch:
> @brianstrauch The `plugin` command is for both Cloud and Platform (the `connect plugin install` is Platform only though)

Since `connect plugin install` is platform-only, it makes sense that `uninstall` is also platform-only, right?

internal/plugin/command_list.go
```diff
@@ -16,6 +16,7 @@ import (
 
 type out struct {
 	PluginName string `human:"Plugin Name" serialized:"plugin_name"`
+	PluginId   string `human:"Plugin Id" serialized:"plugin_id"`
```
brianstrauch, line 19:
```suggestion
	PluginId   string `human:"Plugin ID" serialized:"plugin_id"`
```

brianstrauch: I wonder if we should just say "Name" and "ID" here to save space. The command is named `confluent plugin list` so it's clear that what we're listing are plugins. @sgagniere would need to incorporate this into the v4 release though, since it's a breaking change.

internal/plugin/command_list.go
```diff
@@ -59,6 +60,7 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
 
 		pluginInfo := &out{
 			PluginName: strings.ReplaceAll(strings.ReplaceAll(name, "-", " "), "_", "-"),
+			PluginId:   name,
```
brianstrauch, line 63: Should we change `name` to `id` when it's declared up above to avoid confusion?

tmalikconfluent:
the uninstall works very similar to delete functionality for various other commands, maybe that is why we need it for both cloud and platform? Also have made the other changes you suggested

test/plugin_test.go
```diff
@@ -37,6 +40,39 @@ func (s *CLITestSuite) TestPlugin() {
 	}
 }
 
+func (s *CLITestSuite) TestPluginUninstall() {
+	req := require.New(s.T())
+	filename := "test/fixtures/input/plugin/confluent-test-plugin-uninstall.sh"
+	_, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0777)
```
brianstrauch, line 46: I think there are built-in functions in the Go standard library that set these automatically.

internal/plugin/command_search.go
```diff
@@ -16,13 +16,13 @@ import (
 )
 
 type ManifestOut struct {
-	Name         string `human:"Name" serialized:"name"`
+	Id           string `human:"Id" serialized:"name"`
```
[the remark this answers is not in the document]
tmalikconfluent: the change to id here was to maintain consistency with the `plugin list` command as we added a new column to the list output named ID, which would show the same content as that of this particular column of the `plugin search` command.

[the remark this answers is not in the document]
brianstrauch: Makes sense. Should we show both "ID" and "Name" here in a similar manner, so there's no breaking change. To be clear, the breaking change here is removing the "Name" column?

internal/plugin/command_uninstall.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/confluentinc/cli/v3/pkg/resource"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+func (c *command) newUninstallCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "uninstall <plugin-id-1> [plugin-id-2] ... [plugin-id-n]",
+		Short: "Uninstall official Confluent CLI plugins.",
+		Long:  "Uninstall official Confluent CLI plugins from the confluentinc/cli-plugins repository.",
```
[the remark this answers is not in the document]
tmalikconfluent: users have that option. Removed the long description as discussed with steven earlier

internal/plugin/command_list.go
tmalikconfluent: i have changed the name to ID and Name for now. I will revert back if we decide otherwise

brianstrauch:
> the uninstall works very similar to delete functionality for various other commands, maybe that is why we need it for both cloud and platform? Also have made the other changes you suggested

Hmm :thinking: since we're only able to `install` plugins locally with Confluent Platform, I don't think giving Cloud users the option to uninstall would serve any purpose.

internal/plugin/command_search.go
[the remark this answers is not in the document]
tmalikconfluent: i see. let me discuss this with steven on call and make the change accordingly

internal/plugin/command_uninstall.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if err := deletion.ValidateAndConfirmDeletionYesNo(cmd, args, existenceFunc, resource.Plugin); err != nil {
+		return err
+	}
+
+	deleteFunc := func(name string) error {
+		if err := os.Remove(pluginMap[name][0]); err != nil {
+			return err
+		}
+		return nil
```
brianstrauch, line 37:
```suggestion
		return os.Remove(pluginMap[name][0])
```

internal/plugin/command_search.go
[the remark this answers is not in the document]
sgagniere: I think it's a breaking change either way actually, since the current `Name` field for search is the `ID` for the list, install and uninstall.

We can switch this to ID and add in a new Name field to match the list command's Name field but that would still be a breaking change, so I think we should just merge this into V4

internal/plugin/command_install.go
```diff
@@ -103,16 +103,16 @@ func installPlugin(manifest *Manifest, repositoryDir, installDir string) error {
 	var pluginInstaller plugin.PluginInstaller
 	switch language {
 	case "go":
-		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Name}
+		pluginInstaller = &plugin.GoPluginInstaller{Name: manifest.Id}
```
brianstrauch, line 106: Does this need to be changed to "Id" now?
```suggestion
		pluginInstaller = &plugin.GoPluginInstaller{Id: manifest.Id}
```

internal/plugin/command_list.go
```diff
@@ -51,14 +52,15 @@ func (c *command) list(cmd *cobra.Command, _ []string) error {
...
-	for name, paths := range pluginMap {
+	for id, paths := range pluginMap {
 		path := paths[0]
 		if home != "" && strings.HasPrefix(path, home) {
 			path = filepath.Join("~", strings.TrimPrefix(path, home))
 		}
 
 		pluginInfo := &out{
-			PluginName: strings.ReplaceAll(strings.ReplaceAll(name, "-", " "), "_", "-"),
+			PluginName: strings.ReplaceAll(strings.ReplaceAll(id, "-", " "), "_", "-"),
```
brianstrauch, line 62: Can we use `plugin.ToCommandName()` here?

pkg/plugin/go_plugin_installer.go
```diff
@@ -48,7 +48,7 @@ func (g *GoPluginInstaller) CheckVersion(ver *version.Version) error {
 }
 
 func (g *GoPluginInstaller) Install() error {
-	packageName := fmt.Sprintf("github.com/confluentinc/cli/v3-plugins/%s@latest", g.Name)
+	packageName := fmt.Sprintf("github.com/confluentinc/cli-plugins/%s@latest", g.Id)
```
brianstrauch, line 51: This looks like a bug fix. Should we make a separate PR for this so users don't have to wait for v4 to get these changes?

brianstrauch:
Since this PR is against the `v4` branch, I assume there's at least one breaking change. Can you include that in the release notes section and then work with @sgagniere to get that added to the main v4 PR release notes?

pkg/plugin/go_plugin_installer.go
tmalikconfluent: fixed it on main. https://github.com/confluentinc/cli/pull/2763

sgagniere (approved):

PR #2764, opened 2024-05-16
SCP-2653-update-flink-artifact-with-runtime_lang-arg

NicoleYWDu:
Release Notes
-------------

New Features
- add REQUIRED attribute `runtime-language` for flink artifact, only two choices: python or java

more detailed explanation:
1. input runtime-language: java/python is not case sensitive, code will use toLower() and we store as lowercase in backend. cli will return lowercase type.
2. .zip only allowed for python runtime, .jar only allowed for java runtime

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/internal/flink/command_artifact_create.go
+++ b/internal/flink/command_artifact_create.go
@@ -2,6 +2,7 @@ package flink
 
 import (
 	"fmt"
+	"github.com/confluentinc/cli/v3/pkg/ccloudv2"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


byang-confluent (commented):
i will suggest we updated API before change cli.

internal/flink/command_artifact_create.go
```diff
@@ -32,18 +34,20 @@ func (c *command) newCreateCommand() *cobra.Command {
...
+	cmd.Flags().String("artifact-file", "", "Flink artifact JAR/ZIP file.")
+	c.addRuntimeLanguage(cmd)
 	cmd.Flags().String("description", "", "Description of Flink artifact.")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
 	pcmd.AddOutputFlag(cmd)
 
 	cobra.CheckErr(cmd.MarkFlagRequired("artifact-file"))
-	cobra.CheckErr(cmd.MarkFlagFilename("artifact-file", "jar"))
+	cobra.CheckErr(cmd.MarkFlagFilename("artifact-file", "jar", "zip"))
+	cobra.CheckErr(cmd.MarkFlagRequired("runtime-language"))
```
brianstrauch, line 50: This is a breaking change. Can you create this PR against the `v4` branch?

brianstrauch (changes requested):

PR #2769, opened 2024-05-23
[AM-4684] Add Flink model roles to CLI list/describe commands

lucy-fan:
What
----
Adds the Flink model roles to the list and describe roles commands to support the Flink Model EA. 

References
----------

Test & Review
-------------

### DIFF
--- a/internal/iam/command_rbac.go
+++ b/internal/iam/command_rbac.go
@@ -19,6 +19,7 @@ var (
 	identityNamespace       = optional.NewString("identity")
 	flinkNamespace          = optional.NewString("flink")
 	workloadNamespace       = optional.NewString("workload")
+	flinkModelNamespace     = optional.NewString("flinkmodel")
 )
 
 func newRbacCommand(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command {
--- a/internal/iam/command_rbac_role_describe.go
+++ b/internal/iam/command_rbac_role_describe.go
@@ -64,6 +64,10 @@ func (c *roleCommand) ccloudDescribe(cmd *cobra.Command, role string) error {
 		namespacesList = append(namespacesList, flinkNamespace.Value(), workloadNamespace.Value())
 	}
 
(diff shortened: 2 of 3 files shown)


lihaosky (approved):
Thanks!

internal/iam/command_rbac.go
```diff
@@ -19,6 +19,7 @@ var (
 	identityNamespace       = optional.NewString("identity")
 	flinkNamespace          = optional.NewString("flink")
 	workloadNamespace       = optional.NewString("workload")
+	flinkModelNamespace     = optional.NewString("flinkmodel")
```
brianstrauch, line 22: This is clearly not scaling very well since your team has had to open 9 separate PRs to gradually add all of these namespaces... Has your team considered controlling this list of namespaces in the backend?

lihaosky: cc @AydenAdair

lucy-fan: We can consider looking into controlling this via the backend but often the CLI and UI have different needs (ie. these roles only need to be available in the CLI for the EA) so doing it via the backend may end up being more work. These roles are scheduled to be in EA this week though so we may not have enough time to figure that out in this change.

brianstrauch: Let's merge this PR as is, but in the future it would be great to add this to the backend. I assume you could add a switch to distinguish if a request is coming from the UI vs CLI in the backend!

brianstrauch (approved):

internal/iam/command_rbac.go
brianstrauch: (At the very least, let's create a ticket for this!)

brianstrauch:
Do you want to add any release notes by the way @lucy-fan? If the changes are early access only, you can prefix the release notes with `[Early Access]`.

lucy-fan:
@lihaosky Do you want any release notes for the Flink Model role CLI changes?

lihaosky:
I guess not since this won't forbid users to create the role binding. Not necessary to advocate it I think.

PR #2770, opened 2024-05-24
Fix missing checks on command run requirements

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Some commands set `c.PersistentPreRunE = prerunner.InitializeOnPremKafkaRest(c.AuthenticatedCLICommand)`, which overwrites the usual `PersistentPreRunE`.

However, doing this causes the error from the call to `ErrIfMissingRunRequirement` in `AuthenticatedWithMDS` to not be returned. This means that using `InitializeOnPremKafkaRest` causes annotations to be ignored. This results in misleading errors or even panics if users try to run the commands that the annotations were supposed to prevent them from running.

As a quick fix, I've added the `ErrIfMissingRunRequirement` requirement check to `InitializeOnPremKafkaRest` itself.

References
----------

Test & Review
-------------
Running `confluent kafka broker list` before the change (logged out or logged into Cloud):
```
confluent kafka broker list
Error: Kafka REST URL not found

Suggestions:
    Use the --url flag or set CONFLUENT_REST_URL.
```
And after the change (logged out):
```
confluent kafka broker list                                                     
Error: you must log in to Confluent Platform to use this command

Suggestions:
    Log in to Confluent Platform with confluent login --url <mds-url>.
```
After the change (logged into Cloud):
```
confluent kafka broker list      

Suggestions:
    Log in to Confluent Platform with confluent login --url <mds-url>.
    Use the "--help" flag to see available commands.
```

### DIFF
--- a/pkg/cmd/prerunner.go
+++ b/pkg/cmd/prerunner.go
@@ -538,6 +538,12 @@ func (r *PreRun) createMDSClient(ctx *config.Context, ver *version.Version, unsa
 // Initializes a default KafkaRestClient
 func (r *PreRun) InitializeOnPremKafkaRest(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		// ErrIfMissingRunRequirement is called inside of AuthenticatedWithMDS but we don't return that error
+		// So call it in this func too to avoid skipping the annotation check
+		if err := ErrIfMissingRunRequirement(cmd, r.Config); err != nil {
+			return err
+		}
+
 		// pass mds token as bearer token otherwise use http basic auth
 		// no error means user is logged in with mds and has valid token; on an error we try http basic auth since mds is not needed for RP commands
 		err := r.AuthenticatedWithMDS(command)(cmd, args)
--- a/pkg/cmd/prerunner.go
+++ b/pkg/cmd/prerunner.go
@@ -32,10 +32,6 @@ import (
 
 const autoLoginMsg = "Successful auto-login with non-interactive credentials."
 
-var wrongLoginCommandsMap = map[string]string{
-	"confluent cluster": "confluent kafka cluster",
-}
-
 // PreRun is a helper class for automatically setting up Cobra PersistentPreRun commands
 type PreRunner interface {
 	Anonymous(command *CLICommand, willAuthenticate bool) func(*cobra.Command, []string) error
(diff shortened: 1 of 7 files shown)


brianstrauch (approved):

brianstrauch:
Can we easily write an integration test for this?

sgagniere:
Added some integration tests using the `broker` command as a representative.

Maybe we could do something a bit more comprehensive later like a `TestAnnotations` suite to check all commands w/ annotations.

pkg/cmd/prerunner.go
```diff
@@ -538,6 +538,12 @@ func (r *PreRun) createMDSClient(ctx *config.Context, ver *version.Version, unsa
 // Initializes a default KafkaRestClient
 func (r *PreRun) InitializeOnPremKafkaRest(command *AuthenticatedCLICommand) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		// ErrIfMissingRunRequirement is called inside of AuthenticatedWithMDS but we don't return that error
+		// So call it in this func too to avoid skipping the annotation check
```
brianstrauch, line 542: Any way we can only call it once / why don't we return the error from AuthenticatedWithMDS?

sgagniere: We don't return the error because `InitializeOnPremKafkaRest` uses the existence of an error to set the bool `useMdsToken := err == nil`; and it prompts the user for username/password if `useMdsToken` is false.

Instead of calling `ErrIfMissingRunRequirement` twice, we could instead check if the error from `AuthenticatedWithMDS` contains text from the `ErrIfMissingRunRequirement` error. If that sounds good I can go ahead and make that change.

brianstrauch: If the code won't be too messy, let's go ahead with that!

brianstrauch: > contains text from the ErrIfMissingRunRequirement error

I think the more idiomatic way to do this would be to define a new "RunRequirementError" type and check to see if the error is of that type.

pkg/config/config.go
```diff
@@ -24,42 +24,48 @@ const emptyFieldIndicator = "EMPTY"
...
+		ErrorMsg:       "you must log out of Confluent Cloud to use this command",
+		SuggestionsMsg: "Log out with `confluent logout`.\n",
+	}
+	RequireOnPremLoginErr = &errors.RunRequirementError{
+		ErrorMsg:       "you must log in to Confluent Platform to use this command",
+		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.",
+	}
+	RunningOnPremCommandInCloudErr = &errors.RunRequirementError{
+		ErrorMsg:       "this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
+		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.\n" + `Use the "--help" flag to see available commands.`,
```
brianstrauch, line 61:
```suggestion
		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.\n" + "See available commands with `--help`.",
```

brianstrauch (approved):

pkg/config/config.go
```diff
@@ -58,7 +58,7 @@ var (
 	}
 	RunningOnPremCommandInCloudErr = &errors.RunRequirementError{
 		ErrorMsg:       "this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command",
-		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.\n" + `Use the "--help" flag to see available commands.`,
+		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.\n" + "See available commands with `--help`.",
```
brianstrauch, line 61:
```suggestion
		SuggestionsMsg: "Log in to Confluent Platform with `confluent login --url <mds-url>`.\nSee available commands with `--help`.",
```

brianstrauch (approved):

PR #2782, opened 2024-05-29
Support prod-us-gov recovery region URLs

brianstrauch:
Checklist
---------

What
----
Support production URLs in recovery regions such as prod-us-gov-east-1.

References
----------

Test & Review
-------------


sgagniere (approved):
lgtm

PR #2783, opened 2024-05-30
Fix Alpine dockerfile build issue

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Installing bash and/or bash-completions in the latest version of the `alpine` dockerfile now creates `/etc/bash_completion.d`, which causes `mkdir -p /etc/bash_completion.d` to fail.

Fix is to add `-p` so that mkdir doesn't error if the folder already exists.

References
----------

Test & Review
-------------
Tested manually with dry run.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,6 +30,6 @@ RUN chown $USER:$USER /etc
 # This symbolic link exists for backwards compatibility reasons
 RUN ln -s /bin/confluent /confluent
 
-RUN mkdir -p /etc/bash_completion.d && confluent completion bash > /etc/bash_completion.d/confluent
+RUN mkdir -p /etc/bash_completion.d/ && confluent completion bash > /etc/bash_completion.d/confluent
 
 USER $USER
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -30,6 +30,6 @@ RUN chown $USER:$USER /etc
 # This symbolic link exists for backwards compatibility reasons
 RUN ln -s /bin/confluent /confluent
 
-RUN mkdir /etc/bash_completion.d && confluent completion bash > /etc/bash_completion.d/confluent
+RUN mkdir -p /etc/bash_completion.d && confluent completion bash > /etc/bash_completion.d/confluent
```
brianstrauch, line 33: I like adding trailing slashes so we know it's a directory.
```suggestion
RUN mkdir -p /etc/bash_completion.d/ && confluent completion bash > /etc/bash_completion.d/confluent
```

brianstrauch (approved):

brianstrauch (approved):

PR #2784, opened 2024-05-30
[MATRIX-191] only pass secret for create and alter model

lihaosky:
Release Notes
-------------
bug fix:
   [early access] only pass secrets for create and alter model request

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/flink/config/local_statements.go
+++ b/pkg/flink/config/local_statements.go
@@ -22,6 +22,27 @@ const (
 	KeyServiceAccount = "client.service-account"
 	KeyStatementName  = "client.statement-name"
 	KeyOutputFormat   = "client.output-format"
+
+	/*
+	   Create Alter model regex. Match create|alter model with
+	     1. allow any space,return,tab between them
+	     2. They are not surrounded by unescaped single quote
+	     3. Single quote is escaped by doubling them like ''.
+	     4. Start beginning of line, with space or `;`
+
+	     (?msi) -> enable multiline, . match \n and case insensitive
+	     \s -> matches space, \n, \t, \r, \f
+	     '' -> match escaped single quote
+	     ?: -> ignore group
+	     '(?:''|[^'])*' -> match anything inside single quote including escaped quote
+	     ?P<query> -> give group match a name
+	     (?:^|;|\\s)(?:create|alter)\\s+model\\s+ -> start with beginning of line, spaces or ;. Match create or alter, then multiple \s, then model, then multiple \s
+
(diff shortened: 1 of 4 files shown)


pkg/flink/internal/store/store_utils.go
```diff
@@ -392,6 +392,39 @@ func useError() *types.StatementError {
 	}
 }
 
+func maybeCreateAlterModelStatement(statement string) bool {
+	regex := regexp.MustCompile(config.CreateAlterModelRegex)
+	matches := regex.FindAllStringSubmatch(statement, -1)
+	names := regex.SubexpNames()
+	fmt.Println(names)
+	//return match != nil && len(match) == 2 && len(match[1]) != 0
+	fmt.Println(matches)
```
brianstrauch, line 401: Please don't commit commented code or `Println()` functions

pkg/flink/config/local_statements.go
```diff
@@ -22,6 +22,27 @@ const (
...
+	/*
+	   Create Alter model regex. Match create|alter model with
+	     1. allow any space,return,tab between them
+	     2. They are not surrounded by unescaped single quote
+	     3. Single quote is escaped by doubling them like ''.
+	     4. Start beginning of line, with space or `;`
+
+	     (?msi) -> enable multiline, . match \n and case insensitive
+	     \s -> matches space, \n, \t, \r, \f
+	     '' -> match escaped single quote
```
chmeyers, line 35: Do we have to make sure it's not inside a comment as well?

pkg/flink/internal/store/store.go
```diff
@@ -105,6 +105,9 @@ func (s *Store) ProcessStatement(statement string) (*types.ProcessedStatement, *
 }
 
 func createSqlV1Statement(statement string, statementName string, computePoolId string, properties map[string]string) flinkgatewayv1.SqlV1Statement {
+	if !maybeCreateAlterModelStatement(statement) {
```
chmeyers, line 108: Would it make sense to only send those secret properties that are actually referenced by the SQL, or does ALTER MODEL need them even if they aren't listed?

lihaosky: alter model can also reference it. Make sense to check they are used

Minggedeng (approved):

lihaosky:
Don't need this anymore. Will fix in gateway

PR #2787, opened 2024-05-31
Fix on-prem kafka rest bug

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix a bug requiring an on-premises login to use the on-premises `confluent kafka` commands

Checklist
---------

What
----
Fixing the run requirement bug where `InitializeOnPremKafkaRest` caused all run requirements to be ignored surfaced this issue, where incorrect run requirements were put on various on-prem kafka commands.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/connect/command_plugin_install.go
+++ b/internal/connect/command_plugin_install.go
@@ -51,7 +51,7 @@ func newInstallCommand(prerunner pcmd.PreRunner) *cobra.Command {
 				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configurations $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
 			},
 		),
-		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonCloudLogin},
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNotCloudLogin},
 	}
 
 	cmd.Flags().String("plugin-directory", "", "The plugin installation directory. If not specified, a default will be selected based on your Confluent Platform installation.")
--- a/internal/connect/command_plugin_install.go
+++ b/internal/connect/command_plugin_install.go
@@ -51,7 +51,7 @@ func newInstallCommand(prerunner pcmd.PreRunner) *cobra.Command {
 				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configurations $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
 			},
 		),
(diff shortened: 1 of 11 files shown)


internal/kafka/command_broker.go
```diff
@@ -14,7 +14,7 @@ func newBrokerCommand(prerunner pcmd.PreRunner) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:         "broker",
 		Short:       "Manage Kafka brokers.",
-		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireOnPremLogin},
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonCloudLogin},
```
brianstrauch, line 17: This naming seems misleading. We only require login for Cloud, but not for On-Prem. Should we have an annotation here at all?

sgagniere: The name is definitely misleading. I checked the logic of that annotation, and what it's doing is return an error if the user is logged into Cloud.

It doesn't error if the user is logged out or logged into MDS.

I believe that's what we want here: the broker commands are on-prem only so we want to block Cloud users here

brianstrauch: Yeah exactly, so instead of `RequireNonCloudLogin` we want `RequireNotCloudLogin` lol

internal/connect/command_plugin_install.go
```diff
@@ -51,7 +51,7 @@ func newInstallCommand(prerunner pcmd.PreRunner) *cobra.Command {
 				Code: "confluent connect plugin install confluentinc/kafka-connect-datagen:latest --plugin-directory $CONFLUENT_HOME/plugins --worker-configurations $CONFLUENT_HOME/etc/kafka/connect-distributed.properties",
 			},
 		),
-		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonCloudLogin},
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNotCloudLogin},
```
brianstrauch, line 54: Alternate language suggestion since "Non" and "Not" differ by one letter and might be confusing:
```suggestion
		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogout},
```

brianstrauch (approved):

test/fixtures/output/kafka/broker/list-cloud-login.golden
```diff
@@ -1,5 +1,5 @@
-Error: this is not a Confluent Cloud command. You must log in to Confluent Platform to use this command
+Error: you must log out of Confluent Cloud to use this command
 
 Suggestions:
-    Log in to Confluent Platform with `confluent login --url <mds-url>`.
-    See available commands with `--help`.
+    Log out with `confluent logout`.
+
```
brianstrauch, line 5: 💯 much clearer language

PR #2793, opened 2024-06-06
Feat/KFS-1551 add diagnostics prototype to flink shell

Guttz:
Release Notes
-------------

Checklist
---------

What
----

This is a prototype for the Diagnostics feature for the language server - highlighting errors for the user while he is writing flink sql. 

The language service has this behind a feature flag and this won't affect current users. 

References
----------

Test & Review
-------------

unit test


pkg/flink/internal/controller/input_controller.go
```diff
@@ -31,7 +34,28 @@ type InputController struct {
...
 const defaultWindowSize = 100
 
-func NewInputController(history *history.History, lspCompleter prompt.Completer) types.InputControllerInterface {
+func NewLSPHandler(c types.InputControllerInterface, handlerCh chan *jsonrpc2.Request) {
+	if handlerCh == nil {
+		return
+	}
+
+	go func() {
+		for req := range handlerCh {
```
YannickPferr, line 43: oh I didn't know you could do that, so it ranges until the channel is closed?

Guttz: What chatgpt taught me: 

1 - 
```
go func() {
    for {
        select {
            case request := <-reqCh:
                // process request
            default:
                // nothing to do, loop again
        }
    }
}()
```

2 - 
```
go func() {
    for request := range reqCh {
        // process request
    }
}()
```

In 1, the use of the default case in this way can cause high CPU usage because when there are no messages, it will constantly check for them in a tight loop. A better approach might be to remove the default case and let the goroutine block until there's something to receive like in 2.

Guttz: I added one test and also tested it manually with fake mock data and it works ✅

pkg/flink/internal/controller/input_controller.go
```diff
@@ -31,7 +34,28 @@ type InputController struct {
 
 const defaultWindowSize = 100
 
-func NewInputController(history *history.History, lspCompleter prompt.Completer) types.InputControllerInterface {
+func NewLSPHandler(c types.InputControllerInterface, handlerCh chan *jsonrpc2.Request) {
```
YannickPferr, line 37: I think I'd rename this to something like `startLspRequestHandler` or something like that

Guttz: done

pkg/flink/lsp/lsp_completer_ws.go
```diff
@@ -71,7 +71,7 @@ func (w *WebsocketLSPClient) refreshWebsocketConnection() {
 		}
 
 		// we only update client and conn if there was no error, otherwise we leave them as is
-		if lspClient, conn, err := newLSPConnection(w.baseUrl, w.getAuthToken(), w.organizationId, w.environmentId); err == nil {
+		if lspClient, conn, err := newLSPConnection(w.baseUrl, w.getAuthToken(), w.organizationId, w.environmentId, nil); err == nil { //todo - replace nil at the end
```
YannickPferr, line 74: hmm I think this is a bit tricky, I think we need to pass the channel as well, otherwise it won't work after reconnecting right?

pkg/flink/internal/controller/input_controller.go
YannickPferr: yeah the for pattern I knew:
```
for {
        select {
            case request := <-reqCh:
                // process request
        }
    }
```
but not the range one. That's pretty cool

pkg/flink/lsp/lsp_completer_ws.go
YannickPferr: I think you might need to store the channel in the client when creating it, so that you can reuse it here

pkg/flink/app/application.go
```diff
@@ -66,7 +67,8 @@ func StartApp(gatewayClient ccloudv2.GatewayClientInterface, tokenRefreshFunc fu
 	resultFetcher := results.NewResultFetcher(dataStore)
 
 	// Instantiate LSP
-	lspClient := lsp.NewWebsocketClient(getAuthToken, appOptions.GetLSPBaseUrl(), appOptions.GetOrganizationId(), appOptions.GetEnvironmentId())
+	handlerCh := make(chan *jsonrpc2.Request) // This is the channel used for the messages received by the language to be passed through to the input controller
```
YannickPferr, line 70: one more thing, I think we need to close the channel somewhere. On shutdown maybe?

pkg/flink/lsp/lsp_completer_ws.go
Guttz: ohh, good catch - I had added this only to test yesterday

pkg/flink/app/application.go
Guttz: Hmmm, we just create one channel and the only situation where it isn't necessary is when the shell exists, so I don't think this is necessary, right?

That said, I made the change to close it on shutdown it, since it's simple, just to play safe

pkg/flink/lsp/lsp_completer_ws.go
Guttz: done ✅

YannickPferr (approved):

pkg/flink/app/application.go
YannickPferr: have you? 😄 maybe you haven't pushed, I can't see it

Guttz: good eye, it was a test 👀 pushed it

YannickPferr (approved):

PR #2796, opened 2024-06-08
[CLI-2339,CLI-2495,CLI-2689,CLI-2318] Rename/remove fields from the configuration file

sgagniere:
Release Notes
-------------

Breaking Changes
- The Windows-only configuration field `update_plugins_once` has been renamed to `update_plugins_once_windows`
- Unused configuration fields `disable_updates`, `anonymous_id`, `no_browser`, and `ver` have been removed
- The "Read-Only" field has been removed from the output of `confluent configuration` commands

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
- Rename `disable_plugins_once` -> `disable_plugins_once_windows`
- Remove `disable_updates`
- Remove `anonymous_id`, `no_browser`, `ver`

References
----------

Test & Review
-------------


pkg/config/config.go
```diff
@@ -85,20 +84,18 @@ type Config struct {
...
+	Filename       string            `json:"-"`
+	IsTest         bool              `json:"-"`
+	Version        *pversion.Version `json:"-"`
 
 	overwrittenCurrentContext      string
 	overwrittenCurrentEnvironment  string
 	overwrittenCurrentKafkaCluster string
+
+	// Deprecated
+	DisablePluginsOnce bool `json:"disable_plugins_once,omitempty"`
```
brianstrauch, line 98: Make sure you open a v5 ticket for removing this :)

test/fixtures/output/configuration/list-windows.golden
```diff
@@ -4,5 +4,4 @@
   disable_plugins       | true  | false      
   disable_plugins_once  | true  | false      
   disable_update_check  | false | false
```
brianstrauch, line 6: Would a Homebrew user see this in their output?

pkg/config/config.go
```diff
@@ -85,20 +84,18 @@ type Config struct {
...
-	// The following configurations are not persisted between runs
-
-	IsTest   bool              `json:"-"`
-	Version  *pversion.Version `json:"-"`
-	Filename string            `json:"-"`
+	// The following configurations are not persisted between runs or are read-only
+	DisableUpdates bool              `json:"-"`
+	Filename       string            `json:"-"`
+	IsTest         bool              `json:"-"`
+	Version        *pversion.Version `json:"-"`
```
brianstrauch, line 91: So happy we will no longer have `Ver` and `Version`

internal/configuration/command.go
```diff
@@ -86,8 +86,8 @@ func getWhitelist(cfg *config.Config) map[string]*fieldInfo {
 	return whitelist
 }
 
-func isReadOnly(jsonField string) bool {
-	return jsonField == "disable_updates" && update.IsHomebrew()
+func isReadOnly(_ string) bool {
+	return false
 }
```
brianstrauch, line 91: Looks like this is a function and not a method of a struct... Any reason why we still need it / still need to pass a string to it?

brianstrauch (approved):

internal/configuration/command.go
sgagniere: It's mostly pointless now, actually. It's only used to set the read only column of the output: `readOnly: isReadOnly(jsonTag),`.

I can remove this and just hardcode `false` until/unless we add an actual read-only config to the output.

brianstrauch: > I can remove this and just hardcode false

+1

brianstrauch: Oh I see, every configuration field is editable now. We could consider removing that column then?

sgagniere: Yeah, sounds good. We can always reintroduce it if we ever feel the need to.

brianstrauch (approved):

PR #2811, opened 2024-06-11
Do not pass `""` to backend if `--algorithm` is unspecified

brianstrauch:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Allow `confluent schema-registry dek create` to be called without specifying `--algorithm`

Checklist
---------

What
----
If `--algorithm` was omitted, we previously would pass `""` to the backend which resulted in:
```
Error: Cannot coerce empty String ("") to `DekFormat` value (but could if coercion was enabled using `CoercionConfig`)
```
Now, we do not pass that field.

Test & Review
-------------
Manually tested


sgagniere (approved):
thanks!

PR #2812, opened 2024-06-12
[CLI-2851] `kafka region list` output should match `flink region list`

tmalikconfluent:
Release Notes
-------------

Breaking Changes
- Changed column "Region Name" to "Name", "Cloud ID" to "Cloud ", "Region ID" to "Region" and removed "Cloud Name" in the output of `confluent kafka region list`

Checklist
---------

What
----
`kafka region list` output should match `flink region list`

References
----------
https://confluentinc.atlassian.net/browse/CLI-2851

Test & Review
-------------
Updated Integration tests.

### DIFF
--- a/internal/kafka/command_region_list.go
+++ b/internal/kafka/command_region_list.go
@@ -1,8 +1,6 @@
 package kafka
 
 import (
-	"strings"
-
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
@@ -44,7 +42,7 @@ func (c *regionCommand) list(cmd *cobra.Command, _ []string) error {
 	list := output.NewList(cmd)
 	for _, region := range regions {
 		out := &regionOut{
-			Cloud:  strings.ToUpper(region.CloudId),
+			Cloud:  region.CloudId,
 			Region: region.RegionId,
 			Name:   region.RegionName,
 		}
--- a/internal/kafka/command_region_list.go
+++ b/internal/kafka/command_region_list.go
@@ -1,12 +1,20 @@
 package kafka
 
 import (
+	"strings"
+
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
 	"github.com/confluentinc/cli/v3/pkg/kafka"
 	"github.com/confluentinc/cli/v3/pkg/output"
 )
+
+type regionOut struct {
+	Name   string `human:"Name" serialized:"name"`
+	Cloud  string `human:"Cloud" serialized:"cloud"`
+	Region string `human:"Region" serialized:"region"`
+}
 
 func (c *regionCommand) newListCommand() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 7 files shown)


brianstrauch:
> kafka region list output columns have been changed.

Can you explain what has changed? i.e. Field "A" has changed to "B".

internal/kafka/command_region_list.go
```diff
@@ -35,7 +43,12 @@ func (c *regionCommand) list(cmd *cobra.Command, _ []string) error {
 
 	list := output.NewList(cmd)
 	for _, region := range regions {
-		list.Add(region)
+		out := &regionOut{
+			Cloud:  strings.ToUpper(region.CloudId),
```
brianstrauch, line 47: Do we have a `nil`-safe `region.GetCloudId()` function available here?

test/fixtures/output/kafka/10.golden
```diff
@@ -1,4 +1,4 @@
-  Cloud ID |     Cloud Name      |   Region ID    |       Region Name        
------------+---------------------+----------------+--------------------------
-  aws      | Amazon Web Services | ap-northeast-1 | ap-northeast-1 (Tokyo)   
-  aws      | Amazon Web Services | us-east-1      | us-east-1 (N. Virginia)  
+           Name           | Cloud |     Region      
+--------------------------+-------+-----------------
+  ap-northeast-1 (Tokyo)  | AWS   | ap-northeast-1
```
brianstrauch, line 3: Previously, "AWS" used to be lowercase. Let's change that back so it matches the backend output.

tmalikconfluent:
> > kafka region list output columns have been changed.
> 
> Can you explain what has changed? i.e. Field "A" has changed to "B".

Almost all the columns have changed to make it similar to flink region list, should i just mention that columns are now similar to flink region list?

internal/kafka/command_region_list.go
tmalikconfluent: we don't have that

test/fixtures/output/kafka/10.golden
tmalikconfluent: does this mean changing it to lower case for all test cases and also in the code?

brianstrauch: It looks like we're doing `strings.ToUpper(region.CloudId)`, so let's remove the function and update tests accordingly!

internal/kafka/command_region_list.go
brianstrauch: Huh, I wonder why that is. I was assuming the `region` struct is defined in ccloud-sdk-go-v2, but it looks like maybe not?

brianstrauch:
> > > kafka region list output columns have been changed.
> > 
> > 
> > Can you explain what has changed? i.e. Field "A" has changed to "B".
> 
> Almost all the columns have changed to make it similar to flink region list, should i just mention that columns are now similar to flink region list?

I see three changes and one deletion, so how about:

> Renamed field "A" to "B", "C" to "D", "E" to "F", and deleted field "G".

internal/kafka/command_region_list.go
tmalikconfluent: no its not coming directly from  ccloud-sdk-go-v2. The values are in the region.go class of the cli repo.

brianstrauch: Ah, ok. And hopefully that struct was initialized with `nil`-safe function calls!

brianstrauch (approved):
Thank you!!

internal/kafka/command_region_list.go
tmalikconfluent: yes that is correct!

PR #2823, opened 2024-06-21
[CLI-2584] Remove `context create` and `context update`

sgagniere:
Release Notes
-------------

Breaking Changes
- `confluent context create` and `confluent context update` has been removed

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
The ability to create and maintain custom contexts is obsolete now that `confluent kafka topic produce | consume` works without login for both on-prem and cloud.

References
----------

Test & Review
-------------


internal/context/command_update.go
```diff
@@ -18,11 +15,10 @@ func (c *command) newUpdateCommand() *cobra.Command {
 		RunE:              c.update,
```
brianstrauch, line 15: Wait we should also delete this command, right? I think we have nothing to update if you can't create a Kafka-based context in the first place.

sgagniere: I thought about it, but it still has some functionality: you can set the current Kafka cluster for a context.

Should we remove that in favor of `confluent kafka cluster use`? The only difference is that you'll have to switch contexts first.

brianstrauch: > Should we remove that in favor of confluent kafka cluster use?

Yep, I think so.

brianstrauch: (as long as no functionality is lost, we'd prefer having one standardized way to do this)

pkg/errors/typed.go
```diff
@@ -127,7 +127,7 @@ func NewCorruptedConfigError(format, contextName, configFile string) CLITypedErr
 	return &CorruptedConfigError{
 		errorMsg: fmt.Sprintf("corrupted CLI config: %v", errorWithStackTrace),
 		suggestionsMsg: fmt.Sprintf("Your configuration file \"%s\" is corrupted.\n"+
-			"Remove config file, and run `confluent login` or `confluent context create`.\n"+
+			"Remove config file, and run `confluent login`.\n"+
```
brianstrauch, line 130:
```suggestion
			"Remove it, and run `confluent login`.\n"+
```

brianstrauch (approved):

brianstrauch (changes requested):
Sorry to backpedal here, but I think we still want to support `confluent context update --name <name>`. Not sure what I was thinking when I opened this ticket, but we still want to allow users to rename contexts because they're currently super verbose... and renaming them makes it easier to do `confluent context use <name>` later.

pkg/errors/typed.go
```diff
@@ -127,7 +127,7 @@ func NewCorruptedConfigError(format, contextName, configFile string) CLITypedErr
 	return &CorruptedConfigError{
 		errorMsg: fmt.Sprintf("corrupted CLI config: %v", errorWithStackTrace),
 		suggestionsMsg: fmt.Sprintf("Your configuration file \"%s\" is corrupted.\n"+
-			"Remove config file, and run `confluent login` or `confluent context create`.\n"+
+			"Remove it, and run `confluent login`.\n"+
```
brianstrauch, line 130: Nit: More precise wording
```suggestion
			"Delete it, and run `confluent login`.\n"+
```

brianstrauch (approved):

PR #2824, opened 2024-06-21
[CLI-3140] Allow multiple onprem SSO login refresh

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fix an issue preventing on-premises users from refreshing SSO logins more than once

Checklist
---------

What
----
On-prem refresh tokens remain valid after login extension requests. They still expire, and their expiration is set on the IDP side.

References
----------

Test & Review
-------------
TODO: manual testing

### DIFF
--- a/pkg/auth/auth_token_handler.go
+++ b/pkg/auth/auth_token_handler.go
@@ -150,7 +150,7 @@ func (a *AuthTokenHandlerImpl) GetConfluentToken(mdsClient *mdsv1.APIClient, cre
 	if credentials.IsSSO {
 		if credentials.AuthRefreshToken != "" {
 			if authToken, err := refreshConfluentToken(mdsClient, credentials); err == nil {
-				return authToken, "", nil
+				return authToken, credentials.AuthRefreshToken, nil
 			}
 		}
 


pkg/auth/auth_token_handler.go
```diff
@@ -150,7 +150,7 @@ func (a *AuthTokenHandlerImpl) GetConfluentToken(mdsClient *mdsv1.APIClient, cre
 	if credentials.IsSSO {
 		if credentials.AuthRefreshToken != "" {
 			if authToken, err := refreshConfluentToken(mdsClient, credentials); err == nil {
-				return authToken, "", nil
+				return authToken, credentials.AuthRefreshToken, nil
```
brianstrauch, line 153: Interesting that we re-use the refresh token. Sounds like the refresh endpoint doesn't also refresh the refresh token?

brianstrauch (approved):

pkg/auth/auth_token_handler.go
sgagniere: Nope.

It also doesn't return the current refresh token, which is why I originally assumed that it was one time use.

brianstrauch: I'm curious why that is. Let's merge since this is still an improvement!

sgagniere: Sounds good. Just going to wait a bit on Monday to see if I can get some help from feature team to test this (since my manual testing setup doesn't work anymore...)

PR #2825, opened 2024-06-22
[CLI-2316] Replace CheckFile with a config field

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Track the last update time directly in the config instead of using the timestamp on a separate text file.

References
----------

Test & Review
-------------


pkg/update/client_test.go
```diff
@@ -425,37 +385,12 @@ func TestCheckForUpdates_BehaviorOverTime(t *testing.T) {
...
 
 	// Finally we should check once more
-	clock.Advance(3 * time.Second)
+	clock.Advance(4 * time.Second)
 	repo.Reset()
 	_, _, _ = client.CheckForUpdates("my-cli", "v1.2.3", false)
 	req.True(repo.GetLatestMajorAndMinorVersionCalled())
 }
 
-func TestCheckForUpdates_NoCheckFileGiven(t *testing.T) {
```
sgagniere, line 434: I removed this one because the test condition doesn't make sense anymore without a CheckFile.

sgagniere:
Not really sure what to put for the breaking change release note on this one.

pkg/update/client_test.go
```diff
@@ -425,37 +385,12 @@ func TestCheckForUpdates_BehaviorOverTime(t *testing.T) {
 	}
 
 	// Finally we should check once more
-	clock.Advance(3 * time.Second)
+	clock.Advance(4 * time.Second)
```
brianstrauch, line 388: Shouldn't we be able to advance just 1s here? Why do we need 3 or 4?

internal/command.go
```diff
@@ -75,8 +75,7 @@ func NewConfluentCommand(cfg *config.Config) *cobra.Command {
 	cmd.PersistentFlags().CountP("verbose", "v", "Increase verbosity (-v for warn, -vv for info, -vvv for debug, -vvvv for trace).")
 	cmd.PersistentFlags().Bool("unsafe-trace", false, "Equivalent to -vvvv, but also log HTTP requests and responses which might contain plaintext secrets.")
 
-	disableUpdateCheck := cfg.DisableUpdates || cfg.DisableUpdateCheck
-	updateClient := update.NewClient(pversion.CLIName, disableUpdateCheck)
+	updateClient := update.NewClient(pversion.CLIName, cfg)
```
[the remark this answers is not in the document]
sgagniere: Turns out that the argument wasn't even being used by NewClient...

brianstrauch (approved):
Thanks!

brianstrauch:
> Not really sure what to put for the breaking change release note on this one.

No release notes needed, IMO! The only user-facing changes are low-level enough that users shouldn't need to care about them. In retrospect, this totally could've been part of a minor release. I think the reason it was included in v4 is because originally we were thinking of changing `.confluent/config.json` to `.confluent-config.json` or something.

PR #2829, opened 2024-06-26
[CLI-2622] Update flags for `schema-registry schema compatibility validate`

sgagniere:
Release Notes
-------------

Breaking Changes
- The `--schema` flag for `confluent schema-registry schema compatibility validate` has been replaced with a required argument
- The `--subject` flag is now required for `confluent schema-registry schema compatibility validate`; the `--type` flag is additionally required for the Cloud version of this command

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
- Make `--schema` a required argument instead of a flag
- Make `--subject` required
- Make `--type` required for Cloud

The Cloud backend returns an unprocessable entity error when the `--type` isn't specified. Schema type is optional for on-prem according to the on-prem API docs, though (since it defaults to AVRO).

The ticket also mentions making `--version` default to latest, but I think we'll lose functionality this way. Omitting `--version` calls `POST /compatibility/subjects/(string: subject)/versions`, which checks compatibility against all versions.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/schema-registry/command_schema_compatibility_validate.go
+++ b/internal/schema-registry/command_schema_compatibility_validate.go
@@ -53,7 +53,9 @@ func (c *command) newSchemaCompatibilityValidateCommand(cfg *config.Config) *cob
 	cobra.CheckErr(cmd.MarkFlagFilename("references", "json"))
 
 	cobra.CheckErr(cmd.MarkFlagRequired("subject"))
-	cobra.CheckErr(cmd.MarkFlagRequired("type"))
+	if cfg.IsCloudLogin() {
+		cobra.CheckErr(cmd.MarkFlagRequired("type"))
+	}
 
 	return cmd
 }
--- a/internal/schema-registry/command_schema_compatibility_validate.go
+++ b/internal/schema-registry/command_schema_compatibility_validate.go
@@ -37,8 +37,8 @@ func (c *command) newSchemaCompatibilityValidateCommand(cfg *config.Config) *cob
 	}
 	cmd.Example = examples.BuildExampleString(example)
 
-	pcmd.AddSchemaTypeFlag(cmd)
 	cmd.Flags().String("subject", "", subjectUsage)
+	pcmd.AddSchemaTypeFlag(cmd)
 	cmd.Flags().String("version", "", `Version of the schema. Can be a specific version or "latest".`)
 	cmd.Flags().String("references", "", "The path to the references file.")
 	pcmd.AddContextFlag(cmd, c.CLICommand)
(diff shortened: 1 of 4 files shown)


internal/schema-registry/command_schema_compatibility_validate.go
```diff
@@ -37,7 +37,6 @@ func (c *command) newSchemaCompatibilityValidateCommand(cfg *config.Config) *cob
 	}
 	cmd.Example = examples.BuildExampleString(example)
 
-	cmd.Flags().String("schema", "", "The path to the schema file.")
 	pcmd.AddSchemaTypeFlag(cmd)
 	cmd.Flags().String("subject", "", subjectUsage)
```
brianstrauch, line 41: This needs to go before `--schema-type` since it's always required

test/schema_registry_test.go
```diff
@@ -28,9 +28,9 @@ func (s *CLITestSuite) TestSchemaRegistryCluster() {
 
 func (s *CLITestSuite) TestSchemaRegistryCompatibilityValidate() {
 	tests := []CLITest{
-		{args: fmt.Sprintf("schema-registry schema compatibility validate --subject payments --version 1 --schema %s --environment %s", schemaPath, testserver.SRApiEnvId), fixture: "schema-registry/schema/compatibility/validate.golden"},
-		{args: fmt.Sprintf("schema-registry schema compatibility validate --subject payments --version 1 --schema %s --environment %s -o json", schemaPath, testserver.SRApiEnvId), fixture: "schema-registry/schema/compatibility/validate-json.golden"},
-		{args: fmt.Sprintf("schema-registry schema compatibility validate --subject payments --version 1 --schema %s --environment %s -o yaml", schemaPath, testserver.SRApiEnvId), fixture: "schema-registry/schema/compatibility/validate-yaml.golden"},
+		{args: fmt.Sprintf("schema-registry schema compatibility validate %s --type json --subject payments --version 1 --environment %s", schemaPath, testserver.SRApiEnvId), fixture: "schema-registry/schema/compatibility/validate.golden"},
```
brianstrauch, line 31: How come "JSON" wasn't passed before, but now it's required?

internal/schema-registry/command_schema_compatibility_validate.go
```diff
@@ -21,25 +21,24 @@ type validateOut struct {
 
 func (c *command) newSchemaCompatibilityValidateCommand(cfg *config.Config) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "validate",
+		Use:   "validate <schema-file>",
```
brianstrauch, line 24: Nit: We support full filepaths here, right?
```suggestion
		Use:   "validate <schema-path>",
```

brianstrauch (approved):

PR #2839, opened 2024-07-02
Add prefix for encrypted Windows auth tokens

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -323,10 +323,14 @@ func (c *Config) encryptContextStateTokens(tempAuthToken, tempAuthRefreshToken s
 	}
 
 	// The Confluent Gov environment and the Confluent Platform MDS return a refresh token that does not match `authRefreshTokenRegex` and cannot be distinguished from an already encrypted refresh token.
-	// We prefix encrypted tokens with "AES/GCM/NoPadding" to ensure that they are only encrypted once.
-	isUnencryptedConfluentGov := !strings.HasPrefix(tempAuthRefreshToken, secret.AesGcm) && (strings.Contains(c.Context().PlatformName, "confluentgov.com") || strings.Contains(c.Context().PlatformName, "confluentgov-internal.com"))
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -314,21 +313,16 @@ func (c *Config) encryptContextStateTokens(tempAuthToken, tempAuthRefreshToken s
 		c.Context().GetState().Nonce = nonce
 	}
 
-	if regexp.MustCompile(authTokenRegex).MatchString(tempAuthToken) {
+	// We prefix encrypted tokens with "AES/GCM/NoPadding" to ensure that they are only encrypted once.
+	if !strings.HasPrefix(tempAuthToken, secret.AesGcm) && tempAuthToken != "" {
```
sgagniere, line 317: Should we add a prefix for Windows, too? I think the auth token regex was really the only guard against accidental double encryption on Windows.

sgagniere:
Converting to draft while I figure out the windows test failures

pkg/config/config.go
brianstrauch: Yeah it would be great to stay consistent, but we would need to make sure not to break Windows users' CLIs

brianstrauch: As an aside, I don't think this was a great pattern. Instead of writing logic to avoid double encryption, we should have just made sure encrypt was only called once.

pkg/secret/password_protection_constants.go
```diff
@@ -57,4 +57,5 @@ const (
 const (
 	AesCbc = "AES/CBC/PKCS5Padding"
 	AesGcm = "AES/GCM/NoPadding"
+	DPAPI  = "DPAPI"
```
brianstrauch, line 60: Looks weird but I think this is the "correct" way
```suggestion
	Dpapi  = "DPAPI"
```

brianstrauch (approved):

PR #2841, opened 2024-07-02
Remove `ID` from human and serialized fields

tmalikconfluent:
Release Notes
-------------

Breaking Changes
- 'id' has been removed from most of fields containing 'id' in their name

Checklist
---------

What
----
Remove `ID` from human and serialized fields

References
----------
https://confluentinc.atlassian.net/browse/CLI-3014

Test & Review
-------------
Updated integration tests.


internal/kafka/command_cluster_describe.go
```diff
@@ -35,8 +35,8 @@ type describeStruct struct {
 	Network            string `human:"Network,omitempty" serialized:"network,omitempty"`
 	Status             string `human:"Status" serialized:"status"`
 	Endpoint           string `human:"Endpoint" serialized:"endpoint"`
-	ByokKeyId          string `human:"BYOK Key ID" serialized:"byok_key_id"`
-	EncryptionKeyId    string `human:"Encryption Key ID" serialized:"encryption_key_id"`
```
brianstrauch, line 39: These are both external resources (outside of Confluent). There's a big difference between the "BYOK key" and "BYOK key ID" and both are strings. We might want to re-evaluate this change...

tmalikconfluent: okay, let me revert this one back. Any others that need to still contain ID?

internal/flink/command_artifact.go
```diff
@@ -35,8 +35,8 @@ func printTable(cmd *cobra.Command, plugin connectcustompluginv1.ConnectV1Custom
 	table := output.NewTable(cmd)
 	table.Add(&flinkArtifactOut{
 		Name:          plugin.GetDisplayName(),
-		PluginId:      plugin.GetId(),
-		VersionId:     plugin.GetConnectorClass(),
+		Plugin:        plugin.GetId(),
+		Version:       plugin.GetConnectorClass(),
```
brianstrauch, line 39: @byang-confluent, are you ok with the CLI team making this change to improve standardization across our commands?

I'm also curious why "connector class" maps to "version ID". These seem like two separate things.

internal/kafka/command_cluster_describe.go
brianstrauch: A general rule of thumb would be anything containing "Azure", "Aws", or "Gcp" should keep the "ID". For instance, `AzurePrivateLinkServiceId` should stay as is

brianstrauch: or `AwsVpcEndpoint`

internal/plugin/command_list.go
```diff
@@ -16,7 +16,7 @@ import (
 
 type out struct {
 	PluginName string `human:"Name" serialized:"plugin_name"`
```
brianstrauch, line 18:
```suggestion
	Name string `human:"Name" serialized:"name"`
```

internal/plugin/command_list.go
```diff
@@ -16,7 +16,7 @@ import (
...
-	PluginId   string `human:"ID" serialized:"plugin_id"`
+	PluginId   string `human:"ID" serialized:"id"`
```
brianstrauch, line 19:
```suggestion
	Id   string `human:"ID" serialized:"id"`
```

internal/flink/command_artifact.go
byang-confluent: this is a temporary workaround that versioning api is not GA yet, but flink udf requires this information. 

we are working on versioning api this quarter to deliver this. and update cli after that. please keep this setting for now.

tmalikconfluent: reverting this back then

internal/kafka/command_cluster_describe.go
tmalikconfluent: got it , reverting these ones

internal/flink/command_artifact.go
brianstrauch: > please keep this setting for now.

We're not deleting it, we're just asking if we can change the field names from "Plugin ID" to "Plugin" and "Version ID" to "Version".

byang-confluent: yes, that should OK

byang-confluent (approved):

brianstrauch (changes requested):
It doesn't look like these changes match the PR title

PR #2866, opened 2024-09-03
[APIT-893] Schema Registry on Private Link

tmalikconfluent:
Release Notes
-------------

New Features
- Added `Private Endpoint URL` and `Catalog Endpoint URL` to `confluent schema-registry cluster describe`.

Checklist
---------

What
----
Add schema registry on Private Link

References
----------
https://confluentinc.atlassian.net/browse/DGS-8284

Test & Review
-------------
Wrote integration tests and tested manually


internal/schema-registry/command_cluster_describe.go
```diff
@@ -20,7 +20,9 @@ import (
 type clusterOut struct {
 	Name                string `human:"Name" serialized:"name"`
 	Cluster             string `human:"Cluster" serialized:"cluster"`
-	EndpointUrl         string `human:"Endpoint URL" serialized:"endpoint_url"`
+	EndpointUrl         string `human:"Endpoint URL,omitempty" serialized:"endpoint_url"`
+	PrivateEndpointUrl  string `human:"Private Endpoint URL,omitempty" serialized:"private_endpoint_url"`
+	CatalogEndpointUrl  string `human:"Catalog Endpoint URL,omitempty" serialized:"catalog_endpoint_url"`
```
sgagniere, line 25: I think we can add `omitempty` for the serialized versions of the two new fields.

brianstrauch:
> command_cluster_describe.go

Let's update the release notes to say `confluent schema-registry cluster describe`. These release notes are for customers, not CLI engineers

internal/schema-registry/command_cluster_describe.go
tmalikconfluent: should i also add it to the already existing `EndpointUrl`?

brianstrauch:
> Catalog Endpoint

One more release notes nit: I think this should say "Catalog Endpoint URL"

internal/schema-registry/command_cluster_describe.go
brianstrauch: No, I think that would technically be a breaking change. A user might have a script that explicitly reads that field (even if it is `""`) and hiding it might break their script.

sgagniere: Update for @brianstrauch: We decided to add this anyway on the following reasoning: only one will be non-empty at a time, and this change would only affect users who opt in to private (so any script that uses the public endpoint would fail anyway)

brianstrauch: Makes sense!

brianstrauch:
> Omit the Endpoint URL if it is empty.

Don't include this in the "Breaking Changes" section or our automation will release Confluent CLI v5 :)

brianstrauch (approved):

PR #2868, opened 2024-09-04
[FRT-521] Integrate Confluent CLI with new Flink Artifact API

pgaref:
Release Notes
-------------

New Features
- Introduces the new public `Flink Artifact API` to CLI

Checklist
---------

What
----

References
----------

Test & Review
-------------


brianstrauch:
> Migrates CLI to use the latest public Flink Artifact API

If you list this as a breaking change, our automation will release CLI v5 which is not what's intended here. While there might be breaking changes, it looks like the previous functionality was hidden behind a feature flag, so we can consider this a new feature.

pgaref:
> > Migrates CLI to use the latest public Flink Artifact API
> 
> If you list this as a breaking change, our automation will release CLI v5 which is not what's intended here. While there might be breaking changes, it looks like the previous functionality was hidden behind a feature flag, so we can consider this a new feature.

Makes sense @brianstrauch ! Updated :)

channingdong (approved):

PR #2869, opened 2024-09-04
[CLI-2818] Add flink statement update

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- Add `confluent flink statement update` for updating a statement's principal and compute pool, or stopping a statement

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
Moving over from internal.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/iam/command_rbac_role_binding_delete.go
+++ b/internal/iam/command_rbac_role_binding_delete.go
@@ -32,6 +32,13 @@ func (c *roleBindingCommand) newDeleteCommand() *cobra.Command {
 				Code: "confluent iam rbac role-binding delete --principal User:u-123456 --role ResourceOwner --environment env-123456 --kafka-cluster lkc-123456 --resource Topic:my-topic",
 			},
 		)
+	} else {
+		cmd.Example = examples.BuildExampleString(
+			examples.Example{
+				Text: `Delete the role "ResourceOwner" for the resource "Topic:my-topic" on the specified Kafka cluster:`,
+				Code: "confluent iam rbac role-binding delete --principal User:u-123456 --role ResourceOwner --kafka-cluster 0000000000000000000000 --resource Topic:my-topic",
+			},
+		)
 	}
 
 	cmd.Flags().String("role", "", "Role name of the existing role binding.")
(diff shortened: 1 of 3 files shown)


internal/flink/command_statement_update.go
```diff
@@ -0,0 +1,125 @@
...
+		Args:              cobra.ExactArgs(1),
+		ValidArgsFunction: pcmd.NewValidArgsFunction(c.validStatementArgs),
+		RunE:              c.statementUpdate,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Request to update the principal of statement "my-statement" to service account "sa-123456".`,
+				Code: "confluent flink statement update my-statement --principal sa-123456",
+			},
+			examples.Example{
+				Text: `Request to update statement "my-statement" to be run on compute pool "lfcp-123456".`,
```
brianstrauch, line 27: What if the statement is already running? Should we say "Request to move statement..."?

internal/flink/command_statement_update.go
```diff
@@ -0,0 +1,125 @@
...
+				Code: "confluent flink statement update my-statement --compute-pool lfcp-123456",
+			},
+			examples.Example{
+				Text: `Request to resume statement "my-statement".`,
```
brianstrauch, line 31: Let's add an example for how to stop the statement?

pkg/flink/lsp/lsp_completer_ws.go
```diff
@@ -77,10 +77,10 @@ func (w *WebsocketLSPClient) refreshWebsocketConnection() {
 	}
 }
 
-func NewWebsocketClient(getAuthToken func() string, baseUrl, organizationId, environmentId string, handlerCh chan *jsonrpc2.Request) LspInterface {
+func NewWebsocketClient(getAuthToken func() string, baseUrl, organizationId, environmentId string, handlerCh chan *jsonrpc2.Request) (LspInterface, error) {
```
brianstrauch, line 80: Interesting that this is part of this PR's diff

internal/flink/command_statement_update.go
brianstrauch: I noticed we also have `confluent flink statement stop`... are there going to be two ways to do the same thing?

brianstrauch (commented):
Thanks!!

pkg/flink/lsp/lsp_completer_ws.go
sgagniere: Oh, that's because I moved two commits at once; I think it was the commit right before the flink update one. Or maybe the one right after.

PR #2887, opened 2024-09-18
SCP-3030 Plugin versioning cli updated

NicoleYWDu:
Release Notes
-------------
Add `confluent connect custom-plugin version` commands for managing custom plugin versions

New Features
- support `confluent connect custom-plugin version create` 
- support `confluent connect custom-plugin version update` 
- support `confluent connect custom-plugin version list` 
- support `confluent connect custom-plugin version delete` 
- support `confluent connect custom-plugin version describe` 

Checklist
---------
- [ ] Leave this box unchecked if features are not yet available in production

What
----

References
----------

https://confluentinc.atlassian.net/browse/SCP-2167

Test & Review
-------------

PR #2889, opened 2024-09-24
SCP-3030 Add custom plugin version cmd

NicoleYWDu:
Release Notes
-------------
New Features
- [Early Access] Add `confluent connect custom-plugin version` commands for managing custom connect plugin versions

Checklist
---------

What
----
This PR is the follow up for https://github.com/confluentinc/cli/pull/2887.
In the previous PR, there are conflicts in go.mod and pkg/deletion/deletion.go , which are preventing the merge. So create a new PR contains all the change in https://github.com/confluentinc/cli/pull/2887.

What is in this PR:

- contents in https://github.com/confluentinc/cli/pull/2887
- update deletion cmd to adopt y/n
- update plugin version feature flag to cli.custom-connect.plugin.versioning.enabled

References
----------

Test & Review
-------------

### DIFF
--- a/internal/connect/command_custom_plugin_version_create.go
+++ b/internal/connect/command_custom_plugin_version_create.go
@@ -1,0 +1,146 @@
+package connect
+
+import (
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	connectcustompluginv1 "github.com/confluentinc/ccloud-sdk-go-v2/connect-custom-plugin/v1"
+
+	pcmd "github.com/confluentinc/cli/v3/pkg/cmd"
+	"github.com/confluentinc/cli/v3/pkg/examples"
+	"github.com/confluentinc/cli/v3/pkg/output"
+	"github.com/confluentinc/cli/v3/pkg/utils"
+)
+
+type pluginVersionOut struct {
+	Version             string   `human:"Version" serialized:"version"`
+	VersionNumber       string   `human:"Version Number" serialized:"version_number"`
+	IsBeta              string   `human:"Beta" serialized:"is_beta"`
+	ReleaseNotes        string   `human:"Release Notes" serialized:"release_notes"`
+	SensitiveProperties []string `human:"Sensitive Properties" serialized:"sensitive_properties"`
+}
+
+func (c *customPluginCommand) newVersionCreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a custom connector plugin version.",
+		Args:  cobra.NoArgs,
+		RunE:  c.createCustomPluginVersion,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Create custom connector plugin version for plugin "ccp-123456".`,
+				Code: "confluent connect custom-plugin version create --plugin ccp-123456 --plugin-file datagen.zip --version-number 0.0.1 --beta=true --sensitive-properties=passwords,keys,tokens",
+			},
+		),
+	}
+
+	cmd.Flags().String("plugin", "", "ID of custom connector plugin.")
+	cmd.Flags().String("plugin-file", "", "Custom plugin ZIP or JAR file.")
+	cmd.Flags().String("version-number", "", "Version number of custom plugin version.")
+	cmd.Flags().Bool("beta", false, `Mark the custom plugin version as beta. (default "false")`)
+	cmd.Flags().String("release-notes", "", "Release notes for custom plugin version.")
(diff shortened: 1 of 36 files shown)


brianstrauch:
Judging by the feature flag, this is still Early Access, right? Can we prefix the release notes with `[Early Access]`?

internal/connect/command_custom_plugin_version_create.go
```diff
@@ -33,15 +33,15 @@ func (c *customPluginCommand) newVersionCreateCommand() *cobra.Command {
...
+				Code: "confluent connect custom-plugin version create --plugin ccp-123456 --plugin-file datagen.zip --version-number 0.0.1 --sensitive-properties passwords,keys,tokens",
 			},
 		),
 	}
 
 	cmd.Flags().String("plugin", "", "ID of custom connector plugin.")
 	cmd.Flags().String("plugin-file", "", "Custom plugin ZIP or JAR file.")
 	cmd.Flags().String("version-number", "", "Version number of custom plugin version.")
-	cmd.Flags().Bool("beta", false, `Mark the custom plugin version as beta. (default "false")`)
+	cmd.Flags().Bool("beta", false, `Mark the custom plugin version as beta. (default "false").`)
```
brianstrauch, line 44: Do we really need to specify the default here? You would think that it would be implied since it is a boolean flag, after all.

internal/connect/command_custom_plugin_version_create.go
```diff
@@ -33,15 +33,15 @@ func (c *customPluginCommand) newVersionCreateCommand() *cobra.Command {
 		Example: examples.BuildExampleString(
 			examples.Example{
 				Text: `Create custom connector plugin version for plugin "ccp-123456".`,
-				Code: "confluent connect custom-plugin version create --plugin ccp-123456 --plugin-file datagen.zip --version-number 0.0.1 --beta=true --sensitive-properties=passwords,keys,tokens",
```
brianstrauch: What's the reason for removing this from the examples? Do we no longer support it?

NicoleYWDu: Hi Brian, this is to resolve error while running `go run cmd/lint/main.go -aff-file cmd/lint/en_US.aff -dic-file cmd/lint/en_US.dic`
```
For context "Cloud", 6 errors occurred:
	* confluent connect custom-plugin version create: flag `--beta` must not use "=" in example 1
	* confluent connect custom-plugin version create: flag `--sensitive-properties` must not use "=" in example 1
	* flag usage doesn't end with punctuation for --beta on `confluent connect custom-plugin version create`
	* confluent connect custom-plugin version update: flag `--beta` must not use "=" in example 1
	* confluent connect custom-plugin version update: flag `--sensitive-properties` must not use "=" in example 1
	* flag usage doesn't end with punctuation for --beta on `confluent connect custom-plugin version update`

exit status 1
```

internal/connect/command_custom_plugin_version_create.go
NicoleYWDu: As a new user reading the help documentation ⬇️ , I might be confused about the --beta flag. While most other flags clearly show their type as ‘string,’ the --beta flag doesn’t specify a type. However, the default value of false suggests that this flag is likely a boolean. This could serve as a clue for me to understand its purpose.
```
Flags:
      --plugin string                  REQUIRED: ID of custom connector plugin.
      --version string                 REQUIRED: ID of custom connector plugin version.
      --version-number string          Version number of custom plugin version.
      --beta                           Mark the custom plugin version as beta. (default "false").
      --release-notes string           Release notes for custom plugin version.
      --sensitive-properties strings   A comma-separated list of sensitive property names.
      --context string                 CLI context name.
```

internal/connect/command_custom_plugin_version_create.go
brianstrauch: cmd/lint/main.go allows you to override some of these lints if they don't make sense. For instance, we necessarily need to write `--beta=true` here. Don't sacrifice the quality of the examples to make the linter pass!!

internal/connect/command_custom_plugin_version_create.go
brianstrauch: I think the examples are a great way to communicate the usage of these flags! By the way, all of our other boolean flags have this formatting throughout the CLI. I don't think we necessarily need to make this flag a special case.

brianstrauch: If you disagree, feel free to keep the `(default "false")` here, but let's remove the period at the end to match the output of our other commands

NicoleYWDu: Gotcha, thank you! I totally agree. Previous I thought if we didn't have the beta example (to pass the lint in order to merge pr) it would be confusing.

brianstrauch (approved):

PR #2901, opened 2024-10-08
KGLOBAL-5631: CLI to only have one column of "Remote Cluster Id" and remove "Destination Cluster Id" and "Source Cluster Id" for list link

luyixiao12:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
JIRA ticket: https://confluentinc.atlassian.net/browse/KGLOBAL-5631
- For list link with cluster name, since destination cluster and local cluster is not useful, we only show remote cluster in this case

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/cmd/docs/main.go
+++ b/cmd/docs/main.go
@@ -62,19 +62,12 @@ func main() {
 		panic(err)
 	}
 
-	removeUnreleasedDocs()
-
 	if err := os.Setenv("HOME", home); err != nil {
 		panic(err)
 	}
 }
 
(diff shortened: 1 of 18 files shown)


sgagniere (changes requested):
Removing output fields is a breaking change, so we'll need to wait for v5 for this one.

sgagniere:
Converting to draft since this needs to go in v5.

PR #2926, opened 2024-11-05
Adds support for output command in env

vsantwana:
Release Notes
-------------
This change fixes two major issues:
- `flink environment create`, `flink environment update` and `flink environment describe` did not have an `--output` flag and always printed human output. With this change, we are introducing JSON and YAML output as well.
- So far, even though the `flink application create`, `flink application update` and `flink application describe` command did not have support for `human` output, the help section used to show it like `Specify the output format as "human", "json", or "yaml". (default "json")` which gives false hopes. With this PR, we are  fixing that by changing the registration method. 

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Adds `output` flag support to `flink environment create`, `flink environment describe` and `flink environment update` commands
- Remove misleading help from application commands for human output

Checklist
---------

What
----

References
----------

Test & Review
-------------

### DIFF
--- a/internal/flink/command_environment_create.go
+++ b/internal/flink/command_environment_create.go
@@ -11,6 +11,7 @@ import (
 
 	cmfsdk "github.com/confluentinc/cmf-sdk-go/v1"
 
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
 	"github.com/confluentinc/cli/v4/pkg/output"
 )
 
@@ -26,6 +27,7 @@ func (c *command) newEnvironmentCreateCommand() *cobra.Command {
 	cmd.Flags().String("defaults", "", "JSON string defining the environment's Flink application defaults, or path to a file to read defaults from (with .yml, .yaml or .json extension).")
 
 	addCmfFlagSet(cmd)
+	pcmd.AddOutputFlag(cmd)
 
 	cobra.CheckErr(cmd.MarkFlagRequired("kubernetes-namespace"))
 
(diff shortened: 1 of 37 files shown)


internal/flink/command_environment_create.go
```diff
@@ -26,6 +27,7 @@ func (c *command) newEnvironmentCreateCommand() *cobra.Command {
 	cmd.Flags().String("defaults", "", "JSON string defining the environment's Flink application defaults, or path to a file to read defaults from (with .yml, .yaml or .json extension).")
 
 	addCmfFlagSet(cmd)
+	pcmd.AddOutputFlag(cmd)
```
sgagniere, line 30: I think we might want to create another version of this flag that just has json and yaml outputs.

That way, the flag description won't show human output as option, and the command won't error by default.

vsantwana: Environment allows all the formats i.e JSON, YAML and human.
For application, I have created another registration function `AddOutputFlagWithHumanRestricted` where the allowed values are only JSON and YAML. This ensures that for application command we do not print help like 

```
Specify the output format as "human", "json", or "yaml". (default "json")
``` 
when we do not support human

vsantwana: Logic behind the decision to continue `human` output for env commands: The version included in CP release has only human output create, update and describe environment
While I personally don't like the human output, I would like to not remove things in the future version of the CLI (unless it is v5), so with the change I am making I will simply add support for JSON and YAML and not take away what is already there

vsantwana: @sgagniere I have updated the description, please take a look

sgagniere (approved):

PR #2940, opened 2024-11-14
Update pipeline and linter

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
- Clone the golang-fips/go release matching the current .go-version in the pipeline to catch currently unsupported go versions
- Compare the `.go-version` and the `go.mod` version in the lint target to make sure they match.
- - We don't necessarily need them to match, but we can't have the go.mod version being higher than the .go-version or else we'll have issues with go-licenses

References
----------

Test & Review
-------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,9 @@ lint: lint-go lint-cli
 
 .PHONY: lint-go
 lint-go:
+ifneq ($(shell cat .go-version),$(shell go list -m -f '{{.GoVersion}}'))
+	$(error The go versions in go.mod and .go-version must match)
+endif
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.0 && \
 	golangci-lint run --timeout 10m
 	@echo "✅  golangci-lint"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -79,6 +79,9 @@ lint: lint-go lint-cli
 
 .PHONY: lint-go
 lint-go:
+ifneq ($(shell cat .go-version),$(shell go list -m -f '{{.GoVersion}}'))
+	$(error The go versions in go.mod and .go-version must match)
+endif
```
brianstrauch, line 84: Make a note that when ServiceBot does this for us, we can remove this check?

sgagniere: I don't think we want to remove it. The problem this is designed to block is someone making a manual update to the go.mod version without also changing the .go-version.

If the service bot updates both at once, there's no issue.

PR #2955, opened 2024-11-23
[CLI-3337] Map single-zone and multi-zone to high and low for freight clusters

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- PLACEHOLDER

Checklist
---------

What
----
For freight clusters, `LOW` and `SINGLE_ZONE` are not the same. `LOW` is acceptable but `SINGLE_ZONE` is not. For freight clusters only, we want to map single-zone and multi-zone to low and high.

References
----------

Test & Review
-------------

### DIFF
--- a/internal/kafka/command_cluster.go
+++ b/internal/kafka/command_cluster.go
@@ -14,16 +14,25 @@ const (
 	multiZone        = "multi-zone"
 	lowAvailability  = "SINGLE_ZONE"
 	highAvailability = "MULTI_ZONE"
+	low              = "LOW"
+	high             = "HIGH"
 )
 
 var availabilitiesToHuman = map[string]string{
 	lowAvailability:  singleZone,
 	highAvailability: multiZone,
+	low:              singleZone,
+	high:             multiZone,
 }
 
 var availabilitiesToModel = map[string]string{
 	singleZone: lowAvailability,
 	multiZone:  highAvailability,
+}
+
+var availabilitiesToFreightModel = map[string]string{
+	singleZone: low,
+	multiZone:  high,
 }
 
 type clusterCommand struct {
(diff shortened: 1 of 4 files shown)


internal/kafka/command_cluster.go
```diff
@@ -14,18 +14,27 @@ const (
...
 	highAvailability = "MULTI_ZONE"
+	low              = "LOW"
+	high             = "HIGH"
 )
 
 var availabilitiesToHuman = map[string]string{
 	lowAvailability:  singleZone,
 	highAvailability: multiZone,
+	low:              singleZone,
+	high:             multiZone,
```
sgagniere, line 25: Since this is meant to be an under-the-hood change for now, these lines are to map "LOW" and "HIGH" to the current "single-zone" and "multi-zone" for output

internal/kafka/command_cluster_create.go
```diff
@@ -163,7 +163,12 @@ func (c *clusterCommand) create(cmd *cobra.Command, args []string) error {
 	return c.outputKafkaClusterDescription(cmd, &kafkaCluster, false)
 }
 
-func stringToAvailability(s string) (string, error) {
+func stringToAvailability(s string, sku ccstructs.Sku) (string, error) {
```
channingdong, line 166: Should here be a if - else statement?

```
if sku == ccstructs.Sku_FREIGHT {
	if modelAvailability, ok := availabilitiesToFreightModel[s]; ok {
		return modelAvailability, nil
	}
}
else if modelAvailability, ok := availabilitiesToModel[s]; ok {
	return modelAvailability, nil
}
```

channingdong (approved):

PR #2958, opened 2024-11-27
DGS-19209 AsyncApi export - failing to read from topic and output correct data

kcorman0:
Release Notes
-------------

Bug Fixes
- Prevent `confluent asyncapi export` from always failing when `--consume-examples` is passed

Checklist
---------

What
----

Consume examples flag for asyncapi export had some issues, fixed
- Closing Kafka consumer too early
- Passing incorrect type to schema deserializer

References
----------

Test & Review
-------------

Manually tested to confirm `--consume-examples` works

### DIFF
--- a/internal/asyncapi/command_export.go
+++ b/internal/asyncapi/command_export.go
@@ -111,7 +111,9 @@ func (c *command) export(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	defer accountDetails.consumer.Close()
+	if flags.consumeExamples {
+		defer accountDetails.consumer.Close()
+	}
 	// Servers & Info Section
 	reflector := addServer(accountDetails.kafkaUrl, accountDetails.schemaRegistryUrl, flags.specVersion)
 	log.CliLogger.Debug("Generating AsyncAPI specification")


internal/asyncapi/command_export.go
```diff
@@ -275,7 +278,16 @@ func (c *command) getMessageExamples(consumer *ckgo.Consumer, topicName, content
...
+	token, err := auth.GetDataplaneToken(c.Context)
+	if err != nil {
+		return nil, err
+	}
+	srClusterId, srEndpoint, err := c.GetCurrentSchemaRegistryClusterIdAndEndpoint()
+	if err != nil {
+		return nil, err
+	}
+
+	err = deserializationProvider.InitDeserializer(srEndpoint, srClusterId, "value", "", "", token, nil)
```
brianstrauch, line 290: Hopefully we don't need to pass anything for these other two `""` values. It's hard to tell what's going on when a function takes 7 arguments...

brianstrauch (approved):

internal/asyncapi/command_export.go
kcorman0: Lol +1. It's for apikey / secret, alternative to sending token. Should be separate call I guess

PR #2971, opened 2024-12-12
fix(am-5030): IP Filtering CLI-gate unreleased functionality behind a feature flag

confluent-alec-kendall:
Checklist
---------

- [X] I have successfully built and used a custom CLI binary, with no linting issues identified from this PR.
- [X] I have clearly specified in the `Release Notes` section above whether this PR applies to Confluent Cloud, Confluent Platform, or both. 
- [N/A] I have verified this PR with Confluent Cloud resources in a production environment, if applicable. (N/A as feature flag not turned on in prod). 
- [N/A] I have verified this PR with Confluent Platform resources in an on-premises environment, if applicable.
- [X] I have attached manual CLI verification results or screenshots in the `Test & Review` section below.
- [X] I have added appropriate CLI integration tests for any new or updated commands and functionality.
- [X] I confirm that this PR introduces no breaking changes or backward compatibility issues.
- [X] I have documented the blast radius if something goes unexpectedly by this PR.
- [ ] Leave this box unchecked if features introduced in this PR are not yet available in production.

What
----
1. Fixes default behaviors: i.e resource-group should default to management if flag is turned off, operation groups should do the same. if the flag is turned on, resource group should default to "multiple". 
2. Updates the golden files for integration tests which were failing before + includes cfg.IsTest for all ipSrEnabled checks. Looks like the integration tests were not set up correctly previously and were expecting the LD flag to be off, when they should be testing new behavior (with the flag turned on). 
3. Adds examples if ipSrEnabled is true which are relevant to new functionality. 

Blast Radius
----------
Customers could face issues with iam ip-filter list, create, update commands with CLI. Ip filtering is still of course available via UI. 

References
----------
https://confluentinc.atlassian.net/browse/AM-5030
Follow up to this PR: https://github.com/confluentinc/airlock-cli/pull/170 

Test & Review
-------------
See: https://confluentinc.atlassian.net/wiki/spaces/PM/pages/3879862450/CLI+Bug+Bash for all test cases with LD flag turned on and off.


internal/iam/command_ipfilter_list.go
```diff
@@ -66,14 +67,17 @@ func (c *ipFilterCommand) list(cmd *cobra.Command, _ []string) error {
...
+			ID:            filter.GetId(),
+			Name:          filter.GetFilterName(),
+			ResourceGroup: filter.GetResourceGroup(),
+			IpGroups:      convertIpGroupObjectsToIpGroupIds(filter),
+		}
+		if isSrEnabled {
+			filterOut.ResourceScope = filter.GetResourceScope()
+			filterOut.OperationGroups = filter.GetOperationGroups()
+		}
+		list.Add(filterOut)
```
swathirameshcflt, line 80: this needs to be list.Add(&filterOut), otherwise this breaks the whole list command!

internal/iam/command_ipfilter_create.go
```diff
@@ -19,29 +19,38 @@ import (
...
-	pcmd.AddResourceGroupFlag(cmd)
-	err := cmd.MarkFlagRequired("resource-group")
-	if err != nil {
-		return nil
+	cmd := &cobra.Command{
+		Use:     "create <name>",
+		Short:   "Create an IP filter.",
+		Args:    cobra.ExactArgs(1),
+		RunE:    c.create,
+		Example: exampleString,
```
sgagniere, line 42: Let's rearrange this to follow the pattern used in other parts of the CLI (for example `internal/iam/command_rbac_role_binding_delete.go`:
```
cmd := &cobra.Command{
    ...
}

isSrEnabled := ...

if isSrEnabled {
    cmd.Example = examples.BuildExampleString{...}
} else {
    cmd.Example = examples.BuildExampleString{...}
}
```

This way, we can keep the core cmd definition at the top.

internal/iam/command_ipfilter_create.go
```diff
@@ -19,29 +19,38 @@ import (
...
 	cmd.Flags().StringSlice("ip-groups", []string{}, "A comma-separated list of IP group IDs.")
-	if cfg.IsTest || (cfg.Context() != nil && featureflags.Manager.BoolVariation("auth.ip_filter.sr.cli.enabled", cfg.Context(), featureflags.GetCcloudLaunchDarklyClient(cfg.Context().PlatformName), true, false)) {
+	if isSrEnabled {
 		cmd.Flags().String("environment", "", "Name of the environment for which this filter applies. By default will apply to the organization only.")
 		cmd.Flags().StringSlice("operations", nil, fmt.Sprintf("A comma-separated list of operation groups: %s.", utils.ArrayToCommaDelimitedString([]string{"MANAGEMENT", "SCHEMA"}, "or")))
 		cmd.Flags().Bool("no-public-networks", false, "Use in place of ip-groups to reference the no public networks IP Group.")
 		cmd.MarkFlagsMutuallyExclusive("ip-groups", "no-public-networks")
+		cmd.MarkFlagsMutuallyExclusive("resource-group", "operations")
+	} else {
+		cobra.CheckErr(cmd.MarkFlagRequired("ip-groups"))
```
sgagniere, line 53: Just to double check: when the feature flag is not enabled, are API requests without ip-groups are expected to fail? (In other words, I want to double check that this is not a breaking change)

internal/iam/command_ipfilter_list.go
```diff
@@ -66,14 +67,17 @@ func (c *ipFilterCommand) list(cmd *cobra.Command, _ []string) error {
...
-		})
+		filterOut := ipFilterOut{
+			ID:            filter.GetId(),
+			Name:          filter.GetFilterName(),
+			ResourceGroup: filter.GetResourceGroup(),
+			IpGroups:      convertIpGroupObjectsToIpGroupIds(filter),
+		}
+		if isSrEnabled {
+			filterOut.ResourceScope = filter.GetResourceScope()
+			filterOut.OperationGroups = filter.GetOperationGroups()
```
sgagniere, line 78: Does this also need to be sorted, like in the describe/create/update output?

sgagniere (commented):
Hi, some comments/questions

internal/iam/command_ipfilter_create.go
swathirameshcflt: Yes it's not breaking, I removed it here https://github.com/confluentinc/cli/pull/2929/files#diff-fb319f5596bbb5275cde35c7561e1a57508d5e44188bc23824d7b372407ef94dL34 incorrectly, if the feature flag is turned off --ip-groups should still be a required parameter

swathirameshcflt: They would fail without this passed in regardless due to the way the API behaves

internal/iam/command_ipfilter_list.go
swathirameshcflt: not sure if it's necessary but I'll add it in any case!

internal/iam/command_ipfilter_create.go
swathirameshcflt: Yep done

sgagniere:
Looks good to me; pinging @channingdong for a second pair of eyes.

sgagniere (approved):

PR #3014, opened 2025-01-27
AM-5132: Update RBAC CLI List command to reflect new RBAC List API changes.

AydenAdair:
Bug Fixes
---------
- Ensures all public and production ready RBAC roles are available via the RBAC List command.

Checklist
---------

- [x] I have verified this PR in Confluent Platform on-premises environment, if applicable.
  - [x] Confluent Cloud prod
  - [x] Confluent Cloud stag
  - [x] Confluent Platform

What
----
Some production-ready roles were hidden in the Confluent CLI due to existing outside the explicit namespace list in our CLI code. The [RBAC Roles List API](https://github.com/confluentinc/cc-rbac-services/pull/1643) has been updated so that it will now return all public and production ready roles by default without needing a specific namespace parameter entered. This will allow the CLI to automatically show the correct list of public RBAC roles without needing continuous LD changes or code updates.

Blast Radius
----
The worst-case blast radius here would be a failure to see the correct roles when using the RBAC list command. This would not impact the ability to view existing roles assigned in the organization, or assign new roles to principals in the organization. 

References
----------
https://confluentinc.atlassian.net/browse/AM-5132

Test & Review
-------------
The RBAC List API has been tested with the same parameters being passed into the CLI command here, and the results have been verified to work as expected. As an example, curling the API with an empty namespace parameter and querying for flink provides:
```
            "resourceType": "OwnFlinkEnvironmentRegionApiKey",
            "resourceType": "FlinkEnvironmentRegionApiKey",
            "resourceType": "FlinkArtifact",
            "resourceType": "OwnFlinkEnvironmentRegionApiKey",
            "resourceType": "FlinkArtifact",
            "resourceType": "FlinkEnvironmentRegionApiKey",
    "name": "FlinkAdmin",
            "resourceType": "FlinkEnvironmentRegionApiKey",
            "resourceType": "FlinkArtifact",
            "resourceType": "OwnFlinkEnvironmentRegionApiKey",
              "CreateFlinkStatement",
              "CreateFlinkComputePool"
    "name": "FlinkDeveloper",
              "CreateFlinkStatement"
            "resourceType": "OwnFlinkEnvironmentRegionApiKey",
            "resourceType": "FlinkArtifact",
```
We can see all the respective Flink resources and roles appear by default without additional namespace parameters.

Similarly, when compiling the binary locally and running against stag, I can see the new roles appearing.

### Integration test changes
The [integration test uses a mocked response](https://github.com/confluentinc/cli/blob/3fc90b15e827aed2fc1a59ed7ab0654aba722225/test/test-server/cloud_mds_handlers.go#L112) that simulates a few roles and their associated permissions. To update this to today's roles, it would require manually updating this test with over 4,000 new lines of roles and permissions. Since this is a mock, it is not useful to keep it perpetually up-to-date with the latest role binding information. Instead, this just checks the output format of the roles. This update to the test shows that passing in an empty string, which is the new default behavior of the CLI, still returns the same roles. This validates that the user experience of using the RBAC CLI for the List command is unchanged.

The most useful way to ensure these changes work as expected with the real API is by manually checking with the CLI binary, which we did. Checking that a mock response meets our expectation doesn't actually validate the API changes in any way, so it's important we do both.

### DIFF
--- a/test/test-server/cloud_mds_handlers.go
+++ b/test/test-server/cloud_mds_handlers.go
@@ -3,7 +3,6 @@ package testserver
 import (
 	"encoding/json"
 	"net/http"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


tadsul (commented):
Can we please add some test  to ensure its working as expected

AydenAdair:
> Can we please add some test to ensure its working as expected

I updated the description with my testing reasoning, and the rational behind not changing the `list-cloud` integration test.

test/test-server/cloud_mds_handlers.go
```diff
@@ -13,33 +12,39 @@ import (
...
 	}
 }
 
+func generateAllRolesList() []mdsv2alpha1.Role {
+	var allRoles []mdsv2alpha1.Role
+	/*
+	 * This is the specific order in which roles are added to the list
+	 * during our mocked tests. It includes the same namespace multiple
+	 * times due to some legacy behavior, which prints the same role more
+	 * than once.
```
brianstrauch, line 28: > It includes the same namespace multiple times due to some legacy behavior, which prints the same role more than once.

Even though it's legacy behavior, and it sounds like you're trying to avoid making a breaking change, this is still a bug. Can we clean this up now?

test/test-server/cloud_mds_handlers.go
```diff
@@ -13,33 +12,39 @@ import (
...
+	 *
+	 * TODO:
+	 * In the future, this should be cleaned up. For now, this is intentionally
```
brianstrauch, line 31: In the future, do you think the CLI could grab this list of roles from an endpoint or feature flag so we don't have to keep using `append()` every time there's a new set of roles?

AydenAdair: Similar to my above comment - this is just a weird implementation detail of our mock server for the integration test. The actual API call grabs all roles from our list endpoint, and does not have to do any weird `append()` logic. I'll update this test suite to reflect this.

AydenAdair: These CLI tests were built under the assumption that the user would/could query for different role namespaces via the CLI. That is no longer true today, the CLI just returns all `public` roles, and there is no option to query for individual namespaces. The tests were never updated to reflect that, though.

test/test-server/cloud_mds_handlers.go
brianstrauch: > This duplication bug only happens in our mocked integration tests.

Ah, got it!

brianstrauch (approved):

PR #3035, opened 2025-02-21
Add options for LOW/HIGH availability in cluster create CLI commands

danielayaz:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- PLACEHOLDER

Bug Fixes
- Fixed "For `confluent kafka cluster create command`, it showed as single-zone/multi-zone availability instead of high/low for Freight clusters"

Checklist
---------

  - [ ] Confluent Cloud prod
  - [ ] Confluent Cloud stag
  - [ ] Confluent Platform

What
----

DOCS-28660: Add new availability type attribute value options (LOW/HIGH) for Enterprise clusters
Done
 

The availability attribute of multi-tenant clusters was changed to LOW/HIGH , and earlier it used to be SINGLE_ZONE, & MULTI_ZONE 

Different cluster types have different valid options and even the same cluster type might have different valid options based on when the org was created.

LOW/HIGH options for Standard and Enterprise clusters created after 4/16

LOW option only for Basic clusters created after 4/16

The CLI and Terraform interfaces should support four availability options - LOW, HIGH, SINGLE_ZONE, MULTI_ZONE  and let the backend APIs do the final validation

Blast Radius
----

There are no breaking changes. Changes are localized to the availability flag of confluent kafka cluster create

References
----------

https://confluentinc.atlassian.net/browse/CLI-3152

Test & Review
-------------

**High**:
￼
Screenshot 2025-02-21 at 4 10 09 PM

**LOW**:
￼
Screenshot 2025-02-21 at 4 11 09 PM

**Single-zone:**
￼
Screenshot 2025-02-21 at 4 11 41 PM

**Multi-zone**:
￼
Screenshot 2025-02-21 at 4 12 02 PM

**All in UI:**
Screenshot 2025-02-21 at 4 15 24 PM

￼

### DIFF
--- a/internal/kafka/command_cluster.go
+++ b/internal/kafka/command_cluster.go
@@ -2,37 +2,31 @@ package kafka
 
 import (
 	"fmt"
-
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
 	"github.com/confluentinc/cli/v4/pkg/config"
 )
 
-const (
-	singleZone       = "single-zone"
-	multiZone        = "multi-zone"
-	lowAvailability  = "SINGLE_ZONE"
-	highAvailability = "MULTI_ZONE"
-	low              = "LOW"
-	high             = "HIGH"
-)
-
 var availabilitiesToHuman = map[string]string{
-	lowAvailability:  singleZone,
-	highAvailability: multiZone,
-	low:              singleZone,
-	high:             multiZone,
+	"SINGLE_ZONE": "single-zone",
+	"MULTI_ZONE":  "multi-zone",
+	"LOW":         "low",
+	"HIGH":        "high",
 }
 
 var availabilitiesToModel = map[string]string{
-	singleZone: lowAvailability,
-	multiZone:  highAvailability,
+	"single-zone": "SINGLE_ZONE",
+	"multi-zone":  "MULTI_ZONE",
+	"low":         "LOW",
+	"high":        "HIGH",
 }
 
 var availabilitiesToFreightModel = map[string]string{
-	singleZone: low,
-	multiZone:  high,
+	"single-zone": "LOW",
+	"multi-zone":  "HIGH",
+	"low":         "LOW",
+	"high":        "HIGH",
 }
 
 type clusterCommand struct {
--- a/internal/kafka/command_cluster.go
+++ b/internal/kafka/command_cluster.go
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"fmt"
+
 	"github.com/spf13/cobra"
 
 	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
@@ -16,8 +17,8 @@ var availabilitiesToModel = map[string]string{
 }
 
 var availabilitiesToFreightModel = map[string]string{
-	"single-zone": "LOW",  // TODO: This mapping are deprecated and will be removed in v5
-	"multi-zone":  "HIGH", // TODO: This mapping are deprecated and will be removed in v5
+	"single-zone": "LOW",  // TODO: This mapping is deprecated and will be removed in v5
+	"multi-zone":  "HIGH", // TODO: This mapping is deprecated and will be removed in v5
 	"low":         "LOW",
 	"high":        "HIGH",
 }
--- a/internal/kafka/command_cluster.go
+++ b/internal/kafka/command_cluster.go
@@ -23,8 +23,8 @@ var availabilitiesToModel = map[string]string{
 }
 
 var availabilitiesToFreightModel = map[string]string{
-	"single-zone": "LOW",
-	"multi-zone":  "HIGH",
+	"single-zone": "LOW",  // TODO: This mapping are deprecated and will be removed in v5
+	"multi-zone":  "HIGH", // TODO: This mapping are deprecated and will be removed in v5
 	"low":         "LOW",
 	"high":        "HIGH",
 }
--- a/internal/kafka/command_cluster.go
+++ b/internal/kafka/command_cluster.go
@@ -8,13 +8,6 @@ import (
 	"github.com/confluentinc/cli/v4/pkg/config"
 )
 
-var availabilitiesToHuman = map[string]string{
-	"SINGLE_ZONE": "single-zone",
-	"MULTI_ZONE":  "multi-zone",
-	"LOW":         "low",
-	"HIGH":        "high",
-}
-
 var availabilitiesToModel = map[string]string{
 	"single-zone": "SINGLE_ZONE",
 	"multi-zone":  "MULTI_ZONE",
--- a/internal/kafka/command_cluster_create.go
+++ b/internal/kafka/command_cluster_create.go
@@ -44,6 +44,10 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 				Code: "confluent kafka cluster create my-cluster --cloud aws --region us-west-2 --type dedicated --cku 1 --byok cck-a123z",
 			},
 			examples.Example{
+				Text: "Create a new freight cluster that uses a customer-managed encryption key in AWS:",
+				Code: "confluent kafka cluster create my-cluster --cloud aws --region us-west-2 --type freight --cku 1 --byok cck-a123z --availability high",
+			},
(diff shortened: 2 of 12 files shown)


internal/kafka/command_cluster.go
```diff
@@ -2,37 +2,31 @@ package kafka
...
 var availabilitiesToHuman = map[string]string{
-	lowAvailability:  singleZone,
-	highAvailability: multiZone,
-	low:              singleZone,
-	high:             multiZone,
+	"SINGLE_ZONE": "single-zone",
+	"MULTI_ZONE":  "multi-zone",
+	"LOW":         "low",
+	"HIGH":        "high",
 }
```
brianstrauch, line 16: We can replace this map with `ccloudv2.ToLower()` now 😎

internal/kafka/command_cluster.go
```diff
@@ -2,37 +2,31 @@ package kafka
...
 
 var availabilitiesToModel = map[string]string{
-	singleZone: lowAvailability,
-	multiZone:  highAvailability,
+	"single-zone": "SINGLE_ZONE",
+	"multi-zone":  "MULTI_ZONE",
+	"low":         "LOW",
+	"high":        "HIGH",
 }
```
brianstrauch, line 23: `ccloudv2.ToUpper()`

internal/kafka/command_cluster.go
```diff
@@ -2,37 +2,31 @@ package kafka
...
 
 var availabilitiesToFreightModel = map[string]string{
-	singleZone: low,
-	multiZone:  high,
+	"single-zone": "LOW",
+	"multi-zone":  "HIGH",
```
brianstrauch, line 27: Write a comment here saying that these two mappings are deprecated and will be removed in v5?

brianstrauch: (And make a v5 Jira ticket for this if you haven't already)

internal/kafka/command_cluster_create.go
```diff
@@ -43,6 +43,10 @@ func (c *clusterCommand) newCreateCommand() *cobra.Command {
 				Text: "Create a new dedicated cluster that uses a customer-managed encryption key in AWS:",
 				Code: "confluent kafka cluster create my-cluster --cloud aws --region us-west-2 --type dedicated --cku 1 --byok cck-a123z",
 			},
+			examples.Example{
+				Text: "Create a new freight cluster that uses a customer-managed encryption key in AWS:",
```
brianstrauch, line 47: Nit: Capitalize "Freight" since it's a product name / proper noun

internal/kafka/command_cluster_create.go
```diff
@@ -175,7 +179,7 @@ func stringToAvailability(s string, sku ccstructs.Sku) (string, error) {
 	}
 	return "", errors.NewErrorWithSuggestions(
 		fmt.Sprintf("invalid value \"%s\" for `--availability` flag", s),
-		fmt.Sprintf("Allowed values for `--availability` flag are: %s, %s.", singleZone, multiZone),
+		fmt.Sprintf("Allowed values for `--availability` flag are: %s, %s, %s, %s.", "single-zone", "multi-zone", "low", "high"),
```
brianstrauch, line 182: If you want to be extra user-friendly, this should only say "low" and "high" if `--type freight` is passed

brianstrauch: We also have `utils.ArrayToCommaDelimitedString()` that you could use here

brianstrauch (changes requested):
Nice job!

internal/kafka/command_cluster.go
danielayaz: @brianstrauch It's not exactly a .ToLower(). If you see here it turns an underscore (_) into a hypen (-)

```go
"SINGLE_ZONE": "single-zone" 
```

internal/kafka/command_cluster.go
danielayaz: > It's not exactly a .ToLower(). If you see here it turns an underscore (_) into a hypen (-)
> 
> ```go
> "SINGLE_ZONE": "single-zone" 
> ```

Same situation as above

internal/kafka/command_cluster_create.go
danielayaz: I agree that is more user-friendly, I added the change @brianstrauch.

I decided against using `utils.ArrayToCommaDelimitedString()` as it it would be a bit more complicated than it already is imo.

internal/kafka/command_cluster_create.go
danielayaz: Done, thanks!

internal/kafka/command_cluster.go
danielayaz: Added and I have created a [Jira ticket](https://confluentinc.atlassian.net/browse/CLI-3414).

internal/kafka/command_cluster.go
brianstrauch: `ccloudv2.ToLower()` will do that for you! (You might be thinking of `string.ToLower()` which won't)

internal/kafka/command_cluster_create.go
brianstrauch: All of our other flag descriptions put quotes around the options and include an "and" or "or", and `utils.ArrayToCommaDelimitedString()` will make it easy to conform to that style (and potentially change that style in the future across all commands, if we want to)

linouk23: > If you want to be extra user-friendly, this should only say "low" and "high" if --type freight is pass

I'm not sure whether it's helpful but they're applicable for Enterprise clusters too: https://confluentinc.atlassian.net/browse/DOCS-28660

danielayaz: @brianstrauch After reviewing the other instances, I agree now that it's a good pattern that we use `utils.ArrayToCommaDelimitedString() ` in this case. Please see the latest.

internal/kafka/command_cluster.go
danielayaz: I definitely assumed that was just our own duplicate function of `string.ToLower()` :) I changed it to use `ccloudv2.ToLower()`

internal/kafka/command_cluster.go
danielayaz: @brianstrauch I think `ccloudv2.ToUpper()` might not be right here as we actually need the validation from the ok value returned when the map is accessed to know if the flag passed in was a valid value. 

```go

if modelAvailability, ok := availabilitiesToModel[s]; ok {
	return modelAvailability, nil
}
return "", errors.NewErrorWithSuggestions(
	fmt.Sprintf("invalid value \"%s\" for `--availability` flag", s),
	fmt.Sprintf("Allowed values for `--availability` flag are: %s.", utils.ArrayToCommaDelimitedString(allowedEnterpriseAvailabilities, "or")),
)

```

brianstrauch: That's fair. It may be worth checking if the backend is able to perform the validation, and then we could get rid of this map lookup. But this is fine for now.

internal/kafka/command_cluster.go
```diff
@@ -2,37 +2,24 @@ package kafka
 
 import (
 	"fmt"
-
```
brianstrauch, line 5: I think `make lint` won't like this

internal/kafka/command_cluster.go
```diff
@@ -2,37 +2,24 @@ package kafka
...
+	"multi-zone":  "MULTI_ZONE",
+	"low":         "LOW",
+	"high":        "HIGH",
 }
 
 var availabilitiesToFreightModel = map[string]string{
-	singleZone: low,
-	multiZone:  high,
+	"single-zone": "LOW",  // TODO: This mapping are deprecated and will be removed in v5
+	"multi-zone":  "HIGH", // TODO: This mapping are deprecated and will be removed in v5
```
brianstrauch, line 20:
```suggestion
	"single-zone": "LOW",  // TODO: This mapping is deprecated and will be removed in v5
	"multi-zone":  "HIGH", // TODO: This mapping is deprecated and will be removed in v5
```

brianstrauch (dismissed):

brianstrauch (approved):

PR #3080, opened 2025-04-14
DGS-20003 Asyncapi correctly parse union avro types

kcorman0:
Release Notes
-------------

Bug Fixes
- `asyncapi export` correctly handles Avro union types in examples output

Checklist
---------

  - [x] Confluent Cloud prod
  - [x] Confluent Cloud stag
  - [ ] Confluent Platform

What
----

CC only
If you use consume example flag in asyncapi and have a union type defined in avro, the resulting example will likely not match the asyncapi expected format.
For example, here is a union type definition (meaning it can be of multiple types):
```
        - default: null
          name: exampleField
          type:
          - "null"
          - string
```
Here is what we put in the example:
```
          exampleField:
            string: "exampleString"
```

Asyncapi syntax checker will mark this invalid because of the `string:` part being present. This PR removes all maps which use an avro type as the key, which in this case means just outputting `exampleField: "exampleString"`

There is also a block to deal with arrays for the same reason

Blast Radius
----

CC customers calling `asyncapi export` with `--consume-examples` and `--value-type avro` would be affected

References
----------

Test & Review
-------------

Tested against customer schema with this problem (not gonna post).
Example described in `What` also tested:
Before change output
```
examples:
...
          exampleField:
            string: "exampleString"
```
After change output (passes syntax check on https://studio.asyncapi.com/)
```
examples:
...
          exampleField: "exampleString"
```

### DIFF
--- a/internal/asyncapi/command_export.go
+++ b/internal/asyncapi/command_export.go
@@ -335,7 +335,66 @@ func (c *command) getMessageExamples(consumer *ckgo.Consumer, topicName, content
 	if err := json.Unmarshal([]byte(jsonMessage), &jsonObject); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal JSON message: %v", err)
 	}
+
+	if valueFormat == "avro" {
+		processUnionTypes(jsonObject)
+	}
+
 	return jsonObject, nil
+}
+
+// Valid avro types
+var avroTypes = map[string]bool{
+	"string":  true,
+	"array":   true,
+	"int":     true,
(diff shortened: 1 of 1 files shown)


sgagniere (commented):
Is this for Cloud or on-prem? The descriptions says on-prem but the checklist says Cloud.

Otherwise looks good

kcorman0:
@sgagniere sorry updated, this command is cloud only

sgagniere (approved):

PR #3101, opened 2025-05-17
[MATRIX-871] Add new REST connection type

letaoj:
Release Notes
-------------

New Features
- Add "rest" as a new Flink connection type
- Add "mcp_server" as a new Flink connection type

Checklist
---------

  - [x] Confluent Cloud prod
  - [x] Confluent Cloud stag
  - [ ] Confluent Platform

What
----

Add rest and mcp_server as new flink connection types for Confluent Cloud.

Blast Radius
----

will affect flink connection subcommand.

References
----------

https://confluentinc.atlassian.net/browse/MATRIX-871

Test & Review
-------------
Successfully build the cli
image

rest with basic auth
image

rest with bearer token auth
image

rest with oauth
image

mcp with oauth
image

mcp with api-key
image


pkg/flink/utils.go
```diff
@@ -40,14 +46,20 @@ var (
 		"pinecone":       {"api-key"},
 		"couchbase":      {"username", "password"},
 		"confluent_jdbc": {"username", "password"},
+		"rest":           {},
```
yanand0909, line 49: This wont do any required key validation I guess.

pkg/flink/utils.go
```diff
@@ -15,16 +15,22 @@ var (
 		"pinecone":       {"api-key"},
 		"couchbase":      {"username", "password"},
 		"confluent_jdbc": {"username", "password"},
+		"rest":           {"username", "password", "bearer-token", "oauth-token-endpoint", "oauth-client-id", "oauth-client-secret", "oauth-scope"},
```
yanand0909, line 18: don't we need auth type in connection payload as well?

pkg/flink/utils.go
```diff
@@ -15,16 +15,22 @@ var (
...
-		"username":          {"mongodb", "couchbase", "confluent_jdbc"},
-		"password":          {"mongodb", "couchbase", "confluent_jdbc"},
+		"api-key":              {"openai", "azureml", "azureopenai", "googleai", "elastic", "pinecone"},
+		"aws-access-key":       {"bedrock", "sagemaker"},
+		"aws-secret-key":       {"bedrock", "sagemaker"},
+		"aws-session-token":    {"bedrock", "sagemaker"},
+		"service-key":          {"vertexai"},
+		"username":             {"mongodb", "couchbase", "confluent_jdbc", "rest"},
+		"password":             {"mongodb", "couchbase", "confluent_jdbc", "rest"},
+		"bearer-token":         {"rest"},
```
yanand0909, line 29: Like this user will be able to pass `bearer-token` with `password ` if we don't validate

test/flink_test.go
```diff
@@ -163,6 +163,10 @@ func (s *CLITestSuite) TestFlinkConnectionCreateSuccess() {
...
 		{args: "flink connection create my-connection --cloud aws --region eu-west-1 --type confluent_jdbc --endpoint jdbc:mysql://custom.com:3306/customerdb --username name --password pass", fixture: "flink/connection/create/create-confluent_jdbc.golden"},
+		{args: "flink connection create my-connection --cloud aws --region eu-west-1 --type rest --endpoint https://api.example.com --username name --password pass", fixture: "flink/connection/create/create-rest.golden"},
+		{args: "flink connection create my-connection --cloud aws --region eu-west-1 --type rest --endpoint https://api.example.com --oauth-token-endpoint https://api.example.com/auth --oauth-client-id clientId --oauth-client-secret secret --oauth-scope test_scope", fixture: "flink/connection/create/create-rest.golden"},
+		{args: "flink connection create my-connection --cloud aws --region eu-west-1 --type rest --endpoint https://api.example.com --bearer-token token", fixture: "flink/connection/create/create-rest.golden"},
+		{args: "flink connection create my-connection --cloud aws --region eu-west-1 --type rest --endpoint https://api.example.com", fixture: "flink/connection/create/create-rest.golden"},
```
yanand0909, line 169: If we add test is no flags I think that will pass as well

pkg/flink/utils.go
letaoj: Yeah, I do not want to force the user to provide any auth since they are free to provides no auth to their endpoint.

pkg/flink/utils.go
letaoj: Yep, you are right. But that can be something we can add tho not strictly needed from user. Let me add that. `auth_type` should also be an optional field.

pkg/flink/utils.go
letaoj: should we allow this scenario? To say, if we allow user to pass all the necessary auth information, they can use update command to update the auth_type.

test/flink_test.go
letaoj: sorry, I did not get this. This line is for no-auth (no flag) case which would pass.

letaoj: basically, I've tested all four cases (no_auth, basic, bearer, oauth)

pkg/flink/utils.go
letaoj: But I guess, on the other hand, we should still make sure the user passed at least pairs that could work?

test/flink_test.go
yanand0909: ok but what if user passes `--oauth-token-endpoint https://api.example.com/auth --oauth-client-id clientId ` and forgets `oauth-client-secret` flag, connection will still get created right? And it will fail in Flink runtime. Now the user has to come back and update the connection. 
if we could just use auth type to validate in CLI itself that would save lot of work for user.

letaoj: Yeah. That's what I was talking in one of the comment above that we probably still want to validate that a given pair would work.

yanand0909: I was thinking to implement something like https://github.com/confluentinc/cc-flink-gateway-service-v2/pull/1347/files#diff-6c57fdc06a3956731553777c5b145f2b2d5760c2a682b78b429d61e39cca3d09R82 in CLI as well for both MCP and REST

letaoj: Make sense. I can rewrite the code a bit to match the same validation logic as it is in gateway.

yanand0909:
I had already started working on it but haven't finished it yet. If you want I can complete my PR or you can merge your PR and then I can add MCP on similar line. I have added the PR for reference.

https://github.com/confluentinc/cli/pull/3105/files

test/fixtures/output/flink/connection/create/create-rest-missing-required-secret.golden
```diff
@@ -0,0 +1 @@
+Error: must provide client-id for type rest
```
letaoj, line 1: this error msg probably needs to be improved...

letaoj: but i guess it is fine since it is the current behavior is one of the required options is missing...

yanand0909 (dismissed):
LGTM! Thanks

internal/flink/command_connection.go
```diff
@@ -87,6 +88,12 @@ func AddConnectionSecretFlags(cmd *cobra.Command) {
 	cmd.Flags().String("service-key", "", fmt.Sprintf("Specify service key for the type: %s.", utils.ArrayToCommaDelimitedString(flink.ConnectionSecretTypeMapping["service-key"], "or")))
 	cmd.Flags().String("username", "", fmt.Sprintf("Specify username for the type: %s.", utils.ArrayToCommaDelimitedString(flink.ConnectionSecretTypeMapping["username"], "or")))
 	cmd.Flags().String("password", "", fmt.Sprintf("Specify password for the type: %s.", utils.ArrayToCommaDelimitedString(flink.ConnectionSecretTypeMapping["password"], "or")))
+	cmd.Flags().String("auth-type", "", fmt.Sprintf("Specify authentication type for the type: %s.", utils.ArrayToCommaDelimitedString(flink.ConnectionSecretTypeMapping["auth-type"], "or")))
```
lihaosky, line 91: Should `auth-type` be deduced from the input? e.g. `username:password` -> basic, `api-key` -> bearer. Because current behavior is that user can put username/password without `auth-type` for mongodb

letaoj: The idea is to use `auth_type` to dynamically enforce which flags are required. We are not requiring other connection types to set the auth_type to use basic auth. We could deduce the type but not sure how to handle the case when the customer put both bearer token and username/password

lihaosky: It should be an error if both api-key and username/password are provided. Just inconsistent `auth_type` is only required for rest and mcp

internal/flink/command_connection.go
```diff
@@ -110,28 +116,71 @@ func validateConnectionSecrets(cmd *cobra.Command, connectionType string) (map[s
...
-			return nil, fmt.Errorf("must provide %s for type %s", requiredKey, connectionType)
-		}
-		backendKey, ok := flink.ConnectionSecretBackendKeyMapping[requiredKey]
-		if !ok {
-			return nil, fmt.Errorf(`backend key not found for "%s"`, requiredKey)
+	}
+
+	switch {
+	case len(found) == 0:
+		secretMap["AUTH_TYPE"] = "no_auth"
```
lihaosky, line 157: no_auth is not valid for lots of providers? They must have secret

internal/flink/command_connection.go
lihaosky: Thanks. I wonder if there are library which can define the relationships of flags

internal/flink/command_connection.go
yanand0909: No Auth would come in handy for testing purpose

letaoj: > no_auth is not valid for lots of providers? They must have secret

The logic is a bit tricky but if you check the above method, if there is only one auth type required for a given connection type, then any of the required flag is missing will error out directly. If the code reaches to this point, it means either it requires more than one flag groups or it finds all the requires flags for a given type thus the length of found will be greater than 1.

internal/flink/command_connection.go
letaoj: Yeah, I tried to use the build function of cobra to rewrite the PR. In the new PR, I've introduce a explicitly flag for `no-auth`, let me know what's your thoughts. https://github.com/confluentinc/cli/pull/3108/files

lihaosky (dismissed):
lgtm

internal/flink/command_connection.go
```diff
@@ -150,5 +161,29 @@ func validateConnectionSecrets(cmd *cobra.Command, connectionType string) (map[s
...
+		case "USERNAME":
+			secretMap[authType] = "BASIC"
+			break
+		case "BEARER_TOKEN":
+			secretMap[authType] = "BEARER"
+			break
+		case "OAUTH2_CLIENT_ID":
+			secretMap[authType] = "OAUTH2"
+			break
+		}
```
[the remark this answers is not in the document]
letaoj: sure...updated

yanand0909 (approved):

lihaosky (approved):

internal/flink/command_connection.go
```diff
@@ -150,5 +161,22 @@ func validateConnectionSecrets(cmd *cobra.Command, connectionType string) (map[s
...
+		secretMap[authType] = "API_KEY"
+	} else if _, ok := secretMap["USERNAME"]; ok {
+		secretMap[authType] = "BASIC"
+	} else if _, ok := secretMap["BEARER_TOKEN"]; ok {
+		secretMap[authType] = "BEARER"
+	} else if _, ok := secretMap["OAUTH2_CLIENT_ID"]; ok {
+		secretMap[authType] = "OAUTH2"
+	}
+
+	if secretMap[authType] == "" && slices.Contains(lo.Keys(flink.ConnectionOneOfRequiredSecretsMapping), connectionType) {
```
sgagniere, line 174: Nitpick: the CLI already has a function to do this, that we could use here: https://github.com/confluentinc/cli/blob/main/pkg/types/map.go#L9

internal/flink/command_connection.go
```diff
@@ -150,5 +161,22 @@ func validateConnectionSecrets(cmd *cobra.Command, connectionType string) (map[s
...
 	}
 
+	if _, ok := secretMap["API_KEY"]; ok {
+		secretMap[authType] = "API_KEY"
+	} else if _, ok := secretMap["USERNAME"]; ok {
+		secretMap[authType] = "BASIC"
+	} else if _, ok := secretMap["BEARER_TOKEN"]; ok {
+		secretMap[authType] = "BEARER"
+	} else if _, ok := secretMap["OAUTH2_CLIENT_ID"]; ok {
+		secretMap[authType] = "OAUTH2"
```
sgagniere, line 171: To double check, does the backend require `AUTH_TYPE` in the map for the create/update request body?

pkg/flink/utils.go
```diff
@@ -40,7 +47,15 @@ var (
 		"pinecone":       {"api-key"},
 		"couchbase":      {"username", "password"},
 		"confluent_jdbc": {"username", "password"},
+		"rest":           {},
```
sgagniere, line 50: Is this empty because the user has multiple options for which set of credentials they have to provide?

sgagniere (commented):
Thanks!

I left a nitpick and a few questions.

internal/flink/command_connection.go
letaoj: Updated

pkg/flink/utils.go
letaoj: Yes. For rest and mcp_server, it will use the OneOfRequiredSecret map instead of this requiredAll secret map.

internal/flink/command_connection.go
letaoj: `AUTH_TYPE` is not a required field, it will be put into a map as the secret payload. When using the secret, rest and mcp_server type will use `AUTH_TYPE` in flink runtime to determine the authentication method.

lihaosky (approved):

PR #3320, opened 2026-04-14
 Bump cmf-sdk-go v0.0.5 to v0.0.6 and fix breaking changes

paras-negi-flink:
## What

Bumps `cmf-sdk-go` from **v0.0.5 → v0.0.6** and handles the two breaking changes that ship with it. Applies to Confluent Platform on-prem only.

### Breaking change 1: `ComputePool.Status` type

The generator now produces `*map[string]map[string]interface{}` for `ComputePool.Status` (caused by an inline OpenAPI schema with `additionalProperties: true` instead of a `$ref` to `ComputePoolStatus`). The real CMF API returns a flat status like `{"phase":"RUNNING","message":null}`, which `json.Unmarshal` cannot fit into the declared inner `map[string]interface{}`, so the SDK's `Execute()` fails on every valid 200 response — breaking `compute-pool list/describe/create`.

**Fix**: a fallback in `pkg/flink/cmf_rest_client.go`. When the SDK returns an error but the HTTP status is 2xx, we re-read the buffered response body and parse it ourselves:

- `unmarshalComputePool` embeds `cmfsdk.ComputePool` and shadows `Status` with `json.RawMessage`. Everything except status unmarshals via the default path; status is captured as raw JSON, the phase is extracted, and re-encoded into the nested format (`{"phase":{"value":"RUNNING"}}`) that the existing `extractComputePoolPhase` helper expects.
- `unmarshalComputePoolsPage` does the same for paginated list responses.
- If the fallback parse also fails, the body is re-buffered so `parseSdkError` can still produce a useful error message.

### Breaking change 2: `PostEnvironment.Name` type

Changed from `string` to `*string` with `GetName()`/`SetName()` accessors. Call sites in `command_environment_create.go` and `command_environment_update.go` migrated, with nil-name guards in `CreateEnvironment`/`UpdateEnvironment`.

### Test handler rewrite (forced, not optional)

`test/test-server/flink_onprem_handler.go` had to be modified because two types it used no longer compile against v0.0.6:

- `cmfsdk.ComputePoolStatus{Phase: phase}` — this struct type no longer exists in v0.0.6 (replaced by the generic map).
- `environment.Name` reads — `Name` is now `*string`, so all reads were migrated to `environment.GetName()`.

While rewriting `createComputePool`, output is changed to a raw `map[string]interface{}` with **flat status** (`"phase": "RUNNING"`) so the mock matches the real CMF API response. This ensures the CLI's fallback path is actually exercised in CI rather than being dead code that only runs in production.

### Scope

- ~55 lines of workaround in `pkg/flink/cmf_rest_client.go`, isolated to the 3 compute-pool methods
- No API surface change for consumers of `CmfRestClient`
- Golden file changes are `null` → `{}` only (cosmetic, from test mock defaulting)

Release Notes
-------------

Breaking Changes
- None

New Features
- None

Bug Fixes
- Fixed on-prem Flink environment and compute pool commands to work with cmf-sdk-go v0.0.6 SDK type changes.

Checklist
---------
- [ ] I have verified this PR in Confluent Cloud pre-prod or production environment, if applicable.
- [x] I have verified this PR in Confluent Platform on-premises environment, if applicable.
  - [ ] Confluent Cloud prod
  - [ ] Confluent Cloud stag
  - [x] Confluent Platform

Blast Radius
----
Scope is limited to on-prem Flink compute pool and environment commands; Confluent Cloud commands (which use ccloud-sdk-go-v2, not cmf-sdk-go) are completely unaffected.

Tests
------
```shell
 /Users/parasnegi/cli/dist/confluent_darwin_arm64_v8.0/confluent flink compute-pool
  describe pool --url http://localhost:8080 --environment test
  +---------------+--------------------------+
  | Creation Time | 2026-03-25T14:08:09.371Z |
  | Name          | pool                     |
  | Type          | DEDICATED                |
  | Phase         | DEDICATED                |
  +---------------+--------------------------+

  ---
  /Users/parasnegi/cli/dist/confluent_darwin_arm64_v8.0/confluent flink compute-pool
  describe pool --url http://localhost:8080 --environment test -o json

  {
    "apiVersion": "cmf.confluent.io/v1",
    "kind": "ComputePool",
    "metadata": {
      "creationTimestamp": "2026-03-25T14:08:09.371Z",
      "uid": "7cbd1475-53b4-4b71-ba8d-af3e4d7ee6d7",
      "labels": {},
      "annotations": {}
    },
    "spec": {
      "type": "DEDICATED",
      "clusterSpec": {
        "flinkConfiguration": {
          "execution.checkpointing.interval": "2000",
          "taskmanager.numberOfTaskSlots": "4"
        },
        "flinkVersion": "v1_19",
        "image": "519856050701.dkr.ecr.us-west-2.amazonaws.com/docker/dev/confluentinc/cp
  -flink-sql:0.0119-SNAPSHOT",
        "jobManager": {
          "resource": {
            "cpu": 0.5,
            "memory": "1024m"
          }
        },
        "podTemplate": {
          "spec": {
            "imagePullSecrets": [
              {
                "name": "confluent-registry"
              }
            ]
          }
        },
        "taskManager": {
          "resource": {
            "cpu": 0.5,
            "memory": "1024m"
          }
        }
      }
    },
    "status": {
      "phase": "DEDICATED"
    }
  }
```

### DIFF
--- a/internal/flink/command_compute_pool.go
+++ b/internal/flink/command_compute_pool.go
@@ -61,8 +61,9 @@ func (c *command) validComputePoolArgs(cmd *cobra.Command, args []string) []stri
 }
 
 // extractComputePoolPhase extracts the phase string from the generic status map.
-// ComputePool.Status changed from *ComputePoolStatus to *map[string]map[string]interface{} in SDK v0.0.6.
-// The status is a nested map where phase is at status["phase"]["value"].
+// ComputePool.Status is typed as *map[string]map[string]interface{} in SDK v0.0.6.
+// The unmarshalComputePool workaround in cmf_rest_client.go encodes the flat API
+// status into this nested format, placing the phase at status["phase"]["value"].
 func extractComputePoolPhase(pool cmfsdk.ComputePool) string {
 	if pool.Status == nil {
 		return ""
--- a/internal/flink/command_compute_pool.go
+++ b/internal/flink/command_compute_pool.go
@@ -1,11 +1,14 @@
 package flink
 
 import (
+	"encoding/json"
+
 	"github.com/spf13/cobra"
 
 	cmfsdk "github.com/confluentinc/cmf-sdk-go/v1"
 
 	"github.com/confluentinc/cli/v4/pkg/config"
+	"github.com/confluentinc/cli/v4/pkg/output"
 )
 
 type computePoolOut struct {
@@ -60,6 +63,35 @@ func (c *command) validComputePoolArgs(cmd *cobra.Command, args []string) []stri
 	return c.autocompleteComputePools(cmd, args)
 }
 
+// extractComputePoolPhase extracts the phase string from the generic status map.
+// ComputePool.Status changed from *ComputePoolStatus to *map[string]map[string]interface{} in SDK v0.0.6.
+// The status is a nested map where phase is at status["phase"]["value"].
+func extractComputePoolPhase(pool cmfsdk.ComputePool) string {
+	if pool.Status == nil {
+		return ""
+	}
+	status := pool.GetStatus()
+	if phaseMap, ok := status["phase"]; ok {
+		if value, ok := phaseMap["value"].(string); ok {
+			return value
+		}
+	}
+	// Fallback: try re-parsing as a simpler type in case the API shape varies.
+	raw, err := json.Marshal(pool.Status)
+	if err != nil {
+		output.ErrPrintf(false, "Warning: failed to marshal compute pool status: %v\n", err)
+		return ""
+	}
+	var flat map[string]string
+	if err := json.Unmarshal(raw, &flat); err == nil {
+		if phase, ok := flat["phase"]; ok {
+			return phase
+		}
+	}
+	output.ErrPrintf(false, "Warning: compute pool has status but phase could not be extracted\n")
+	return ""
+}
+
 func convertSdkComputePoolToLocalComputePool(sdkComputePool cmfsdk.ComputePool) LocalComputePool {
 	localPool := LocalComputePool{
 		ApiVersion: sdkComputePool.ApiVersion,
@@ -77,9 +109,9 @@ func convertSdkComputePoolToLocalComputePool(sdkComputePool cmfsdk.ComputePool) 
 		},
 	}
 
-	if sdkComputePool.Status != nil {
+	if phase := extractComputePoolPhase(sdkComputePool); phase != "" {
 		localPool.Status = &LocalComputePoolStatus{
-			Phase: sdkComputePool.Status.Phase,
+			Phase: phase,
 		}
 	}
 
--- a/internal/flink/command_compute_pool.go
+++ b/internal/flink/command_compute_pool.go
@@ -1,14 +1,11 @@
 package flink
 
 import (
-	"encoding/json"
-
 	"github.com/spf13/cobra"
 
 	cmfsdk "github.com/confluentinc/cmf-sdk-go/v1"
 
 	"github.com/confluentinc/cli/v4/pkg/config"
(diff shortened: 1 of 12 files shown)


rmetzger:
What caused breaking change 2?

internal/flink/command_compute_pool.go
```diff
@@ -60,6 +60,21 @@ func (c *command) validComputePoolArgs(cmd *cobra.Command, args []string) []stri
...
+		return ""
+	}
+	status := pool.GetStatus()
+	if phaseMap, ok := status["phase"]; ok {
+		if value, ok := phaseMap["value"].(string); ok {
+			return value
+		}
+	}
+	return ""
+}
```
rmetzger, line 76: Wouldn't it make sense to move this utility into the SDK?

internal/flink/command_compute_pool.go
```diff
@@ -77,9 +92,9 @@ func convertSdkComputePoolToLocalComputePool(sdkComputePool cmfsdk.ComputePool)
 		},
 	}
 
-	if sdkComputePool.Status != nil {
+	if phase := extractComputePoolPhase(sdkComputePool); phase != "" {
```
rmetzger, line 95: Can we always call this method, also when the CLI runs against older CMF versions? Probably yes? because the response has not changed, only the SDK.

pkg/flink/cmf_rest_client.go
```diff
@@ -570,6 +589,81 @@ func (cmfClient *CmfRestClient) DeleteCatalog(ctx context.Context, catalogName s
...
+		if err != nil {
+			return nil, err
+		}
+		pools = append(pools, pool)
+	}
+	return pools, nil
+}
+
+// readFallbackBody returns the HTTP response body when the SDK's Execute() returned
+// an error despite a successful 2xx status — the ComputePool deserialization bug.
```
rmetzger, line 637: What is "the ComputePool deserialization bug"? I don't think this is a bug?
It's just that the SDK's interfaces have changed

rmetzger:
`err != nil && statusCode in [200,300)` fires on *any* decode failure, not just the known Status-shape bug. Scenarios it could mask:

1. SDK adds a new broken field unrelated to Status → fallback "succeeds" by ignoring that field, we silently return a partially-correct pool
2. Response truncation on a 2xx → fallback parse fails, we fall through to `parseSdkError` which reports the wrong cause


**narrow the error signature.** The v0.0.6 bug produces a specific `*json.UnmarshalTypeError` on the `Status` field. Match on that:

```go
var typeErr *json.UnmarshalTypeError
if !errors.As(err, &typeErr) || typeErr.Field != "status" {
    return nil   // not the bug we're working around
}
```

Brittle if the SDK changes error wrapping, but **explicitly scoped to the known bug**. Any other decode failure now surfaces correctly.

internal/flink/command_compute_pool.go
paras-negi-flink: My understanding is that  we don't do human code push in cmf-go-sdk, but if we want to : [Adds a custom UnmarshalJSON method on the ComputePool type (in a new file v1/model_compute_pool_custom.go )](https://github.com/confluentinc/cmf-sdk-go/pull/65) seems to be the better fix, which completely removes "custom unmarshaling with json.RawMessage embedding" from cli repo

rmetzger: > My understanding is that we don't do human code push in cmf-go-sdk, but if we want to

My only worry about human pushes in the sdk repo is that the code generator is overwriting human written code / utils.
But when we started the go sdk repo, I actually had additional, non generated code for shared utilities in mind.

PR #3342, opened 2026-05-01
Add per-platform smoke test pipeline with OTLP metric emittance

danielayaz:
## Summary

Mirrors the smoke-tests pattern from `terraform-provider-confluent` for the CLI: a scheduled, lightweight liveness check against Confluent Cloud that emits a pass/fail OTLP gauge to Heracles. Unlike the TF provider (one binary, runs on Linux), the CLI binary is platform-specific, so the pipeline runs once per platform and labels the metric with `{os, arch}` so each platform has its own pass/fail series.

**Platforms covered:** `linux/arm64` and `windows/amd64` — the two platforms whose PR builds were skipped in #3325 to save CI time. `linux/amd64` is already covered by the daily `live-tests.yml`; macOS coverage stays on the manual `macos.yml` promotion; the s390x build (#3338) is intentionally out of scope.

**Cadence:** every 3 hours (`0 */3 * * *`) on `main`. Each run emits 2 series, so ~16 platform-runs/day.

## Changes

| File | Change |
|---|---|
| `cmd/otel-smoke-metric/main.go` | NEW — OTLP gRPC emitter for `cli_smoke_test_result{os,arch,command,branch,job_id}`. Always exits 0 (fail-open). |
| `test/live/environment_smoke_live_test.go` | NEW — smoke test under build tag `live_test && (all \|\| smoke)`. Runs `environment list -o json` (validates auth + API + JSON shape). |
| `.semaphore/smoke-tests.yml` | NEW — two parallel blocks, no inter-block dependencies, so a Linux failure doesn't gate the Windows block. |
| `service.yml` | EDIT — adds `CLI Smoke Tests - PROD` task on `0 */3 * * *`. |
| `Makefile` | EDIT — adds `live-test-smoke` and `build-otel-smoke-metric` targets. |
| `go.mod` / `go.sum` | EDIT — promotes `otlpmetricgrpc`, `sdk`, `sdk/metric` to direct deps at `v1.39.0` (matches the existing `otel` version, no upgrades). |
| `.gitignore` | EDIT — adds `/bin/` (where `build-otel-smoke-metric` writes). |

## Reference

- TF-provider implementation this mirrors: `terraform-provider-confluent/cmd/otel-smoke-metric/main.go` and `.semaphore/smoke-tests.yml`.
- Existing dashboard panels for the TF metric live in `cc-terraform-monitoring/teams/apie/module/json/apie-ops-overview-dashboard.tpl.json`. CLI panels are out of scope for this PR; the metric will be queryable in Grafana ad-hoc as soon as data flows.

## Test plan

- [ ] Local build of the emitter passes: `go build ./cmd/otel-smoke-metric`
- [ ] `go vet -tags='live_test,smoke' ./test/live/` is clean
- [ ] Emitter fail-opens correctly: bogus endpoint exits 0, bad CLI arg exits 0
- [ ] After merge, manually promote `CLI Smoke Tests - PROD` once before the schedule fires to validate both platform blocks pass on real Semaphore agents
- [ ] After 1–2 successful scheduled runs, query Heracles Prometheus for `cli_smoke_test_result{repo="cli"}` and confirm two series exist (one per `{os,arch}`) with value `1`
- [ ] Inject a failure on a branch (assert a non-existent JSON field), promote, confirm Slack pings and metric reports `0`


.semaphore/smoke-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+          type: s1-prod-ubuntu24-04-arm64-1
+      jobs:
+        - name: "Smoke linux/arm64"
+          commands:
+            - checkout
+            - sem-version go $(cat .go-version)
+            - export PATH=$(go env GOPATH)/bin:$PATH
+
+            - . vault-sem-get-secret v1/ci/kv/apif/cli/live-testing-data
+            - . vault-sem-get-secret v1/ci/kv/apif/cli/slack-notifications-live-testing
```
sgagniere, line 30: We'll need to replace `vault-sem-get-secret` (v1) with `vault-get-secret` (v3) and store these new secrets in the v3 vault.

danielayaz: Will do so this Monday!

sgagniere (approved):

PR #3344, opened 2026-05-02
chore: make Windows smoke block resilient to vault + setup failures

danielayaz:
## Summary

Follow-up to #3342. The first scheduled run of the smoke pipeline against `main` had:

- ✅ **linux/arm64** — passed cleanly, metric emitted
- ❌ **windows/amd64** — failed at `vault kv get -field=CONFLUENT_CLOUD_EMAIL …` with `Field "CONFLUENT_CLOUD_EMAIL" not present in secret`. Because the failure was on a regular YAML command line, the job aborted before reaching the metric emitter, so the windows panel showed "No data" instead of `0`.

This PR fixes both issues.

### 1. Dynamic vault field handling

Replaced 3 hardcoded `vault kv get -field=…` calls with a `Set-VaultFields` PowerShell helper that:

- Pulls each secret with `vault kv get -format=json`
- **Logs every field name it finds** — so the next run will print e.g. `Fields in v1/ci/kv/apif/cli/live-testing-data: email, password` and we'll know exactly what's there
- Exports each field under **both** its original name AND an `UPPER_SNAKE_CASE` variant, covering every common Vault → env-var convention (`email` / `EMAIL`, `confluent-cloud-email` / `CONFLUENT_CLOUD_EMAIL`, `confluent_cloud_email`, etc.) at once

This mirrors what `vault-sem-get-secret` does on Linux, which is why Linux works without specifying field names. `vault-sem-get-secret` ships with the Semaphore Linux agent toolbox and is not available on Windows agents.

### 2. Always emit the metric

The emitter binary is now built **before** the failable section. The vault auth + CLI build + smoke test are wrapped in one PowerShell `try / catch / finally`, and the `finally` clause **always** calls `otel-smoke-metric.exe` with the captured `$RESULT`:

- `$RESULT = "0"` is the default
- Only set to `"1"` after a clean test pass

So vault failures, build failures, test failures, or any thrown exception all report `cli_smoke_test_result{os="windows",arch="amd64"} = 0` to Heracles. The windows panel in [cc-terraform-monitoring#9639](https://github.com/confluentinc/cc-terraform-monitoring/pull/9639) will now go red on infra failures instead of going silent.

## Test plan

- [ ] Manually promote the smoke pipeline once after merge
- [ ] Look at the windows job log — confirm the `Fields in v1/ci/kv/apif/cli/live-testing-data: …` line lists field names that include something matching `CONFLUENT_CLOUD_EMAIL` / `CONFLUENT_CLOUD_PASSWORD` (after upper-snake-case normalization)
- [ ] If the test runs and passes: confirm `cli_smoke_test_result{os="windows",arch="amd64"} = 1` shows up in Heracles
- [ ] If the test still fails (e.g. fields really aren't there): confirm `cli_smoke_test_result{os="windows",arch="amd64"} = 0` shows up — the new always-emit guarantee makes the failure visible

## Out of scope

- Linux block is unchanged.
- Two-set-of-graphs in Grafana is already handled by the merged dashboard PR ([cc-terraform-monitoring#9639](https://github.com/confluentinc/cc-terraform-monitoring/pull/9639)) — separate panels per `{os, arch}` label, which is the standard Prometheus per-platform pattern.

### DIFF
--- a/.semaphore/smoke-tests.yml
+++ b/.semaphore/smoke-tests.yml
@@ -60,14 +60,6 @@ blocks:
             - $Env:PATH += ";C:\Program Files\Git\bin"
             - checkout
 
-            # Vault auth (vault-sem-get-secret is Linux-only; use native vault CLI on Windows,
-            # following the pattern from cli-release/.semaphore/4-release-cli.yml).
-            - $Env:VAULT_ADDR = "https://vault.cireops.gcp.internal.confluent.cloud"
-            - vault login -no-print token=$(vault write -field=token "auth/semaphore_self_hosted/login" role="default" jwt="$Env:SEMAPHORE_OIDC_TOKEN")
-            - $Env:CONFLUENT_CLOUD_EMAIL = (vault kv get -field=CONFLUENT_CLOUD_EMAIL v1/ci/kv/apif/cli/live-testing-data)
-            - $Env:CONFLUENT_CLOUD_PASSWORD = (vault kv get -field=CONFLUENT_CLOUD_PASSWORD v1/ci/kv/apif/cli/live-testing-data)
-            - $Env:SLACK_WEBHOOK_URL = (vault kv get -field=SLACK_WEBHOOK_URL v1/ci/kv/apif/cli/slack-notifications-live-testing)
-
             # Install Go (matches the pattern in semaphore.yml; chocolatey is community-maintained)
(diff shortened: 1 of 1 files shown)


sgagniere:
> Logs every field name it finds — so the next run will print e.g. Fields in v1/ci/kv/apif/cli/live-testing-data: email, password and we'll know exactly what's there

Can we guarantee that we won't also log the secret values themselves?

PR #3361, opened 2026-05-19
[APIE-1040] Add generic --wait framework; refactor Flink statement create

danielayaz:
Release Notes
-------------

New Features
- Introduced a unified polling framework for the `--wait flag`, improving transitional state tracking for Flink commands and increasing the default timeout to 6 hours to align with Terraform.

Checklist
---------
- [ ] I have verified this PR in Confluent Cloud pre-prod or production environment, if applicable. _(verified via integration mocks; live verification pending if reviewer requests)_
- [ ] I have verified this PR in Confluent Platform on-premises environment, if applicable. _(verified via integration mocks; live verification pending if reviewer requests)_
- [x] I confirm that this PR introduces no breaking changes or backward compatibility issues. (The `--wait-timeout` default changes from 1m → 6h to match `terraform-provider-confluent`; scripts that depended on the 1m circuit-breaker should pass `--wait-timeout 1m` explicitly. The flag itself is unchanged.)
  - [x] Confluent Cloud prod _(N/A — refactor preserves polling semantics; new framework is additive)_
  - [x] Confluent Cloud stag _(N/A — same reason)_
  - [x] Confluent Platform _(N/A — same reason)_

What
----
**Applies to:** Confluent Cloud AND Confluent Platform.

**Problem.** `--wait` exists today on exactly two commands (`flink statement create` cloud + on-prem) as near-duplicate inline `retry.Retry(...)` blocks. Both check `Status.Phase != "PENDING"` to decide that a statement has settled, which is wrong: a statement in `FAILING` or `STOPPING` is also still in transition, and the CLI today prints it as if it had reached a terminal state. Many other 202-style create commands (`network/*`, `flink compute-pool/connection`, `kafka cluster`, `ksql cluster`, `tableflow`, `ccpm`) lack `--wait` entirely. Copy-pasting the inline pattern would compound the duplication.

**Approach.**

1. **`pkg/wait`** — a generic `Poll[T](ctx, Options[T])` with pluggable `IsTerminal` / `IsFailed` predicates, configurable `Tick` and `Timeout`, sentinel errors (`ErrTimeout`, `ErrFailed`), and context-cancellation support. A direct loop (not a wrapper around `pkg/retry`) because `retry.Retry`'s `func() error` signature can't return the polled value or distinguish "still pending" from "permanently failed" from "fetch error" cleanly.

2. **`pkg/wait.PollPhases` + `PhaseSet`** — declarative wrapper for the common case where readiness is a single status-phase string. Callers supply `PendingPhases` and `FailedPhases` slices; `PollPhases` builds the `IsTerminal` / `IsFailed` closures internally. This is the shape the cli-terraform-generator can emit deterministically from `AsyncConfig.PendingStates` / `FailedStates` (see the companion generator PR confluentinc/cli-terraform-generator#197), so future async resources adopt `--wait` without touching `pkg/wait` again.

3. **`pkg/cmd/flags.go`** — `AddWaitFlag` and `AddWaitTimeoutFlag(cmd, defaultTimeout)` helpers so future adoptions are one-liners. `pkg/cmd/flags.go` is the only file in `pkg/cmd/` in CLAUDE.md's "edit freely" zone; the rest of `pkg/cmd/` is untouched.

4. **Reference adoption.** `internal/flink/command_statement_create.go` (1s tick) and `internal/flink/command_statement_create_onprem.go` (2s tick) refactored onto `PollPhases` with phase sets sourced from the Flink statement OpenAPI enum: pending = `{PENDING, FAILING, STOPPING, DELETING}`, failed = `{FAILED}`. `IsFailed` is wired so the call site can distinguish a failed-state landing from a timeout — the failed path now returns `statement "x" entered failed phase "FAILED": <detail>` with an `Inspect the statement with` `confluent flink statement describe x` `.` suggestion, while the timeout path keeps the `Increase --wait-timeout` suggestion.

5. **TF-aligned default.** `--wait-timeout` default raised from 1m to 6h to match `terraform-provider-confluent`'s `statementsAPICreateTimeout`. Users who want a short circuit-breaker still pass `--wait-timeout` explicitly. Other async resources adopting the framework will surface their own per-resource defaults via `AsyncConfig.CreateTimeout` (handled in the generator PR).

**Notable detail (pflag gotcha).** `pflag.Duration` interprets the first backtick-quoted word in a flag description as a type-name override. `AddWaitTimeoutFlag`'s description therefore uses plain `--wait` (no backticks) so the help renders as `--wait-timeout duration` instead of `--wait-timeout --wait`. CLAUDE.md's backtick convention still applies in `Short`/`Long`/error messages.

Blast Radius
----
- **Confluent Cloud customers using `confluent flink statement create --wait`** — semantic change: statements in `FAILING` / `STOPPING` / `DELETING` are now correctly treated as still in transition, where they were previously mis-reported as terminal. Mitigated by unit tests (`TestPollPhases_AllPendingPhasesContinuePolling`) and an integration test against a new `failed-statement` mock that exercises the differentiated error path.
- **Customers scripting around the previous 1m default** — the new 6h default is more permissive; any script that depended on `--wait` timing out within ~1m should pass `--wait-timeout 1m` (or shorter) explicitly. The flag name and value type are unchanged.
- **Confluent Platform customers using `confluent flink statement create --wait` (on-prem)** — same semantic change, same mitigation.
- **All commands using `pkg/cmd/flags.go`** — additive helpers, no existing call site touched.
- **`pkg/retry` callers** — unchanged; the package still exists with 5+ other call sites in `internal/asyncapi`, `internal/kafka`, `internal/connect`, `pkg/auth`. Flink statement create just no longer imports it.

References
----------
- JIRA: [APIE-1040](https://confluentinc.atlassian.net/browse/APIE-1040)
- Companion generator PR: confluentinc/cli-terraform-generator#197

Test & Review
-------------
**Manual test: `--wait` on `confluent flink statement create` (APIE‑1040)**

Verified `--wait` blocks until the statement leaves the pending set, and returns the right errors on timeout/failure. Three live runs + the unit-tested failure path:

**1. Baseline (no `--wait`) → returns immediately in `PENDING`:**
```
$ confluent flink statement create wait-demo-a-… --sql "SELECT 1;" --cloud aws --region us-west-2 -o json | grep status
  "status": "PENDING"
```

**2. `--wait` → blocks until terminal (`COMPLETED`), ~9.6s:**
```
$ time confluent flink statement create wait-demo-b-… --sql "SELECT 1;" --cloud aws --region us-west-2 --wait
+---------------+------------------------+
| Name          | wait-demo-b-1782500402 |
| Statement     | SELECT 1;              |
| Compute Pool  | lfcp-stgc91j885        |
| Status        | COMPLETED              |
+---------------+------------------------+
… 9.558 total
```
Same statement as #1, but `--wait` turned the instant `PENDING` return into a 9.56s block ending in `COMPLETED`.

**3. `--wait --wait-timeout 1ms` → timeout error + suggestion:**
```
$ confluent flink statement create wait-demo-c-… --sql "SELECT 1;" --cloud aws --region us-west-2 --wait --wait-timeout 1ms
Error: wait timed out: statement "wait-demo-c-1782500517" is still in phase "PENDING"
Suggestions:
    Increase `--wait-timeout` or omit `--wait`.
```

**4. Failed-phase path** (hard to force live) — covered deterministically by `pkg/wait.TestPollPhases_FailedPhase`; full `go test ./pkg/wait/` suite passes.

**Conclusion:** all three runtime outcomes — success-after-poll, timeout, failed — behave correctly. Good to go.

**Automated tests run locally** (Go 1.25.7; repo pins `.go-version` to 1.26.1 which isn't installed locally, so 1.25.7 was used — CI will pick up 1.26.1):

| Suite | Result |
|---|---|
| `go test ./pkg/wait/ ./pkg/cmd/ ./internal/flink/` (unit) | PASS — 11 wait tests (7 baseline + 4 new: `PhaseSet`, `PollPhases_TerminalSuccess`, `PollPhases_FailedPhase`, `PollPhases_AllPendingPhasesContinuePolling` covering FAILING/STOPPING/DELETING) |
| `make lint` (golangci-lint + hunspell spell check) | PASS |
| `make integration-test -run TestCLI/TestFlinkStatementCreate$` (cloud) | PASS — includes existing `--wait` case (unchanged golden), `--wait-timeout 100ms` timeout case, and new `failed-statement --wait` case exercising the differentiated `ErrFailed` error path |
| `make integration-test -run TestCLI/TestHelp/flink/statement/create` (help) | PASS — both cloud and on-prem help goldens regenerated for the 6h default |

**Pre-existing test failure not introduced by this PR:**
- `pkg/flink/internal/controller/TestInteractiveOutputControllerTestSuite` fails in non-TTY environments with `open /dev/tty: device not configured` (tview/tcell needs a real terminal). Reproduced on baseline (`origin/main` without these changes). Expected to pass in Semaphore CI which provides a pseudo-TTY.

**Golden file changes** (all intentional):
- `create-wait.golden` — **unchanged** (parity proof for the success path).
- `create-wait-timeout.golden` — **unchanged** (timeout error path, error message identical).
- `create-wait-failed.golden` — **new** (covers the wired-up `IsFailed` path with the differentiated error message).
- `create-help.golden` / `create-help-onprem.golden` / `create-missing-sql-failure.golden` / `create-missing-compute-pool-failure.golden` — regenerated to reflect the new `--wait-timeout default 6h0m0s`. Diff is exactly one line in each.

**Backwards compatibility.** Per CLAUDE.md compatibility rules: adding `--wait-timeout` was non-breaking (additive flag). Changing the *default* from 1m to 6h is technically a behavior change but doesn't break the flag contract — see Blast Radius. The flag name, value type, and existing `--wait` boolean are all unchanged. Existing scripts that pass `--wait` continue to work; their wait window simply expands to match TF's.

[APIE-1040]: https://confluentinc.atlassian.net/browse/APIE-1040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/flink/command_statement_create.go
+++ b/internal/flink/command_statement_create.go
@@ -187,13 +187,19 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
 			Timeout:       timeout,
 		})
 		if err != nil {
-			if errors.Is(err, wait.ErrFailed) {
+			switch {
+			case errors.Is(err, wait.ErrFailed):
 				return clierrors.NewErrorWithSuggestions(
 					fmt.Sprintf(`statement "%s" entered failed phase %q: %s`, name, statement.Status.GetPhase(), statement.Status.GetDetail()),
 					fmt.Sprintf("Inspect the statement with `confluent flink statement describe %s`.", name),
 				)
+			case errors.Is(err, wait.ErrTimeout):
+				return clierrors.NewErrorWithSuggestions(err.Error(), "Increase `--wait-timeout` or omit `--wait`.")
+			default:
+				// Fetch error or context cancellation — surface the underlying
+				// cause unmodified; suggesting a longer timeout would mislead.
+				return err
 			}
-			return clierrors.NewErrorWithSuggestions(err.Error(), "Increase `--wait-timeout` or omit `--wait`.")
 		}
 	}
 
--- a/internal/flink/command_statement_create.go
+++ b/internal/flink/command_statement_create.go
@@ -1,7 +1,6 @@
 package flink
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -16,7 +15,7 @@ import (
 	"github.com/confluentinc/cli/v4/pkg/flink/types"
 	"github.com/confluentinc/cli/v4/pkg/output"
 	"github.com/confluentinc/cli/v4/pkg/properties"
-	"github.com/confluentinc/cli/v4/pkg/retry"
+	"github.com/confluentinc/cli/v4/pkg/wait"
 )
 
 func (c *command) newStatementCreateCommand() *cobra.Command {
@@ -41,7 +40,8 @@ func (c *command) newStatementCreateCommand() *cobra.Command {
 	c.addComputePoolFlag(cmd)
 	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
 	c.addDatabaseFlag(cmd)
-	cmd.Flags().Bool("wait", false, "Block until the statement is running or has failed.")
+	pcmd.AddWaitFlag(cmd)
+	pcmd.AddWaitTimeoutFlag(cmd, time.Minute)
 	cmd.Flags().StringSlice("property", []string{}, "A mechanism to pass properties in the form key=value when creating a Flink statement.")
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddContextFlag(cmd, c.CLICommand)
@@ -151,25 +151,27 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	wait, err := cmd.Flags().GetBool("wait")
+	shouldWait, err := cmd.Flags().GetBool("wait")
 	if err != nil {
 		return err
 	}
-	if wait {
-		err := retry.Retry(time.Second, time.Minute, func() error {
-			statement, err = client.GetStatement(environmentId, name, c.Context.LastOrgId)
-			if err != nil {
-				return err
-			}
-
-			if statement.Status.GetPhase() == "PENDING" {
-				return fmt.Errorf(`statement phase is "%s"`, statement.Status.GetPhase())
-			}
-
-			return nil
-		})
+	if shouldWait {
+		timeout, err := cmd.Flags().GetDuration("wait-timeout")
 		if err != nil {
 			return err
+		}
+		statement, err = wait.Poll(cmd.Context(), wait.Options[flinkgatewayv1.SqlV1Statement]{
+			Fetch: func() (flinkgatewayv1.SqlV1Statement, error) {
+				return client.GetStatement(environmentId, name, c.Context.LastOrgId)
+			},
+			IsTerminal: func(s flinkgatewayv1.SqlV1Statement) bool {
+				return s.Status.GetPhase() != "PENDING"
+			},
+			Tick:    time.Second,
+			Timeout: timeout,
+		})
+		if err != nil {
+			return errors.NewErrorWithSuggestions(err.Error(), "Increase `--wait-timeout` or omit `--wait`.")
 		}
 	}
 
--- a/internal/flink/command_statement_create.go
+++ b/internal/flink/command_statement_create.go
@@ -1,6 +1,8 @@
 package flink
 
 import (
+	"errors"
+	"fmt"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -9,7 +11,7 @@ import (
 
 	"github.com/confluentinc/cli/v4/pkg/ccloudv2"
 	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
-	"github.com/confluentinc/cli/v4/pkg/errors"
+	clierrors "github.com/confluentinc/cli/v4/pkg/errors"
 	"github.com/confluentinc/cli/v4/pkg/examples"
 	"github.com/confluentinc/cli/v4/pkg/flink/config"
 	"github.com/confluentinc/cli/v4/pkg/flink/types"
@@ -17,6 +19,20 @@ import (
 	"github.com/confluentinc/cli/v4/pkg/properties"
 	"github.com/confluentinc/cli/v4/pkg/wait"
 )
+
+// Phase enum from flink-gateway/v1 SqlV1StatementStatus.Phase. PENDING /
+// FAILING / STOPPING / DELETING are transitioning states; FAILED is the only
+// terminal failure; RUNNING / COMPLETED / STOPPED / DEGRADED are terminal
+// success. The generator will source the same sets from AsyncConfig.
+var (
+	flinkStatementPendingPhases = []string{"PENDING", "FAILING", "STOPPING", "DELETING"}
+	flinkStatementFailedPhases  = []string{"FAILED"}
+)
+
+// statementsAPICreateTimeout aligns with terraform-provider-confluent's
+// statementsAPICreateTimeout in internal/provider/constants.go. CLI users can
+// shorten it with --wait-timeout; TF has no equivalent override.
+const flinkStatementCreateWaitTimeout = 6 * time.Hour
 
 func (c *command) newStatementCreateCommand() *cobra.Command {
 	cmd := &cobra.Command{
@@ -41,7 +57,7 @@ func (c *command) newStatementCreateCommand() *cobra.Command {
 	pcmd.AddServiceAccountFlag(cmd, c.AuthenticatedCLICommand)
 	c.addDatabaseFlag(cmd)
 	pcmd.AddWaitFlag(cmd)
-	pcmd.AddWaitTimeoutFlag(cmd, time.Minute)
+	pcmd.AddWaitTimeoutFlag(cmd, flinkStatementCreateWaitTimeout)
 	cmd.Flags().StringSlice("property", []string{}, "A mechanism to pass properties in the form key=value when creating a Flink statement.")
 	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
 	pcmd.AddContextFlag(cmd, c.CLICommand)
@@ -62,7 +78,7 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
 
 	environment, err := c.V2Client.GetOrgEnvironment(environmentId)
 	if err != nil {
-		return errors.NewErrorWithSuggestions(err.Error(), "List available environments with `confluent environment list`.")
+		return clierrors.NewErrorWithSuggestions(err.Error(), "List available environments with `confluent environment list`.")
 	}
 
 	computePool := c.Context.GetCurrentFlinkComputePool()
@@ -78,7 +94,7 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
 
 	if computePool == "" {
 		if cloud == "" || region == "" {
-			return errors.New("Flink cloud and region flags are required when compute pool is not specified.")
+			return clierrors.New("Flink cloud and region flags are required when compute pool is not specified.")
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


internal/flink/command_statement_create.go
```diff
@@ -151,26 +151,28 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
...
+		timeout, err := cmd.Flags().GetDuration("wait-timeout")
 		if err != nil {
 			return err
 		}
+		statement, err = wait.Poll(cmd.Context(), wait.Options[flinkgatewayv1.SqlV1Statement]{
+			Fetch: func() (flinkgatewayv1.SqlV1Statement, error) {
+				return client.GetStatement(environmentId, name, c.Context.LastOrgId)
+			},
+			IsTerminal: func(s flinkgatewayv1.SqlV1Statement) bool {
+				return s.Status.GetPhase() != "PENDING"
```
channingdong: `"PENDING"` may not be the only intermediate status, there are `"FAILING"` and `"STOPPING"` for Flink statement:
https://docs.confluent.io/cloud/current/ccloud/get-sqlv-1-statement/?_highlight=statement

If we think even one level higher, is there a way we can retrieve information from OpenAPI spec to determine which states are pending states, and which states are terminal states, this is the question we need to give answer to ourselves.

For example, a naive idea would be, if the status ends with `"ING"`, we need to wait, otherwise we should terminate.
https://github.com/confluentinc/api/blob/master/flink-gateway/v1/openapi.yaml#L3313

Can you check if such idea can be applied to other async resources?

pkg/wait/wait.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"errors"
+	"time"
+)
+
+// Options describes a single Poll invocation. T is the polled resource type.
+type Options[T any] struct {
+	Fetch      func() (T, error)
+	IsTerminal func(T) bool
+	IsFailed   func(T) bool
```
channingdong: `IsFailed` is not wired up to the command source file, so we can't tell if the successful wait status or failed wait status, can we handle this better?

internal/flink/command_statement_create_onprem.go
```diff
@@ -124,22 +125,24 @@ func (c *command) statementCreateOnPrem(cmd *cobra.Command, args []string) error
...
 			return err
 		}
+		finalStatement, err = wait.Poll(cmd.Context(), wait.Options[cmfsdk.Statement]{
+			Fetch: func() (cmfsdk.Statement, error) {
+				return client.GetStatement(c.createContext(), environment, name)
+			},
+			IsTerminal: func(s cmfsdk.Statement) bool {
+				return s.GetStatus().Phase != "PENDING"
+			},
+			Tick:    2 * time.Second,
```
channingdong: `2 * time.Second` may work for Flink statement but not work for others, a short interval increases the chance of being rate limited, a long interval may upset users.

Also, please investigate on how TF handles the polling interval and timeout, is it purely manual setting, is it semi-manual setting with default, is it something we can derive from OpenAPI spec, is it something we can write to registry.yaml as override value, once you have a proposal, please sync with @linouk23 for the unified implementation.

internal/flink/command_statement_create_onprem.go
```diff
@@ -36,7 +36,8 @@ func (c *command) newStatementCreateCommandOnPrem() *cobra.Command {
 	cmd.Flags().String("catalog", "", "The name of the default catalog.")
 	cmd.Flags().String("database", "", "The name of the default database.")
 	cmd.Flags().String("flink-configuration", "", "The file path to hold the Flink configuration for the statement.")
-	cmd.Flags().Bool("wait", false, "Boolean flag to block until the statement is running or has failed.")
+	pcmd.AddWaitFlag(cmd)
+	pcmd.AddWaitTimeoutFlag(cmd, time.Minute)
```
channingdong: TF doesn't have the user configurable wait timeout, as it's purely handled inside an internal function with hardcoded value, since we want to have unified UX for CLI and TF users, when designing please take this into consideration, we need to either alter the existing TF behavior or let CLI follow the CLI behavior.

pkg/wait/wait.go
```diff
@@ -0,0 +1,65 @@
...
+// returns true (ErrFailed), IsTerminal returns true (success), opts.Timeout
+// elapses (ErrTimeout), ctx is cancelled (ctx.Err()), or Fetch returns a
+// non-nil error. Always returns the most recent successfully-fetched T.
+func Poll[T any](ctx context.Context, opts Options[T]) (T, error) {
+	var last T
+
+	check := func() (T, bool, error) {
+		v, err := opts.Fetch()
+		if err != nil {
+			return last, true, err
```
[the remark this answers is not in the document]
danielayaz: Addressed in 1b38ccbc1. `Poll` now retries fetch errors until timeout (matching the original `retry.Retry` behavior); when the timeout fires while the most recent fetch errored, that error is returned in place of `ErrTimeout` so the user sees the real cause. New test `TestPoll_FetchErrorThenSuccess` covers the mid-poll recovery path.

internal/flink/command_statement_create.go
```diff
@@ -151,26 +151,28 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
...
+			},
+			Tick:    time.Second,
+			Timeout: timeout,
+		})
+		if err != nil {
+			return errors.NewErrorWithSuggestions(err.Error(), "Increase `--wait-timeout` or omit `--wait`.")
```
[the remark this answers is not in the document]
danielayaz: Addressed in 1b38ccbc1. The cloud call site is now a 3-way switch: `ErrFailed` returns the describe-suggestion, `ErrTimeout` keeps the "Increase `--wait-timeout`" suggestion, and any other error (fetch errors propagated by the new Poll-retries-fetch-errors behavior, ctx cancellation) is returned unmodified.

internal/flink/command_statement_create_onprem.go
```diff
@@ -124,22 +125,24 @@ func (c *command) statementCreateOnPrem(cmd *cobra.Command, args []string) error
...
+				return client.GetStatement(c.createContext(), environment, name)
+			},
+			IsTerminal: func(s cmfsdk.Statement) bool {
+				return s.GetStatus().Phase != "PENDING"
+			},
+			Tick:    2 * time.Second,
+			Timeout: timeout,
+		})
+		if err != nil {
+			return errors.NewErrorWithSuggestions(err.Error(), "Increase `--wait-timeout` or omit `--wait`.")
```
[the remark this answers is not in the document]
danielayaz: Addressed in 1b38ccbc1 (same fix as the cloud call site). The on-prem path also uses the 3-way switch — only `ErrTimeout` carries the "Increase `--wait-timeout`" suggestion now.

internal/flink/command_statement_create_onprem.go
```diff
@@ -124,22 +125,24 @@ func (c *command) statementCreateOnPrem(cmd *cobra.Command, args []string) error
...
-			// Update the finalStatement with the completed state
-			finalStatement = polledStatement
-			return nil
-		})
+	if shouldWait {
+		timeout, err := cmd.Flags().GetDuration("wait-timeout")
 		if err != nil {
 			return err
 		}
+		finalStatement, err = wait.Poll(cmd.Context(), wait.Options[cmfsdk.Statement]{
```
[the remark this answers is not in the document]
danielayaz: Addressed in 1b38ccbc1. Added `running-wait-stmt` to the on-prem test table; a matching branch in `flink_onprem_handler.go` returns RUNNING on GET while POST keeps returning PENDING, so the test exercises the full PENDING→RUNNING polling transition. Golden: `create-wait-onprem.golden`.

internal/flink/command_statement_create.go
danielayaz: Addressed in 711f60d3e. The hand-written reference now uses an explicit phase set: pending = `{PENDING, FAILING, STOPPING, DELETING}`, failed = `{FAILED}` — sourced directly from the SqlV1StatementStatus.Phase OpenAPI enum. Channing's broader question about deriving these from the spec is answered in the companion generator PR confluentinc/cli-terraform-generator#197: `classifyPhase` does the categorization from `x-extensible-enum`, and FAILING is explicitly moved out of `phaseClassFailed` and into `phaseClassPending` per this comment. The generator then emits the resulting Go slices into the create template; verified output on flink_compute_pool, networking_access_point, rtce_topic, etc.

pkg/wait/wait.go
danielayaz: Addressed in 711f60d3e (and tightened in 1b38ccbc1). Both Flink call sites now pass `FailedPhases: {FAILED}` to the new `PollPhases` wrapper, which builds the `IsFailed` closure internally. The error handler differentiates: `ErrFailed` returns "`statement \"x\" entered failed phase \"FAILED\": <detail>`" with an "Inspect the statement with `confluent flink statement describe x`" suggestion; `ErrTimeout` keeps the increase-timeout suggestion. See `create-wait-failed.golden` for the new integration coverage.

internal/flink/command_statement_create_onprem.go
danielayaz: Investigated and partially addressed; the cross-codebase unification is in the companion generator PR. Tick stays at 2s for the hand-written on-prem reference (existing Flink behavior) but the generator-emitted code defaults to 30s (the conservative TF-friendly default; matches `pollIntervalNormal` in terraform-provider-confluent). Per-resource override surface is in place via `AsyncConfig` — registry.yaml hooks for tick overrides will land in a follow-up. `--wait-timeout` default was raised from 1m → 6h to match TF's `statementsAPICreateTimeout` so the timeout side is also unified at the data layer.

internal/flink/command_statement_create_onprem.go
danielayaz: Working assumption in this PR: CLI keeps `--wait-timeout` as a user override (CI/shell users want a short circuit-breaker) and the *default* unifies with TF via OpenAPI-derived `AsyncConfig.CreateTimeout`. Flink statement default raised from 1m → 6h to match TF's `statementsAPICreateTimeout`; TF stays hardcoded. So the unification is at the data layer (one source of truth in the OpenAPI spec / generator) rather than the UX layer. Happy to flip to either of the alternatives (drop `--wait-timeout` to match TF exactly, or push TF to accept a user override) — flagging for separate discussion.

linouk23 (approved):

PR #3382, opened 2026-06-12
[ORC-10127] Add switchover CLI commands (pair, endpoint)

wyuzheng:
Release Notes
-------------

New Features
- Add `confluent switchover pair` and `confluent switchover endpoint` command groups for Kafka Disaster Recovery (Switchover): create, list, describe, update, delete, and `pair trigger-switch` (failover/switchback). **[Early Access]**

Bug Fixes
- N/A

Checklist
---------
- [x] I have clearly specified in the `What` section below that this PR applies to **Confluent Cloud**.
- [x] I have verified this PR in a Confluent Cloud pre-prod environment (**devel** for the full lifecycle; **stag** for authentication).
- [ ] I have verified this PR in Confluent Platform on-premises environment. _(N/A — Cloud only)_
- [x] I have attached manual CLI verification results in the `Test & Review` section below.
  - [ ] Confluent Cloud prod
  - [x] Confluent Cloud stag
  - [ ] Confluent Platform
  - [x] Check this box if the feature is enabled for certain organizations only _(Early Access allowlist: `switchover.api.enable` flag → `switchover-enabled-orgs` LaunchDarkly segment, enforced by `cc-switchover` as a 403)_

What
----
Adds Confluent CLI support for the **Switchover** API (Kafka Disaster Recovery), applicable to **Confluent Cloud** only. Two command groups:

- `confluent switchover pair` — `create` / `list` / `describe` / `update` / `delete` / `trigger-switch` (PLANNED/UNPLANNED/RESTORE). Members are passed as `--member name=<name>,crn=<member-crn>` (×2).
- `confluent switchover endpoint` — `create` / `list` / `describe` / `update` / `delete`. Endpoint sides are passed as `--endpoint name=<name>,type=<private|public>[,network=…|,access-point=…]` (×2).

Consumes the **public** `github.com/confluentinc/ccloud-sdk-go-v2/switchover v0.1.0` SDK (delete returns 202-with-body; endpoint conditions use the dedicated `SwitchoverV1SwitchoverEndpointCondition` type).

### Authentication

The Switchover API is served by **frontdoor-api-gateway**, which accepts Cloud / Global API keys and regional customer access tokens but **not** raw login-session JWTs (the credential types it validates are the `CredentialType` enum in [cc-auth-dataplane `token_platform.proto`](https://github.com/confluentinc/cc-auth-dataplane/blob/master/pkg/api/cc_auth_dataplane/v2/token_platform.proto#L205)). Sending the session token directly yields `401 Unauthorized`. The switchover client therefore resolves credentials from the CLI's **existing** mechanisms — the login session and the `api-key` keystore — in this order (`pkg/ccloudv2/switchover.go`, `switchoverApiContext`):

| # | Credential | Set by | Sent as | Role |
|---|---|---|---|---|
| 1 | **Active Global API key** in the local keystore | `api-key create --resource global` / `api-key use <key>` | HTTP Basic | Explicit override. A key the user deliberately selected wins over ambient login state, so the acting principal can be pinned (e.g. a service account for a failover) and the token exchange bypassed. Break-glass path. |
| 2 | **Login session** → regional customer access token | `confluent login` | Bearer | **Default** for logged-in users. The session token is exchanged via `auth.GetRegionalToken` (#3513, merged) — the `regional_token` the existing `POST /api/access_tokens` call already returns — the same way the Kafka REST / Flink clients authenticate. |
| 3 | Cloud API key stored in the context by an API-key login | — | HTTP Basic | Contexts with no session. |
| 4 | Raw session JWT | — | Bearer | Last resort; frontdoor rejects it today. Kept so a failure surfaces as a 401 rather than a silent unauthenticated request. |

No switchover-specific environment variables. (An interim revision of this branch read `CONFLUENT_CLOUD_API_KEY` / `CONFLUENT_CLOUD_API_SECRET`; those are used by no other CLI command and were replaced by the keystore. The keystore holds Kafka-cluster and **Global** keys, so the key path is specifically the Global API key — also the single credential the DR design wants valid across both members of a pair.)

Test & Review
-------------
**Build & lint:** `make build` + `make lint` (both `lint-go` and the `lint-cli` spell-check) pass cleanly.

**Integration tests** (`test/switchover_test.go` + test-server handler in `test/test-server/switchover_handler.go`, routes wired in `ccloudv2_router.go`) — cover pair & endpoint create/list/describe/update/delete, `pair trigger-switch`, JSON output, and not-found error paths:
```
$ make integration-test INTEGRATION_TEST_ARGS="-run TestCLI/TestSwitchover"
ok  github.com/confluentinc/cli/v4/test
```
Golden files under `test/fixtures/output/switchover/**` (help + CRUD). The `/api/access_tokens` exchange used by tier 2 is unit-tested in `pkg/auth/auth_test.go` (landed with #3513).

**Live testing against devel** (custom binary, Cloud API key, `env-devc7mdr2w`, frontdoor `api.devel.cpdev.cloud`) — the **full lifecycle was executed end-to-end** on a real DR-linked cluster pair (`lkc-devcnvx9rgd` us-west-2 / `lkc-devc81voy9q` us-east-2):

1. **`pair create`** → pair reached `READY_TO_FAILOVER` ✓
2. **`endpoint create`** (private sides, PrivateLink-attachment access-points `gateway=<platt>/access-point=<plattc>`) → endpoint reached `READY` ✓
3. **`pair trigger-switch --failover-type PLANNED`** → **real failover**: active member flipped `east → west`, phase `READY_TO_FAILOVER → UPDATING → READY_TO_FAILOVER`, `First Active` retained ✓
4. **`pair describe`** → post-failover active member confirmed `west` ✓
5. **`endpoint delete` then `pair delete`** → async 202-with-body; the 409 parent-child ordering ("endpoint must be deleted first") is enforced correctly ✓

Also verified live: `pair list` / `describe` / `describe -o json`, `endpoint list` / `describe`, `pair update` (real rename mutation + revert), and `pair describe` correctly surfacing `PROVISIONING` / `READY_TO_FAILOVER` / `FAILED` phases with the `ResourceValidationComplete` condition. All test resources were cleaned up afterward.

**Authentication, live against stag** (2026-09-16, `confluent login --url https://stag.cpdev.cloud`, org `51dbc82f-…`, env `env-stgc13kyr3`, pair `sw-e4444bc76c52`; `--unsafe-trace` used to confirm which credential was sent):

| Scenario | Credential sent | Result |
|---|---|---|
| Plain login, no Global key active — `pair list`, `pair describe`, `endpoint list` | Bearer (regional token; `POST /api/access_tokens` observed first) | **200** — previously 401 |
| Global key stored but **not** active + logged in | Bearer (regional token) | 200 — a stored-but-inactive key does not override login |
| `api-key use <global-key>` + logged in — same three commands | Basic; **no** `/api/access_tokens` exchange | **200** — also confirms frontdoor accepts a **Global** key, not only a `cloud` key |
| Active key cleared | Bearer (regional token) | 200 — back to the login default |
| Raw login-session JWT (manual `curl`; same token returns 200 on `/cmk/v2`) | Bearer (session JWT) | 401 — the rejection is the credential *type*, not the user |

Full command/response log and auth analysis: **[Switchover CLI — Testing Report](https://confluentinc.atlassian.net/wiki/spaces/~7120209995ae3a4a734547acfbc8d1a84aa646/pages/6178045953/Switchover+CLI+Testing+Report)** (see "Authentication findings").

Blast Radius
------------
New commands only; **no changes to existing commands or output**. The Switchover API is **Early Access** and org-gated, so exposure is limited to allowlisted organizations. Worst case if something regresses is confined to the new `switchover` commands for those orgs.

The one behavior reviewers should consciously accept: when a **Global API key is active** in the keystore, switchover commands run as that key's principal rather than the logged-in user (tier 1 above). This is deliberate — an explicitly selected credential beats ambient session state, which is what makes the key usable as a break-glass / pinned-principal path for `trigger-switch` — but it means a user who keeps a Global key active for Kafka commands will see switchover calls attributed to the key. `--unsafe-trace` shows which credential was used (`Basic` vs `Bearer`).

References
----------
- #3513 — `Expose regional_token from the /api/access_tokens exchange` (merged; prerequisite for tier 2)
- [Token Platform: Client Switchover LOEs](https://confluentinc.atlassian.net/wiki/spaces/CANP/pages/6116081973) — the remaining OAuth/STS credential-type work, which affects Terraform / customer automation, not `confluent login`
- [frontdoor-api-gateway `registration/cpi-team/cc-switchover-service.yaml`](https://github.com/confluentinc/frontdoor-api-gateway/blob/master/registration/cpi-team/cc-switchover-service.yaml) — route registration

### DIFF
--- a/test/switchover_test.go
+++ b/test/switchover_test.go
@@ -1,0 +1,27 @@
+package test
+
+func (s *CLITestSuite) TestSwitchover() {
+	tests := []CLITest{
+		// SwitchoverPair
+		{args: "switchover pair create prod-kafka-dr --member west=lkc-111111 --member east=lkc-222222 --active-member west", fixture: "switchover/pair/create.golden"},
+		{args: "switchover pair list", fixture: "switchover/pair/list.golden"},
+		{args: "switchover pair describe sw-123456", fixture: "switchover/pair/describe.golden"},
+		{args: "switchover pair update sw-123456 --name renamed-dr", fixture: "switchover/pair/update.golden"},
+		{args: "switchover pair failover sw-123456 --member east --type CLEAN", fixture: "switchover/pair/failover.golden"},
+		{args: "switchover pair delete sw-123456 --force", fixture: "switchover/pair/delete.golden"},
+		{args: "switchover pair describe sw-000000", fixture: "switchover/pair/describe-not-found.golden", exitCode: 1},
+		{args: "switchover pair describe sw-123456 --output json", fixture: "switchover/pair/describe-json.golden"},
+
+		// SwitchoverEndpoint
+		{args: "switchover endpoint create prod-endpoint --switchover-pair sw-123456 --endpoint name=west-platt,resource-id=lkc-111111,type=PRIVATE --endpoint name=east-platt,resource-id=lkc-222222,type=PRIVATE", fixture: "switchover/endpoint/create.golden"},
+		{args: "switchover endpoint list", fixture: "switchover/endpoint/list.golden"},
+		{args: "switchover endpoint describe se-123456", fixture: "switchover/endpoint/describe.golden"},
+		{args: "switchover endpoint activate se-123456", fixture: "switchover/endpoint/activate.golden"},
+		{args: "switchover endpoint delete se-123456 --force", fixture: "switchover/endpoint/delete.golden"},
+	}
+
+	for _, test := range tests {
+		test.login = "cloud"
+		s.runIntegrationTest(test)
+	}
+}
--- a/cmd/lint/main.go
+++ b/cmd/lint/main.go
@@ -375,6 +375,7 @@ var vocabWords = []string{
 	"whitelist",
 	"wikipedia",
 	"workspace",
+	"switchover",
 	"yaml",
 	"yml",
 	"zstd",
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/charmbracelet/lipgloss v0.11.0
 	github.com/client9/gospell v0.0.0-20160306015952-90dfc71015df
 	github.com/confluentinc/ccloud-sdk-go-v1-public v0.0.0-20250521223017-0e8f6f971b52
+	github.com/confluentinc/ccloud-sdk-go-v2-internal/switchover v0.0.0-20260612194002-842615b031fd
 	github.com/confluentinc/ccloud-sdk-go-v2/ai v0.1.0
 	github.com/confluentinc/ccloud-sdk-go-v2/apikeys v0.4.0
 	github.com/confluentinc/ccloud-sdk-go-v2/billing v0.3.0
--- a/go.sum
+++ b/go.sum
@@ -172,6 +172,8 @@ github.com/cncf/udpa/go v0.0.0-20200629203442-efcf912fb354/go.mod h1:WmhPx2Nbnht
 github.com/cncf/udpa/go v0.0.0-20201120205902-5459f2c99403/go.mod h1:WmhPx2Nbnhtbo57+VJT5O0JRkEi1Wbu0z5j0R8u5Hbk=
 github.com/confluentinc/ccloud-sdk-go-v1-public v0.0.0-20250521223017-0e8f6f971b52 h1:19qEGhkbZa5fopKCe0VPIV+Sasby4Pv10z9ZaktwWso=
 github.com/confluentinc/ccloud-sdk-go-v1-public v0.0.0-20250521223017-0e8f6f971b52/go.mod h1:62EMf+5uFEt1BJ2q8WMrUoI9VUSxAbDnmZCGRt/MbA0=
+github.com/confluentinc/ccloud-sdk-go-v2-internal/switchover v0.0.0-20260612194002-842615b031fd h1:+cvg0ZJFpRo9NhF0hYCov6zCJ5BXPbrjvcXrHD7iVuE=
+github.com/confluentinc/ccloud-sdk-go-v2-internal/switchover v0.0.0-20260612194002-842615b031fd/go.mod h1:75XYxEMix/kXHlKza4pP9P2in9w1tB3vM2TY3G4mHl4=
 github.com/confluentinc/ccloud-sdk-go-v2/ai v0.1.0 h1:zSF4OQUJXWH2JeAo9rsq13ibk+JFdzITGR8S7cFMpzw=
 github.com/confluentinc/ccloud-sdk-go-v2/ai v0.1.0/go.mod h1:DoxqzzF3JzvJr3fWkvCiOHFlE0GoYpozWxFZ1Ud9ntA=
 github.com/confluentinc/ccloud-sdk-go-v2/apikeys v0.4.0 h1:8fWyLwMuy8ec0MVF5Avd54UvbIxhDFhZzanHBVwgxdw=
--- a/internal/command.go
+++ b/internal/command.go
@@ -44,6 +44,7 @@ import (
 	"github.com/confluentinc/cli/v4/internal/secret"
 	servicequota "github.com/confluentinc/cli/v4/internal/service-quota"
 	streamshare "github.com/confluentinc/cli/v4/internal/stream-share"
+	"github.com/confluentinc/cli/v4/internal/switchover"
 	"github.com/confluentinc/cli/v4/internal/tableflow"
 	unifiedstreammanager "github.com/confluentinc/cli/v4/internal/unified-stream-manager"
 	"github.com/confluentinc/cli/v4/internal/update"
@@ -138,6 +139,7 @@ func NewConfluentCommand(cfg *config.Config) *cobra.Command {
 	cmd.AddCommand(streamshare.New(prerunner))
 	cmd.AddCommand(tableflow.New(prerunner))
 	cmd.AddCommand(unifiedstreammanager.New(cfg, prerunner))
+	cmd.AddCommand(switchover.New(prerunner))
 	cmd.AddCommand(update.New(cfg, prerunner))
 	cmd.AddCommand(version.New(prerunner, cfg.Version))
 	// cli-tfgen:cli-commands — DO NOT REMOVE (verified by TestCliTfgenMarkers)
--- a/internal/switchover/command.go
+++ b/internal/switchover/command.go
@@ -1,0 +1,29 @@
+package switchover
+
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+)
+
+type command struct {
+	*pcmd.AuthenticatedCLICommand
+}
+
+// New returns the parent `confluent switchover` command, which groups the
+// Kafka Disaster Recovery resources: switchover pairs and switchover endpoints.
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "switchover",
+		Short:       "Manage Kafka disaster recovery switchover resources.",
+		Long:        "Manage Kafka disaster recovery switchover pairs and switchover endpoints in Confluent Cloud.",
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireCloudLogin},
+	}
+
+	c := &command{AuthenticatedCLICommand: pcmd.NewAuthenticatedCLICommand(cmd, prerunner)}
+
+	cmd.AddCommand(c.newPairCommand())
+	cmd.AddCommand(c.newEndpointCommand())
+
+	return cmd
+}
--- a/internal/switchover/command_endpoint.go
+++ b/internal/switchover/command_endpoint.go
@@ -1,0 +1,138 @@
+package switchover
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	switchoverv1 "github.com/confluentinc/ccloud-sdk-go-v2-internal/switchover/v1"
+
+	"github.com/confluentinc/cli/v4/pkg/ccloudv2"
+	"github.com/confluentinc/cli/v4/pkg/output"
+)
+
+func (c *command) newEndpointCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "endpoint",
+		Short: "Manage switchover endpoints.",
+		Long:  "Manage endpoint-level Disaster Recovery switchover endpoints for Kafka.",
+		Args:  cobra.NoArgs,
+	}
+
+	cmd.AddCommand(c.newEndpointActivateCommand())
+	cmd.AddCommand(c.newEndpointCreateCommand())
+	cmd.AddCommand(c.newEndpointDeleteCommand())
+	cmd.AddCommand(c.newEndpointDescribeCommand())
+	cmd.AddCommand(c.newEndpointListCommand())
+	cmd.AddCommand(c.newEndpointUpdateCommand())
+
+	return cmd
+}
+
+// parseEndpoint parses an `--endpoint` flag value into a SwitchoverV1EndpointConfig.
+// Format: comma-separated key=value pairs, for example:
+//
+//	name=west-platt,resource-id=lkc-west,type=PRIVATE,cloud=AWS,region=us-west-2,gateway=n-ccn-west-1
+//
+// `name`, `resource-id`, and `type` are required; `cloud`, `region`, `gateway`,
+// and `access-point` are optional.
+func parseEndpoint(raw string) (switchoverv1.SwitchoverV1EndpointConfig, error) {
+	fields := map[string]string{}
+	for _, pair := range strings.Split(raw, ",") {
+		if strings.TrimSpace(pair) == "" {
+			continue
+		}
+		kv := strings.SplitN(pair, "=", 2)
+		if len(kv) != 2 {
+			return switchoverv1.SwitchoverV1EndpointConfig{}, fmt.Errorf(`invalid --endpoint value %q: each field must be "key=value"`, raw)
+		}
+		fields[strings.TrimSpace(kv[0])] = strings.TrimSpace(kv[1])
+	}
+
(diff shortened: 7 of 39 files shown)


KetanA:
Auth prerequisite split out into #3513 (expose `regional_token` from the `/api/access_tokens` exchange). Once that lands on main, this branch will be rebased and the switchover client will switch its login-path credential from the raw session JWT to `auth.GetRegionalToken` — frontdoor accepts `CREDENTIAL_TYPE_REGIONAL_CUSTOMER_ACCESS_TOKEN` but not session JWTs, so this is what makes plain `confluent login` work for these commands (verified E2E in stag). Cloud API key (Basic auth) stays as the first-preference credential and the ORC-10314 incident fallback.

internal/switchover/endpoint/command.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// out is the detailed human-readable shape used by create/describe/update.
+// Machine-readable output (`-o json` / `-o yaml`) is emitted straight from the
+// SDK object so it matches the API response verbatim.
+type out struct {
+	Id             string `human:"ID"`
+	DisplayName    string `human:"Display Name"`
+	SwitchoverPair string `human:"Switchover Pair"`
+	Environment    string `human:"Environment"`
```
sgagniere:
```suggestion
	EnvironmentCrn    string `human:"Environment CRN"`
```

Field names should be the same (other than casing) between human and serialized formats.

internal/switchover/endpoint/command.go
```diff
@@ -0,0 +1,141 @@
...
+	*pcmd.AuthenticatedCLICommand
+}
+
+// out is the detailed human-readable shape used by create/describe/update.
+// Machine-readable output (`-o json` / `-o yaml`) is emitted straight from the
+// SDK object so it matches the API response verbatim.
+type out struct {
+	Id             string `human:"ID"`
+	DisplayName    string `human:"Display Name"`
+	SwitchoverPair string `human:"Switchover Pair"`
```
sgagniere:
```suggestion
	ParentResourceCrn string `human:"Parent Resource CRN"`
```
Same reason for this one.

internal/switchover/endpoint/command.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// listOut is the compact per-row shape for `list` (human output only).
+type listOut struct {
+	Id             string `human:"ID"`
+	DisplayName    string `human:"Display Name"`
+	SwitchoverPair string `human:"Switchover Pair"`
+	Environment    string `human:"Environment"`
+	Phase          string `human:"Phase"`
+}
```
sgagniere: While having a lot of fields/columns can make the list a bit unreadable, we don't want to split the outputs between describe & list in general.

When we replace this with our automation down the line, it will automatically replace these with a common struct for both.

internal/switchover/endpoint/command.go
```diff
@@ -0,0 +1,141 @@
...
+
+func New(prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "endpoint",
+		Short: "Manage switchover endpoints.",
+		Long:  "Manage switchover endpoints. This API is not yet implemented on the backend; commands will fail against a live Confluent Cloud environment.",
```
sgagniere:
```suggestion
```
We can add this to the release note instead of in the description; this is something we could accidentally forget to remove and it would be confusing later.

internal/switchover/endpoint/command.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return err
+		}
+		output.Print(false, string(out))
+		return nil
+	}
+
+	output.Print(false, string(pretty.Pretty(data)))
+	return nil
+}
```
sgagniere: This is a good general tool, let's move it to pkg/output/serialized_output.go.

There's a similarly named function in there already (I don't think it handles this situation though) so we'll probably need to rename this one a bit to distinguish them more.

internal/switchover/endpoint/command_create.go
```diff
@@ -0,0 +1,143 @@
...
+		RunE:  c.create,
+		Example: examples.BuildExampleString(
+			examples.Example{
+				Text: `Create switchover endpoint "prod-kafka-dr-endpoint" for switchover pair "sw-123456".`,
+				Code: `confluent switchover endpoint create prod-kafka-dr-endpoint --switchover-pair sw-123456 --endpoint name=west-platt,type=private --endpoint name=east-platt,type=private`,
+			},
+		),
+	}
+
+	cmd.Flags().String("switchover-pair", "", "The ID of the switchover pair this endpoint is bound to.")
```
sgagniere: While this is sensible from a UX perspective, when we switch to auto-generated code later this will end up as `parent_resource_crn` because that's all that can be deduced from the API spec.

internal/switchover/pair/command.go
```diff
@@ -0,0 +1,131 @@
...
+	*pcmd.AuthenticatedCLICommand
+}
+
+// out is the detailed human-readable shape used by create/describe/update/
+// trigger-switch. Machine-readable output (`-o json` / `-o yaml`) is emitted
+// straight from the SDK object so it matches the API response verbatim.
+type out struct {
+	Id           string `human:"ID"`
+	DisplayName  string `human:"Display Name"`
+	Environment  string `human:"Environment"`
```
sgagniere: Same comment here as in the other resource.

internal/switchover/pair/command.go
```diff
@@ -0,0 +1,131 @@
...
+
+// listOut is the compact per-row shape for `list` (human output only).
+type listOut struct {
+	Id           string `human:"ID"`
+	DisplayName  string `human:"Display Name"`
+	Environment  string `human:"Environment"`
+	ActiveMember string `human:"Active Member"`
+	FailoverType string `human:"Failover Type"`
+	Phase        string `human:"Phase"`
+}
```
sgagniere: Same comment about describe/list output split here.

internal/switchover/pair/command_trigger_switch.go
```diff
@@ -0,0 +1,83 @@
...
+	switchoverv1 "github.com/confluentinc/ccloud-sdk-go-v2/switchover/v1"
+
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+	"github.com/confluentinc/cli/v4/pkg/deletion"
+	"github.com/confluentinc/cli/v4/pkg/examples"
+)
+
+func (c *command) newTriggerSwitchCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "trigger-switch <id>",
```
sgagniere:
```suggestion
		Use:   "failover <id>",
```

Let's rename this to `failover` to align with the custom method name on this api path: `/switchover/v1/switchover-pairs/{id}:failover`.

test/switchover_test.go
```diff
@@ -0,0 +1,28 @@
...
+		{args: "switchover pair update sw-123456 --display-name renamed-dr", fixture: "switchover/pair/update.golden"},
+		{args: "switchover pair trigger-switch sw-123456 --active-member east --failover-type UNPLANNED --force", fixture: "switchover/pair/trigger-switch.golden"},
+		{args: "switchover pair delete sw-123456 --force", fixture: "switchover/pair/delete.golden"},
+		{args: "switchover pair describe sw-000000", fixture: "switchover/pair/describe-not-found.golden", exitCode: 1},
+		{args: "switchover pair describe sw-123456 --output json", fixture: "switchover/pair/describe-json.golden"},
+
+		// SwitchoverEndpoint
+		{args: `switchover endpoint create prod-endpoint --switchover-pair sw-123456 --endpoint name=west-platt,type=private,network=n-111111 --endpoint name=east-platt,type=private,network=n-222222`, fixture: "switchover/endpoint/create.golden"},
+		{args: "switchover endpoint list", fixture: "switchover/endpoint/list.golden"},
+		{args: "switchover endpoint describe se-123456", fixture: "switchover/endpoint/describe.golden"},
```
sgagniere, line 19: Let's add a json output test on one of these too.

And ideally a yaml one for both since these commands use a special output function for json & yaml

SWITCHOVER_LOCAL_DEMO.md
```diff
@@ -0,0 +1,217 @@
+# Local CLI build with Switchover support
```
sgagniere: This looks like a development file, is it safe to remove this one from the PR?

sgagniere (commented):
Thanks for the PR! Here's some comments:

internal/switchover/endpoint/command.go
KetanA: +1, this should be `EnvironmentCrn` as the output gives a crn.
Screenshot 2026-09-22 at 7 51 46 PM
cc: @ngaur99

internal/switchover/endpoint/command.go
KetanA: +1, SWE API returns `parent_resource_crn` field, so we can be consistent here.
`ParentResourceCrn string `human:"Parent Resource CRN"`` makes sense.
Screenshot 2026-09-22 at 7 57 14 PM
In above response, it should return as `ParentResourceCrn`
cc: @ngaur99

internal/switchover/endpoint/command.go
KetanA: +1 @ngaur99 we can drop the `listOut` struct for both pairand endpoints and use `out` consistently across all the (`list`, `describe`, `create`..) commands

internal/switchover/endpoint/command.go
KetanA: @ngaur99 Actually this is incorrect long description. Good catch @sgagniere .
The `endpoint` APIs are implemented and are available in prod. We need to change the long description to something appropriate or remove altogether (same as in pair - only keep short)

internal/switchover/endpoint/command.go
KetanA: +1

internal/switchover/endpoint/command_create.go
KetanA: Yes, let's make the change above as you suggested and reflect the same here.
Would it be `parent-resource-crn` or `parent_resource_crn`?

KetanA: I think it'll be `parent-resource-crn` as the flags in other commands also use `-` and not `_`

internal/switchover/pair/command.go
KetanA: +1 @ngaur99 
It'd be `EnvironmentCrn    string `human:"Environment CRN"`

internal/switchover/pair/command.go
KetanA: +1 @ngaur99 let's get rid of `listOut` struct and use `out` consistently

internal/switchover/pair/command_trigger_switch.go
KetanA: +1

SWITCHOVER_LOCAL_DEMO.md
KetanA: yeah, should be removed

internal/switchover/endpoint/command.go
ngaur99: Done in 9478b08fb — `EnvironmentCrn string `human:"Environment CRN"``, so the serialized name matches the human label.

internal/switchover/endpoint/command.go
ngaur99: Done in 9478b08fb — `ParentResourceCrn string `human:"Parent Resource CRN"``, matching the API's `parent_resource_crn`.

internal/switchover/endpoint/command.go
ngaur99: Done in 9478b08fb — dropped `listOut`; `list` uses the same `out` struct as the other commands; also disables table auto-wrap so the multi-line `Endpoints` cell stays intact.

internal/switchover/endpoint/command.go
ngaur99: Done in 9478b08fb — removed the `Long` entirely (the endpoint API is live), keeping only `Short`, same as `pair`.

internal/switchover/endpoint/command.go
ngaur99: Done in 9478b08fb — moved to `pkg/output/serialized_output.go` as `SerializedOutputFromJsonTags`, beside the existing `SerializedOutput`.

internal/switchover/endpoint/command_create.go
ngaur99: Done in 9478b08fb — `--parent-resource-crn <crn>`, passed straight through as `spec.parent_resource_crn` (verified live on devel); `list --switchover-pair` kept since it maps to the API query param.

internal/switchover/pair/command.go
ngaur99: Done in 9478b08fb — `EnvironmentCrn string `human:"Environment CRN"`` on the pair struct as well.

internal/switchover/pair/command.go
ngaur99: Done in 9478b08fb — `listOut` removed for pairs too; `list` uses `out` via `newPairOut`.

internal/switchover/pair/command_trigger_switch.go
ngaur99: Done in 9478b08fb — renamed to `confluent switchover pair failover <id>` (file `command_failover.go`, goldens renamed). Flags are unchanged.

test/switchover_test.go
ngaur99: Done in 9478b08fb — added `--output json` for endpoint and `--output yaml` for both pair and endpoint, with goldens.

SWITCHOVER_LOCAL_DEMO.md
ngaur99: Done in 9478b08fb — `SWITCHOVER_LOCAL_DEMO.md` removed from the PR.

sgagniere (dismissed):
+1

KetanA (dismissed):

PR #3419, opened 2026-07-30
[FSE-1855] Read Flink statement warnings from status.warnings

raminqaf:
### Release Notes

New Features
- Show Flink SQL statement warnings with their severity, reason, and timestamp in the Flink SQL shell, on dry runs, and in confluent flink statement describe and confluent flink statement create. Warnings are listed most severe first.
- Add a warnings field to the serialized output of confluent flink statement describe, confluent flink statement create, and confluent flink statement list so -o json and -o yaml consumers can read each warning's severity, reason, message, and creation time.

Bug Fixes
- Fix Flink SQL statement warnings disappearing shortly after a statement was submitted, and multiple warnings being joined into a single unreadable line.

Checklist

- [x] I have clearly specified in the What section below whether this PR applies to Confluent Cloud, Confluent Platform, or both.
- [ ] I have verified this PR in Confluent Cloud pre-prod or production environment, if applicable. Not yet done. Verified against the integration test server only, see Test & Review.
- [ ] I have verified this PR in Confluent Platform on-premises environment, if applicable. Not applicable, see What.
- [x] I have attached manual CLI verification results or screenshots in the Test & Review section below.
- [x] I have indicated the potential customer impact if something goes wrong in the Blast Radius section below.
  - [x] Confluent Cloud prod
  - [x] Confluent Cloud stag
  - [ ] Confluent Platform

The CLI only renders the warnings the API returns. Where a statement has no warnings, output is unchanged, so no CLI-side enablement is needed.

### What

Applies to Confluent Cloud only. The warnings field is part of the Confluent Cloud Flink statement status. The Confluent Platform on-premises code paths are untouched.

The CLI displayed statement warnings by printing status.detail as a warning banner. That field is a single flat string. It has no severity, no machine-readable reason, and no timestamp, and it is rewritten as a statement moves through its lifecycle. Two consequences for users: warnings vanished from view shortly after submission, and several warnings arrived concatenated into one long line that was hard to read and impossible to render as a list.

Statement warnings are now read from the structured status.warnings array. Each warning is rendered with its severity, reason, and creation time, most severe first:
```
Warnings:

CRITICAL [HIGH_STATE_OPERATOR_WITHOUT_TTL] (Logged: 2022-01-01T00:00:00Z)
Your query includes one or more highly state-intensive operators but does not set a time-to-live (TTL) value.

MODERATE [MISSING_WINDOW_START_END] (Logged: 2022-01-01T00:00:00Z)
The GROUP BY clause contains only `window_start` with no corresponding `window_end`.
```
### Notes on the approach:

- Surfaces covered. The Flink SQL shell, dry runs, confluent flink statement describe, and confluent flink statement create. confluent flink statement list carries the warnings in serialized output but does not render a block per row, which would bury the table.
- Serialized output. A new warnings field is added to describe, create, and list. It is omitempty, so output for statements without warnings is byte-for-byte unchanged. This is an additive field only.
- Human output keeps prose out of the table. A warning message is a paragraph, so it is printed below the table rather than in a cell.
- status.detail is still printed when a statement has no warnings, and always for a failed statement, where the detail carries the failure reason. Skipping it when warnings are present avoids showing the same warning twice, since the detail can still repeat the warning text.
- Unrecognized severities are displayed, not dropped. Severity is an extensible enum, so a value this version of the CLI does not know is still shown and sorted last.

### Blast Radius

Confluent Cloud customers using confluent flink shell, confluent flink statement describe, confluent flink statement create, or confluent flink statement list.

If something goes wrong, the worst case is a display defect: warnings are not shown, are shown in the wrong order, or the Status Detail line is omitted for a statement that has warnings. Statement submission, execution, and results are not touched, so no customer would be blocked from running a statement. Serialized output gains a field and loses none, so existing scripts that parse -o json or -o yaml continue to work.

Confluent Platform on-premises users are unaffected.

### References

- FSE-1855

### Test & Review

Automated:

- Unit tests for the conversion from the API type, severity ordering, unrecognized severities, timestamp rendering in UTC, and the header format. See pkg/flink/types/statement_warning_test.go.
- Unit tests for the print rules, including that a failed statement still shows its detail alongside its warnings, and that a statement without warnings prints exactly what it did before. See pkg/flink/types/processed_statement_test.go.
- A snapshot test for the shell path in pkg/flink/internal/controller/statement_controller_test.go. Writing this caught a real bug: the formatter only produced sorted output when the slice came from the converter, so a caller-built slice printed unsorted. The formatter now sorts a copy of its input.
- Integration tests for describe in human and YAML output, with new golden files. The test server returns warnings only for a dedicated statement name, so no existing golden file changed.

Commands run:
```
go test ./internal/... ./pkg/...
golangci-lint run
make integration-test INTEGRATION_TEST_ARGS="-run TestCLI/TestFlinkStatement"
```

Verified output, human format:
```
+--------------------------+---------------------------------------------------------+
| Creation Date            | 2022-01-01 00:00:00 +0000 UTC                           |
| Name                     | my-statement-with-warnings                              |
| Statement                | CREATE TABLE test;                                      |
| Compute Pool             | lfcp-123456                                             |
| Status                   | COMPLETED                                               |
| Status Detail            | SQL statement is completed                              |
| Latest Offsets           | customers_source=partition:0,offset:9223372036854775808 |
| Latest Offsets Timestamp | 2022-01-01 00:00:00 +0000 UTC                           |
| Properties               | sql.current-catalog=default                             |
|                          | sql.current-database=my-cluster                         |
| Principal                | u-123456                                                |
+--------------------------+---------------------------------------------------------+

Warnings:

CRITICAL [HIGH_STATE_OPERATOR_WITHOUT_TTL] (Logged: 2022-01-01T00:00:00Z)
Your query includes one or more highly state-intensive operators but does not set a time-to-live (TTL) value.

MODERATE [MISSING_WINDOW_START_END] (Logged: 2022-01-01T00:00:00Z)
The GROUP BY clause contains only `window_start` with no corresponding `window_end`.
```

Verified output, -o yaml:
```
creation_date: 2022-01-01T00:00:00Z
name: my-statement-with-warnings
statement: CREATE TABLE test;
compute_pool: lfcp-123456
status: COMPLETED
status_detail: SQL statement is completed
warnings:
    - severity: CRITICAL
      reason: HIGH_STATE_OPERATOR_WITHOUT_TTL
      message: Your query includes one or more highly state-intensive operators but does not set a time-to-live (TTL) value.
      created_at: 2022-01-01T00:00:00Z
    - severity: MODERATE
      reason: MISSING_WINDOW_START_END
      message: The GROUP BY clause contains only `window_start` with no corresponding `window_end`.
      created_at: 2022-01-01T00:00:00Z
latest_offsets:
    customers_source: partition:0,offset:9223372036854775808
latest_offsets_timestamp: 2022-01-01T00:00:00Z
properties:
    sql.current-catalog: default
    sql.current-database: my-cluster
principal: u-123456
```

Built and tested locally

Describe output:
image

From Flink Shell
image

### DIFF
--- a/internal/flink/command_statement.go
+++ b/internal/flink/command_statement.go
@@ -24,13 +24,12 @@ func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarnin
 }
 
 type statementOut struct {
-	CreationDate time.Time `human:"Creation Date" serialized:"creation_date"`
-	Name         string    `human:"Name" serialized:"name"`
-	Statement    string    `human:"Statement" serialized:"statement"`
-	ComputePool  string    `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
-	Status       string    `human:"Status" serialized:"status"`
-	StatusDetail string    `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
-	// Rendered below the table, since a warning message is too long for a cell.
+	CreationDate           time.Time                `human:"Creation Date" serialized:"creation_date"`
+	Name                   string                   `human:"Name" serialized:"name"`
+	Statement              string                   `human:"Statement" serialized:"statement"`
+	ComputePool            string                   `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
+	Status                 string                   `human:"Status" serialized:"status"`
+	StatusDetail           string                   `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
 	Warnings               []types.StatementWarning `human:"-" serialized:"warnings,omitempty"`
 	LatestOffsets          map[string]string        `human:"Latest Offsets" serialized:"latest_offsets"`
 	LatestOffsetsTimestamp *time.Time               `human:"Latest Offsets Timestamp" serialized:"latest_offsets_timestamp"`
--- a/internal/flink/command_statement.go
+++ b/internal/flink/command_statement.go
@@ -4,17 +4,36 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+
+	"github.com/confluentinc/cli/v4/pkg/flink/types"
+	"github.com/confluentinc/cli/v4/pkg/output"
 )
 
+// printStatementWarnings renders warnings below the table, on stderr so that stdout stays the
+// command's data. Serialized output already carries them in the warnings field.
+func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarning) {
+	if output.GetFormat(cmd) != output.Human {
+		return
+	}
+
+	if block := types.FormatStatementWarnings(warnings); block != "" {
+		output.ErrPrintln(false, "")
+		output.ErrPrintln(false, block)
+		output.ErrPrintln(false, "")
+	}
+}
+
 type statementOut struct {
-	CreationDate           time.Time         `human:"Creation Date" serialized:"creation_date"`
-	Name                   string            `human:"Name" serialized:"name"`
-	Statement              string            `human:"Statement" serialized:"statement"`
-	ComputePool            string            `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
-	Status                 string            `human:"Status" serialized:"status"`
-	StatusDetail           string            `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
-	LatestOffsets          map[string]string `human:"Latest Offsets" serialized:"latest_offsets"`
-	LatestOffsetsTimestamp *time.Time        `human:"Latest Offsets Timestamp" serialized:"latest_offsets_timestamp"`
+	CreationDate time.Time `human:"Creation Date" serialized:"creation_date"`
+	Name         string    `human:"Name" serialized:"name"`
+	Statement    string    `human:"Statement" serialized:"statement"`
+	ComputePool  string    `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
+	Status       string    `human:"Status" serialized:"status"`
+	StatusDetail string    `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
+	// Rendered below the table, since a warning message is too long for a cell.
+	Warnings               []types.StatementWarning `human:"-" serialized:"warnings,omitempty"`
+	LatestOffsets          map[string]string        `human:"Latest Offsets" serialized:"latest_offsets"`
+	LatestOffsetsTimestamp *time.Time               `human:"Latest Offsets Timestamp" serialized:"latest_offsets_timestamp"`
 }
 
 func (c *command) newStatementCommand() *cobra.Command {
--- a/internal/flink/command_statement.go
+++ b/internal/flink/command_statement.go
@@ -27,13 +27,12 @@ func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarnin
 }
 
 type statementOut struct {
-	CreationDate time.Time `human:"Creation Date" serialized:"creation_date"`
-	Name         string    `human:"Name" serialized:"name"`
-	Statement    string    `human:"Statement" serialized:"statement"`
-	ComputePool  string    `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
-	Status       string    `human:"Status" serialized:"status"`
-	StatusDetail string    `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
-	// Rendered below the table, since a warning message is too long for a cell.
+	CreationDate           time.Time                `human:"Creation Date" serialized:"creation_date"`
+	Name                   string                   `human:"Name" serialized:"name"`
+	Statement              string                   `human:"Statement" serialized:"statement"`
+	ComputePool            string                   `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
+	Status                 string                   `human:"Status" serialized:"status"`
+	StatusDetail           string                   `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
 	Warnings               []types.StatementWarning `human:"-" serialized:"warnings,omitempty"`
 	LatestOffsets          map[string]string        `human:"Latest Offsets" serialized:"latest_offsets"`
 	LatestOffsetsTimestamp *time.Time               `human:"Latest Offsets Timestamp" serialized:"latest_offsets_timestamp"`
--- a/internal/flink/command_statement.go
+++ b/internal/flink/command_statement.go
@@ -8,17 +8,35 @@ import (
 	cmfsdk "github.com/confluentinc/cmf-sdk-go/v1"
 
 	"github.com/confluentinc/cli/v4/pkg/config"
+	"github.com/confluentinc/cli/v4/pkg/flink/types"
+	"github.com/confluentinc/cli/v4/pkg/output"
 )
 
+// printStatementWarnings renders warnings below the table, on stderr so that stdout stays the
+// command's data. Serialized output already carries them in the warnings field.
+func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarning) {
+	if output.GetFormat(cmd) != output.Human {
+		return
+	}
+
+	if block := types.FormatStatementWarnings(warnings); block != "" {
+		output.ErrPrintln(false, "")
+		output.ErrPrintln(false, block)
+		output.ErrPrintln(false, "")
+	}
+}
+
 type statementOut struct {
-	CreationDate           time.Time         `human:"Creation Date" serialized:"creation_date"`
-	Name                   string            `human:"Name" serialized:"name"`
-	Statement              string            `human:"Statement" serialized:"statement"`
-	ComputePool            string            `human:"Compute Pool,omitempty" serialized:"compute_pool,omitempty"`
-	Status                 string            `human:"Status" serialized:"status"`
-	StatusDetail           string            `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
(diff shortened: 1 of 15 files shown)


internal/flink/command_statement_describe.go
```diff
@@ -58,6 +61,8 @@ func (c *command) statementDescribe(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	warnings := types.NewStatementWarnings(statement.Status.GetWarnings())
+
```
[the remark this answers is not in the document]
raminqaf: Good catch on the duplication, but I'd rather not match on the [Warning] prefix. That string parsing is exactly what this change removes, and it would become dead code as soon as the server stops inlining warnings into detail, so we'd be adding a shim with a known expiry date.
Checked against two real statements to see how often this actually bites. In both, status.detail was already empty by the time it was read: one is a SHOW CREATE TABLE, whose handler never wrote warnings into detail at all, and the other is a SELECT where the controller had overwritten detail within a few minutes of submission. So the overlap window is short, and it's mostly reachable on create right after submission rather than on describe.
I also want describe to stay a faithful dump of what the API returned, rather than hiding a populated field based on its contents. If we do decide the surfaces should behave identically, the fix I'd pick is presence-based, drop the Status Detail row when warnings is non-empty and the phase isn't FAILED, which matches what the shell already does and needs no string matching. Happy to do that in a follow-up if reviewers prefer it.

internal/flink/command_statement_create.go
```diff
@@ -173,6 +173,8 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	warnings := types.NewStatementWarnings(statement.Status.GetWarnings())
```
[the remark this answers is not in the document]
raminqaf: Duplicate of the comment above

pkg/flink/types/statement_warning.go
```diff
@@ -0,0 +1,92 @@
...
+// severityOrder ranks severities for display, most severe first. Severity is an extensible enum, so
+// an unrecognized value is still displayed, sorted last.
+var severityOrder = []string{"CRITICAL", "MODERATE", "LOW"}
+
+// StatementWarning is a non-fatal issue reported for a statement.
+type StatementWarning struct {
+	Severity  string    `json:"severity" yaml:"severity"`
+	Reason    string    `json:"reason" yaml:"reason"`
+	Message   string    `json:"message" yaml:"message"`
+	CreatedAt time.Time `json:"created_at" yaml:"created_at"`
```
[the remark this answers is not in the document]
raminqaf: addressed

internal/flink/command_statement.go
```diff
@@ -8,17 +8,35 @@ import (
...
+func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarning) {
+	if output.GetFormat(cmd) != output.Human {
+		return
+	}
+
+	if block := types.FormatStatementWarnings(warnings); block != "" {
+		output.Println(false, "")
+		output.Println(false, block)
+		output.Println(false, "")
+	}
```
sgagniere, line 23: Would it work for the ticket to log these to `-v` (warn), at least for the non-shell outputs?

It's only in the human readable output so it's not necessarily an issue, but it does deviate from the usual pattern for CLI outputs.

If we do want these to always be displayed, let's write it to stderr instead.

sgagniere (commented):
Thanks!

I left a comment. Also, could we get some manual test results from the flink shell?

internal/flink/command_statement.go
raminqaf: Good point on the convention, moved to stderr for the non-shell commands. `stdout` is now just the table, so redirecting or piping gives you clean data, and `-o json/-o yaml` are untouched since they carry the warnings in the payload itself.
I left the shell on stdout. Everything there is already conversational output on stdout, including the status messages and the old `Details`: banner, so splitting only the warnings onto stderr would be inconsistent with the surrounding lines.
On `-v`: I'd rather not hide them behind a verbosity flag. The requirement for this feature is that all warnings are always shown, which is the whole reason for moving them out of `status.detail`, so making them opt-in would undo that.
Goldens are unchanged since the integration harness captures combined output.

internal/flink/command_statement_describe.go
```diff
@@ -8,20 +8,23 @@ import (
...
-	LatestOffsetsTimestamp *time.Time        `human:"Latest Offsets Timestamp" serialized:"latest_offsets_timestamp"`
-	Properties             map[string]string `human:"Properties" serialized:"properties"`
-	Principal              string            `human:"Principal" serialized:"principal"`
+	CreationDate time.Time `human:"Creation Date" serialized:"creation_date"`
+	Name         string    `human:"Name" serialized:"name"`
+	Statement    string    `human:"Statement" serialized:"statement"`
+	ComputePool  string    `human:"Compute Pool" serialized:"compute_pool"`
+	Status       string    `human:"Status" serialized:"status"`
+	StatusDetail string    `human:"Status Detail,omitempty" serialized:"status_detail,omitempty"`
+	// Rendered below the table, since a warning message is too long for a cell.
```
channingdong: nit: remove this comment from table and run `make fmt` to align the text

internal/flink/command_statement_list.go
```diff
@@ -100,6 +101,7 @@ func (c *command) statementList(cmd *cobra.Command, _ []string) error {
 			ComputePool:            statement.Spec.GetComputePoolId(),
 			Status:                 statement.Status.GetPhase(),
 			StatusDetail:           statement.Status.GetDetail(),
+			Warnings:               types.NewStatementWarnings(statement.Status.GetWarnings()),
```
channingdong, line 104: For the `list` command, can you try and see if the warnings can fit into the CLI console output? I am just a bit concerned that the warnings can sometimes be long, and we may just limit it to the `describe` command.

internal/flink/command_statement_create.go
```diff
@@ -181,7 +183,13 @@ func (c *command) statementCreate(cmd *cobra.Command, args []string) error {
...
+		Warnings:     warnings,
 	})
-	table.Filter([]string{"CreationDate", "Name", "Statement", "ComputePool", "Status", "StatusDetail"})
-	return table.Print()
+	table.Filter([]string{"CreationDate", "Name", "Statement", "ComputePool", "Status", "StatusDetail", "Warnings"})
+	if err := table.Print(); err != nil {
+		return err
+	}
+
+	printStatementWarnings(cmd, warnings)
```
channingdong, line 193: Are the warnings immediately available after the create of the statement if we decide not to `wait`?

internal/flink/command_statement_describe.go
```diff
@@ -66,10 +71,16 @@ func (c *command) statementDescribe(cmd *cobra.Command, args []string) error {
 		ComputePool:            statement.Spec.GetComputePoolId(),
 		Status:                 statement.Status.GetPhase(),
 		StatusDetail:           statement.Status.GetDetail(),
+		Warnings:               warnings,
```
channingdong, line 73: This means having the `warnings` in both regular table printout, as well as an additional rendered warnings below the table, just want to make sure we want this duplicated information here.

raminqaf: These are never both shown. `Warnings` is tagged `human:"-"` so it is excluded from the human table, and `printStatementWarnings` returns early unless the format is human. So human output is table plus the stderr block, and `-o json`/`-o yaml` is the payload field only. Matches the summary you posted in the thread.

internal/flink/command_statement_describe.go
raminqaf: Removed

internal/flink/command_statement_list.go
raminqaf: I agree with limiting the full block to `describe`. That is what the PR does today. list does not call `printStatementWarnings`, and `human:"-"` keeps the field out of the table.

One thing worth confirming: `list -o json` does carry warnings, I verified that the list endpoint populates status.warnings. So serialized consumers already get them.

If you want `list` users to see that warnings exist, a scalar column would fit the table where prose does not. Either a count or the highest severity, so you know which statement to go `describe`. Happy to add that if you think it earns the column width as a follow-up

internal/flink/command_statement_create.go
raminqaf: Yes. `--wait` defaults to `false`, and I verified against real devel that the warnings come back on the create response itself. A statement created without `--wait` printed its `MISSING_WINDOW_START_END` warning immediately. Validation-time warnings are attached before the statement leaves `PENDING`.

manner:
Hi! Is there anything that still needs to be done before this PR can be merged? I have a [follow up PR](https://github.com/confluentinc/flink-sql-service/pull/2674) that should only be merged after this PR.

internal/flink/command_statement.go
```diff
@@ -8,17 +8,34 @@ import (
 	cmfsdk "github.com/confluentinc/cmf-sdk-go/v1"
 
 	"github.com/confluentinc/cli/v4/pkg/config"
+	"github.com/confluentinc/cli/v4/pkg/flink/types"
+	"github.com/confluentinc/cli/v4/pkg/output"
 )
 
+// printStatementWarnings renders warnings below the table, on stderr so that stdout stays the
+// command's data. Serialized output already carries them in the warnings field.
+func printStatementWarnings(cmd *cobra.Command, warnings []types.StatementWarning) {
```
channingdong, line 14: It's ok to leave the dedicated human output function as is, but I may move it to a dedicated utils file so that we can automate all future CLI changes.

internal/flink/command_statement_create.go
channingdong, line 176: This line can be merged with line#186, but don't worry, I will make this change.

sgagniere:
Hi, please resolve the merge conflicts on this PR! We should be good to go after that.

channingdong (approved):

PR #3424, opened 2026-07-31
[CF-4208] Add `--page-size` to on-prem Flink list commands

paras-negi-flink:
Release Notes
-------------

New Features
- Added a `--page-size` flag to the on-prem (Confluent Platform / CMF) `confluent flink ... list` commands, to fetch large result sets in fewer, larger API requests. Defaults to 100 — existing behavior is unchanged.

Checklist
---------
- [x] I have successfully built and used a custom CLI binary, without linter issues from this PR.
- [x] I have clearly specified in the `What` section below whether this PR applies to Confluent Cloud, Confluent Platform, or both.
- [x] I have attached manual CLI verification results in the `Test & Review` section below.
- [x] I have added appropriate CLI integration or unit tests for any new or updated commands and functionality.
- [x] I confirm that this PR introduces no breaking changes or backward compatibility issues.
- [x] I have indicated the potential customer impact if something goes wrong in the `Blast Radius` section below.
- [ ] I have put checkmarks below confirming that the feature associated with this PR is enabled in:
  - [ ] Confluent Cloud prod
  - [ ] Confluent Cloud stag
  - [x] Confluent Platform
  - [ ] Check this box if the feature is enabled for certain organizations only

What
----
**Confluent Platform (CMF on-prem) only** — part of **CF-4208**; Confluent Cloud `flink` commands are untouched.

On-prem Flink list commands paginate at a hardcoded `size=100` and always fetch every page, so large environments cost many serial round trips (~51 requests for 5000 applications — CF-4202).

- Adds `--page-size` to all paginated on-prem list commands (`application` / `event` / `instance` / `catalog` / `catalog database` / `compute-pool` / `detached-savepoint` / `environment` / `savepoint` / `secret` / `secret-mapping` / `statement` list). Shared `addPageSizeFlag`/`getPageSize` in `command.go` validate `0..MaxInt32`; `0` = the default 100.
- Refactors the 12 duplicated pagination loops in `pkg/flink/cmf_rest_client.go` into one generic `listAllPages` helper (net −84 lines). `statement exception list` is untouched (non-paginated).

Blast Radius
----
- Additive optional flag; `--page-size 0` reproduces today's output exactly. Non-breaking, easy to revert.
- The shared pagination refactor is guarded by a `listAllPages` unit test plus multi-page integration coverage.

References
----------
- Jira: [CF-4208](https://confluentinc.atlassian.net/browse/CF-4208); motivation [CF-4202](https://confluentinc.atlassian.net/browse/CF-4202).
[...]

PR #3428, opened 2026-08-02
[CF-4208] Add `--filter` to on-prem Flink application list

paras-negi-flink:
Release Notes
-------------

New Features
- Added a `--filter` flag to the on-prem (Confluent Platform / CMF) `confluent flink application list` command, to filter large environments server-side instead of "list everything then grep".

Checklist
---------
- [x] I have attached manual CLI verification results in the `Test & Review` section below.
  - [ ] Confluent Cloud prod
  - [ ] Confluent Cloud stag
  - [x] Confluent Platform

> Builds on #3424 (`--page-size`), now merged into `main`; this PR targets `main` directly and contains only the filtering change.

What
----
**Confluent Platform (CMF on-prem) only** — filtering half of **CF-4208**; Confluent Cloud `flink` commands are untouched.

`flink application list` had no filtering, so at scale the only pattern was "list everything then grep" (CF-4202). This adds a single **`--filter`** flag that takes a CMF filter expression and applies it server-side:

- Comma-separated `key=value` terms, e.g. `--filter name=my-app*,state=RUNNING`.
- `name` supports a trailing `*` wildcard and its value is case-sensitive (Kubernetes resource names).
- `state` filters by Flink job state; its value is **case-insensitive** (`state=running` behaves like `state=RUNNING`).

This mirrors the generic `filter` shape the CLI code generator will produce for CMF resources, rather than inventing per-field flags — per review feedback on this PR. `ListApplications` gains a `filter` argument applied before pagination. The CLI upper-cases only `state=` values before sending (CMF matches the job-state enum case-sensitively); every other term is passed through verbatim, and there is no client-side validation — an unknown state is forwarded and simply matches nothing.

Blast Radius
----
- Scoped to `application list`; with no `--filter` set, behavior is unchanged (no `filter` param). No `CmfClientInterface`/mock change. Non-breaking, easy to revert.

References
----------
- Jira: [CF-4208](https://confluentinc.atlassian.net/browse/CF-4208); motivation [CF-4202](https://confluentinc.atlassian.net/browse/CF-4202).
- Builds on #3424 (`--page-size`), now merged.

Test & Review
-------------
- Unit `TestNormalizeApplicationFilter`: `state` folded to upper case (value and key case-insensitive), `name` and unknown keys passed through verbatim.
- Integration: `--filter name=` exact + wildcard, `--filter state=` in lower- and upper-case resolving to the same result (proves case-folding), unknown state → empty, combined `name=...,state=...`; help golden regenerated. `make lint` clean.

```shell
# name wildcard + state; state is case-insensitive (lower-case works)
$ confluent flink application list --environment default --filter name=default-application-1*,state=reconciling
          Name          | Environment |     Job Name      | Job Status
------------------------+-------------+-------------------+--------------
  default-application-1 | default     | State machine job | RECONCILING

# unknown state is forwarded as-is and simply matches nothing (no client-side validation)
$ confluent flink application list --environment default --filter state=bogus -o json
[]
```

[CF-4208]: https://confluentinc.atlassian.net/browse/CF-4208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/flink/command_application_list.go
```diff
@@ -18,6 +27,8 @@ func (c *command) newApplicationListCommand() *cobra.Command {
 	}
 
 	cmd.Flags().String("environment", "", "Name of the Flink environment.")
+	cmd.Flags().String("name", "", `Filter the Flink applications by name. Supports wildcards, for example "my-app*".`)
+	cmd.Flags().String("status", "", "Filter the Flink applications by status.")
```
sgagniere: I think we should name this `state` to match the filter key.

internal/flink/command_application_list.go
```diff
@@ -2,13 +2,22 @@ package flink
...
 	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
 	"github.com/confluentinc/cli/v4/pkg/output"
+	"github.com/confluentinc/cli/v4/pkg/utils"
 )
 
+// allowedApplicationStatuses lists the Flink job states recognized by the CMF applications
+// "state=" filter, per the cmf-sdk-go GetApplications filter documentation. Unknown values are
+// still forwarded (the server returns no matches rather than erroring); this list only drives
+// the advisory --status warning.
+var allowedApplicationStatuses = []string{"RUNNING", "FINISHED", "FAILED", "CANCELED", "RECONCILING", "COMPLETED", "UNKNOWN"}
```
sgagniere: Just to confirm: are all of these statuses accepted filter arguments? The spec's description for filterParam implies that only `RUNNING` or `FAILED` are valid; so the description may be out of date for the spec.

sgagniere (commented):
Hi, I have a few comments:

PR #3436, opened 2026-08-10
Migrate USM commands to generated versions

sgagniere:
Release Notes
-------------

Breaking Changes
- PLACEHOLDER

New Features
- `confluent unified-stream-manager [connect | kafka] [register | deregister]` have been renamed to `confluent usm [connect-cluster | kafka-cluster] [create | delete]` and the previous names have been added as aliases

Bug Fixes
- PLACEHOLDER

Checklist
---------

- [ ] I have indicated the potential customer impact if something goes wrong in the `Blast Radius` section below.
  - [x] Confluent Cloud prod
  - [ ] Confluent Cloud stag
  - [ ] Confluent Platform

What
----
Migrate USM commands to new generated versions.

To align with the spec, commands have been renamed and the former names have been added as aliases to prevent breaking changes:
- `unified-stream-manager` --> `usm`
- `connect` --> `connect-cluster`
- `kafka` --> `kafka-cluster`

The generated `usm connect-cluster` commands also retrieve `usm_kafka_cluster_id` directly from the API response now, whereas before it was a constructed field. For context, the previous implementation was done before `usm_kafka_cluster_id` was added to the API response.

Blast Radius
----
Blast radius would be limited to the USM commands. Commands have been renamed, but aliases have been added to preserve the previous names so there should not be any breaking change.

References
----------

Test & Review
-------------
The integration tests and their associated test handlers have been replaced with the generated versions.

Manual testing of the new commands and the new aliases: https://docs.google.com/document/d/1ao0caPEUF4w8r8MhIVM-PKfaJkSPeJifr88wuZWWSUg/edit?usp=sharing


internal/usm/command.go
```diff
@@ -0,0 +1,26 @@
...
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+	"github.com/confluentinc/cli/v4/pkg/config"
+)
+
+func New(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "usm",
+		Aliases: []string{"unified-stream-manager"},
+		Short:   "Manage Unified Stream Manager(USM).",
+	}
```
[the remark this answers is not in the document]
sgagniere: This is typo carried over from the spec.

internal/usm/command_connect_cluster_create.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// Required flags
+	cmd.Flags().String("confluent-platform-kafka-cluster", "", "The unique identifier of the metadata Kafka cluster for the Connect Cluster.")
+	cobra.CheckErr(cmd.MarkFlagRequired("confluent-platform-kafka-cluster"))
+
+	// Optional flags
+	pcmd.AddCloudFlag(cmd)
+	cmd.Flags().String("region", "", "The home region of the Confluent Platform Connect cluster where the metadata should be stored. This field is optional. If provided, 'cloud' must also be provided. If neither 'cloud' nor 'region' are provided, the home region of the associated metadata Kafka cluster (identified by 'kafka_cluster_id') will be used as a fallback.")
+	pcmd.AddEnvironmentFlag(cmd, c.AuthenticatedCLICommand)
```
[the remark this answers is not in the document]
sgagniere: We'll have to think about how to consistently generate this kind of validation.

But it shouldn't be a breaking change since we are not adding a new restriction, and this simply changes how the command fails.

internal/usm/command_kafka_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	*pcmd.AuthenticatedCLICommand
+}
+
+type kafkaClusterOut struct {
+	ID                              string `human:"ID" serialized:"id"`
+	Name                            string `human:"Name" serialized:"name"`
+	ConfluentPlatformKafkaClusterId string `human:"Confluent Platform Kafka Cluster Id" serialized:"confluent_platform_kafka_cluster_id"`
+	Cloud                           string `human:"Cloud" serialized:"cloud"`
+	Region                          string `human:"Region" serialized:"region"`
+	Environment                     string `human:"Environment" serialized:"environment"`
```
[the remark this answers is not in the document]
sgagniere: This is not a breaking change since the change is only in the human readable field name, but maybe we should add `ID` as an acronym.

What do you think @channingdong?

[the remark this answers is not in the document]
channingdong: Not a breaking change, I've come across this many times, as long as the serialized field name looks ok then it's fine.

`ID` should already be handled in terms of acronym detection, if the lint fails we can fix it somewhere.

internal/usm/command.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+	"github.com/confluentinc/cli/v4/pkg/config"
+)
+
+func New(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "usm",
```
channingdong, line 14: So the switch between `Use` and `Aliases` are expected, and aligns with the OpenAPI spec, correct?

internal/usm/command_connect_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	Environment                     string `human:"Environment" serialized:"environment"`
+	UsmKafkaClusterId               string `human:"USM Kafka Cluster Id" serialized:"usm_kafka_cluster_id"`
+	Cloud                           string `human:"Cloud" serialized:"cloud"`
+	Region                          string `human:"Region" serialized:"region"`
+}
+
+func newConnectClusterCommand(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command { //nolint:unparam
+	cmd := &cobra.Command{
+		Use:         "connect-cluster",
+		Aliases:     []string{"connect"},
```
channingdong, line 32: This is really nice, we can optionally remove the `Aliases` in v5 without too much efforts.

internal/usm/command_connect_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+		Short:       "Manage Confluent Cloud USM Connect clusters.",
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
channingdong, line 34: Is the `Annotations` here expected? The original requirement is:
```
pcmd.RequireCloudLogin
```

internal/usm/command_connect_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil
+	}
+	connectClusters, err := c.V2Client.ListUsmConnectClusters(environmentId)
+	if err != nil {
+		return nil
+	}
+
+	suggestions := make([]string, len(connectClusters))
+	for i, connectCluster := range connectClusters {
+		suggestions[i] = connectCluster.GetId()
```
channingdong, line 94: Is there any concern we alter the suggestion messages here? If this is the correct behavior (I believe so since auto-complete should focus on the connect cluster IDs), then it's totally fine.

internal/usm/command_connect_cluster_create.go
```diff
@@ -0,0 +1,80 @@
...
+
+	usmv1 "github.com/confluentinc/ccloud-sdk-go-v2/usm/v1"
+
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+	"github.com/confluentinc/cli/v4/pkg/errors"
+)
+
+func (c *connectClusterCommand) newCreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "register <confluent-platform-connect-cluster-id>",
```
channingdong: I remember the concept of `register` was during the initial project bringing up, according to the latest OpenAPI spec, maybe we can add an option to follow our convention to put `register` as Aliases?

This can be a follow-up though.

internal/usm/command_connect_cluster_create.go
```diff
@@ -0,0 +1,80 @@
...
+		Short: "Create a USM Connect cluster.",
+		Args:  cobra.ExactArgs(1),
+		RunE:  c.create,
```
channingdong: Any recommendation on the removed Example? Is it a good idea to manually copy the existing ones back on top of the generated code before we enable the example generation capability?

internal/usm/command_connect_cluster_create.go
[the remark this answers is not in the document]
channingdong: Exactly, `MarkFlagsRequiredTogether()` should be put into consideration for our client side validation bucket.

internal/usm/command_connect_cluster_delete.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	pcmd "github.com/confluentinc/cli/v4/pkg/cmd"
+	"github.com/confluentinc/cli/v4/pkg/deletion"
+)
+
+func (c *connectClusterCommand) newDeleteCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "deregister <id-1> [id-2] ... [id-n]",
```
channingdong: Same here and register.

internal/usm/command_connect_cluster_list.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return err
+	}
+
+	connectClusters, err := c.V2Client.ListUsmConnectClusters(environmentId)
+	if err != nil {
+		return err
+	}
+
+	list := output.NewList(cmd)
```
channingdong, line 42: Is the removal of `usmKafkaClusterId, ok := onPremToCloudKafkaIdMap[cluster.GetKafkaClusterId()]` only impacting the filtering of this list? 

I suppose it's UX behavior change instead of functional behavior change, right?
https://github.com/confluentinc/cli/blob/main/internal/unified-stream-manager/command_connect_list.go#L51

internal/usm/command_kafka_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func newKafkaClusterCommand(cfg *config.Config, prerunner pcmd.PreRunner) *cobra.Command { //nolint:unparam
+	cmd := &cobra.Command{
+		Use:         "kafka-cluster",
+		Aliases:     []string{"kafka"},
+		Short:       "Manage Confluent Cloud USM Kafka clusters.",
+		Annotations: map[string]string{pcmd.RunRequirement: pcmd.RequireNonAPIKeyCloudLogin},
```
channingdong, line 35: The same Annotation question applies here.

internal/usm/command_kafka_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil
+	}
+	kafkaClusters, err := c.V2Client.ListUsmKafkaClusters(environmentId)
+	if err != nil {
+		return nil
+	}
+
+	suggestions := make([]string, len(kafkaClusters))
+	for i, kafkaCluster := range kafkaClusters {
+		suggestions[i] = fmt.Sprintf("%s\t%s", kafkaCluster.GetId(), kafkaCluster.GetDisplayName())
```
channingdong, line 94: The suggestions here look different than the Connect clusters counterpart.

test/test-server/ccloudv2_router.go
```diff
@@ -132,10 +132,10 @@ var ccloudV2Routes = []route{
 	{"/tableflow/v1/catalog-integrations", handleCatalogIntegrations},
 	{"/tableflow/v1/tableflow-topics/{display_name}", handleTableflowTopic},
 	{"/tableflow/v1/tableflow-topics", handleTableflowTopics},
-	{"/usm/v1/kafka-clusters", handleUsmKafkaClusters},
-	{"/usm/v1/kafka-clusters/{id}", handleUsmKafkaCluster},
-	{"/usm/v1/connect-clusters", handleUsmConnectClusters},
-	{"/usm/v1/connect-clusters/{id}", handleUsmConnectCluster},
+	{"/usm/v1/connect-clusters", handleUsmV1ConnectClusters},
```
channingdong, line 135: Question: the USM generated integration test looks good in terms of the arbitrary ID issue?

internal/usm/command.go
sgagniere: Yup. `usm` is the API group instead of `unified-stream-manager`, hence the switch.

internal/usm/command_connect_cluster.go
sgagniere: Yeah. There's no actual difference between these two annotations since we don't support API key login yet.

There's a lot of references to API key logins in the CLI code, but it's ancient and unfinished.

internal/usm/command_connect_cluster.go
sgagniere: The automation dropped the descriptive bit because this resource doesn't have a name field, and I think I just picked something that seemed close when I originally implemented the old version.

It should not be a breaking change since autocomplete is for human users and not scripts/automation, so I think it's fine.

internal/usm/command_connect_cluster_list.go
sgagniere: It shouldn't be a change at all for the user since it was added to the backend response so that we didn't have to construct it on our end.

internal/usm/command_kafka_cluster.go
sgagniere: Yup that's because this one has a name field, so it's a natural description of the ID.

internal/usm/command_connect_cluster_create.go
sgagniere: Done!

channingdong (approved):

PR #3484, opened 2026-08-28
[DGS-25204] Do not send empty encrypted-key-material on dek create

karthikeyanas:
## Summary

https://confluentinc.atlassian.net/browse/DGS-25204

`confluent schema-registry dek create` returned HTTP 500 on shared KEKs when `--encrypted-key-material` was not provided.

`--encrypted-key-material` defaults to `""`, but the field was set unconditionally on `CreateDekRequest`, so an unset flag serialized as `"encryptedKeyMaterial": ""` — a present-but-empty JSON string. Schema Registry treats a present-but-empty value as caller-supplied ciphertext: it skips DEK generation, then fails the KMS round-trip (Base64-decoding `""` → empty bytes → KMS unwrap rejects), surfacing as `50070` / HTTP 500. Only shared KEKs hit this path; non-shared KEKs return a clean 4xx.

The sibling `--algorithm` field was already guarded with `cmd.Flags().Changed(...)` (#2811, "Do not pass \"\" to backend if --algorithm is unspecified") — the exact same bug, fixed for one field and missed for the other. This applies the identical guard.

## Changes

- `command_dek_create.go`: set `EncryptedKeyMaterial` only when the flag was explicitly changed.
- Integration test: added a `dek create` case with no `--encrypted-key-material` (the customer scenario).
- Mock server: assert the DEK POST request omits the field rather than sending an empty string — this is what distinguishes the bug from the fix, since the output table renders the field either way.

## Testing

- `TestSchemaRegistryDek` passes.
- Negative check: reverting only the one-line fix makes the new test fail with `Should NOT be empty, but was`, confirming it guards the regression.

## Scope

CLI half only. The two Schema Registry hardening items from triage (normalize empty→absent at the `createDek` boundary; classify KMS-rejected input as 4xx not 500) belong to the SR/DGov team and are tracked separately.

### DIFF
--- a/internal/schema-registry/command_dek_create.go
+++ b/internal/schema-registry/command_dek_create.go
@@ -77,13 +77,16 @@ func (c *command) dekCreate(cmd *cobra.Command, _ []string) error {
 	}
 
 	createReq := srsdk.CreateDekRequest{
-		Subject:              srsdk.PtrString(subject),
-		Version:              srsdk.PtrInt32(version),
-		EncryptedKeyMaterial: srsdk.PtrString(encryptedKeyMaterial),
+		Subject: srsdk.PtrString(subject),
+		Version: srsdk.PtrInt32(version),
 	}
 
 	if cmd.Flags().Changed("algorithm") {
 		createReq.Algorithm = srsdk.PtrString(algorithm)
+	}
+
+	if cmd.Flags().Changed("encrypted-key-material") {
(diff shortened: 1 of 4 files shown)


rayokota (approved):
Thanks @karthikeyanas  , LGTM

linouk23 (approved):
Thank you! Could you also add a manual test against devel / stag?

PR #3490, opened 2026-09-02
Rebase APIE-1608 branch onto current main

davidadas:
Rebases this branch onto current main. Not a normal rebase — this branch predates the agentdetect package's own merge to main (#3472), so it contains the full original agentdetect development history (20 commits), all of which conflict against main's squashed version of the same work.

Resolved by identifying the 6 commits actually unique to this branch (everything after the agentdetect package itself was already on main) and cherry-picking just those onto current main:

- b3da52279 Wire agentdetect.Detect into the usage-event path
- 0c99c1df4 Assign agent-detect attributes onto CliV1Usage
- 1d985b3ce Omit chain_shape when empty, not as ""
- acb5154f1 Recover panics inside the walk goroutine, not the caller
- 3b804929a Test the walk-goroutine panic recovery, flag schema risk
- 424d26861 reword comments for clarity

All 6 applied cleanly (one trivial auto-merge in internal/command.go, no manual conflict resolution needed), original authorship preserved.

Still doesn't compile — same single expected failure as before, unrelated to this rebase: the 9 CliV1Usage fields this branch assigns onto don't exist in ccloud-sdk-go-v2/cli yet. That's confluentinc/api#2710 (APIE-1607), still open.

PR #3491, opened 2026-09-02
[APIE-1608] Wire agentdetect into the usage-report path (rebased)

davidadas:
Supersedes #3487 (draft, stale — merge-base predates agentdetect's own merge to main via #3472, so it couldn't be rebased cleanly; see closed #3490 for why a PR into that branch didn't work either).

Same content as #3487, cherry-picked cleanly onto current main: the 6 commits that are actually unique to that branch (the agentdetect package itself is already on main via #3472, this is just the wiring on top).

- Wire agentdetect.Detect into the usage-event path
- Assign agent-detect attributes onto CliV1Usage
- Omit chain_shape when empty, not as ""
- Recover panics inside the walk goroutine, not the caller
- Test the walk-goroutine panic recovery, flag schema risk
- reword comments for clarity

Original authorship preserved throughout (all commits are Noel Cothren's).

Still doesn't compile — same single expected reason as #3487: the 9 CliV1Usage fields this assigns onto don't exist in ccloud-sdk-go-v2/cli yet. Blocked on confluentinc/api#2710 (APIE-1607, currently mergeable, pending review) landing and the SDK regenerating from it.

Also has 4 open Copilot findings carried over from #3487 (unaddressed): panic value logged raw at pkg/usage/usage.go:55, a test stricter than its own production contract at pkg/agentdetect/proc_gopsutil_test.go:35, a non-stoppable time.After on the hot path at pkg/agentdetect/detect.go:340, and an optionalStrings/Flags nil-handling inconsistency at pkg/usage/usage.go:85.

### DIFF
--- a/internal/command.go
+++ b/internal/command.go
@@ -195,6 +195,7 @@ func reportUsage(cmd *cobra.Command, cfg *config.Config, u *usage.Usage) error {
 		if err != nil {
 			return err
 		}
+		u.CollectAgentDetect()
 		u.Report(ccloudv2.NewClient(cfg, unsafeTrace))
 	}
 	return nil
--- a/internal/flink/command_shell.go
+++ b/internal/flink/command_shell.go
@@ -360,6 +360,9 @@ func reportUsage(cmd *cobra.Command, cfg *config.Config, unsafeTrace bool) func(
 
 	return func() {
 		u := ppanic.CollectPanic(cmd, nil, cfg)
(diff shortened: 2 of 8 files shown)


noeldevelops:
@davidadas I will close this in favor of https://github.com/confluentinc/cli/pull/3515 , since I couldn't write updates to your forked branch. TY for your help here !

